PR #118, opened 2025-08-13
fix: use correct path to junie mcp config

mattlewis92:
Thanks for creating this package, it's really helpful! When testing it, I noticed it was using the wrong mcp file path for junie: https://www.jetbrains.com/help/junie/model-context-protocol-mcp.html#bksdkr_21


cm-dyoshikawa:
Good catch! Thank you so much 👍

PR #126, opened 2025-08-14
fix: actually fix junie MCP path

mattlewis92:
Follow up to https://github.com/dyoshikawa/rulesync/pull/118 where the junie MCP path was not changed in all places


dyoshikawa (approved):
thank you!

PR #608, opened 2025-11-28
fix: write frontmatter when generating skill files

mkanakana:
writeAiDirs was only writing mainFile.body without serializing the frontmatter, causing generated skill files to lack required YAML frontmatter (name, description, etc.)

- Use stringifyFrontmatter to serialize body with frontmatter in DirFeatureProcessor.writeAiDirs
- Add round-trip tests to verify frontmatter is written and readable


cm-dyoshikawa (approved):
Thank you!

PR #623, opened 2025-12-01
refactor: add isDeletable() polymorphic method to AiFile

mkanakana:
## Summary
- Add `isDeletable()` polymorphic method to `AiFile` base class for deletion control
- Override in specific classes: `ClaudecodeMcp`, `OpencodeMcp`, `ClaudecodeIgnore`
- Replace `instanceof` checks with polymorphic method calls in processors

## Test plan
- [x] Existing tests pass (2917 tests)
- [x] Added unit tests for `isDeletable()` behavior in each class
- [x] Verified deletion filtering works correctly in processors

### DIFF
--- a/src/features/mcp/claudecode-mcp.test.ts
+++ b/src/features/mcp/claudecode-mcp.test.ts
@@ -36,6 +36,30 @@ describe("ClaudecodeMcp", () => {
     });
   });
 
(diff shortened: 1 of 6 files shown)


cm-dyoshikawa:
It looks very good. I'll do detailed review later.

PR #693, opened 2025-12-17
fix: add forDeletion pattern to handle files with incompatible formats

mkanakana:
## Summary
- Add `forDeletion` static method to all tool base classes (ToolRule, ToolCommand, ToolSubagent, ToolMcp, ToolIgnore, ToolSkill)
- Implement `forDeletion` in all concrete subclasses to create minimal instances without parsing file content
- Modify all processors to use `forDeletion` when loading files/dirs for deletion
- Add tests for broken frontmatter scenarios

## Background
When using the `--delete` option with `rulesync generate`, `loadToolFiles`/`loadToolDirs` would parse existing files. If files had old/incompatible formats (e.g., broken YAML frontmatter), parsing would fail and deletion would fail.

## Test plan
- [x] Add tests for broken frontmatter in rules-processor.test.ts
- [x] Add tests for broken frontmatter in subagents-processor.test.ts
- [x] Add tests for broken frontmatter in skills-processor.test.ts
- [x] Run `pnpm cicheck:code` - all 3034 tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #707, opened 2025-12-22
fix: add forDeletion pattern to mcp feature

mkanakana:
## Summary
- Add `forDeletion` static method to ToolMcp base class and all concrete mcp subclasses
- Update McpProcessor to use `forDeletion` when loading files for deletion
- Update existing tests to use forDeletion mocks

## Background
When using the `--delete` option with `rulesync generate`, `loadToolFiles` would parse existing files. If files had old/incompatible formats, parsing would fail and deletion would fail.

This is part 4 of splitting #693 into smaller PRs for easier review.

## Test plan
- [x] All 489 mcp tests pass
- [x] Run `pnpm vitest run src/features/mcp/`


cm-dyoshikawa:
@mkanakana Thank you for your PRs!

PR #847, opened 2026-01-20
fix(import): default subagent targets to wildcard

TanishqChavan10:
Fix: Default imported subagents to wildcard targets
Summary

This PR updates the rulesync import conversion for subagents so that imported .rulesync/subagents/*.md files are tool-agnostic by default.

Previously, the import converters hard-coded the source tool (e.g. claudecode, opencode, copilot) into the targets field. This caused imported subagents to be unnecessarily restricted to a single tool.


cm-dyoshikawa:
@TanishqChavan10 Good catch, thank you. It was a bug certainly.

PR #859, opened 2026-01-21
fix: a warn log when there is not a root rule file  #445

### LINKED ISSUE #445: Output a warn log when there is not a root rule file
To recommend a user to set a root file when using Rulesync rule.

karthik-nair-20:
Screenshot 2026-01-22 at 14 43 56

FIX #445

### DIFF
--- a/src/features/rules/rules-processor.ts
+++ b/src/features/rules/rules-processor.ts
@@ -612,7 +612,7 @@ export class RulesProcessor extends FeatureProcessor {
 
     if (rootRules.length === 0 && rulesyncRules.length > 0) {
       logger.warn(
-        `No root rulesync rule file found. Consider adding 'root: true' to one of your rule files in .rulesync/rules/.`
+        `No root rulesync rule file found. Consider adding 'root: true' to one of your rule files in .rulesync/rules/.`,
       );
     }
 


cm-dyoshikawa:
[review]

dyoshikawa:
Thanks for the PR! I have some feedback:

### 1. Location of the warning

The current implementation is in `loadToolFiles()`, but I believe the intent of Issue #445 is to warn when there's no root file in the **rulesync rule files** (`.rulesync/rules/`).

`loadToolFiles()` loads generated tool files (`.cursorrules`, `CLAUDE.md`, etc.), so the warning should be moved to `loadRulesyncFiles()`.

### 2. Suggested implementation

Add to `loadRulesyncFiles()` method (around line 609, after the multiple root rules check):

```typescript
const rootRules = rulesyncRules.filter((rule) => rule.getFrontmatter().root);

if (rootRules.length > 1) {
  throw new RulesProcessorError(/* existing error */);
}

// Add here
if (rootRules.length === 0 && rulesyncRules.length > 0) {
  logger.warn(
    \`No root rulesync rule file found. Consider adding 'root: true' to one of your rule files in .rulesync/rules/.\`
  );
}
```

### 3. Issues with current implementation

- Warning fires even for tools where `settablePaths.root` is undefined (no root file support)
- Warning fires during deletion operations (`forDeletion: true`)
- Warning may be logged multiple times when generating for multiple tools

### 4. Tests

Could you also add tests to verify the warning behavior?

cm-dyoshikawa:
@karthik-nair-20 Thank you!

Issue #831, opened 2026-01-18
Support any AI agent's hooks


cm-dyoshikawa:

land0r:
+1 for that feature, as we have hooks in [ClaudeCode](https://code.claude.com/docs/en/hooks-guide) and [Cursor](https://cursor.com/docs/agent/hooks).

Thanks!

PR #901, opened 2026-01-29
Hooks support for Cursor + Claude

TylorS:
First off, thank you for this project!  Cursor and I made an attempt to start including support for hooks.

Closes: https://github.com/dyoshikawa/rulesync/issues/831

Behavior at a glance
- Generate (Cursor): Uses shared hooks (only Cursor-supported events) + cursor.hooks; Cursor-only events can live in shared or in cursor.hooks.
- Generate (Claude): Uses shared hooks (only Claude-supported events) + claudecode.hooks; keys become PascalCase; non-absolute commands get $CLAUDE_PROJECT_DIR/; Claude-only events can live in shared or in claudecode.hooks.
- Import: Cursor/Claude tool files are converted back to canonical camelCase and written into the shared hooks block (overrides are for authoring; import normalizes into shared).
- Skipped events: If the config (shared or overrides) mentions events the current target doesn’t support, one warning per target is logged listing those events.

### DIFF
--- a/src/features/hooks/claudecode-hooks.test.ts
+++ b/src/features/hooks/claudecode-hooks.test.ts
@@ -195,7 +195,7 @@ describe("ClaudecodeHooks", () => {
       const rulesyncHooks = claudecodeHooks.toRulesyncHooks();
       const json = rulesyncHooks.getJson();
       expect(json.hooks.sessionStart).toHaveLength(1);
-      expect(json.hooks.sessionStart[0].command).toContain("echo.sh");
+      expect(json.hooks.sessionStart?.[0]?.command).toContain("echo.sh");
       expect(json.hooks.stop).toHaveLength(1);
     });
   });
--- a/src/features/hooks/cursor-hooks.test.ts
+++ b/src/features/hooks/cursor-hooks.test.ts
@@ -168,9 +168,9 @@ describe("CursorHooks", () => {
       const backToRulesync = loaded.toRulesyncHooks();
       const json = backToRulesync.getJson();
       expect(json.hooks.sessionStart).toHaveLength(1);
-      expect(json.hooks.sessionStart[0].command).toBe(".rulesync/hooks/session-start.sh");
+      expect(json.hooks.sessionStart?.[0]?.command).toBe(".rulesync/hooks/session-start.sh");
       expect(json.hooks.postToolUse).toHaveLength(1);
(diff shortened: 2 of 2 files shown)


dyoshikawa:
@TylorS 

Thank you for this contribution! 🙏 This is a great addition to Rulesync, and the overall approach is solid.

The implementation follows the existing codebase patterns well (similar structure to `McpProcessor`, `CommandsProcessor`, etc.), includes comprehensive tests, and the documentation is clear with good examples. The security model is also appropriate - Rulesync acts as a configuration generator, not an executor, which correctly delegates command execution security to the downstream tools.

---

### Items to Address Before Merge

#### 1. README.md "Supported Tools and Features" table needs update

Per `feature-change-guidelines.md`, the table should include a "hooks" column to stay synchronized with implemented functionality.

#### 2. Redundant conditional in `ClaudecodeHooks.getSettablePaths()`

**File:** `src/features/hooks/claudecode-hooks.ts:546-550`

```typescript
static getSettablePaths({ global }: { global?: boolean } = {}): ToolHooksSettablePaths {
  if (global) {
    return { relativeDirPath: ".claude", relativeFilePath: "settings.json" };
  }
  return { relativeDirPath: ".claude", relativeFilePath: "settings.json" };
}
```

Both branches return the same value. Either remove the conditional or implement different paths for global mode if needed.

#### 3. Missing try-catch for JSON parsing

**File:** `src/features/hooks/claudecode-hooks.ts`

```typescript
const settings: Record<string, unknown> = JSON.parse(existingContent);
```

If `existingContent` is malformed JSON, this will throw an uncaught exception. Consider wrapping with try-catch and using `formatError` for error logging.

#### 4. Type guard is too permissive

**File:** `src/features/hooks/claudecode-hooks.ts`

```typescript
function isClaudeMatcherEntry(x: unknown): x is ClaudeMatcherEntry {
  return x !== null && typeof x === "object";
}
```

This type guard allows any non-null object through. Consider validating the structure more thoroughly (e.g., checking for expected properties).

---

### Suggestions (Nice to have)

- Consider whether `gitignore.ts` needs to include `.cursor/hooks.json`
- The unused `rulesyncHooks` parameter in constructor is destructured and discarded - consider removing it from the type or using it
- Consider adding JSDoc comments to public APIs

---

Overall, this is well-structured work. Looking forward to seeing these minor items addressed! 👍

PR #908, opened 2026-01-31
Fix Claude code paths/globs to be an array

boazy:
This PR fixes Claude code paths/globs to be an array

According to the [documentation](https://code.claude.com/docs/en/memory#path-specific-rules), Claude Code `paths` values should be arrays of strings, not a comma-separated string.

It seems like the syntax on the documentation was either [initially incorrect or has changed](https://github.com/anthropics/claude-code/issues/13905).


cm-dyoshikawa:
@boazy Thank you for your contribution! Please check the ci failure. After that, I will merge it.

PR #976, opened 2026-02-06
fix: update opencode-mcp global path

### LINKED ISSUE #835: OpenCode MCP output path mismatch (global mode)
When running:

```bash
rulesync generate --targets opencode --features mcp --global
```

Rulesync writes the OpenCode MCP config to `~/opencode.json`.

But OpenCode reads its config from `~/.config/opencode/opencode.json` (Windows: `%USERPROFILE%\.config\opencode\opencode.json`). As a result, the generated MCP servers are not picked up unless the file is copied/moved.

Suggested fix: in global mode, write OpenCode MCP output to `~/.config/opencode/opencode.json` (align with other OpenCode global outputs like skills/commands).

r1bilski:
Currently, we're setting `relativeDirPath` to `.` for global config.

This updates global `relativeDirPath` to the correct `.config/opencode` path.

closes #835

### DIFF
--- a/src/features/mcp/opencode-mcp.test.ts
+++ b/src/features/mcp/opencode-mcp.test.ts
@@ -32,7 +32,7 @@ describe("OpencodeMcp", () => {
     it("should return correct paths for global mode", () => {
       const paths = OpencodeMcp.getSettablePaths({ global: true });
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
Thank you!

PR #1001, opened 2026-02-09
feat: implement opencode hooks

### LINKED ISSUE #928: Support opencode plugins as a hooks generation target
In Opencode hooks are called plugins. We could probably create a rulesync format that can generate hooks in claude and cursor and plugins in opencode

r1bilski:
This:
- implements hooks feature for opencode using opencode's plugin system.
- restructures existing hooks implementations: current code was based on the assumption that the only agents with hooks were claude code and cursor - refactored to make it more general: for example removed the list of common hook events - instead each implementation declares the full list of supported events
- added hook type: command and prompt and added supportedHookTypes to hooks implementations. Opencode does not support injecting content into the prompt, so it only supports commands.
- added a warning notifying the user if a hook has an unsupported type for the selected target

**Notes:**
- Opencode does not have a hook system like claude code or cursor, instead it has a full plugin system. That means that the hooks are not translated into a simple json file, but a js plugin that 'emulates' hook behavior found in the other implementations. This makes it impossible to implement importing rulesync hooks config from opencode.
- I wasn't sure about how to implement the fact that we can't import hooks from opencode - I added a new flag `supportsImport`, since it didn't seem to fit the `simulated` pattern for me.
- This PR implements only events already present - opencode does support more events that are not in cursor or claude code, which I plan to introduce in another PR.
- Additional refactors to the base hooks functionality could be made to decouple it from Cursor being used as the baseline (e.g. changing `CURSOR_TO_*_EVENT_NAMES` to a more generic nomenclature - we already describe these events as 'canonical' events in some places.)

closes #928

### DIFF
--- a/config-schema.json
+++ b/config-schema.json
@@ -49,16 +49,7 @@
           "type": "array",
           "items": {
             "type": "string",
-            "enum": [
-              "rules",
-              "ignore",
-              "mcp",
-              "subagents",
-              "commands",
-              "skills",
-              "hooks",
-              "*"
-            ]
+            "enum": ["rules", "ignore", "mcp", "subagents", "commands", "skills", "hooks", "*"]
           }
         },
         {
@@ -70,16 +61,7 @@
             "type": "array",
             "items": {
               "type": "string",
-              "enum": [
-                "rules",
-                "ignore",
-                "mcp",
-                "subagents",
-                "commands",
-                "skills",
-                "hooks",
-                "*"
-              ]
+              "enum": ["rules", "ignore", "mcp", "subagents", "commands", "skills", "hooks", "*"]
             }
           }
         }


cm-dyoshikawa:
[ai] review

dyoshikawa:
@r1bilski Thank you. I'm worried about how we should handle opencode plugins beacuse it is quite diffrent from the others. I might take over this PR. Even in that case, I'll pull your commit into the main branch.

r1bilski:
@dyoshikawa No worries, I don't mind that. I hope this acts as a good starting point for you.
Opencode's implementation is quite different, it fits into the 'simulated' kind a little bit, because while the plugins have access to hooks and can be used to emulate claude's / cursor's hooks, they're much more than that.

If you need any help with that or want to me to implement any changes, let me know!

r1bilski:
@dyoshikawa No worries, I don't mind that. I hope this acts as a good starting point for you.
Opencode's implementation is quite different, it fits into the 'simulated' kind a little bit, because while the plugins have access to hooks and can be used to emulate claude's / cursor's hooks, they're much more than that.

If you need any help with that or want to me to implement any changes, let me know!

cm-dyoshikawa:
@r1bilski Thank you!

PR #973, opened 2026-02-06
feat: declarative skill sources with lockfile and curated directory

mrfelton:
## Summary

Adds **Declarative Skill Sources** — a new feature allowing users to declare remote GitHub skill repositories in `rulesync.jsonc`. During `generate`, these sources are automatically fetched, SHA-locked, and placed into `.rulesync/skills/.curated/` for use alongside local skills.

## Motivation

Currently, fetching skills from external repos requires manually running `rulesync fetch` for each source. This feature makes skill sourcing declarative and reproducible: declare once in config, and `generate` handles the rest — with lockfile-based determinism and local-skill precedence.

### Core Implementation
- **`src/lib/sources.ts`** — Main orchestrator: resolves source entries, fetches remote skill directories into `.curated/`, handles local-skill precedence, duplicate detection, and error recovery
- **`src/lib/sources-lock.ts`** — Lockfile operations: read/write `.rulesync/sources-lock.json` with Zod schema validation, immutable set/get helpers
- **`src/lib/github-client.ts`** — New `resolveRefToSha()` method to resolve branch/tag/SHA to commit SHA via `repos.getCommit()`

### Integration
- **`src/lib/generate.ts`** — Calls `resolveAndFetchSources()` before skill generation when `sources` is configured
- **`src/cli/commands/generate.ts`** — Wires `--skip-sources` and `--update-sources` CLI flags; validates mutual exclusivity
- **`src/cli/index.ts`** — Registers `--skip-sources` and `--update-sources` CLI options
- **`src/config/config.ts`** — Adds `sources` field to config schema with `SourceEntry` type
- **`src/config/config-resolver.ts`** — Resolves `sources` from config file (not CLI); type narrowed with `Omit<ConfigParams, 'sources'>`
- **`src/features/skills/skills-processor.ts`** — `loadRulesyncDirs()` now discovers curated skills in `.curated/` subdirectory with local-precedence filtering
- **`src/cli/commands/gitignore.ts`** — Adds `.rulesync/skills/.curated/` to gitignore entries
- **`src/constants/rulesync-paths.ts`** — New constants for curated skills dir and sources-lock path

### Security
- Path traversal validation on `skillDir.name` (rejects `..`, `/`, `\` characters)
- Path traversal validation on file paths within skill directories via `checkPathTraversal()`
- File size limits enforced via `MAX_FILE_SIZE` (10MB)

### Tests (27 new tests)
[...]

PR #1032, opened 2026-02-11
feat: add `rulesync install` command for declarative skill sources

mrfelton:
## Summary

Introduces a new `rulesync install` command for declarative skill sources, keeping `generate` pure (no network I/O) and `fetch` unchanged.

Note, this is an alternative approach to #973 that supersedes that PR.

## Motivation

Teams want to declare shared skill sources in `rulesync.jsonc` and have them automatically resolved and fetched. The current `fetch` command is designed for one-shot pulls, which does not cover the use case of keeping a curated set of skills in sync across a team.

### Security concerns with fetched skills

Since skills contain instructions and prompts that directly influence AI tool behavior, fetching untrusted code introduces multiple prompt injection and supply chain risks:

- **Skill content injection**: A malicious upstream skill could contain prompts designed to bypass security policies, exfiltrate data, or manipulate AI decision-making across your entire team. Skills are not sandboxed and can influence how AI assistants behave during code generation and analysis.
- **Silent override attacks**: Without separation, a malicious skill matching a trusted local skill's name could silently override it, affecting all team members who pull it. This is especially dangerous in team settings where shared curated skills are distributed.
- **Supply chain risk**: Depending on external repositories means trusting their maintainers, infrastructure, and that they never get compromised. Popular skills could be a high-value attack target.
- **Lack of auditability**: Without a lockfile, you cannot tell which version was fetched, when it changed, or if it was modified. This breaks the ability to investigate security incidents or roll back compromises.

This PR addresses these concerns by:

- **Separating curated skills** into `.rulesync/skills/.curated/`, making it clear which skills come from external sources
- **Prioritizing local skills** -- local skill definitions always take precedence over curated ones, allowing teams to shield themselves from upstream changes
- **Creating an immutable lockfile** -- `rulesync.lock` (at the project root) pins exact commit SHAs and integrity hashes (SHA-256) for each fetched skill, enabling reproducible builds and tampering detection
- **Enabling CI verification** -- `--frozen` mode fails if the lockfile is missing or stale, ensuring teams control when and how skills are updated

Related: vercel-labs/skills#283

### Separate `install` command instead of extending `generate`

Per maintainer feedback on #973, network I/O does not belong in `generate`. The new `rulesync install` command is the only entry point for source resolution and fetching. `generate` remains a pure local operation.

### Lockfile (`rulesync.lock` at project root)

A lockfile pins resolved SHAs, integrity hashes (SHA-256), and timestamps. This provides critical supply chain security and auditability:

- **Reproducible installs** -- the same lockfile always produces the same skill files, enabling bit-for-bit reproducible builds across machines and team members
- **Tampering detection** -- per-skill integrity hashes allow verification that fetched skills haven't been modified between fetch and generate operations
- **`--frozen` mode** for CI -- fails if the lockfile is missing or stale, never writes. This ensures CI systems cannot silently update to new upstream versions without explicit approval
- **Auditability** -- `requestedRef`, `resolvedRef`, and `resolvedAt` fields provide a complete audit trail of what was fetched, when, and from where. Enables security incident investigation and rollback decisions
- **Supply chain transparency** -- the lockfile can be committed to version control, allowing teams to review and approve updates to external skill sources

Without a lockfile, every `install` would hit the GitHub API and could silently pick up upstream changes, breaking reproducibility and making it impossible to detect supply chain compromises.

### Curated skills directory (`.rulesync/skills/.curated/`)

Skills fetched from sources are written to `.curated/` inside the skills directory. During `generate`, the skills processor merges curated and local skills, with **local skills always winning on name conflicts**. This two-tier approach provides supply chain isolation:

- **Transparency**: Team members can immediately see which skills are external (in `.curated/`) vs. internal (in `.skills/`)
- **Override capability**: Teams can protect themselves by defining local skills with the same name as upstream skills, ensuring only trusted code influences AI tools
- **Gitignore separation**: Curated skills are automatically added to `.gitignore`, preventing accidental commits of downloaded code and keeping git history clean
- **Controlled updates**: Unlike `fetch` which immediately overwrites, `install` fetches into `.curated/` where they don't affect `generate` unless explicitly accepted (by not defining a conflicting local skill)

### Lockfile placement at project root

The lockfile is placed at the project root (next to `rulesync.jsonc`) following industry standard conventions where lockfiles live alongside their manifests:

- **npm/pnpm**: `package.json` + `package-lock.json` / `pnpm-lock.yaml`
- **Cargo**: `Cargo.toml` + `Cargo.lock`
- **Poetry**: `pyproject.toml` + `poetry.lock`

This placement ensures:

- **Clear visibility**: The lockfile immediately appears in team diffs and code reviews, making supply chain changes impossible to miss
- **Standard expectations**: Developers familiar with npm, cargo, poetry, etc. will immediately know where the lockfile is
- **Git workflow clarity**: Clean version control without awkward `.gitignore` patterns to exclude runtime directories while preserving the lockfile
- **Security auditability**: Lockfile updates (critical for supply chain integrity) are prominently logged in git history and require explicit approval

### Source parser extraction

`parseSource`, `parseUrl`, and `parseShorthand` were extracted from `fetch.ts` into a new `source-parser.ts` module so both `fetch` and `install` can share the parsing logic. `fetch.ts` re-exports these functions for backward compatibility.

## Usage

```jsonc
// rulesync.jsonc
{
  "sources": [
    // Fetch all skills from a repository
    { "source": "owner/repo" },

    // Fetch only specific skills by name
    { "source": "anthropics/skills", "skills": ["skill-creator"] },

    // Pin to a ref and specify a subdirectory path
    { "source": "owner/repo@v1.0.0:path/to/skills" }
  ]
}
```

Each source entry is an object with:

- `source` (string, required) -- repository reference using the same format as the `fetch` command (`owner/repo`, `owner/repo@ref`, `github:owner/repo@ref:path`, etc.)
- `skills` (string[], optional) -- list of skill names to fetch. If omitted, all skills are fetched.

```bash
# Install skills using locked refs
rulesync install

# Force re-resolve all refs to latest
rulesync install --update

# CI mode: fail if lockfile is missing or stale
rulesync install --frozen

# Typical workflow
rulesync install && rulesync generate
```

## Testing

- 3786 tests across 141 files, all passing
- `pnpm cicheck` passes (fmt, oxlint, eslint, typecheck, test, cspell, secretlint)
- New test files cover source parsing, lockfile operations, source resolution, install command
- Existing tests updated for gitignore `.curated/` entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ rulesync init
 
 # Install official skills (recommended)
 rulesync fetch dyoshikawa/rulesync --features skills
+
+# Or add skill sources to rulesync.jsonc and run 'rulesync install' (see "Declarative Skill Sources")
 ```
 
 On the other hand, if you already have AI tool configurations:
@@ -241,6 +243,18 @@ rulesync generate --dry-run --targets claudecode --features rules
 
 # Check if files are up to date (for CI/CD pipelines)
 rulesync generate --check --targets "*" --features "*"
+
+# Install skills from declarative sources in rulesync.jsonc
+rulesync install
+
+# Force re-resolve all source refs (ignore lockfile)
+rulesync install --update
+
+# Fail if lockfile is missing or out of sync (for CI)
+rulesync install --frozen
+
+# Install then generate (typical workflow)
+rulesync install && rulesync generate
 
 # Add generated files to .gitignore
 rulesync gitignore
@@ -399,6 +413,13 @@ Example:
   "simulateCommands": false, // Generate simulated commands
   "simulateSubagents": false, // Generate simulated subagents
   "simulateSkills": false, // Generate simulated skills
+
+  // Declarative skill sources — installed via 'rulesync install'
+  // See the "Declarative Skill Sources" section for details.
+  // "sources": [
+  //   { "source": "owner/repo" },
+  //   { "source": "org/repo", "skills": ["specific-skill"] },
+  // ],
 }
 ```
 
@@ -697,20 +718,26 @@ You can use global mode via Rulesync by enabling `--global` option. It can also 
 Currently, supports rules and commands generation for Claude Code. Import for global files is supported for rules and commands.
 
 1. Create an any name directory. For example, if you prefer `~/.aiglobal`, run the following command.
+
    ```bash
    mkdir -p ~/.aiglobal
    ```
+
 2. Initialize files for global files in the directory.
+
    ```bash
    cd ~/.aiglobal
    rulesync init
    ```
+
 3. Edit `~/.aiglobal/rulesync.jsonc` to enable global mode.
+
    ```jsonc
    {
      "global": true,
    }
    ```
+
 4. Edit `~/.aiglobal/.rulesync/rules/overview.md` to your preferences.
 
    ```md
@@ -724,6 +751,7 @@ Currently, supports rules and commands generation for Claude Code. Import for gl
    ```
 
 5. Generate rules for global settings.
+
    ```bash
    # Run in the `~/.aiglobal` directory
    rulesync generate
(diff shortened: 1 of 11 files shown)


cm-dyoshikawa:
@mrfelton Thank you, I've added some fixes and merged!

PR #1044, opened 2026-02-12
fix: enable GitHub installs by committing dist and auto-rebuilding on commit

mrfelton:
### Problem

When this package is installed directly from GitHub (e.g., `npm install github:dyoshikawa/rulesync#<commit-sha>`), the `rulesync` CLI command is unavailable because:

1. The `dist/` directory is gitignored and not committed to the repository
2. npm's GitHub install flow does not run build scripts (`prepare`/`prepack`) on the dependency — it simply extracts the repo tarball and applies the `files` filter from package.json
3. Since `dist/` doesn't exist in the repo, the downloaded package contains no built CLI executable

This breaks workflows where teams need to use unreleased code without manually cloning and building locally.

### Solution

Commit the built `dist/` artifacts to version control, and automate keeping them in sync via a pre-commit hook.

### Changes

1. **Unignore `dist/` in `.gitignore`** so built artifacts can be committed
2. **Commit `dist/` directory** with current build output
3. **Auto-rebuild on commit** via the existing `simple-git-hooks` pre-commit hook: when any `src/` files are staged, `pnpm build` runs and `dist/` is automatically re-staged — no manual steps required
4. **Add `prepack` script** as a safety net that rebuilds before `npm publish`
5. **Make `prepare` graceful** with `|| true` so git installs don't fail when pnpm isn't globally available

### Workflow Impact

**No changes to daily development workflow.** The pre-commit hook handles everything automatically:

- When committing changes that include `src/` files → the hook detects this, runs `pnpm build`, and stages the updated `dist/` alongside the source changes
- When committing changes that don't touch `src/` (e.g., docs, config) → no rebuild, no delay
- `npm publish` → the `prepack` hook rebuilds as an additional safety net

Contributors do not need to remember to run `pnpm build` manually. The hook takes care of it.

### Validation

Tested with a consumer project installing via GitHub reference:
- `npm install github:LN-Zap/rulesync#<commit>` installs successfully
- `npx rulesync install && npx rulesync generate` executes without errors
- `rulesync --version` returns the correct version
- Verified pre-commit hook auto-rebuilds `dist/` when `src/` files are staged

### Backwards Compatibility

No breaking changes. Fully compatible with npm registry installs. The only difference is that `dist/` is now tracked in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,8 @@ out
 # Nuxt.js build / generate output
 .nuxt
 dist
+!/dist/
+!/dist/**
 
 # Gatsby files
 # Comment in the public line in if your project uses Gatsby and not Next.js
--- a/dist/index.cjs
+++ b/dist/index.cjs
@@ -1,0 +1,18660 @@
+#!/usr/bin/env node
+"use strict";
+var __create = Object.create;
+var __defProp = Object.defineProperty;
+var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
+var __getOwnPropNames = Object.getOwnPropertyNames;
+var __getProtoOf = Object.getPrototypeOf;
+var __hasOwnProp = Object.prototype.hasOwnProperty;
+var __copyProps = (to, from, except, desc) => {
+  if (from && typeof from === "object" || typeof from === "function") {
+    for (let key of __getOwnPropNames(from))
+      if (!__hasOwnProp.call(to, key) && key !== except)
+        __defProp(to, key, { get: () => from[key], enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable });
+  }
+  return to;
+};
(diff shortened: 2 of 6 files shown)


cm-dyoshikawa:
@mrfelton Thanks for all your hard work on this! I'm listening, but I have some reservations about this suggestion as it's not a standard practice for me.

We already support npm, Homebrew, and binary downloads via GitHub Releases, which should cover most use cases. My main concern is that committing the dist/* folder makes reviewing diffs a nightmare.

mrfelton:
I definitely understand the reservation. That said, it is a pretty common thing to do and many popular libraries take this approach. It lowers the barrier to contribution and enables a more ways to source in the library.

My main motivation for pushing this was to enable me to fully test and integrate my changes from https://github.com/dyoshikawa/rulesync/pull/1032 before it made it's way into an official release.

I'm happy for it to be closed if you don't want to support the git installation method.

dyoshikawa:
@mrfelton Okay, sorry, I'll close this to keep current simple operation.

Thank you for the suggestion!

PR #1093, opened 2026-02-16
fix(install): allow --frozen to fetch missing locked skills

mrfelton:
## Summary
- Adjust `rulesync install --frozen` behavior to match lockfile-driven install semantics
- Keep strict lockfile coverage checks in frozen mode
- Allow missing curated skills to be fetched using locked refs instead of failing
- Keep frozen mode non-mutating for lockfile writes

## Why
Current frozen mode fails when locked skills are missing on disk. This makes first-run/clean-environment installs fail even when lockfile state is valid. The new behavior keeps deterministic lock semantics while allowing bootstrap from the lockfile.

## Changes
- `src/lib/sources.ts`
  - Remove frozen-mode disk-presence hard failure
  - Keep lockfile coverage validation
- `src/lib/sources.test.ts`
  - Replace failing-missing-skill frozen test with success path that fetches missing locked skill
  - Assert no lockfile writes in frozen mode
- `src/cli/index.ts`
  - Clarify `--frozen` help text
- Docs updated to reflect behavior:
  - `docs/guide/declarative-sources.md`
  - `docs/reference/cli-commands.md`
  - `docs/getting-started/quick-start.md`
  - mirrored rulesync skill docs under `skills/rulesync/*`

## Validation
- `pnpm vitest src/lib/sources.test.ts src/cli/commands/install.test.ts`
- `pnpm test`

### DIFF
--- a/docs/guide/declarative-sources.md
+++ b/docs/guide/declarative-sources.md
@@ -47,11 +47,11 @@ When `rulesync install` runs and `sources` is configured:
 
 The `install` command accepts these flags:
 
-| Flag              | Description                                                                           |
-| ----------------- | ------------------------------------------------------------------------------------- |
-| `--update`        | Force re-resolve all source refs, ignoring the lockfile (useful to pull new updates). |
+| Flag              | Description                                                                                                                                                  |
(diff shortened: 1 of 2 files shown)


cm-dyoshikawa:
@mrfelton Thank you!

PR #1109, opened 2026-02-18
feat(mcp): Add local mode support for Codex CLI

### LINKED ISSUE #1103: Add support for local mode mcp for codex
Codex MCP can be configured via `config.toml`, but currently we only support generating global MCP.
See: https://developers.openai.com/codex/mcp

## Feature request
Add support for local mode mcp for codex.

r1bilski:
This implements generating project-level config for mcp for codexcli.

closes #1103

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -239,6 +239,7 @@ docs/.vitepress/cache
 **/.codex/memories/
 **/.codex/skills/
 **/.codex/agents/
+**/.codex/config.toml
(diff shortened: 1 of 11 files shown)


r1bilski:
Whoops, I forgot to update the docs :)
I resolved 1 and 2.

3. I'd say is good as is. We're not using the options so I think it's slightly clearer to ignore the whole parameter instead of destructuring.

PR #1140, opened 2026-02-20
fix(ignore): write Claude deny rules to settings.json for team versioning

### LINKED ISSUE #1094: `.aiignore` get output to `.claude/settings.local.json` instead of `.claude/settings.json`
@dyoshikawa is this intended? My guess would be that is should go to the `.claude/settings.json` by default since that is version controlled by default and applied to other users of the same project?

CillianMyles:
## Summary
- write claude ignore output to `.claude/settings.json` instead of `.claude/settings.local.json` so deny rules can be version-controlled and shared across the team
- update unit/integration/e2e tests for the new path

## Breaking Change
users upgrading from a previous version may have stale deny rules remaining in `.claude/settings.local.json`. claude code loads both files, so rules will be duplicated until the old file is manually cleaned up. I decided not to do any migration since doing so could expose people's personal, and non-version controlled, claude settings.

## Testing
- [x] pnpm cicheck
- [x] pnpm build

Fixes #1094

### DIFF
--- a/.serena/project.yml
+++ b/.serena/project.yml
@@ -14,7 +14,7 @@
 # The first language is the default language and the respective language server will be used as a fallback.
 # Note that when using the JetBrains backend, language servers are not used and this list is correspondingly ignored.
 languages:
-  - typescript
+- typescript
 
 # the encoding used by text files in the project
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@CillianMyles Thank you!

Issue #832, opened 2026-01-18
Support any AI agent's permissions


cm-dyoshikawa:

r1bilski:
Hey, this is the last thing that is missing for me to be able to migrate our configs to rulesync.
Would you mind if I took this on? This obviously requires some design to handle discrepancies between how different agents handle permissions and how to handle permissions for common tools like webfetch, edit, cli, etc. which are often named differently in different agents.
I can either come up with a general design or even publish a draft PR, just wanted to confirm if you're fine with new feature contributions.

cm-dyoshikawa:
@r1bilski Of course yes! Thank you for your usual work.

CillianMyles:
FWIW I am a Claude Code and Codex user...

Claude has more features in this space, while Codex is still yet to add ignore/deny/permissions ability.

Working in a large company, it basically rules a tool out entirely if we can't be sure it won't read sensitive info like secrets etc.

I worked on a spike for how it would be possible to support Claude's `Read`, `Write`, `Edit` permissions: https://github.com/dyoshikawa/rulesync/pull/1141

Admittedly I didn't think about other tools (because Codex doesn't support this yet), but I think it does make sense to see if the industry lands on a some sort of standard for this stuff, although that might take a while.

Happy to discuss further or work on more spikes/implementations later if it helps. 

What do others want/need from this? What tools do you use? What are the outcomes you need? Be specific :-)

r1bilski:
(FYI: I somewhat started working on this some time ago, but got busy. If you want I can push what I have right now for reference.)

I think opencode's format is the most flexible one to be used as the canonical format - it's also a little bit easier to parse than the `Tool(matcher)` format.
IMO for sure the format shouldn't assume the possible keys (e.g. opencode allows setting permissions for any tool, including mcp ones), and should support three possible values: allow, deny, ask. Some tools support only 'deny', so we'd need to have something in the likes of supportedEvents or supportedTypes for hooks.

We'd also need to support translating common tool names, like edit, read, websearch between different tools. 

I asked claude to generate a table of tools that support permissions, to see what would make the most sense:

<details>
<summary>List of Agents supporting permissions</summary>

| Agent | Docs Link | Per-Tool Config Permissions | Supported Levels | Format |
|---|---|---|---|---|
[...]

PR #1141, opened 2026-02-20
[SPIKE] feat(ignore): add ignore.yaml source with action rules

CillianMyles:
## Summary
- add .rulesync/ignore.yaml as the primary rulesync ignore source with per-rule actions (read, write, edit)
- add a parser/converter layer for plain text + YAML ignore rules and project unsupported action semantics safely for non-Claude tools
- update ignore processing, init scaffolding, fetch mapping, docs, and tests to support YAML-first with legacy fallback

## Notes
- keeps backward compatibility by still reading legacy .rulesync/.aiignore / .rulesyncignore when YAML is absent
- includes warnings when non-read actions are projected into tools that only support path-style ignore rules

## Testing
- [x] pnpm cicheck
- [x] pnpm build

Spike exploring new feature suggestion in --> #1094


dyoshikawa:
@CillianMyles Thanks a lot for the thoughtful PR — I really appreciate the effort here 🙏

After reviewing it, I think this is a bit too feature-rich for an **ignore file** scope.
For now, I’d like to keep ignore files simple/minimal, so I’m going to close this PR.

That said, I do think your idea has value. It seems like a better fit for the broader design discussion in https://github.com/dyoshikawa/rulesync/issues/832

If you’re open to it, I’d love to continue the conversation there and shape this into that larger proposal.

Thanks again for the contribution!

CillianMyles:
@dyoshikawa no worries, this was mostly an experiment - added thoughts to the issue!

PR #1191, opened 2026-02-25
feat: implement hooks for copilot

### LINKED ISSUE #1006: [Feature request]: Github Copilot Hooks
Github Copilot has support for hooks as well: https://docs.github.com/en/copilot/concepts/agents/coding-agent/about-hooks

Two things to consider are:
- Copilot has no support for "Prompt" type hooks (I already refactored the code to handle that case in #1001)
- Copilot allows for declaring powershell or bash scripts for each hook. We'd need to make a decision on how to handle that.

Some of the refactorings in #1001 like removing the assumption that the only two implementations are Claude and Cursor would also be useful to have in for this.

@dyoshikawa  I'd be happy to implement this if you provide your input on the two considerations above, and whether we want to tackle this before #1001 is merged (causing conflicts) or after.

r1bilski:
This:
- adds hooks feature for copilot

Notes:
Copilot has an unique hooks format, where it allows for defining the command through `bash` or `powershell` fields. Depending on the platform you're on, either `powershell` or `bash` are required. There are different ways in which this mismatch could be resolved. This chooses to read the user's platform and generate bash or powershell accordingly. If the user wants to maintain both, then they'll need to use copilot-specific overrides.

Closes #1006

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ See [Quick Start guide](https://dyoshikawa.github.io/rulesync/getting-started/qu
 | Codex CLI          | codexcli     | ✅ 🌏 |        | ✅ 🌏 🔧 |    🌏    |    ✅     | ✅ 🌏  |       |
 | Gemini CLI         | geminicli    | ✅ 🌏 |   ✅   |  ✅ 🌏   |  ✅ 🌏   |    🎮     | ✅ 🌏  |       |
 | Goose              | goose        | ✅ 🌏 |        |          |          |           |        |       |
-| GitHub Copilot     | copilot      | ✅ 🌏 |        |    ✅    |    ✅    |    ✅     |   ✅   |       |
+| GitHub Copilot     | copilot      | ✅ 🌏 |        |    ✅    |    ✅    |    ✅     |   ✅   |  ✅   |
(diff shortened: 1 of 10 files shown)


r1bilski:
@dyoshikawa This one is a little bit more complicated due to copilot's unusual format.

I was considering different approaches, but in the end landed on the one implemented, as described in the commit message.

One thing worth noting is that I decided to pass through unknown fields when generating, instead of skipping them, to allow users to define Copilot's `bash` and `powershell` fields in `config.copilot.hooks`.

It seems that currently handling of translation between rulesync's format and agent's format is inconsistent: depending on the feature or agent, we're either passing these fields through or ignoring them when exporting. When importing we're mostly ignoring unknown fields, except for rules, where we put them under agent specific configuration - this approach is not possible to achieve with other features because of how overrides work there.

I think it'd be worthwhile to consolidate these approaches, but that'd probably require rethinking how we do agent-specific overrides for all agents. I do have some ideas, so I might write up an issue if you want, but at the same time, changing this now would be a big breaking change.

Let me know if you find anything in this PR or even would like to approach Copilot's format differently. Thanks!

dyoshikawa:
Thank you!

cm-dyoshikawa:
@r1bilski Thank you!

PR #1202, opened 2026-02-27
fix(copilot hooks): Filter out unsupported schema keys

r1bilski:
Copilot passes through unknown keys to allow user's to include copilot-specific keys in their configs.
We were filtering out unsuported Rulesync keys through a hardcoded array and missed `loop_limit`.

This updates the implementation to use the schema's shape to avoid the issue of new keys being added and not filtered out.

This fixes #1191


r1bilski:
@dyoshikawa This is a fix to something I missed in the previous PR.

r1bilski:
I moved the line out of the loop as the review suggested

dyoshikawa:
@r1bilski Thank you!

PR #1228, opened 2026-03-01
feat(hooks): Add hooks support for geminicli

### LINKED ISSUE #1197: Feature Request: Support for Gemini CLI hooks
Hello,
I always appreciate rulesync.

I found some new information regarding Gemini CLI, and I'd like to share it as a feature request.

## Feature Request

Gemini CLI has recently introduced support for hooks.
It would be greatly appreciated if rulesync could support generating hooks for Gemini CLI.

I am aware that you are currently working on supporting various other hooks, such as those for GitHub Copilot, as mentioned in the issue below:
https://github.com/dyoshikawa/rulesync/issues/1006

There is no rush at all. It would be wonderful if you could consider implementing this in the future when your current tasks settle down.

## Reference Information

It seems that this feature has been officially enabled by default starting from v0.26.0.

- Documentation: https://geminicli.com/docs/hooks/
- Blog: https://developers.googleblog.com/tailor-gemini-cli-to-your-workflow-with-hooks/
- Discussion: https://github.com/google-gemini/gemini-cli/discussions/17812

r1bilski:
This:
- Adds hooks support for geminicli
- Updates cursor hooks implementation to select explicit fields, since gemini introduces two new canonical fields (`name` and `description`) that are not supported by cursor - this change matches the implementation of all the other agents.

Notes:
I have used the same pattern Claude Code does to inject `$GEMINI_PROJECT_DIR` into the command. This seems to be potentially buggy in cases where commands (not files) are used, e.g. `echo foo` becomes `$GEMINI_PROJECT_DIR/echo foo`. I decided to keep this in to maintain parity with Claude Code, but this probably needs to be fixed in another PR.

closes #1197

### DIFF
--- a/src/features/hooks/cursor-hooks.ts
+++ b/src/features/hooks/cursor-hooks.ts
@@ -75,7 +75,14 @@ export class CursorHooks extends ToolHooks {
     const mappedHooks: HooksConfig["hooks"] = {};
     for (const [eventName, defs] of Object.entries(mergedHooks)) {
       const cursorEventName = CANONICAL_TO_CURSOR_EVENT_NAMES[eventName] ?? eventName;
-      mappedHooks[cursorEventName] = defs;
+      mappedHooks[cursorEventName] = defs.map((def) => ({
+        ...(def.type !== undefined && def.type !== null && { type: def.type }),
+        ...(def.command !== undefined && def.command !== null && { command: def.command }),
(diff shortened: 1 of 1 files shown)


r1bilski:
It looks like the problem I mentioned in the PR (which would now get backported to gemini hooks) got an issue #1229.

See the suggested alternative approaches there. If we decide on any of them, I can update this PR to match.

PR #1300, opened 2026-03-09
fix(sources): drop renamed/deleted skills from lockfile on re-fetch

### LINKED ISSUE #1299: install: renamed/deleted skills persist in lockfile and .curated directory
## Description

When using `rulesync install`, if a skill is renamed or deleted in the upstream source, both the old and new skill entries persist in `rulesync.lock` and the `.rulesync/skills/.curated/` directory after re-running install.

## Root cause

`buildLockUpdate` in `src/lib/sources.ts` unconditionally merges all previously locked skill entries back into the lockfile when they are not in the freshly fetched set. This was intended to preserve entries for skills that were skipped during install (due to local precedence, already-fetched deduplication, etc.), but it also preserves entries for skills that no longer exist in the remote.

When a skill is renamed upstream (e.g. `old-skill` becomes `new-skill`):
1. `rulesync install` fetches `new-skill` from the remote
2. `old-skill` is not in the fetched set, so the merge-back logic copies it from the previous lockfile
3. Both `old-skill` and `new-skill` end up in `rulesync.lock` and `.rulesync/skills/.curated/`

## Expected behavior

Skills that no longer exist in the remote source should be dropped from the lockfile when `rulesync install` is run, so only the current set of remote skills is tracked.

## Reproduction

1. Configure a source in `rulesync.jsonc`
2. Run `rulesync install` to fetch skills
3. Rename or delete a skill in the upstream source
4. Run `rulesync install` again
5. Observe both the old and new skill names in `.rulesync/skills/.curated/` and `rulesync.lock`

wfscot:
Fixes #1299

## Summary

- Added `remoteSkillNames` parameter to `buildLockUpdate` so only skills that still exist in the remote are preserved; renamed/deleted skills are dropped
- Updated both GitHub and git transport call sites

## Test plan

- [x] Added test: "should drop renamed/deleted skills from lockfile when upstream removes them"
- [x] All 4306 existing tests pass
- [x] `pnpm cicheck` passes

### DIFF
--- a/src/lib/sources.test.ts
+++ b/src/lib/sources.test.ts
@@ -953,4 +953,42 @@ describe("resolveAndFetchSources", () => {
     expect(logger.error).toHaveBeenCalledWith(expect.stringContaining("not installed"));
     expect(logger.info).toHaveBeenCalledWith(expect.stringContaining("Hint"));
   });
+
+  it("should drop renamed/deleted skills from lockfile when upstream removes them", async () => {
+    const { readLockFile, writeLockFile } = await import("./sources-lock.js");
+    const { resolveDefaultRef, fetchSkillFiles } = await import("./git-client.js");
+
+    // Lock has "old-skill" from a previous install
+    vi.mocked(readLockFile).mockResolvedValue({
+      lockfileVersion: 1,
+      sources: {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@wfscot Thank you!

PR #1306, opened 2026-03-12
feat(sources): install all directory features from sources

wfscot:
## Summary

Extends `rulesync install` to fetch all directory-based features (rules, commands, subagents) from declared sources, not just skills. Closes the feature request in Discussion #1280.

- Adds `features` field to source entries in `rulesync.jsonc` (`["skills", "rules", "commands", "subagents"]` or `"*"` for all). Defaults to `["skills"]` for backward compatibility.
- Generalizes the fetch loop to handle both subdirectory-based items (skills) and file-based items (rules, commands, subagents) with a single code path per transport.
- Extends the lockfile schema to track integrity hashes for all feature types.
- Adds `fetchDirectoryFiles()` to the git client for multi-path sparse-checkout in a single clone.
- Updates all three processors (rules, commands, subagents) to load remote items from `.remote/` subdirectories with local precedence.
- Updates `.gitignore` generation to include new `.remote/` directories.
- Adds type-safe feature lookup maps and defensive error handling for unknown features.

### Note on remote subdirectory naming

Skills continue to use .curated/ as the remote subdirectory for backwards compatibility, while new features (rules, commands, subagents) use .remote/ as it's a bit more conventional for the usecase. This naming is currently hardcoded. We recommend renaming skills to .remote/ in the next breaking change release to unify the convention. The constant FEATURE_REMOTE_SUBDIR_NAMES in rulesync-paths.ts is the single place to change this.

If you'd rather we used .curated everywhere, we can make that change.

### Not yet supported in install

Single-file features (mcp, hooks, ignore) are already supported by the `fetch` command but are not yet supported by `install`. These features are each represented by a single file (e.g. `mcp.json`, `hooks.json`, `.aiignore`) rather than a directory of items, so they need merge strategies to combine local and remote content. This can be addressed in a follow-up.

## Test plan

- [x] `pnpm cicheck` passes (all 4312 tests, formatting, linting, types, spelling, secrets)
- [x] New multi-feature install tests for both GitHub and git transports
- [x] Backward compatibility: sources without `features` still default to skills-only
- [x] Local precedence: local items override remote items of the same name
- [x] Lockfile tracks per-feature integrity hashes
- [x] Code review and security review completed and findings addressed

Ref: #1280


wfscot:
@dyoshikawa lmk if you need me to split this up. It's going to be messy.

wfscot:
Actually, I'm exploring another option where we cache all features by source. This is probably cleaner and simpler than this approach of extending the "cache by feature" approach.

Please DO NOT MERGE this yet. Thanks!

Issue #1320, opened 2026-03-16
Issues found in PR #1315: --json flag implementation needs follow-up fixes


dyoshikawa:
## Background

PR #1315 added a global `--json` flag for structured JSON output to all CLI commands. A code review and security review were performed before merging. While no security blockers were found, several implementation issues were identified that should be addressed in follow-up work.

### Critical Issues

**1. `fetch` command positional argument is silently lost**

In `src/cli/index.ts`, the `wrapCommand` helper extracts `options` as `args[args.length - 2]`, but for `fetch <source>`, `source` is a positional argument passed as `args[0]`. The wrapper never extracts it, meaning `rulesync fetch owner/repo` will fail at runtime.

**2. `wrapCommand` error handler discards `CLIError.code`**

The catch block calls `formatError(error)` which returns a string, losing the `CLIError`'s `code` property (e.g., `RULESYNC_DIR_NOT_FOUND`). The generic command-level `errorCode` parameter is used instead. This defeats the purpose of structured error codes for programmatic consumption. Additionally, since `formatError` returns a string, `message instanceof Error` is always false, so stack traces are never included in JSON output even with `--verbose`.

**3. Duplicate test cases in `fetch.test.ts`**

Lines for "should handle GitHubClientError with 403 status", "should handle GitHubClientError with 404 status", and "should handle generic errors" are duplicated in the same `describe` block. Copy-paste oversight.

### Moderate Issues

**4. `logger.error()` signature change breaks backward compatibility**

The signature changed from `(message: string, ...args: unknown[])` to `(message: string | Error, code?: string, ...args: unknown[])`. The deprecated singleton `logger` is still exported, so existing callers (e.g., `logGitHubAuthHints` in `github-client.ts`) may have their second argument misinterpreted as an error code.

**5. JSON can potentially be output multiple times**

`error()` calls `outputJson(false, ...)` each time it is invoked. If `logger.error()` is called multiple times during a command, multiple JSON objects are written. There is no "already output" guard to ensure exactly one JSON output per command invocation.

**6. ANNOUNCEMENT hook removed without explanation**

The `postAction` hook for announcements was deleted. If this is intentional, the `ANNOUNCEMENT` constant and its import should also be removed as dead code.

**7. DRY violation: repeated `mockLogger` creation in 7 test files**

Every test file constructs an identical `mockLogger` object with `as unknown as Logger`. A shared `createMockLogger()` test helper would eliminate this duplication.

[...]

PR #1323, opened 2026-03-16
fix: address --json flag follow-up issues (#1320)

dyoshikawa:
## Summary

- **Stack trace in JSON output**: Pass `Error` objects directly to `logger.error()` instead of pre-formatting with `formatError()`, so `--verbose` mode includes stack traces in JSON output
- **Duplicate JSON output guard**: Add `_jsonOutputDone` flag to `Logger` to prevent `outputJson()` from emitting multiple JSON blocks when `error()` is called more than once
- **Inline eslint disable comments**: Replace file-level `no-type-assertion: off` override for `index.ts` with per-line `eslint-disable-next-line` comments at each type assertion
- **Dead code cleanup**: Remove empty `announcements.ts`, duplicate `--json` per-command options, double `configure()` calls, and extract shared `createMockLogger()` to `__test__/mock-logger.ts`
- **CLIError handling**: Propagate `CLIError.code` and `exitCode` correctly, use `Boolean()` for type narrowing, fix fetch positional argument handling

## Test plan

- [x] `pnpm cicheck` passes (lint, typecheck, 4333 tests, cspell, secretlint)
- [ ] Manual: `rulesync generate --json` outputs valid JSON
- [ ] Manual: `rulesync generate --json --verbose` on error includes stack trace
- [ ] Manual: Multiple errors don't produce duplicate JSON output

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -89,7 +89,7 @@ export default defineConfig([
   },
 
   {
-    files: ["src/**/*.test.ts", "src/**/__test__/**/*.ts", "scripts/**/*.test.ts"],
+    files: ["src/**/*.test.ts", "src/test-utils/**/*.ts", "scripts/**/*.test.ts"],
     languageOptions: {
       parser: tseslint.parser,
       parserOptions: {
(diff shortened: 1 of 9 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1328, opened 2026-03-17
refactor: split Logger into interface + ConsoleLogger / JsonLogger

### LINKED ISSUE #1325: Consider splitting Logger into interface + separate implementations (StandardLogger / JsonLogger)
## Background

The current `Logger` class in `src/utils/logger.ts` handles both human-readable output and JSON-structured output internally via a `_jsonMode` flag. This means the class contains branching logic for two fundamentally different output strategies, increasing complexity and making it harder to test each mode in isolation.

This was noted during the review of PR #1323, which further centralized logger configuration into `wrapCommand`.

## Details

Currently, `Logger` uses conditional checks like `if (this._jsonMode)` throughout its methods to switch behavior. This approach:

- Violates the Single Responsibility Principle — one class handles two distinct output formats
- Makes testing harder — mock loggers need to account for both modes
- Increases cognitive load when modifying output behavior for one mode

Additionally, the current `ConsoleLogger` implementation depends on [consola](https://github.com/unjs/consola) for terminal output. This dependency should be removed as part of this refactor, replacing it with direct `console.*` or `process.stdout/stderr` calls. This reduces the dependency footprint and gives full control over output formatting.

## Solution / Next Steps

Refactor into:

1. **`Logger` interface** — defines the contract (`info`, `warn`, `error`, `success`, `outputJson`, etc.)
2. **`ConsoleLogger`** — human-readable terminal output (replaces consola with direct console/process calls)
3. **`JsonLogger`** — structured JSON output to stdout/stderr

The appropriate implementation would be selected at DI time (in `wrapCommand`) based on the `--json` flag.

### Additional scope

- **Remove `consola` dependency** — replace with native Node.js output (`console.*`, `process.stdout.write`, or a lightweight chalk-based approach for colors if needed)

### Things to consider

- The mock logger type safety issue (#1324) would also benefit from this refactor — mocking an interface is straightforward
- The `_jsonOutputDone` guard and `outputJson()` would naturally live only in `JsonLogger`
- `verbose` / `silent` flags could be shared via a base class or handled per-implementation
- Check all consola usages to ensure feature parity (log levels, colors, etc.) when replacing

Related: #1324, PR #1323

dyoshikawa:
## Summary

- Refactor `Logger` class into a `Logger` type (interface) with two implementations: `ConsoleLogger` (human-readable terminal output) and `JsonLogger` (structured JSON output)
- Remove `consola` dependency, replacing with native `console.*` calls
- Update `wrapCommand` to select implementation based on `--json` flag at DI time

## Details

The old `Logger` class handled both human-readable and JSON output via `_jsonMode` branching, violating SRP. This refactor:

1. **`Logger` type** — defines the contract (`info`, `warn`, `error`, `success`, `debug`, `captureData`, `outputJson`, etc.)
2. **`ConsoleLogger`** — human-readable output using native `console.log/warn/error`
3. **`JsonLogger`** — structured JSON output to stdout/stderr
4. **`mock-logger.ts`** — no longer requires `as unknown as Logger` cast (type-safe mock)
5. **Removed `consola`** from dependencies, `cspell.json`

## Test plan

- [ ] All 4344 existing tests pass
- [ ] `pnpm cicheck` (code + content) passes
- [ ] Console output behavior unchanged for non-JSON mode
- [ ] JSON output behavior unchanged for `--json` mode
- [ ] No `consola` imports remain in codebase

Closes #1325

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -6,7 +6,7 @@ import { ALL_FEATURES, RulesyncFeatures } from "../types/features.js";
 import { FetchOptions } from "../types/fetch.js";
 import { CLIError } from "../types/json-output.js";
 import { formatError } from "../utils/error.js";
-import { createConsoleLogger, createJsonLogger, Logger } from "../utils/logger.js";
+import { ConsoleLogger, JsonLogger, Logger } from "../utils/logger.js";
 import { fetchCommand } from "./commands/fetch.js";
 import { generateCommand, GenerateOptions } from "./commands/generate.js";
 import { gitignoreCommand } from "./commands/gitignore.js";
@@ -41,8 +41,8 @@ function wrapCommand(
     const positionalArgs = args.slice(0, -2);
     const globalOpts = command.parent?.opts() ?? {};
     const logger: Logger = globalOpts.json
-      ? createJsonLogger({ command: name, version: getVersion() })
-      : createConsoleLogger();
+      ? new JsonLogger({ command: name, version: getVersion() })
+      : new ConsoleLogger();
     logger.configure({
       verbose: Boolean(globalOpts.verbose) || Boolean(options.verbose),
       silent: Boolean(globalOpts.silent) || Boolean(options.silent),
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1369, opened 2026-03-24
feat: Add initial Rovodev support

nyanshak:


src/features/rules/rules-processor.ts
```diff
@@ -528,6 +541,10 @@ export class RulesProcessor extends FeatureProcessor {
         if (!factory.class.isTargetedByRulesyncRule(rulesyncRule)) {
           return null;
         }
+        // Rovodev supports only the root rule (AGENTS.md); skip non-root rulesync rules
+        if (this.toolTarget === "rovodev" && !rulesyncRule.getFrontmatter().root) {
```
dyoshikawa: This rovodev-specific guard should live inside `RovodevRule.isTargetedByRulesyncRule()` instead — have it return `false` for non-root rules. No other tool adds a tool-specific `if` block directly in the generic processor, and doing so sets a precedent that erodes the encapsulation the factory pattern provides.

src/features/mcp/rovodev-mcp.ts
```diff
@@ -0,0 +1,174 @@
...
+  ToolMcpFromFileParams,
+  ToolMcpFromRulesyncMcpParams,
+  ToolMcpParams,
+  ToolMcpSettablePaths,
+} from "./tool-mcp.js";
+
+/**
+ * Type guard to check if a value is a valid McpServers object
+ */
+function isMcpServers(value: unknown): value is McpServers {
```
dyoshikawa: This `isMcpServers` function is a verbatim copy of the same guard in `cursor-mcp.ts` (and a near-copy in `roo-mcp.ts`). Worth extracting to a shared utility at some point, but at minimum worth noting as tech debt rather than silently adding another copy.

src/features/mcp/rovodev-mcp.ts
```diff
@@ -0,0 +1,174 @@
...
+    return {
+      relativeDirPath: ".rovodev",
+      relativeFilePath: "mcp.json",
+    };
+  }
+
+  static getToolTargetsGlobal(): ToolMcpSettablePaths {
+    return {
+      relativeDirPath: ".rovodev",
+      relativeFilePath: "mcp.json",
```
dyoshikawa: `getToolTargetsGlobal()` appears to be dead code — it is not called by any processor or factory. Other MCP implementations using the factory pattern (Cursor, Roo, etc.) do not define this method either. Consider removing it.

src/features/mcp/rovodev-mcp.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  getJson(): Record<string, unknown> {
+    return this.json;
+  }
+
+  override isDeletable(): boolean {
+    return !this.global;
+  }
+
```
dyoshikawa, line 50: Since Rovodev MCP is global-only (`supportsProject: false`), `this.global` will always be `true` for valid instances, so this always returns `false`. The logic is correct but reads as if project-scoped deletion were possible. A comment clarifying the intent (or just `return false`) would make this less confusing.

.gitignore
```diff
@@ -299,5 +299,8 @@ rulesync.local.jsonc
 **/.rooignore
 **/.roo/mcp.json
 **/.roo/subagents/
+**/.rovodev/
```
dyoshikawa, line 307: `**/.rovodev/` already covers everything under `.rovodev/`, making the next two lines (`**/.rovodev/subagents/` and `**/.rovodev/skills/`) redundant. Either remove the blanket pattern and list specific subdirectories (like the `roo` entries do), or remove the specific ones.

src/cli/commands/gitignore-entries.ts
```diff
@@ -131,6 +131,11 @@ export const GITIGNORE_ENTRY_REGISTRY: ReadonlyArray<GitignoreEntryTag> = [
   { target: "roo", feature: "mcp", entry: "**/.roo/mcp.json" },
   { target: "roo", feature: "subagents", entry: "**/.roo/subagents/" },
 
+  // Rovodev (Atlassian Rovo Dev CLI) - AGENTS.md already covered by agentsmd; MCP is global-only (~/.rovodev/mcp.json)
+  { target: "rovodev", feature: "rules", entry: "**/.rovodev/" },
+  { target: "rovodev", feature: "subagents", entry: "**/.rovodev/subagents/" },
```
dyoshikawa, line 186: Tagging this entry with `feature: "rules"` is semantically misleading — rovodev rules use `AGENTS.md` at repo root (covered by `agentsmd`), not `.rovodev/`. This entry is really for the `.rovodev/` directory structure in general. Consider a more accurate feature tag.

src/features/skills/rovodev-skill.ts
```diff
@@ -0,0 +1,203 @@
...
+    const result = RovodevSkillFrontmatterSchema.parse(this.requireMainFileFrontmatter());
+    return result;
+  }
+
+  getBody(): string {
+    return this.mainFile?.body ?? "";
+  }
+
+  validate(): ValidationResult {
+    if (!this.mainFile) {
```
dyoshikawa, line 88: This `validate()` is missing the name-directory match check that other skill implementations include (e.g. `RooSkill.validate()` verifies `frontmatter.name === this.getDirName()`). Since Rovodev skills use the same `SKILL.md` + directory convention, this check should be here too. Same gap exists in `fromDir()`.

nyanshak:
@dyoshikawa Thanks for your review. I made an attempt to address all the feedback, including the isMcpServers refactoring (latest commit). Happy to drop this if you'd prefer to handle it separately.

nyanshak:
Thanks @JiwaniZakir - I'll have a look when I have a bit more time & update/ try to answer :)

dyoshikawa:
@nyanshak Thank you!

JiwaniZakir:
The content-based comparison approach makes sense for JSON and YAML/Frontmatter alike — normalizing before diffing avoids false positives from formatter-induced whitespace or key ordering differences. For the Rovodev integration, it would be worth ensuring the rule output format aligns with what Rovodev expects natively, rather than relying on a shared intermediate representation that might need reformatting downstream.

PR #1376, opened 2026-03-26
feat(mcp): support GitHub Copilot CLI for MCP sync (#1340)

### LINKED ISSUE #1340: Preserve Copilot CLI MCP fields during sync
## Background
When syncing rules to the Copilot CLI MCP format, we should avoid losing valid configuration fields that users may already rely on. GitHub’s Copilot CLI MCP config allows more fields than the current conversion keeps, so a round trip can unintentionally drop data.

## Details
The current copilotcli MCP conversion rebuilds each server using only `command`, `args`, `env`, and `type`. This can drop other valid Copilot CLI fields. GitHub’s documentation indicates `mcpServers` may include additional fields such as `tools`, and HTTP/SSE servers can include fields like `url` and `headers`. As a result, existing or future supported fields can be lost during sync or round-trip conversion.

## Solution / Next Steps
- Preserve Copilot CLI-specific and unknown fields when reading and writing `mcp-config.json` instead of reconstructing server objects from a narrow subset.
- At minimum, retain `tools`.
- Prefer a keep-and-augment approach where rulesync only adds/removes the required `type` field for compatible conversions while retaining other fields whenever possible.

shanliuling:
## What is the problem?
GitHub Copilot CLI requires a specific MCP configuration format in `.copilot/mcp-config.json`, notably the mandatory `"type": "stdio"` field for each server. The current `rulesync` implementation did not support this target, and a naive conversion could lead to the loss of Copilot-specific fields like `tools`, `url`, or `headers` during synchronization.

## How did I fix it?
- Created a specialized `CopilotcliMcp` class to handle the unique requirements of the Copilot CLI.
- Implemented a **"Keep-and-augment"** approach: instead of reconstructing server objects from a limited schema, we now spread the original object to ensure all existing/unknown fields are preserved.
- Automatically injects `"type": "stdio"` as required by the GitHub documentation.
- Integrated `copilotcli` into the `McpProcessor` for seamless usage via `rulesync mcp sync --target copilotcli`.

## How was this tested?
- **Unit Tests**: 30 new test cases in `copilotcli-mcp.test.ts`.
- **Integration Tests**: 45 tests in `mcp-processor.test.ts`.
- **Total**: 74 tests passed.

Closes #1340

### DIFF
--- a/docs/reference/supported-tools.md
+++ b/docs/reference/supported-tools.md
@@ -10,7 +10,7 @@ Rulesync supports both **generation** and **import** for All of the major AI cod
 | Codex CLI           | codexcli     | ✅ 🌏 |        | ✅ 🌏 🔧 |    🌏    |   ✅ 🌏   | ✅ 🌏  |       |
 | Gemini CLI          | geminicli    | ✅ 🌏 |   ✅   |  ✅ 🌏   |  ✅ 🌏   |    🎮     | ✅ 🌏  | ✅ 🌏 |
 | GitHub Copilot      | copilot      | ✅ 🌏 |        |    ✅    |    ✅    |    ✅     |   ✅   |  ✅   |
-| GitHub Copilot CLI | copilotcli   |       |        |  ✅ 🌏   |          |           |        |       |
+| GitHub Copilot CLI  | copilotcli   |       |        |  ✅ 🌏   |          |           |        |       |
 | Goose               | goose        | ✅ 🌏 |   ✅   |          |          |           |        |       |
 | Cursor              | cursor       |  ✅   |   ✅   |  ✅ 🌏   |  ✅ 🌏   |   ✅ 🌏   | ✅ 🌏  |  ✅   |
 | Factory Droid       | factorydroid | ✅ 🌏 |        |  ✅ 🌏   |    🎮    |    🎮     |   🎮   | ✅ 🌏 |
--- a/skills/rulesync/supported-tools.md
+++ b/skills/rulesync/supported-tools.md
@@ -10,7 +10,7 @@ Rulesync supports both **generation** and **import** for All of the major AI cod
 | Codex CLI           | codexcli     | ✅ 🌏 |        | ✅ 🌏 🔧 |    🌏    |   ✅ 🌏   | ✅ 🌏  |       |
 | Gemini CLI          | geminicli    | ✅ 🌏 |   ✅   |  ✅ 🌏   |  ✅ 🌏   |    🎮     | ✅ 🌏  | ✅ 🌏 |
 | GitHub Copilot      | copilot      | ✅ 🌏 |        |    ✅    |    ✅    |    ✅     |   ✅   |  ✅   |
-| GitHub Copilot CLI | copilotcli   |       |        |  ✅ 🌏   |          |           |        |       |
+| GitHub Copilot CLI  | copilotcli   |       |        |  ✅ 🌏   |          |           |        |       |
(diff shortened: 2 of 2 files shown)


dyoshikawa-claw:
The main issues are:

- `pr-description.md` was added to the repository, so the PR description appears to have been committed by mistake and should be removed.
- `CopilotcliMcpConfig` has been widened to `Record<string, unknown>`, which weakens type safety. It would be better to keep known fields like `type` and `command` explicit while allowing extension where needed.
- `README.md` was updated for `copilotcli`, but `docs/reference/supported-tools.md` and `skills/rulesync/supported-tools.md` were not updated accordingly.
- `docs/reference/file-formats.md` does not document `.copilot/mcp-config.json`, so the docs are still out of sync with the implementation.
- The comment on `addTypeField` is slightly out of sync with the actual normalization behavior.
- The PR description miscounts the tests (30 + 45 = 75).

dyoshikawa (commented):
Overall this is a solid addition — the "keep-and-augment" approach for preserving unknown fields makes sense for Copilot CLI support. Tests are thorough. A few things worth addressing before merge, mainly around type safety and the spread logic in `addTypeField`. Also, `pr-description.md` should be removed from the commit.

pr-description.md
```diff
@@ -0,0 +1,18 @@
+## What is the problem?
```
dyoshikawa: This file duplicates the PR body and shouldn't be committed to the repo. Please remove it before merging.

src/features/mcp/copilotcli-mcp.ts
```diff
@@ -17,12 +17,7 @@ export type CopilotcliMcpParams = ToolMcpParams;
 
 type CopilotcliMcpConfig = {
   mcpServers?: {
-    [key: string]: {
-      type: "stdio";
-      command: string;
-      args?: string[];
-      env?: Record<string, string>;
-    };
+    [key: string]: Record<string, unknown>;
```
dyoshikawa: Changing this to `Record<string, unknown>` loses all compile-time guarantees on the known fields (`type`, `command`, `args`, `env`). You can preserve unknown fields while keeping type safety with an intersection type:

```typescript
[key: string]: {
  type: "stdio";
  command: string;
  args?: string[];
  env?: Record<string, string>;
} & Record<string, unknown>;
```

This way the compiler still catches mistakes on known fields, and unknown fields pass through freely.

src/features/mcp/copilotcli-mcp.ts
```diff
@@ -64,10 +59,10 @@ function addTypeField(mcpServers: McpServers): CopilotcliMcpConfig["mcpServers"]
     }
 
     result[name] = {
+      ...server, // server is already compatible with the target structure
```
dyoshikawa: Two concerns with spreading `server` (the raw input) instead of `parsed` (the Zod-validated output):

1. If `McpServerSchema.parse()` ever normalizes or transforms values, those transformations are silently discarded here since you're spreading the pre-validation object.

2. When the normalized `args` is `undefined` (e.g. the command was a simple string with no extra args, and `parsed.args` is undefined), the conditional spread `...(args && { args })` won't fire, so any stale `args` from the original `server` object leaks through.

Consider using `...parsed` instead of `...server` to stay consistent with the validation pipeline:

```typescript
result[name] = {
  ...parsed,
  type: "stdio",
  command,
  ...(args && { args }),
};
```

dyoshikawa:
@shanliuling Thank you!

Issue #1319, opened 2026-03-15
Generate `.copilot/mcp-config.json` for GitHub Copilot CLI


sergeyd108:
### Summary

When `copilot` is set as a target, rulesync currently generates `.vscode/mcp.json` — which targets **GitHub Copilot in VS Code** (the extension). However, **GitHub Copilot CLI** (`@github/copilot`) uses a completely different config file with a different schema, and does not pick up `.vscode/mcp.json`.

As a result, MCP servers defined in `.rulesync/mcp.json` are silently ignored when using Copilot CLI.

---

### The Problem

Copilot CLI reads MCP config from:
- `~/.copilot/mcp-config.json` (global, per-user)
- `.copilot/mcp-config.json` (project-level, in the repo root)

The schema also differs from VS Code's format:

| | VS Code (`.vscode/mcp.json`) | Copilot CLI (`.copilot/mcp-config.json`) |
|---|---|---|
| Top-level key | `"servers"` | `"mcpServers"` |
| Server type field | not required | `"type": "stdio"` or `"local"` |

**Reference:** [GitHub Docs — Adding MCP servers for GitHub Copilot CLI](https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-mcp-servers)

---

### Expected Behavior

When `copilot` is listed as a target, rulesync should generate **both**:
1. `.vscode/mcp.json` — for Copilot in VS Code (already works)
2. `.copilot/mcp-config.json` — for Copilot CLI

Or alternatively, introduce a way to distinguish between the two targets, e.g. `copilot` vs `copilot-cli`.

---

### Suggested Schema for `.copilot/mcp-config.json`

```json
{
  "mcpServers": {
    "my-server": {
      "type": "stdio",
      "command": "npx",
      "args": ["my-mcp-package"]
    }
  }
}
```

---

### Environment

- rulesync version: 7.18.1
- GitHub Copilot CLI (`@github/copilot`) version: 1.0.5
- OS: Windows 11 Pro

sergeyd108:
Seems like copilot cli doesn't respect per-project `.copilot` directory and mcp file. If there is nothing to do with this, just close the issue.

cm-dyoshikawa:
@sergeyd108 It looks a good suggestion. I hope it is supported.

PR #1386, opened 2026-03-27
feat/copilot cli mcp support

dyoshikawa-claw:
- **feat(mcp): add GitHub Copilot CLI MCP support (#1319)**
- **fix(copilot-cli-mcp): address Copilot review feedback**
- **fix: rename copilot-cli to copilotcli**
- **fix: rename copilot-cli files to copilotcli**
- **fix: import order and unused import**
- **fix: add type keyword to import**
- **fix: fix CI failures and rename copilot-cli to copilotcli**
- **fix: fix eslint unused var and reformat files**
- **test: fix EACCES by using dynamic test directory instead of absolute paths**
- **chore: add copilotcli to cspell dictionary**
- **feat(opencode): honor agentsmd subproject path**

### DIFF
--- a/src/features/rules/opencode-rule.test.ts
+++ b/src/features/rules/opencode-rule.test.ts
@@ -311,6 +311,130 @@ describe("OpenCodeRule", () => {
       expect(opencodeRuleWithValidation.getFileContent()).toContain("# Validation Test");
       expect(opencodeRuleWithoutValidation.getFileContent()).toContain("# Validation Test");
     });
+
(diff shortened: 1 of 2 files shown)


dyoshikawa-claw:
Self-review (OpenCode): the change looks sound overall. One caveat: when  is set,  now emits  in the subproject directory rather than a file under . That matches the current shared helper behavior and Codex CLI parity, but it may be surprising on round-trip. If intentional, it would be worth calling out in docs/tests; otherwise we may want to preserve the non-root filename/path semantics for OpenCode specifically.

PR #1408, opened 2026-03-31
fix(mcp): tighten copilotcli stdio handling (#1401)

### LINKED ISSUE #1401: Review findings for PR #1376: Copilot CLI MCP support
## Background

PR #1376 adds GitHub Copilot CLI as a new MCP sync target (`copilotcli`), generating `.copilot/mcp-config.json`. A code review and security review were conducted, revealing several findings ranging from mid to low severity. This issue consolidates those findings for tracking and follow-up.

### Code Review Findings

**#1 (mid) — Double type assertion and redundant spread in `addTypeField`**
In `copilotcli-mcp.ts`, `addTypeField` spreads both `server` (raw input) and `parsed` (Zod-validated output). This is redundant since `McpServerSchema` uses `z.looseObject()` and `parsed` already preserves unknown fields. The double `eslint-disable` and `as` casts weaken type safety unnecessarily.

**#2 (mid) — `parsed.type ?? "stdio"` contradicts function contract**
The function name and JSDoc state it adds `"type": "stdio"`, but the new code preserves whatever `type` the input already has. The original code unconditionally set `type: "stdio"`. This is a behavioral regression — if Copilot CLI only supports stdio, the original hard-coded approach was correct. If other types are supported, the `command` validation should be conditional on the type.

**#3 (mid) — Missing `rules-processor.ts` integration check**
Per `feature-change-guidelines.md`, new tool targets should be checked against `rules-processor.ts` and `additionalConvention`. The PR diff does not show changes to `rules-processor.ts`.

**#4 (low) — Dead code in `getSettablePaths`**
Both the `global` and non-global branches return identical values (`{ relativeDirPath: ".copilot", relativeFilePath: "mcp-config.json" }`). The `if (global)` branch is dead code.

**#5 (low) — `removeTypeField` unconditionally strips `type`**
If non-stdio types are now preserved by `addTypeField` (per #2), `removeTypeField` would lose that information when converting back to rulesync format.

**#6 (low) — Documentation section placement in `file-formats.md`**
The new `.copilot/mcp-config.json` section is inserted before the general "optional override keys" paragraph about hooks, which may cause confusion.

**#7 (low) — PR description test counts are misleading**
The counts (30 + 45 = 74) refer to total tests in each file, not newly added tests.

**#8 (low) — Missing negative test for command-less servers with unknown fields**
No test verifies that a server with unknown fields but missing `command` still throws an error correctly.

### Security Review Findings

**#9 (low) — Theoretical prototype pollution risk in object spread**
Spreading raw `serverRecord` before Zod-validated `parsed` has a theoretical prototype pollution concern. In practice, JS object literal spread does not trigger this, and `parsed` overrides critical fields. Actual risk is minimal.

**#10 (low) — Type assertion bypasses defense-in-depth**
The `as` casts suppress TypeScript type checking in the data transformation code, weakening a defense-in-depth layer.

**#11 (low) — No sanitization of `command`/`args` fields**
Expected behavior for a config file generator operating on user-controlled local files. Consistent with all other tool targets in the codebase.

**No malicious code, backdoors, obfuscated payloads, or suspicious dependency additions were detected.**

## Solution / Next Steps

1. **Address #2 first (highest impact):** Decide whether Copilot CLI supports only stdio or multiple transport types. If stdio-only, revert to hard-coded `type: "stdio"`. If multiple types, make `command` validation conditional on `type`.
2. **Simplify #1:** Spread only `parsed` (not raw `server`) to eliminate redundant type assertions.
3. **Verify #3:** Check `rules-processor.ts` and `additionalConvention` for the new `copilotcli` target.
4. **Clean up #4:** Remove the dead `if (global)` branch or add a TODO explaining the placeholder.
5. **Add test for #8:** Add a negative test case for servers with unknown fields but no `command`.
6. **Fix #6:** Adjust section placement in `file-formats.md` to avoid confusion with the hooks paragraph.

dyoshikawa-claw:
## What is the problem?

Review findings from PR #1376 (tracked in #1401):
- `addTypeField` spread raw `server` instead of parsed, with redundant type assertions
- `type` was conditional (`parsed.type ?? "stdio"`) but Copilot CLI only supports stdio
- `removeTypeField` stripped type unconditionally
- Dead `if(global)` branch in `getSettablePaths`
- Missing `copilotcli` target in rules-processor
- No negative test for command-less servers with unknown fields

## How did I fix it?

- Simplify `addTypeField`: spread `parsed` only, hard-code `type: "stdio"`, remove eslint-disable
- `removeTypeField`: only strip type when it is `"stdio"`
- Remove dead `if(global)` branch in `getSettablePaths`
- Add `copilotcli` to `rulesProcessorToolTargets` and `toolRuleFactories`
- Add `CopilotcliRule` class extending `CopilotRule`
- Add tests for forced stdio, non-stdio preservation, and command-less negative case

## How was this tested?

- All 4625 tests pass, `pnpm cicheck` clean

Closes #1401

### DIFF
--- a/src/features/mcp/copilotcli-mcp.test.ts
+++ b/src/features/mcp/copilotcli-mcp.test.ts
@@ -391,6 +391,60 @@ describe("CopilotcliMcp", () => {
         },
       });
     });
+
+    it("should force stdio type even when another type is provided", async () => {
+      const inputMcpServers = {
+        "typed-server": {
+          type: "http" as const,
+          command: "node",
+          args: ["server.js"],
+          url: "http://localhost:3000/mcp",
+        },
+      };
+      const rulesyncMcp = new RulesyncMcp({
+        relativeDirPath: RULESYNC_RELATIVE_DIR_PATH,
+        relativeFilePath: "mcp.json",
+        fileContent: JSON.stringify({ mcpServers: inputMcpServers }),
+      });
+
+      const copilotCliMcp = await CopilotcliMcp.fromRulesyncMcp({
+        rulesyncMcp,
+      });
+
+      expect(copilotCliMcp.getJson()).toEqual({
+        mcpServers: {
+          "typed-server": {
+            type: "stdio",
+            command: "node",
+            args: ["server.js"],
+            url: "http://localhost:3000/mcp",
+          },
+        },
+      });
+    });
+
+    it("should throw error when server has unknown fields but no command", async () => {
+      const inputMcpServers = {
+        "unknown-fields-no-command": {
+          url: "http://localhost:3000/mcp",
+          headers: {
+            Authorization: "Bearer test-token",
+          },
+          unknown_field: "value",
+        },
+      };
+      const rulesyncMcp = new RulesyncMcp({
+        relativeDirPath: RULESYNC_RELATIVE_DIR_PATH,
+        relativeFilePath: "mcp.json",
+        fileContent: JSON.stringify({ mcpServers: inputMcpServers }),
+      });
+
+      await expect(
+        CopilotcliMcp.fromRulesyncMcp({
+          rulesyncMcp,
+        }),
+      ).rejects.toThrow('MCP server "unknown-fields-no-command" is missing a command');
+    });
   });
 
   describe("toRulesyncMcp", () => {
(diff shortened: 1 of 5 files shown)


dyoshikawa-claw:
I found three issues in PR #1408.

#1 [high] `src/features/rules/copilotcli-rule.ts:23`
`CopilotcliRule` delegates to `CopilotRule.fromRulesyncRule()` and does not override `getSettablePaths()`, so `copilotcli` rules use the same `.github/copilot-instructions.md` and `.github/instructions/**` paths as `copilot`. A repo that defines different rules for the two targets can have one generation or deletion run overwrite the other target files.

#2 [mid] `src/features/mcp/copilotcli-mcp.ts:59`
`addTypeField()` now forces every server to `type: "stdio"`, even when the Rulesync source intentionally declares `http` or `sse`. On a shared `mcp.json`, that either raises the new missing-command error or emits an invalid stdio config for a non-stdio server. This should reject unsupported transports explicitly instead of coercing them.

#3 [low] `src/features/rules/rules-processor.ts:296`
This enables `copilotcli` rules, but the public support matrix still says `copilotcli` only supports MCP in `README.md` and `docs/reference/supported-tools.md`. Please update the docs in the same change so users are not misled.

PR #1409, opened 2026-03-31
fix(geminicli): generate subagents to .gemini/agents/ as native (not simulated)

lgaudez:
## Summary

- Fix Gemini CLI subagents output path: `.gemini/subagents/` → `.gemini/agents/` (where Gemini CLI actually expects them)
- Make `GeminiCliSubagent` a **native** subagent (extends `ToolSubagent` directly) instead of simulated — no `--simulate-subagents` flag needed
- Remove automatic injection of `experimental.enableAgents` into `.gemini/settings.json` — users should set this manually
- Update docs note (dated 2026-04-01) to inform users to set `"experimental": { "enableAgents": true }` in `.gemini/settings.json` manually
- Forward `global` param in `fromRulesyncSubagent` and `fromFile` for consistency with other native subagents
- Update README/docs: geminicli subagents column `🎮` → `✅`

## Unrelated change included

`.lintstagedrc.js`: changed `tsx scripts/sync-skill-docs.ts` → `node --import tsx/esm scripts/sync-skill-docs.ts`. This was needed because the `tsx` CLI creates a Unix IPC socket at `/tmp/tsx-*/...pipe` which was blocked by the sandbox during development. Included here since it was required to run the pre-commit hook locally.

## Test plan

- [ ] Unit tests for `GeminiCliSubagent` updated to reflect native behavior (`toRulesyncSubagent()` now works)
- [ ] E2E test: `geminicli` subagents generate to `.gemini/agents/planner.md` without `--simulate-subagents`
- [ ] `pnpm cicheck` passes (174 test files, 4603 tests)

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -149,7 +149,7 @@ Based on the user's instruction, create a plan while analyzing the related files
 Attention, again, you are just the planner, so though you can read any files and run any commands for analysis, please don't write any code.
 ```
 
-> **Gemini CLI note:** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
+> **Gemini CLI note (as of 2026-04-01):** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
 
 ## `.rulesync/skills/*/SKILL.md`
 
--- a/skills/rulesync/file-formats.md
+++ b/skills/rulesync/file-formats.md
@@ -149,7 +149,7 @@ Based on the user's instruction, create a plan while analyzing the related files
 Attention, again, you are just the planner, so though you can read any files and run any commands for analysis, please don't write any code.
 ```
 
-> **Gemini CLI note:** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
+> **Gemini CLI note (as of 2026-04-01):** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
 
 ## `.rulesync/skills/*/SKILL.md`
 
--- a/src/features/subagents/geminicli-subagent.ts
+++ b/src/features/subagents/geminicli-subagent.ts
@@ -91,6 +91,7 @@ export class GeminiCliSubagent extends ToolSubagent {
     baseDir = process.cwd(),
     rulesyncSubagent,
     validate = true,
+    global = false,
   }: ToolSubagentFromRulesyncSubagentParams): ToolSubagent {
     const rulesyncFrontmatter = rulesyncSubagent.getFrontmatter();
     const geminicliSection = rulesyncFrontmatter.geminicli ?? {};
@@ -105,7 +106,7 @@ export class GeminiCliSubagent extends ToolSubagent {
     const fileContent = stringifyFrontmatter(body, geminicliSubagentFrontmatter, {
       avoidBlockScalars: true,
     });
-    const paths = this.getSettablePaths();
+    const paths = this.getSettablePaths({ global });
 
     return new GeminiCliSubagent({
       baseDir,
@@ -115,6 +116,7 @@ export class GeminiCliSubagent extends ToolSubagent {
       relativeFilePath: rulesyncSubagent.getRelativeFilePath(),
       fileContent,
       validate,
+      global,
     });
   }
 
@@ -147,8 +149,9 @@ export class GeminiCliSubagent extends ToolSubagent {
     baseDir = process.cwd(),
     relativeFilePath,
     validate = true,
+    global = false,
   }: ToolSubagentFromFileParams): Promise<GeminiCliSubagent> {
-    const paths = this.getSettablePaths();
+    const paths = this.getSettablePaths({ global });
     const filePath = join(baseDir, paths.relativeDirPath, relativeFilePath);
     const fileContent = await readFileContent(filePath);
     const { frontmatter, body: content } = parseFrontmatter(fileContent, filePath);
@@ -166,6 +169,7 @@ export class GeminiCliSubagent extends ToolSubagent {
       body: content.trim(),
       fileContent,
       validate,
+      global,
     });
   }
 


docs/reference/file-formats.md
```diff
@@ -172,6 +149,8 @@ Based on the user's instruction, create a plan while analyzing the related files
 Attention, again, you are just the planner, so though you can read any files and run any commands for analysis, please don't write any code.
 ```
 
+> **Gemini CLI note:** Subagents are generated to `.gemini/agents/`. Additionally, rulesync automatically injects `"experimental": { "enableAgents": true }` into `.gemini/settings.json` when generating Gemini CLI subagents, which is required for the agents feature to work.
```
lgaudez: This I don't want it here, I just want to mention it for gemini user. I don't want rulesync to automatically enableAgents or modify the settings gemini file.

skills/rulesync/file-formats.md
```diff
@@ -172,6 +149,8 @@ Based on the user's instruction, create a plan while analyzing the related files
 Attention, again, you are just the planner, so though you can read any files and run any commands for analysis, please don't write any code.
 ```
 
+> **Gemini CLI note:** Subagents are generated to `.gemini/agents/`. Additionally, rulesync automatically injects `"experimental": { "enableAgents": true }` into `.gemini/settings.json` when generating Gemini CLI subagents, which is required for the agents feature to work.
```
lgaudez: Do not modify this file, revert

src/features/subagents/geminicli-subagent.ts
```diff
@@ -10,21 +16,124 @@ import {
...
+      relativeDirPath: RULESYNC_SUBAGENTS_RELATIVE_DIR_PATH,
+      relativeFilePath: this.getRelativeFilePath(),
+      validate: true,
+    });
+  }
+
+  static fromRulesyncSubagent({
+    baseDir = process.cwd(),
+    rulesyncSubagent,
+    validate = true,
```
dyoshikawa, line 93: Missing `global = false` destructuring here. All other native subagents (`ClaudecodeSubagent`, `CursorSubagent`, `CopilotSubagent`) destructure `global` from `ToolSubagentFromRulesyncSubagentParams` and forward it to `getSettablePaths({ global })` and the constructor. Without it, if global support is ever added for geminicli, this method will silently ignore the flag.

```suggestion
  static fromRulesyncSubagent({
    baseDir = process.cwd(),
    rulesyncSubagent,
    validate = true,
    global = false,
  }: ToolSubagentFromRulesyncSubagentParams): ToolSubagent {
```

src/features/subagents/geminicli-subagent.ts
```diff
@@ -34,7 +143,45 @@ export class GeminiCliSubagent extends SimulatedSubagent {
     });
   }
 
-  static forDeletion(params: ToolSubagentForDeletionParams): GeminiCliSubagent {
-    return new GeminiCliSubagent(this.forDeletionDefault(params));
+  static async fromFile({
+    baseDir = process.cwd(),
+    relativeFilePath,
+    validate = true,
```
dyoshikawa, line 151: Same issue as `fromRulesyncSubagent` — `global` is not destructured or forwarded to `getSettablePaths()`. Other native subagents pass `this.getSettablePaths({ global })` here.

```suggestion
  static async fromFile({
    baseDir = process.cwd(),
    relativeFilePath,
    validate = true,
    global = false,
  }: ToolSubagentFromFileParams): Promise<GeminiCliSubagent> {
    const paths = this.getSettablePaths({ global });
```

.lintstagedrc.js
```diff
@@ -1,7 +1,7 @@
 export default {
   "*": ["npx secretlint"],
   "package.json": ["npx sort-package-json"],
-  "docs/**/*.md": ["tsx scripts/sync-skill-docs.ts", "git add skills/rulesync/"],
+  "docs/**/*.md": ["node --import tsx/esm scripts/sync-skill-docs.ts", "git add skills/rulesync/"],
```
dyoshikawa, line 4: This change from `tsx` to `node --import tsx/esm` looks unrelated to the geminicli subagents fix. Consider splitting it into its own commit or at least noting it in the PR description so it does not get lost in the diff.

docs/reference/file-formats.md
```diff
@@ -172,6 +149,8 @@ Based on the user's instruction, create a plan while analyzing the related files
 Attention, again, you are just the planner, so though you can read any files and run any commands for analysis, please don't write any code.
 ```
 
+> **Gemini CLI note:** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
```
lgaudez: Put the date of today for this declaration because this may evolve in the futur

lgaudez:
Regarding the `.lintstagedrc.js` change: this was introduced because the `tsx` CLI binary creates a Unix IPC socket at `/tmp/tsx-*/...pipe` during execution. In my local environment (Claude Code's sandbox), access to `/tmp/` is restricted, which caused the pre-commit hook to fail on every commit attempt.

Switching to `node --import tsx/esm` avoids the socket entirely and unblocked the hook for me. That said, this is specific to my sandbox setup — `tsx` works fine in a normal terminal and the change has no effect for other contributors.

Happy to revert it if you'd prefer to keep the diff clean. Alternatively, `node --import tsx/esm` is slightly more explicit and avoids spawning an extra subprocess, so it could be kept as a minor improvement. Your call.

src/features/subagents/geminicli-subagent.ts
lgaudez: Done

src/features/subagents/geminicli-subagent.ts
lgaudez: Done

.lintstagedrc.js
lgaudez: Regarding the .lintstagedrc.js change: this was introduced because the tsx CLI binary creates a Unix IPC socket at /tmp/tsx-*/...pipe during execution. In my local environment (Claude Code's sandbox), access to /tmp/ is restricted, which caused the pre-commit hook to fail on every commit attempt.

Switching to node --import tsx/esm avoids the socket entirely and unblocked the hook for me. That said, this is specific to my sandbox setup — tsx works fine in a normal terminal and the change has no effect for other contributors.

Happy to revert it if you'd prefer to keep the diff clean. Alternatively, node --import tsx/esm is slightly more explicit and avoids spawning an extra subprocess, so it could be kept as a minor improvement. Your call.

dyoshikawa:
@lgaudez Thank you!

PR #1452, opened 2026-04-08
feat(claudecode): add ignore fileMode option via per-feature config

### LINKED ISSUE #1374: Clarify Claude ignore output vs gitignore behaviour
After upgrading a repo from `7.12.1` to `7.22.0`, I tripped on the current Claude ignore flow and had to read the compiled output to understand what was happening.

Current behaviour:
- the ignore feature writes Claude deny rules to `.claude/settings.json`
- `rulesync gitignore` emits `**/.claude/settings.local.json`, but not `**/.claude/settings.json`

I do not think this is necessarily a bug.

From reading prior issues, it looks like there is real design history here:
- #1094 moved Claude ignore output from `settings.local.json` to `settings.json` for team versioning
- #13 makes it clear `settings.json` is treated carefully because it can contain real user config

So I think the actual problem is discoverability.

If someone enables the ignore feature and then runs `rulesync gitignore`, the generated file that shows up in the repo is not the one the gitignore output leads you to expect. It is easy to assume this is accidental repo noise, even if the current behaviour is intentional.

I think one of these would help:
1. Document explicitly that Claude ignore writes to shared `.claude/settings.json` by design.
2. Add an explicit mode or flag so users can choose between shared `settings.json` and local `settings.local.json` output.

I do not think automatically adding `**/.claude/settings.json` to the generated gitignore is obviously right, since that could hide intentional shared Claude config.

Main ask: make the intended behaviour more obvious in the docs or CLI output so users do not have to inspect the implementation to understand it.

Happy to put up a PR if useful.

dyoshikawa:
## Summary

Closes #1374.

The Claude Code ignore feature currently always writes deny rules to the shared `.claude/settings.json`. That decision is intentional (see #1094), but it tripped up users who expected `rulesync gitignore` to handle the generated file. This PR keeps the shared default but adds a way to opt out.

- Extend `features` to support an object form per target where each feature value may be `true`, `false`, or an options object (`{ ignore: { fileMode: "local" } }`). The existing array form keeps working.
- Add `Config#getFeatureOptions(target, feature)` and forward per-feature options through `IgnoreProcessor` to `ClaudecodeIgnore`.
- `ClaudecodeIgnore` now resolves its output file from `fileMode`: `"shared"` (default → `.claude/settings.json`) or `"local"` (→ `.claude/settings.local.json`, which `rulesync gitignore` already excludes).
- Document the rationale and the new option in `docs/reference/file-formats.md` and `docs/guide/configuration.md` so users do not need to read the implementation to understand the behaviour. (`skills/rulesync/*` were auto-synced.)

Example:

```jsonc
{
  "targets": ["claudecode"],
  "features": {
    "claudecode": {
      "ignore": { "fileMode": "local" }
    }
  }
}
```

## Test plan

- [x] `pnpm cicheck` (code + content)
- [x] New unit tests cover `Config#getFeatureOptions`, the per-feature object form (truthy/false/wildcard), and `ClaudecodeIgnore` writing to `settings.local.json` when `fileMode: "local"` is supplied.


dyoshikawa:
@dyoshikawa Thank you!

PR #1478, opened 2026-04-14
feat(config): support object form on root.targets; deprecate object form on root.features

dyoshikawa:
## Summary
- Extend `root.targets` to accept the same object form that `root.features` accepts today, so users can declare the target set and its per-tool feature configuration in a single place: `targets: { claudecode: { rules: true, ignore: { fileMode: "local" } } }`.
- Deprecate the object form under `root.features` (still parsed; emits a one-shot warning). Motivation: combining `targets: ["claudecode"]` with `features: { claudecode: {...} }` double-defined the target set.
- Enforce schema-level mutual exclusivity: `targets` object + any `features` now throws at config load, and vice versa. The only valid combinations are array × array, `targets` object alone, or (legacy) `features` object alone.

## Implementation notes
- New `RulesyncConfigTargetsSchema` (union of array | object) is used for `root.targets` inside `rulesync.jsonc`. The original `RulesyncTargetsSchema` (array-only) is preserved for frontmatter parsers (rules / mcp / commands / subagents / skills).
- `assertTargetsFeaturesExclusive` is called by `ConfigResolver` after file load, before defaults are merged in, so defaults cannot trigger a false positive.
- `Config.getTargets` / `getFeatures` / `getFeatureOptions` / `hasPerTargetFeatures` now consult the `targets` object before falling back to the array / legacy `features` paths.
- Docs and JSON schema regenerated.

## Breaking change
The previously-valid combo `targets: [...] + features: { <object> }` now throws at config load. Migration: move the per-target config under `targets` and remove `features`.

## Test plan
- [x] `pnpm cicheck` (format, oxlint, eslint, typecheck, 4876 tests, cspell, secretlint)
- [x] New unit coverage for object-form `targets`: key extraction, per-target feature resolution, per-feature options, wildcard, conflict detection on object keys
- [x] New unit coverage for `assertTargetsFeaturesExclusive` (all valid/invalid combinations)
- [x] Existing object-form `features` tests preserved — behavior unchanged


dyoshikawa:
@dyoshikawa Thank you!

PR #1488, opened 2026-04-15
Always emit agentsmd gitignore entries when gitignoreTargetsOnly is true

dyoshikawa:
## Summary

- When `gitignoreTargetsOnly` is `true` (the default), `rulesync gitignore` now always appends `agentsmd` to the resolved target list even when it is not listed in the config's `targets`. `AGENTS.md` is a de facto standard file read by many AI coding tools regardless of the selected target set, so its `.gitignore` entries must always be emitted to prevent accidental commits of generated rule files.
- Added a note to `docs/reference/cli-commands.md` and `docs/guide/configuration.md` (and the synchronized `skills/rulesync/` copies) explaining the unconditional inclusion and the opt-out via an explicit `--targets` flag.
- Fix `.lintstagedrc.js` to invoke `scripts/sync-skill-docs.ts` through the project's custom `strip-types-loader.mjs` instead of the removed `tsx` dependency, so the docs/skills sync in the pre-commit hook works again.

## Test plan

- [x] `pnpm cicheck` passes locally.
- [x] `pnpm dev gitignore` regenerates the project `.gitignore` without errors (AGENTS.md entries are now present even though `agentsmd` is not in the project's `targets`).
- [x] New unit tests cover the appended-agentsmd case, the already-listed case, and the wildcard case in `src/cli/commands/resolve-gitignore-targets.test.ts`.


dyoshikawa:
@dyoshikawa Thank you!

PR #1494, opened 2026-04-15
fix(ci): handle lightweight tags and verify remote SHA in publish-assets

### LINKED ISSUE #1485: publish-assets.yml: lightweight tag detection fails in remote tag check
## Background

In PR #1481, the `Create and push tag` step of `.github/workflows/publish-assets.yml` was hardened to detect whether the target tag already exists locally or remotely, and to verify that any pre-existing tag points to the current commit. This improves safety against publishing releases against an incorrect commit.

However, the remote-side detection uses `git ls-remote --tags origin "refs/tags/$TAG^{}"`, which only resolves to the dereferenced (peeled) object of an **annotated** tag. For a **lightweight** tag, the `^{}` form returns no output, so `REMOTE_SHA` ends up empty.

## Details

Current code:

```sh
REMOTE_SHA="$(git ls-remote --tags origin "refs/tags/$TAG^{}" | awk '{print $1}' | head -n 1)"
```

Behavior:

- Annotated tag on remote: `^{}` ref exists → SHA is returned → check works.
- Lightweight tag on remote: `^{}` ref does NOT exist → empty result → the script falls through to `git push origin "$TAG"`, which then fails because the tag already exists remotely.

Note that the local side (`git rev-list -n 1 "$TAG"`) resolves both annotated and lightweight tags correctly, so the logic is asymmetric.

Additionally, if the local tag matches `CURRENT_SHA` the script exits early, so a remote tag pointing at a different SHA would not be detected in that path either.

## Solution / Next Steps

- Query both the peeled and plain refs and fall back when the peeled form is empty. Example:

  ```sh
  REMOTE_SHA="$(git ls-remote --tags origin "refs/tags/$TAG^{}" "refs/tags/$TAG" \
    | awk '$2 ~ /\^\{\}$/ {peeled=$1} $2 !~ /\^\{\}$/ {plain=$1} END {print (peeled ? peeled : plain)}')"
  ```

- Consider verifying the remote SHA even when the local tag matches, so that a divergent remote tag is always surfaced.
- Optionally extract the tag-verification block into a shell function (or a dedicated script) to reduce duplication and make it testable (e.g. via bats).

Related: PR #1481.

dyoshikawa:
## Summary

- Query both peeled (`^{}`) and plain refs in `git ls-remote` so lightweight tags on origin are detected. The previous `^{}`-only form returned empty for lightweight tags, causing `git push` to fail with a duplicate-tag error.
- Always verify the remote tag SHA even when the local tag matches, so a divergent remote tag is surfaced instead of being silently skipped.

Closes #1485

## Test plan
- [ ] Dispatch `Publish Assets` on a tag that already exists on origin as a lightweight tag and verify the workflow reports the existing tag instead of failing on push.
- [ ] Dispatch on a tag that exists locally (from checkout) but not on origin and verify it pushes successfully.
- [ ] Dispatch on a tag that exists on origin pointing to a different commit and verify the workflow fails with a clear error.

### DIFF
--- a/.github/workflows/publish-assets.yml
+++ b/.github/workflows/publish-assets.yml
@@ -99,23 +99,33 @@ jobs:
         run: |
           CURRENT_SHA="$(git rev-parse HEAD)"
 
+          LOCAL_TAG_EXISTS=0
           if git rev-parse -q --verify "refs/tags/$TAG" >/dev/null; then
+            LOCAL_TAG_EXISTS=1
             TAG_SHA="$(git rev-list -n 1 "$TAG")"
             if [ "$TAG_SHA" != "$CURRENT_SHA" ]; then
               echo "::error::Tag $TAG already exists locally but points to $TAG_SHA instead of $CURRENT_SHA"
               exit 1
             fi
-            echo "Tag $TAG already exists locally and points to the current commit, skipping creation"
-            exit 0
           fi
 
-          REMOTE_SHA="$(git ls-remote --tags origin "refs/tags/$TAG^{}" | awk '{print $1}' | head -n 1)"
+          # Query both peeled (annotated) and plain (lightweight) refs, preferring
+          # the peeled form when present so annotated tags resolve to their commit.
+          REMOTE_SHA="$(git ls-remote --tags origin "refs/tags/$TAG^{}" "refs/tags/$TAG" \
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1497, opened 2026-04-15
fix(install): support single-skill repos with SKILL.md at path root

lgaudez:
## Summary

- `fetchSourceViaGit` silently skipped all files when a remote repo exposes a single skill at the root of `skillsPath` (e.g. `SKILL.md` directly at path `""`), because every file has `relativePath.indexOf("/") === -1` and was skipped by the existing loop
- Added a fallback that groups root-level files under the requested skill name when: `skillFileMap` is empty after the main loop, `skills` filter contains exactly one non-wildcard name, and root-level files exist
- Wildcard (`*`) is intentionally excluded from the fallback since the skill name would be ambiguous

Note that I opened the PR because I tried to add this skill in mu rylesync.jsonc -> "https://github.com/blader/humanizer.git"

rulesync.jsonc sample : 
``` 
"sources": [
  {
    "source": "https://github.com/blader/humanizer.git",
    "transport": "git",
    "ref": "8b3a17889fbf12bedae20974a3c9f9de746ed754",
    "path": "",
    "skills": ["humanizer"],
    "outputPath": ".rulesync/skills/.curated"
  },
]
```

## Test plan

- [x] New test: `should install single-skill repo with SKILL.md at path root` - verifies the happy path
- [x] New test: `should still handle classic subdirectory skill structure` - regression, existing behavior unchanged
- [x] New test: `should not install root-level files when skills filter is wildcard` - wildcard stays at 0 skills
- [x] All 33 existing unit tests pass


lgaudez:
HI @dyoshikawa, can you have a look at this PR when you get some time ?

Thx !

lgaudez:
HI @dyoshikawa , not interested in supporting single skill from repo root ? It would be very helpful for me

src/lib/sources.ts
```diff
@@ -628,6 +628,22 @@ async function fetchSourceViaGit(params: {
     skillFileMap.set(name, arr);
   }
 
+  // Fallback: single-skill repo where all files sit at the root of skillsPath.
+  // Only applies when the skills filter names exactly one non-wildcard skill.
+  // Wildcard is intentionally excluded — the skill name would be ambiguous.
+  if (skillFileMap.size === 0 && !isWildcard && skillFilter.length === 1) {
```
dyoshikawa, line 634: The guard `!isWildcard && skillFilter.length === 1` means users with `skills: ["*"]` or multiple named skills still silently get 0 skills — which is the exact debugging experience this PR is trying to fix for one specific case. Consider emitting a `logger.warn` when `skillFileMap.size === 0` and the fallback can't apply, similar to the existing `logger.warn("No skills/ directory found ...")` in `fetchSource`. It would help the next user diagnose the problem on their own.

src/lib/sources.ts
```diff
@@ -628,6 +628,22 @@ async function fetchSourceViaGit(params: {
...
+    const [singleSkillName] = skillFilter;
+    if (singleSkillName !== undefined) {
+      const rootFiles = remoteFiles.filter((f) => f.relativePath.indexOf("/") === -1);
```
dyoshikawa, line 637: Minor: `!f.relativePath.includes("/")` reads more naturally than `indexOf("/") === -1`. If you extract it into a small `isRootFile` helper it also makes the fallback's intent self-documenting.

src/lib/sources.test.ts
```diff
@@ -1008,4 +1008,81 @@ describe("resolveAndFetchSources", () => {
     expect(sourceEntry.skills).toHaveProperty("new-skill");
     expect(sourceEntry.skills).not.toHaveProperty("old-skill");
   });
+
+  it("should install single-skill repo with SKILL.md at path root", async () => {
```
dyoshikawa, line 1012: `singleSkillName` now flows from user-supplied config (`sourceEntry.skills[0]`) into `skillFileMap` as a key. The existing `shouldSkipSkill` check should still reject names containing `..`, `/`, or `\`, but since `SourceEntrySchema.skills` is only `z.array(z.string())` with no refinements, an explicit test asserting that `skills: ["../evil"]` produces 0 writes would nail down the guarantee for this new code path.

dyoshikawa (commented):
Solid, targeted fix with good tests. Two things worth considering before merging:

1. The same root-layout problem exists in the GitHub transport path (`fetchSource` around `src/lib/sources.ts:461-464`) — `filteredDirs` only includes directory entries, so a GitHub-transport config with `path: ""` pointing at a single-skill repo will still silently return 0 skills. Users who hit this on GitHub will file the same bug. Worth either extending the fallback to both paths or noting why it's git-only.

2. `docs/guide/declarative-sources.md` describes `path` semantics but doesn't mention that `path: ""` combined with a single-entry `skills` filter now installs root-layout repos. A short note there would make the feature discoverable.

Line-level comments below are minor.

dyoshikawa:
@lgaudez Thank you!

PR #1514, opened 2026-04-20
feat: add --input-root flag to decouple source rules from output directory

maxime-bentin-cko:
## Problem

Currently `rulesync generate` always reads source rules from `.rulesync/` inside the current working directory. This forces users to duplicate their rules in every repository.

## Solution

Add a `--rulesync-dir <path>` CLI flag that points to the **parent** of the `.rulesync/` directory. When set, all feature processors read source files from `<rulesyncDir>/.rulesync/` while generated output still lands in the working directory (or `--base-dirs` paths).

This enables a single central rules store — e.g. `~/.aiglobal/.rulesync/` — shared across any number of repositories:

```bash
rulesync generate --rulesync-dir ~/.aiglobal --targets '*'
```

## Changes

- **CLI** (`src/cli/index.ts`): adds `--rulesync-dir <path>` option to the `generate` command; also fixes `--base-dir` → `--base-dirs` (plural, consistent with existing code)
- **Config** (`src/config/config.ts`): adds `rulesyncDir` to `ConfigParams` / `Config`; `getRulesyncDir()` falls back to `process.cwd()` when not set (no breaking change)
- **All feature processors** (`rules`, `ignore`, `mcp`, `commands`, `subagents`, `skills`): accept and propagate `rulesyncDir` so source-file resolution is decoupled from the output base directory
- **Types** (`src/types/dir-feature-processor.ts`, `feature-processor.ts`): expose `rulesyncDir` on processor option types
- **Tests**: existing generate/import tests updated; `rulesyncDir` threaded through `generate.ts` integration layer

## Checklist

- [x] All existing tests pass (`pnpm cicheck`)
- [x] No breaking change — flag is optional, default behaviour unchanged
- [x] Works with `--base-dirs`, `--dry-run`, `--check`, and all `--features`


docs/guide/separate-rulesync-dir.md
```diff
@@ -0,0 +1,52 @@
...
+| **Changes**  | Source location (where `.rulesync/` is read from) | Output location (writes to user-scope config paths, e.g. `~/.claude/`) |
+| **Use when** | Your rule definitions live in a non-CWD directory | You want the output to go to the tool's global (user-scope) config     |
+
+They can be combined. For example, to read rules from `~/.aiglobal` and write them to Claude Code's global settings:
+
+```bash
+rulesync generate --rulesync-dir ~/.aiglobal --global --targets claudecode --features rules
+```
+
+> **Note:** Passing `--rulesync-dir` does not automatically enable `--global`. If the `rulesync.jsonc` inside `--rulesync-dir` sets `"global": true`, that value is respected — but if you are running from a project directory and want project-scope output, omit `--global`.
```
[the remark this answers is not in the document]
maxime-bentin-cko: [0c43acc](https://github.com/dyoshikawa/rulesync/pull/1514/commits/0c43accdf68c3576050e2280e503aa73c479194e)

src/cli/index.ts
```diff
@@ -215,6 +215,7 @@ const main = async () => {
       "--simulate-skills",
       "Generate simulated skills. This feature is only available for copilot, cursor and codexcli.",
     )
+    .option("--rulesync-dir <path>", "Custom path to the .rulesync source directory")
```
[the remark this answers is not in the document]
maxime-bentin-cko: [0c43acc](https://github.com/dyoshikawa/rulesync/pull/1514/commits/0c43accdf68c3576050e2280e503aa73c479194e)

src/config/config-resolver.ts
```diff
@@ -156,7 +160,11 @@ export class ConfigResolver {
       );
     }
 
-    const resolvedGlobal = global ?? configByFile.global ?? getDefaults().global;
+    // When --rulesync-dir is explicitly provided the user is decoupling source
+    // from output, so "global: true" from the config file must not apply unless
+    // the caller also explicitly passes --global.
+    const configGlobal = rulesyncDir !== undefined ? false : configByFile.global;
+    const resolvedGlobal = global ?? configGlobal ?? getDefaults().global ?? false;
```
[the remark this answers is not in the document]
maxime-bentin-cko: [f39b126](https://github.com/dyoshikawa/rulesync/pull/1514/commits/f39b1267687ad06b7c0c7fce9dea3639ad15ff3f)

src/lib/generate.test.ts
```diff
@@ -98,6 +99,7 @@ describe("generate", () => {
       getSimulateSubagents: vi.fn().mockReturnValue(false),
       getSimulateSkills: vi.fn().mockReturnValue(false),
       isPreviewMode: vi.fn().mockReturnValue(false),
+      getRulesyncDir: vi.fn().mockReturnValue(process.cwd()),
     };
```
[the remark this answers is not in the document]
maxime-bentin-cko: 0c43acc

src/test-utils/test-directories.ts
```diff
@@ -14,7 +14,9 @@ export async function setupTestDirectory({ home }: { home: boolean } = { home: f
   testDir: string;
   cleanup: () => Promise<void>;
 }> {
-  const testsDir = join(originalCwd, "tmp", "tests");
+  // Use TMPDIR environment variable to ensure writes are sandboxed-friendly
+  const baseTmpDir = process.env.TMPDIR || join(originalCwd, "tmp", "tests");
+  const testsDir = join(baseTmpDir, "rulesync-tests");
```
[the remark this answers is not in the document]
maxime-bentin-cko: 0c43acc

src/cli/index.ts
```diff
@@ -195,7 +195,7 @@ const main = async () => {
     )
     .option("--delete", "Delete all existing files in output directories before generating")
     .option(
-      "-b, --base-dir <paths>",
+      "-b, --base-dirs <paths>",
```
dyoshikawa: Renaming `-b, --base-dir` to `-b, --base-dirs` is a breaking change for anyone already scripting `--base-dir` (commander will reject the old flag as unknown). The PR description says "default behaviour unchanged", which isn't quite true with this rename bundled in. Could we either keep the old name as a deprecated alias (commander supports defining both) or split the rename into its own release-noted PR?

src/config/config-resolver.ts
```diff
@@ -215,6 +223,7 @@ export class ConfigResolver {
         getDefaults().gitignoreDestination,
       dryRun: dryRun ?? configByFile.dryRun ?? getDefaults().dryRun,
       check: check ?? configByFile.check ?? getDefaults().check,
+      rulesyncDir: rulesyncDir ?? configByFile.rulesyncDir ?? getDefaults().rulesyncDir,
```
dyoshikawa: `configByFile.rulesyncDir` looks unreachable here: `mergeConfigs` (~lines 81-104) does not copy the `rulesyncDir` field through, so putting `"rulesyncDir": "..."` in `rulesync.jsonc` is silently dropped. Either extend `mergeConfigs` and add a test that covers config-file sourcing, or explicitly reject `rulesyncDir` at the schema level and document that this is CLI/programmatic only.

src/config/config-resolver.ts
```diff
@@ -156,7 +160,11 @@ export class ConfigResolver {
       );
     }
 
-    const resolvedGlobal = global ?? configByFile.global ?? getDefaults().global;
+    // When --rulesync-dir is explicitly provided the user is decoupling source
+    // from output, so "global: true" from the config file must not apply unless
+    // the caller also explicitly passes --global.
+    const configGlobal = rulesyncDir !== undefined ? false : configByFile.global;
```
dyoshikawa: This silently drops `configByFile.global` whenever `rulesyncDir` is set — with no warning. A user whose central `rulesync.jsonc` sets `"global": true` and then runs `rulesync generate --rulesync-dir ~/.aiglobal` will find the output scope has changed out from under them. Consider emitting a `logger.warn` when we override, or — since the file is explicitly their central config — honouring the file value when `--global` isn't passed explicitly.

src/utils/file.ts
```diff
@@ -284,7 +284,11 @@ export function validateBaseDir(baseDir: string): void {
     throw new Error("baseDir cannot be an empty string");
   }
 
-  checkPathTraversal({ relativePath: baseDir, intendedRootDir: process.cwd() });
+  // Traversal check only applies to relative paths; absolute paths are
+  // explicitly provided by the caller and may point anywhere on the filesystem.
+  if (!isAbsolute(baseDir)) {
```
dyoshikawa: Before this change `validateBaseDir` rejected absolute paths that escape CWD. After this change any absolute path is accepted — including one passed via `--base-dirs` (since `config-resolver.ts` normalises all baseDirs to absolute before calling `validateBaseDir`). Net effect: the traversal check is effectively disabled for all baseDirs in practice. At minimum, please add unit tests in `file.test.ts` covering both the allowed absolute-path case and pathological inputs like `/foo/../../etc`. Consider resolving + normalising the path first and verifying the result still matches intent.

src/test-utils/test-directories.ts
```diff
@@ -14,7 +14,9 @@ export async function setupTestDirectory({ home }: { home: boolean } = { home: f
   testDir: string;
   cleanup: () => Promise<void>;
 }> {
-  const testsDir = join(originalCwd, "tmp", "tests");
+  // Use TMPDIR environment variable to ensure writes are sandboxed-friendly
+  const root = process.env.TMPDIR || originalCwd;
```
dyoshikawa: This change is unrelated to `--rulesync-dir` and affects every test in the repo (the default root flips from `./tmp/tests` to `$TMPDIR`). It also conflicts with `.claude/rules/testing-guidelines.md`, which pins the convention to `./tmp/tests/projects/{RANDOM}`. If the sandbox filesystem restriction is real, fixing it by rewriting the shared helper feels like the wrong layer — could we either set `TMPDIR` in the sandbox env itself, or at minimum split this out into its own PR so it can be reviewed on its own merits?

src/features/ignore/rulesync-ignore.ts
```diff
@@ -38,8 +38,9 @@ export class RulesyncIgnore extends RulesyncFile {
     };
   }
 
-  static async fromFile(): Promise<RulesyncIgnore> {
-    const baseDir = process.cwd();
+  static async fromFile({
+    baseDir = process.cwd(),
+  }: { baseDir?: string } = {}): Promise<RulesyncIgnore> {
```
dyoshikawa: Minor consistency nit: every other `RulesyncXxx.fromFile` in the repo uses a named parameter type (`RulesyncFileFromFileParams`, or a `Pick<…>` of it — see `rulesync-mcp.ts`, `rulesync-subagent.ts`). An inline anonymous type here breaks that pattern. Worth defining `RulesyncIgnoreFromFileParams = Pick<RulesyncFileFromFileParams, "baseDir">` and using it.

src/lib/generate.ts
```diff
@@ -304,6 +305,7 @@ async function generateIgnoreCore(params: {
       try {
         const processor = new IgnoreProcessor({
           baseDir: baseDir === process.cwd() ? "." : baseDir,
+          rulesyncDir: config.getRulesyncDir(),
```
dyoshikawa: `rulesyncDir` is threaded into the processors here and in the other `generateXxxCore` functions, but `generateHooksCore` and `generatePermissionsCore` (further down in this file) still construct their processors without it. `HooksProcessor.loadRulesyncFiles` and `PermissionsProcessor.loadRulesyncFiles` still call `RulesyncHooks.fromFile({ baseDir: process.cwd() })` / `RulesyncPermissions.fromFile({ baseDir: process.cwd() })`. Running with `--rulesync-dir ~/.aiglobal` will therefore read hooks/permissions from CWD while reading everything else from the central dir — a subtle, silent split. Please extend both processors and their matching `RulesyncHooks.fromFile` / `RulesyncPermissions.fromFile` the same way.

docs/guide/separate-rulesync-dir.md
```diff
@@ -0,0 +1,52 @@
+# Separate Rulesync Directory
```
dyoshikawa: This page isn't registered in `docs/.vitepress/config.ts` sidebar, so two things follow: (1) the page won't appear in the docs site navigation, and (2) `scripts/sync-skill-docs.ts` iterates the sidebar, so `skills/rulesync/separate-rulesync-dir.md` will never be generated. That breaks the repo's `docs/` ↔ `skills/rulesync/` sync invariant. Please add an entry under the Guide section and re-run the sync script.

Also: the "global is not applied" note at the end of this file is load-bearing behaviour — worth promoting it out of a blockquote into a `::: warning` admonition so readers don't miss it.

dyoshikawa (commented):
Thanks for this — decoupling the source `.rulesync/` from the output dir is a real quality-of-life win for teams with shared rules. A few concerns before merge:

1. `rulesyncDir` is threaded into rules/ignore/mcp/commands/subagents/skills but `HooksProcessor` and `PermissionsProcessor` still read from `process.cwd()`, so a user running `rulesync generate --rulesync-dir ~/.aiglobal` will get a split-brain state where hooks and permissions silently come from CWD while everything else comes from the central dir. See comment on `src/lib/generate.ts`.

2. Renaming `-b, --base-dir` to `-b, --base-dirs` is a breaking CLI change that's unrelated to this feature and contradicts the PR description's "default behaviour unchanged" claim. See comment on `src/cli/index.ts`.

3. `configByFile.rulesyncDir` appears to be unreachable because `mergeConfigs` does not copy the field through — so users who put `"rulesyncDir"` in `rulesync.jsonc` will have it silently dropped. See comment on `src/config/config-resolver.ts`.

4. The new guide `docs/guide/separate-rulesync-dir.md` is not registered in the VitePress sidebar, which means it won't render on the docs site and — because `scripts/sync-skill-docs.ts` iterates the sidebar — it will never be synced to `skills/rulesync/`. That breaks the repo invariant.

5. The `test-directories.ts` change touches the whole test suite and looks unrelated to the feature. Worth splitting off.

6. No e2e coverage was added. The project convention is to keep an end-to-end happy-path per Tool × Feature matrix entry; an e2e that proves "read from dir A, write to dir B" through the actual CLI would be valuable.

On the security side nothing critical stood out, but `validateBaseDir` is now a no-op for absolute paths — see that comment.

dyoshikawa:
Follow-up on the naming discussion — proposing this split:

**In scope for this PR**: rename `rulesyncDir` → `inputRoot` everywhere it was introduced. Rationale: the parameter is brand-new to this PR, so renaming it now costs nothing, and keeping `rulesyncDir` would carry an inferior name into every processor constructor, `Config` accessor, and file-class `fromFile` signature. The CLI flag would become `--input-root <path>`.

**Deferred to a separate issue**: rename the existing `baseDir` / `baseDirs` → `outputRoot` / `outputRoots` for naming symmetry. `baseDir` is pervasive (~4385 occurrences across ~310 files) and the CLI flag / `rulesync.jsonc` field are public API, so it deserves its own review pass with a proper deprecation path. Tracked in #1519.

This means the two names will be temporarily asymmetric (`inputRoot` alongside `baseDir`) on `main` until #1519 lands, which I think is acceptable — the asymmetry is visible and motivating rather than silently wrong, and #1519 is trivially mechanical once started.

Does this split work for you?

maxime-bentin-cko:
@dyoshikawa Yes that works for me. I will have time to work on that only on Friday. Cheers 😄

maxime-bentin-cko:
Taken over by @dyoshikawa

PR #1523, opened 2026-04-21
feat: add create-issue-with-websearch slash command

dyoshikawa:
## Summary
- Add new `/create-issue-with-websearch` slash command under `.rulesync/commands/`.
- Targets the case where the user only has a vague idea and needs heavy web research (WebSearch / WebFetch) to sharpen the spec before filing.
- Modeled on the existing `create-issue` skill, but adds an explicit research phase, a synthesis step, and issue sections for `Background from Research`, `Open Questions`, and `References`.

## Test plan
- [x] `pnpm cicheck` (code + content) passes
- [ ] Manually invoke `/create-issue-with-websearch` on a fuzzy topic and confirm it produces a well-scoped issue with cited sources

### DIFF
--- a/.rulesync/commands/create-issue-with-websearch.md
+++ b/.rulesync/commands/create-issue-with-websearch.md
@@ -1,0 +1,133 @@
+---
+targets:
+  - "*"
+description: >-
+  Create a GitHub issue from a vague idea by thoroughly researching the topic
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1525, opened 2026-04-21
docs(commands): strengthen reference URL requirements in create-issue-with-websearch

dyoshikawa:
## Summary

- Require every non-trivial claim in `Background from Research` to be backed by an inline link to its source.
- Add a dedicated "Reference URL Requirements" subsection with explicit rules for the `References` section (full URLs, annotations, primary-source preference, version/SHA pinning, inline cross-linking, empty-case handling).
- Ensures issues filed via this command always ship with auditable source URLs so reviewers can verify each claim.

## Test plan

- [x] `pnpm cicheck` passes (fmt, lint, typecheck, tests, cspell, secretlint).
- [ ] Next time the `create-issue-with-websearch` command is run, verify the generated issue includes annotated reference URLs inline and in the References section.

### DIFF
--- a/.rulesync/commands/create-issue-with-websearch.md
+++ b/.rulesync/commands/create-issue-with-websearch.md
@@ -81,7 +81,7 @@ The problem or opportunity, grounded in what the research revealed (user impact,
 
 ## Background from Research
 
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1527, opened 2026-04-21
feat: add draft-issue-to-external-repository slash command

dyoshikawa:
## Summary

- Add a new slash command `draft-issue-to-external-repository` that captures the workflow for proposing changes to repositories outside the user's own owner (e.g. microsoft/apm).
- Codifies the steps we used while drafting an issue to add rulesync as an APM install target: inspect target repo conventions (templates, labels, recent titles), research the topic against primary sources + deepwiki, ground claims in this repo's state, align scope/tone with the user, draft in Japanese for review then translate to English, and fall back to saving the body under `tmp/external-repository-issues/*.md` when `gh` auth lacks permission on the target.

## Motivation

Filing a well-scoped, conventional-looking issue on a foreign repo requires meaningfully different steps from creating an issue on our own repo — mostly around auth fallback, template conformance, and tone. The existing `create-issue` and `create-issue-with-websearch` commands assume ownership and do not cover any of that. This command makes the workflow reproducible.

## Test plan

- [x] The new command shows up in the available-skills list after generation
- [ ] Exercise the command end-to-end on another third-party repo and confirm the fallback path writes a valid metadata-headed file

### DIFF
--- a/.rulesync/commands/draft-issue-to-external-repository.md
+++ b/.rulesync/commands/draft-issue-to-external-repository.md
@@ -1,0 +1,120 @@
+---
+targets:
+  - "*"
+description: >-
+  Draft a GitHub issue targeting a third-party repository (outside this
+  project's owner). Researches the target repo's conventions and the topic in
+  depth, iterates with the user on scope and tone, and produces a final body
+  saved to tmp/external-repository-issues/*.md for the user to post manually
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1530, opened 2026-04-21
fix(install): align skill source fallback across transports

dyoshikawa:
### Motivation
- Fix transport asymmetry where root-level `SKILL.md` fallback existed only for `git` transport but not for the GitHub API transport, causing inconsistent behavior for the same user input. 
- Prevent misclassification of nested files when paths use Windows backslashes by making path-separator logic separator-agnostic.

### Description
- Add a small `RemoteSkillFile` type and helper `getFirstPathSeparatorIndex` to detect both `/` and `\` separators. 
- Introduce `groupRemoteFilesBySkillRoot` and use it from both `fetchSourceViaGit` and `fetchSource` to unify fallback semantics for single-skill repos whose files live at the configured skills base path root. 
- Extend GitHub transport to handle the single-skill root-level fallback when `skills` specifies exactly one non-wildcard skill, collecting root files and writing them under that skill name. 
- Replace ad-hoc two-pass path scanning with the shared grouping helper and make git-transport grouping separator-agnostic to correctly handle Windows `\` paths. 
- Add two regression tests in `src/lib/sources.test.ts` covering (1) GitHub transport root-level single-skill fallback and (2) backslash-separated git paths being treated as nested skill files.

### Testing
- Ran `pnpm vitest run src/lib/sources.test.ts` and the updated test file passed (`35 tests`).
- Ran the full CI check `pnpm cicheck` (which runs formatting, lints, typecheck and the full test suite) and it succeeded; the full test run shows all tests passed (`4952 tests`).
- Fixed formatting where required with `pnpm oxfmt` during the rollout before re-running `pnpm cicheck`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69e74f57a3ec8330ae0bf52162d0f8bf)

### DIFF
--- a/src/lib/sources.test.ts
+++ b/src/lib/sources.test.ts
@@ -1085,4 +1085,63 @@ describe("resolveAndFetchSources", () => {
     expect(result.fetchedSkillCount).toBe(0);
     expect(writeFileContent).not.toHaveBeenCalled();
   });
+
+  it("should install single-skill repo with SKILL.md at path root via github transport", async () => {
+    mockClientInstance.listDirectory.mockImplementation(
+      async (_owner: string, _repo: string, path: string) => {
+        if (path === "skills") {
+          return [
+            { name: "SKILL.md", path: "skills/SKILL.md", type: "file", size: 50 },
+            { name: "README.md", path: "skills/README.md", type: "file", size: 20 },
+          ];
+        }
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1548, opened 2026-04-22
chore: enable takt as a generation target

dyoshikawa:
## Summary
- Add `takt` to the targets in `rulesync.jsonc` so that rules, commands, subagents, and skills are generated for it.
- Extend `.gitignore` to ignore the generated `.takt/` artifacts (facets, runs, tasks, cache, and config) that should not be checked in.

## Test plan
- [x] `pnpm cicheck`


dyoshikawa:
@dyoshikawa Thank you!

PR #1561, opened 2026-04-24
fix(copilot): preserve no-globs state in root rule roundtrip

### LINKED ISSUE #1543: [Bug]: Glob pattern data loss in CopilotRule roundtrip conversion

## Problem

I found a data loss bug in `CopilotRule.toRulesyncRule()`. When converting a root rule with `applyTo: "**"`, the value is normalized to `"**/*"` during conversion to a `RulesyncRule`, but this original value is never restored on the way back, causing the roundtrip to be non-idempotent.

## Evidence

In `src/features/rules/copilot-rule.ts` around line 117–118:

```typescript
if (this.isRoot()) {
  globs = ["**/*"];  // Always hardcoded — original "**" is lost
}
```

And in `fromRulesyncRule()` around line 160:

```typescript
applyTo: rulesyncFrontmatter.globs?.length
  ? rulesyncFrontmatter.globs.join(",")
  : undefined,
```

The roundtrip results in:

| Step | Value |
|------|-------|
| Input (`applyTo`) | `"**"` |
| After `toRulesyncRule` (`globs`) | `["**/*"]` |
| After `fromRulesyncRule` (`applyTo`) | `"**/*"` |

The original `"**"` is permanently replaced with `"**/*"` after a single conversion cycle.

## Reproduction Steps

1. Create a `CopilotRule` with `applyTo: "**"` in frontmatter
2. Call `.toRulesyncRule()` → observe `globs: ["**/*"]`
3. Call `CopilotRule.fromRulesyncRule()` on the result → observe `applyTo: "**/*"` (not `"**"`)

## Expected Behavior

A roundtrip conversion should preserve the original glob value, or at minimum, the two values (`"**"` and `"**/*"`) should be treated as equivalent throughout the pipeline.

## Actual Behavior

`"**"` is silently replaced with `"**/*"`, causing inconsistency when re-importing or re-syncing rules.

dyoshikawa:
## Summary

- Fixes #1543.
- `CopilotRule.toRulesyncRule()` previously hardcoded `globs: ['**/*']` for every root rule, even when the source Copilot file had no `applyTo` frontmatter. Re-importing a generated `.github/copilot-instructions.md` therefore produced a rulesync rule with bogus `globs`, yielding two root rules and breaking subsequent `generate` with `Multiple root rulesync rules found`.
- Now only populate `globs` when `applyTo` is actually set on the Copilot frontmatter. Root Copilot files with empty frontmatter round-trip back to rulesync rules with no `globs`. `applyTo: '**'` is preserved verbatim rather than silently normalized to `'**/*'`.

## Changes

- `src/features/rules/copilot-rule.ts` — dropped the unconditional `isRoot()` globs injection in `toRulesyncRule()`; `fromRulesyncRule()` already emits `applyTo: undefined` when globs are empty, so symmetry holds.
- `src/features/rules/copilot-rule.test.ts` — updated one assertion (root + `applyTo: '**'` now round-trips to `globs: ['**']`) and added two regression tests:
  1. Root CopilotRule with empty frontmatter → imported RulesyncRule has no `globs`.
  2. Full `fromRulesyncRule` → `toRulesyncRule` cycle for a root rule with no globs preserves the no-globs state (reproduces the #1543 scenario).

## Test plan

- [x] `pnpm cicheck` passes locally (209 test files / 5260 tests)
- [x] New regression tests reproduce the original #1543 bug and now pass
- [ ] Manually verify the scenario from #1543: `rulesync generate --targets copilot` → `rulesync import --targets copilot` → subsequent `generate --dry-run` no longer errors with "Multiple root rulesync rules found"


dyoshikawa:
@dyoshikawa Thank you!

PR #1564, opened 2026-04-24
fix: directory-aware root detection in CopilotRule.fromFile

dyoshikawa:
### Motivation
- Prevent misclassification of non-root Copilot instruction files that share the same filename as the root file by respecting the provided directory context when available.

### Description
- Make `CopilotRule.fromFile()` detect root vs non-root using `relativeDirPath + relativeFilePath` when `relativeDirPath` is provided, and fall back to filename-only detection for backward compatibility when it's not.
- Resolve and use `resolvedRelativeDirPath` for reading non-root files so the directory context is preserved during file I/O and instance creation.
- Align `CopilotRule.forDeletion()` root detection to the same directory-aware logic.
- Add tests to ensure that a file named `copilot-instructions.md` in a non-root directory is treated as non-root, and that root is detected only when both directory and filename match.

### Testing
- Ran `pnpm vitest run src/features/rules/copilot-rule.test.ts` and the test file passed (62/62 tests passed).
- Ran full checks with `pnpm cicheck`, which completed successfully (format/lint/typecheck passed and the full test suite passed: 5262 tests).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb571142a08330b86e00e07c1eb0d4)

### DIFF
--- a/src/features/rules/copilot-rule.test.ts
+++ b/src/features/rules/copilot-rule.test.ts
@@ -594,6 +594,55 @@ This is test rule content from file.`;
       expect(copilotRule.isRoot()).toBe(true);
     });
 
+    it("should use relativeDirPath to distinguish non-root files named copilot-instructions.md", async () => {
+      const instructionsDir = join(testDir, ".github", "instructions");
+      await ensureDir(instructionsDir);
+
+      const fileContent = `---
+description: "Same filename as root"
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1567, opened 2026-04-25
chore(security): add cooldown for supply chain protection

saitota:
## Summary

Adds 7-day cooldown to mitigate supply chain attacks (Trivy/LiteLLM/axios in March 2026).

- `pnpm-workspace.yaml`: `minimumReleaseAge: 10080` (7 days, in minutes)
- `.github/dependabot.yml`: `cooldown.default-days` 1 → 7 (npm + github-actions)

## Refs

- Article: https://zenn.dev/dely_jp/articles/supply-chain-kowai
- pnpm setting: https://pnpm.io/settings#minimumreleaseage (requires pnpm 10.16+, this repo uses 10.33.2)
- Supply chain guide: https://pnpm.io/supply-chain-security

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 1
+      default-days: 7
     open-pull-requests-limit: 2
     groups:
       all-dependencies:
@@ -38,7 +38,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 1
+      default-days: 7
     open-pull-requests-limit: 1
     groups:
       all-actions:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 7
+      default-days: 1
     open-pull-requests-limit: 2
     groups:
       all-dependencies:
@@ -38,7 +38,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 7
+      default-days: 1
     open-pull-requests-limit: 1
     groups:
       all-actions:
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,0 +1,1 @@
+minimumReleaseAge: 10080
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,1 +1,1 @@
-minimumReleaseAge: 10080
+minimumReleaseAge: 10080 # 7 days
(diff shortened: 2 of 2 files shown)


cm-dyoshikawa:
@saitota 

Thanks for putting this together. The CI action bumps look good to me, please go ahead with those.

On the cooldown side though, I think 7 days is too long for this project. The main concern is that it also delays security fixes by the same amount — if a CVE drops on a transitive dep, we'd be sitting on the patched version for a week before pnpm even considers it. pnpm's own docs note that "in most cases malicious releases are removed from the registry within an hour," and their example config uses 1 day (1440 minutes), so 7 days is well above what the upstream guidance suggests.

Also worth remembering that `minimumReleaseAge` is a floor, not a ceiling — any downstream user who wants a longer window can extend it on their side without us baking it into the project default.

I'd like to drop this to 1 day (1440) at least for now. Same change to `cooldown.default-days` in dependabot.yml — 1 day there feels right too. Happy to revisit if we see a concrete incident that a longer window would have caught.

dyoshikawa (commented):
Thanks for the PR — the motivation is solid and the references are helpful. I'm on board with bumping the GitHub Actions cooldown to 7 days since those updates are less urgent.

However, I think 7 days is too aggressive for the npm side. The pnpm docs themselves note that "in most cases, malicious releases are discovered and removed from the registry within an hour" (https://pnpm.io/settings#minimumreleaseage). A 7-day gate could actually hurt us by delaying adoption of critical security fixes. A 1-day cooldown should be more than enough for this project, and users who want stricter protection can always extend it in their own setup.

pnpm-workspace.yaml
```diff
@@ -0,0 +1 @@
+minimumReleaseAge: 10080 # 7 days
```
dyoshikawa: I would suggest setting this to `1440` (1 day) rather than `10080` (7 days). The pnpm docs say: "In most cases, malicious releases are discovered and removed from the registry within an hour" (https://pnpm.io/settings#minimumreleaseage). Seven days risks blocking timely security patches without a proportional safety gain. A 1-day window already covers the vast majority of supply chain threats while keeping us responsive to legitimate fixes.

.github/dependabot.yml
```diff
@@ -6,7 +6,7 @@ updates:
     schedule:
       interval: "weekly"
     cooldown:
-      default-days: 1
+      default-days: 7
```
dyoshikawa: Same reasoning as the pnpm setting — for npm dependencies, I think `default-days: 1` is sufficient here. Most malicious packages are caught well within that window, and a 7-day delay could slow down important security updates.

saitota: Updated both to 1 day. Thanks for the review!

dyoshikawa:
@saitota Thank you!

PR #1572, opened 2026-04-27
refactor(output-root): rename baseDir to outputRoot for naming symmetry with inputRoot

### LINKED ISSUE #1519: refactor: rename baseDir to outputRoot for naming symmetry with inputRoot
## Summary

Rename the pervasive `baseDir` / `baseDirs` concept to `outputRoot` / `outputRoots` so it reads symmetrically against the new `inputRoot` concept being introduced in #1514.

## Motivation / Purpose

PR #1514 decouples the source `.rulesync/` directory from the output directory by adding a new parameter. During review we agreed to land that new parameter as `inputRoot` for clarity. The natural counterpart is to rename the existing `baseDir` (which in practice is the *output* root — where generated tool configs are written to) to `outputRoot`.

Keeping the names asymmetric (`inputRoot` ↔ `baseDir`) would be confusing and erode the clarity benefit of the original rename. However, `baseDir` is pervasive across the public API and shipping the full rename inside #1514 would bloat that PR and make it harder to review behavior changes in isolation. Splitting it out into a dedicated refactor is the right call.

## Deprecation policy (coexistence period)

The old names **must stay functional** during a coexistence period so existing users are not broken. Concretely:

- **CLI flag**: `--base-dirs` / `-b` continues to work as an alias of `--output-roots`. When the old flag is used, the CLI emits a deprecation warning log (e.g. `logger.warn("--base-dirs is deprecated; use --output-roots instead. It will be removed in a future release.")`).
- **Config file field**: `baseDirs` in `rulesync.jsonc` continues to be accepted as an alias of `outputRoots`. When the old field is present, the config loader emits a deprecation warning log on startup.
- **Public programmatic API**: `Config.getBaseDirs()` continues to exist alongside `getOutputRoots()` and returns the same value, but emits a deprecation warning the first time it is called per process (to avoid spam). The underlying storage is `outputRoots`.
- **Zod schema**: accept both `baseDirs` and `outputRoots`; if both are present, prefer `outputRoots` and warn; if only `baseDirs` is present, map it internally to `outputRoots` and warn.

Removal timeline: keep the deprecated surface for at least one full minor release cycle, then remove in the next major version bump. Update `CHANGELOG.md` and the docs to call out the deprecation at the time of the rename and the removal at the time of the major bump.

Internal-only parameters (processor constructors, file-class `fromFile` / `fromDir` signatures) are *not* part of the documented public API and can be renamed directly without aliasing. If a downstream user is importing these internals they will see a TypeScript error on upgrade, which is acceptable.

## Details

Scope: rename `baseDir` → `outputRoot` (and `baseDirs` → `outputRoots`) everywhere, preserving behavior. There is no functional change in this issue.

Affected surfaces (non-exhaustive; grep currently shows ~4385 `baseDir` occurrences across ~310 files):

- **CLI flag**: `-b, --base-dirs <paths>` → `-b, --output-roots <paths>` in `src/cli/index.ts` (keep old name as deprecated alias)
- **Config schema**: `ConfigParamsSchema.baseDirs` → `outputRoots` in `src/config/config.ts`, plus handling for the field in `rulesync.jsonc` (keep old field as deprecated alias)
- **Config class**: `Config.getBaseDirs()` → `getOutputRoots()` and the private `baseDirs` field (keep old accessor as deprecated alias)
- **Config resolver**: `getBaseDirsInLightOfGlobal` and related helpers in `src/config/config-resolver.ts`
- **Processors**: `FeatureProcessor` / `DirFeatureProcessor` constructor param and protected field in `src/types/feature-processor.ts` and `src/types/dir-feature-processor.ts`, plus every subclass under `src/features/**/*-processor.ts` (internal — direct rename)
- **File classes**: every `RulesyncFile` / tool-file `fromFile` / `fromDir` signature that accepts `baseDir` (internal — direct rename)
- **Utilities**: `validateBaseDir` → `validateOutputRoot` in `src/utils/file.ts` (internal — direct rename)
- **Tests**: all test files that reference these names, plus new tests exercising the deprecation-warning paths
- **Docs**: `docs/**/*.md`, `skills/rulesync/*.md` (synced by `scripts/sync-skill-docs.ts`), and `README.md`

### Acceptance criteria

- All production code uses `outputRoot` / `outputRoots` exclusively for the new surface.
- `--base-dirs` and the `baseDirs` field in `rulesync.jsonc` still work, and produce a visible deprecation warning log each time they are used.
- `Config.getBaseDirs()` still works and produces a deprecation warning the first time it is called per process.
- A regression test confirms each deprecated alias still maps to the new behavior **and** emits the expected warning.
- Existing tests are updated to use the new names.
- Docs under `docs/` and `skills/rulesync/` reflect the new naming, and explicitly mention the deprecation of the old names.
- `CHANGELOG.md` notes both the new name and the deprecation.
- `pnpm cicheck` passes.

### Out of scope

- Any behavior change. This is purely a rename + deprecation path.
- The `rulesyncDir` → `inputRoot` rename, which is handled in #1514.
- Removing the deprecated names (that happens in a separate major-bump PR later).

## Additional Context

- Related PR: #1514

dyoshikawa:
## Summary

- Renames pervasive identifier `baseDir` → `outputRoot` (and `baseDirs` → `outputRoots`) across the codebase so the naming is symmetric with the `inputRoot` concept introduced in #1514.
- Preserves public API via deprecated aliases: `--base-dir` CLI flag, `baseDirs` config-file field, programmatic `Config.getBaseDirs()`, and `ConfigResolver.resolve({ baseDirs })`. Each emits a one-shot deprecation warning the first time it is used per process and can be silenced with `RULESYNC_SILENT_DEPRECATION=1`.
- When both new and deprecated names are supplied, the new `outputRoots` wins; the resolver/CLI also surfaces an override warning when the values differ as a set.
- Updates docs (`docs/api/programmatic-api.md`, `docs/guide/configuration.md`) and the synced `skills/rulesync/` mirror to document the rename and the deprecation path.
- Adds regression tests for each deprecation path (`Config.getBaseDirs()`, `baseDirs` config field, programmatic `baseDirs` resolver param, `--base-dir` CLI flag) covering: alias still maps to new behavior, warning is emitted, warning fires once per process, warning is silenced via `RULESYNC_SILENT_DEPRECATION`.

Closes #1519. Related to #1514.

## Test plan

- [x] `pnpm cicheck` (5318 tests pass; lint, typecheck, fmt, cspell, secretlint clean)
- [x] Programmatic `Config.getBaseDirs()` returns the same array as `getOutputRoots()` and emits the deprecation warning exactly once per process
- [x] Config-file `baseDirs` is mapped to `outputRoots` with warning; when both are present `outputRoots` wins
- [x] CLI `--base-dir` is still accepted, mapped to `outputRoots`, and emits a deprecation warning whenever used
- [x] `RULESYNC_SILENT_DEPRECATION=1` suppresses every new deprecation warning

### DIFF
--- a/.claude/rules/coding-guidelines.md
+++ b/.claude/rules/coding-guidelines.md
@@ -13,7 +13,7 @@ paths:
 - TypeScript file names should be in kebab-case, even for class implementation files.
 - Don't create ballel files. Please always direct import the implementation file.
   - The maintainer thinks that ballel files are harmful to tree-shaking and import path transparency.
-- The default value of `baseDir` should be `process.cwd()` because it is easier to mock in tests compared to hardcoding `"."`. However, the default value of `relativeDirPath` should be `"."` because it should be relative path to concatenate with `baseDir`.
+- The default value of `outputRoot` should be `process.cwd()` because it is easier to mock in tests compared to hardcoding `"."`. However, the default value of `relativeDirPath` should be `"."` because it should be relative path to concatenate with `outputRoot`.
 - When logging errors, you must use `formatError` function in `src/utils/error.ts` to format the error message.
 - When writing any filesystem path, you must always use `join` function in `node:path` to join the path because it must support both Windows and Unix-like paths.
 - When writing non-filesystem paths (e.g., API paths, gitignore entries, generated file content), use `path.posix.join` or `toPosixPath` from `src/utils/file.ts` to ensure forward slashes regardless of platform.
--- a/.github/instructions/coding-guidelines.instructions.md
+++ b/.github/instructions/coding-guidelines.instructions.md
@@ -13,7 +13,7 @@ applyTo: '**/*.ts'
 - TypeScript file names should be in kebab-case, even for class implementation files.
 - Don't create ballel files. Please always direct import the implementation file.
   - The maintainer thinks that ballel files are harmful to tree-shaking and import path transparency.
-- The default value of `baseDir` should be `process.cwd()` because it is easier to mock in tests compared to hardcoding `"."`. However, the default value of `relativeDirPath` should be `"."` because it should be relative path to concatenate with `baseDir`.
+- The default value of `outputRoot` should be `process.cwd()` because it is easier to mock in tests compared to hardcoding `"."`. However, the default value of `relativeDirPath` should be `"."` because it should be relative path to concatenate with `outputRoot`.
 - When logging errors, you must use `formatError` function in `src/utils/error.ts` to format the error message.
 - When writing any filesystem path, you must always use `join` function in `node:path` to join the path because it must support both Windows and Unix-like paths.
 - When writing non-filesystem paths (e.g., API paths, gitignore entries, generated file content), use `path.posix.join` or `toPosixPath` from `src/utils/file.ts` to ensure forward slashes regardless of platform.
--- a/.github/instructions/testing-guidelines.instructions.md
+++ b/.github/instructions/testing-guidelines.instructions.md
@@ -27,7 +27,7 @@ applyTo: '**/*.test.ts,src/e2e/**/*.spec.ts'
       it("Test Case", async () => {
         // Run test using testDir
         await RulesyncRule.fromFile({
-          baseDir: testDir,
+          outputRoot: testDir,
           relativeFilePath: "test.md",
         });
         // ...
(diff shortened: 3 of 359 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1584, opened 2026-04-29
fix: normalize copilot root path comparisons

### LINKED ISSUE #1544: [Bug]: CopilotRule.fromFile() determines root by filename alone, ignoring directory context

## Problem

`CopilotRule.fromFile()` determines whether a file is a "root" rule by comparing only `relativeFilePath`, without considering `relativeDirPath`. This may cause incorrect root detection in edge cases, especially in global mode.

## Evidence

In `src/features/rules/copilot-rule.ts` around line 208:

```typescript
const isRoot = relativeFilePath === paths.root.relativeFilePath;
```

The comparison is performed against only the file name portion (e.g., `"copilot-instructions.md"`), without verifying that the directory also matches. In project mode, the root file is `.github/copilot-instructions.md`, and in global mode it is `.copilot/copilot-instructions.md`. If a non-root file happens to be named `copilot-instructions.md` in a different directory, it could be incorrectly classified as a root rule.

## Expected Behavior

Root detection should use the full relative path (directory + filename) to unambiguously distinguish root from non-root files.

## Proposed Solution

Compare the full path including directory:

```typescript
const isRoot =
  path.join(relativeDirPath, relativeFilePath) ===
  path.join(paths.root.relativeDirPath, paths.root.relativeFilePath);
```

zerone0x:
## Summary
Fixes #1544

Normalize Copilot rule relative paths before comparing explicit paths against the root Copilot instruction path. This preserves directory-aware root detection while handling mixed path separators from Windows/WSL-style inputs.

## Changes
- Add shared relative-path normalization for Copilot root detection.
- Use it in both `fromFile` and `forDeletion` root checks.
- Add regression coverage for same-filename non-root detection and mixed-separator root detection.

## Test Plan
- `npx -y -p node@22 -p pnpm@10 pnpm exec vitest run src/features/rules/copilot-rule.test.ts --silent=true`
- `npx -y -p node@22 -p pnpm@10 pnpm run fmt:check`
- `npx -y -p node@22 -p pnpm@10 pnpm run typecheck`
- `npx -y -p node@22 -p pnpm@10 pnpm run oxlint -- src/features/rules/copilot-rule.ts src/features/rules/copilot-rule.test.ts`
- `npx -y -p node@22 -p pnpm@10 pnpm run eslint -- src/features/rules/copilot-rule.ts src/features/rules/copilot-rule.test.ts`

Note: the local host has Node v20.19.2, while the repo declares Node >=22.0.0, so checks were run via `node@22`/`pnpm@10` from npx.

### DIFF
--- a/src/features/rules/copilot-rule.test.ts
+++ b/src/features/rules/copilot-rule.test.ts
@@ -620,6 +620,24 @@ This should be treated as a non-root rule.`;
       expect(copilotRule.getBody()).toBe("This should be treated as a non-root rule.");
     });
 
+    it("should normalize separators when detecting explicit root paths", async () => {
+      const githubDir = join(testDir, ".github");
+      await ensureDir(githubDir);
+
+      const rootContent = "Root detected with mixed separators.";
+      await writeFileContent(join(githubDir, "copilot-instructions.md"), rootContent);
+
+      const copilotRule = await CopilotRule.fromFile({
+        outputRoot: testDir,
+        relativeDirPath: ".github\\",
+        relativeFilePath: "copilot-instructions.md",
+        validate: true,
+      });
+
+      expect(copilotRule.isRoot()).toBe(true);
+      expect(copilotRule.getBody()).toBe(rootContent);
+    });
+
     it("should detect root only when both relativeDirPath and filename match", async () => {
       const githubDir = join(testDir, ".github");
       await ensureDir(githubDir);
--- a/src/features/rules/copilot-rule.ts
+++ b/src/features/rules/copilot-rule.ts
@@ -48,6 +48,13 @@ export type CopilotRuleSettablePathsGlobal = ToolRuleSettablePathsGlobal & {
   };
 };
 
+const normalizeRelativePath = (path: string): string =>
+  path.replace(/\\/g, "/").replace(/\/+/g, "/");
+
+const sameRelativePath = (leftDir: string, leftFile: string, rightDir: string, rightFile: string) =>
+  normalizeRelativePath(join(leftDir, leftFile)) ===
+  normalizeRelativePath(join(rightDir, rightFile));
+
 /**
  * Rule generator for GitHub Copilot
  *
@@ -209,8 +216,12 @@ export class CopilotRule extends ToolRule {
   }: ToolRuleFromFileParams): Promise<CopilotRule> {
     const paths = this.getSettablePaths({ global });
     const isRoot = relativeDirPath
-      ? join(relativeDirPath, relativeFilePath) ===
-        join(paths.root.relativeDirPath, paths.root.relativeFilePath)
+      ? sameRelativePath(
+          relativeDirPath,
+          relativeFilePath,
+          paths.root.relativeDirPath,
+          paths.root.relativeFilePath,
+        )
       : relativeFilePath === paths.root.relativeFilePath;
     const resolvedRelativeDirPath =
       relativeDirPath ??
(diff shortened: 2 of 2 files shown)


src/features/rules/copilot-rule.ts
```diff
@@ -48,6 +48,13 @@ export type CopilotRuleSettablePathsGlobal = ToolRuleSettablePathsGlobal & {
   };
 };
 
+const normalizeRelativePath = (path: string): string =>
+  path.replace(/\\/g, "/").replace(/\/+/g, "/");
```
dyoshikawa, line 52: `src/utils/file.ts` already exports `toPosixPath`, which does the same backslash-to-slash normalization and is used elsewhere in the codebase (e.g. `src/types/ai-file.ts`). Could we reuse it here instead of defining a local helper? Also, is the `\/+/g` collapsing actually needed? `path.join` shouldn't emit `//` from these inputs, so unless there's a case I'm missing it looks like dead defense — and if it is needed, a test for that input would be helpful.

src/features/rules/copilot-rule.ts
```diff
@@ -48,6 +48,13 @@ export type CopilotRuleSettablePathsGlobal = ToolRuleSettablePathsGlobal & {
...
+
+const sameRelativePath = (leftDir: string, leftFile: string, rightDir: string, rightFile: string) =>
+  normalizeRelativePath(join(leftDir, leftFile)) ===
+  normalizeRelativePath(join(rightDir, rightFile));
```
dyoshikawa, line 56: Heads-up: the `join(relativeDirPath, relativeFilePath) === join(...)` pattern this replaces also exists in other rule classes (e.g. `rovodev-rule.ts`). They'd hit the same bug on Windows/WSL inputs. Consider promoting `sameRelativePath` to `tool-rule.ts` so all rules benefit, or filing a follow-up issue to track the horizontal fix.

src/features/rules/copilot-rule.test.ts
```diff
@@ -620,6 +620,24 @@ This should be treated as a non-root rule.`;
...
+      await writeFileContent(join(githubDir, "copilot-instructions.md"), rootContent);
+
+      const copilotRule = await CopilotRule.fromFile({
+        outputRoot: testDir,
+        relativeDirPath: ".github\\",
+        relativeFilePath: "copilot-instructions.md",
+        validate: true,
+      });
+
+      expect(copilotRule.isRoot()).toBe(true);
```
dyoshikawa, line 637: Nice regression test. Could you add a parallel case for `forDeletion` as well? The fix touches both code paths but only `fromFile` is exercised here, so the `forDeletion` branch could silently regress. A brief comment noting that `".github\\"` simulates Windows/WSL-style input would also help future readers parse the intent.

dyoshikawa (commented):
Targeted fix for the mixed-separator root-detection issue, with a regression test that pins down the bug. Implementation is small and correct. Two things worth considering before merging: (1) the repo already has a `toPosixPath` helper in `src/utils/file.ts` that does the backslash-to-slash conversion, so the new `normalizeRelativePath` ends up duplicating it; reusing the existing helper would keep the codebase consistent. (2) The same `join(...) === join(...)` comparison pattern exists in other rule classes (e.g. `rovodev-rule.ts`), so the same Windows/WSL input could trip them up too — worth a follow-up issue at minimum, or lifting the helper into `tool-rule.ts` if you'd rather fix it everywhere now. Test coverage is also one-sided (only the `fromFile` path); a quick `forDeletion` case would round it out.

dyoshikawa:
@zerone0x Thank you!

PR #1591, opened 2026-05-01
test: cover claudecode global mcp deletion guard

### LINKED ISSUE #1275: Bug: ClaudecodeMcp.fromFile and fromRulesyncMcp do not pass global flag to constructor
\`ClaudecodeMcp.fromFile()\` and \`ClaudecodeMcp.fromRulesyncMcp()\` both accept a \`global\` parameter and use it to resolve paths via \`getSettablePaths({ global })\`, but neither passes it through to the \`ClaudecodeMcp\` constructor.

This means instances created by these factory methods always have \`this.global = false\` (the default), causing \`isDeletable()\` to incorrectly return \`true\` even in global mode. In global mode, \`~/.claude/.claude.json\` may contain other user settings and should not be deleted.

```typescript
// src/features/mcp/claudecode-mcp.ts (lines 61–67)
return new ClaudecodeMcp({
  baseDir,
  relativeDirPath: paths.relativeDirPath,
  relativeFilePath: paths.relativeFilePath,
  fileContent: JSON.stringify(newJson, null, 2),
  validate,
  // missing: global,
});
```

The same issue exists in `fromRulesyncMcp` (lines 86–92).

For reference, `CursorMcp` (added in #1273) correctly passes `global` in both factory methods.

**Fix:** Add `global` to the constructor argument object in both `fromFile` and `fromRulesyncMcp`.

zerone0x:
## Summary
- Add explicit regression coverage that `ClaudecodeMcp.fromFile({ global: true })` keeps global Claude MCP config non-deletable when initializing a missing global config file.
- This guards the issue where factory-created global instances must preserve the constructor's `global` flag so `isDeletable()` remains false.

Fixes #1275

## Tests
- `pnpm --config.engine-strict=false vitest run src/features/mcp/claudecode-mcp.test.ts --silent=true`
- `NPM_CONFIG_ENGINE_STRICT=false pnpm --config.engine-strict=false run check`
- `NPM_CONFIG_ENGINE_STRICT=false pnpm --config.engine-strict=false run test`

Note: local Node is v20.19.2 while the package declares Node >=22, so commands were run with pnpm engine-strict disabled.

## Orbit assessment
- Complexity: XS
- Risk: Low; test-only regression assertion in the existing Claudecode MCP test suite
- Routing: safe direct PR

### DIFF
--- a/src/features/mcp/claudecode-mcp.test.ts
+++ b/src/features/mcp/claudecode-mcp.test.ts
@@ -351,6 +351,7 @@ describe("ClaudecodeMcp", () => {
       expect(claudecodeMcp).toBeInstanceOf(ClaudecodeMcp);
       expect(claudecodeMcp.getJson()).toEqual({ mcpServers: {} });
       expect(claudecodeMcp.getFilePath()).toBe(join(testDir, ".claude/.claude.json"));
+      expect(claudecodeMcp.isDeletable()).toBe(false);
     });
 
     it("should preserve non-mcpServers properties in global mode", async () => {


dyoshikawa:
@zerone0x Thank you!

PR #1604, opened 2026-05-07
feat(skills): support paths in Claude Code skills

jornetsimon:


dyoshikawa-claw (commented):
This is a clean, well-tested addition. The schema, conversion, and round-trip all follow the existing patterns for `allowed-tools`, `model`, and `disable-model-invocation` exactly. Test coverage is thorough — string form, array form, validation rejects, and round-trip preservation are all covered.

One small thing: the `fromDir` method at line 201–225 already handles `paths` correctly since it delegates to the schema — but no test explicitly exercises importing a SKILL.md from disk with a `paths` field. The existing round-trip and conversion tests give good confidence, though.

Otherwise, looks good to merge.

src/features/skills/claudecode-skill.ts
```diff
@@ -24,6 +24,7 @@ export const ClaudecodeSkillFrontmatterSchema = z.looseObject({
   "allowed-tools": z.optional(z.array(z.string())),
   model: z.optional(z.string()),
   "disable-model-invocation": z.optional(z.boolean()),
+  paths: z.optional(z.union([z.string(), z.array(z.string())])),
```
dyoshikawa-claw, line 27: In `claudecode-rule.ts`, `paths` accepts only `string[]`, while here it accepts `string | string[]`. A brief comment explaining this difference would help future maintainers understand it is intentional, not an oversight.

dyoshikawa-claw (commented):
Clean addition that follows the existing patterns well. The schema, both conversion directions, and the rulesync type are all updated consistently. The test coverage for string/array variants in object conversion and round-trip is thorough.

Two things to consider adding:

- A `fromDir` test for `paths`, similar to the existing `should load skill with model` test, so that YAML parsing of comma-separated strings vs arrays is covered through the I/O path too.
- A test that combines `paths` with another claudecode field like `model` or `allowed-tools`, to verify they don't interfere with each other during the spread-based conversion.

PR #1607, opened 2026-05-08
chore: update opencode model selections and add deepseek option

dyoshikawa:
## Summary
- Update model identifiers in the `select-opencode-model` GitHub Action to newer versions (`gpt-5.4`, `glm-5.1`, `kimi-k2.6`)
- Add a new `/deepseek` keyword that selects `openrouter/deepseek/deepseek-v4-pro`
- Change the default model to `openrouter/deepseek/deepseek-v4-pro`

## Test plan
- [ ] Verify the action correctly selects the new model identifiers when triggered with each keyword
- [ ] Verify the default branch returns `openrouter/deepseek/deepseek-v4-pro` when no keyword matches


dyoshikawa:
@dyoshikawa Thank you!

PR #1602, opened 2026-05-06
fix(codexcli): emit project-root filesystem globs

zxyasfas:
## Summary
- Emit relative Codex CLI filesystem globs under the `:project_roots` named root instead of the top-level filesystem table.
- Preserve absolute paths, home paths, and existing named roots at the top level.
- Add `glob_scan_max_depth = 8` when project-root rules contain unbounded `**` patterns.
- Import nested Codex project-root filesystem rules back into Rulesync permissions and document the generated layout.

Fixes #1597

## Testing
- `npx pnpm@10.25.0 exec vitest run --silent=true src/features/permissions/codexcli-permissions.test.ts`
- `npx pnpm@10.25.0 exec oxfmt --check src/features/permissions/codexcli-permissions.ts src/features/permissions/codexcli-permissions.test.ts src/e2e/e2e-permissions.spec.ts docs/reference/file-formats.md skills/rulesync/file-formats.md`
- `npx pnpm@10.25.0 exec oxlint src/features/permissions/codexcli-permissions.ts src/features/permissions/codexcli-permissions.test.ts src/e2e/e2e-permissions.spec.ts --max-warnings 0`
- `npx pnpm@10.25.0 run typecheck`

Note: `npx pnpm@10.25.0 exec vitest run --config vitest.e2e.config.ts --silent=false src/e2e/e2e-permissions.spec.ts` currently fails before assertions in this Windows environment because the E2E helper spawns `node_modules/.bin/tsx` without the Windows `.cmd` shim.

PR #1621, opened 2026-05-11
fix(codexcli): address review findings from #1602

### LINKED ISSUE #1619: Review findings for #1602 (codexcli project-root filesystem globs)
## Background

PR #1602 (`fix(codexcli): emit project-root filesystem globs`) changes how Codex CLI filesystem permissions are generated. Relative file globs (`src/**`, `**/*.tf`) are now emitted under `:project_roots` instead of the top-level filesystem table, because Codex expects top-level filesystem keys to be absolute paths, home paths, or named roots. The PR also sets `glob_scan_max_depth = 8` when project-root rules contain `**` patterns, and supports importing nested Codex project-root rules back into Rulesync.

Two review passes (code review + security review) were performed. All findings are below.

### Code Review

| # | Severity | File | Line | Issue |
|---|----------|------|------|-------|
| 1 | Low | `codexcli-permissions.ts` | ~25 | `CodexFilesystem` type is `Record<string, Access \| Table \| number>` — any key can hold any value type. Runtime guards exist so no actual bug. |
| 2 | Low | `codexcli-permissions.ts` | `toFilesystemRecord` | A nested TOML table named `glob_scan_max_depth` could be mis-parsed as a rule table. `key === "glob_scan_max_depth"` should require a number, else skip. |
| 3 | Low | `skills/rulesync/file-formats.md` | — | Manual edit instead of auto-generation via `scripts/sync-skill-docs.ts`. |
| 4 | Low | `codexcli-permissions.test.ts` | 176 | `join(".codex", "rules")` change is good but unrelated to this PR. |
| 5 | Low | `codexcli-permissions.ts` | 292-300 | `canBeCodexFilesystemRoot` has overlapping `~` checks; `pattern.startsWith("~/")` already covers `~`. |
| 6 | Low | `docs/reference/file-formats.md` | 475 | No explanation of why `glob_scan_max_depth = 8`. |
| 7 | **Mid** | `codexcli-permissions.test.ts` | — | No negative test: `glob_scan_max_depth` should NOT be set when only single-level globs (`src/*`) exist. |
| 8 | Low | PR description | — | Missing issue link per `feature-change-guidelines.md`. |

### Security Review

| # | Severity | Issue |
|---|----------|-------|
| S-001 | Low | Key collision: if user defines `read: { ":project_roots": "allow" }`, the string value gets overwritten by the projectRootFilesystem object assignment. |
| S-002 | Low | Empty string `""` is not filtered and gets emitted under `:project_roots`. Codex interpretation of empty glob is unclear. |
| S-003 | Low | Named root context (`:home`, etc.) is lost during import — everything gets flattened into Rulesync format. |
| S-004 | Low | `glob_scan_max_depth` is only set when `**` patterns exist. Other project-root patterns get no depth limit. |
| S-005 | Low | `isCodexFilesystemRuleTable` returns true for empty tables (`every` vacuous truth). No practical harm. |
| S-006 | Low | Glob patterns are not sanitized before TOML serialization. `smol-toml` handles escaping but consistency with geminicli-permissions validation would be nice. |
| S-007 | Low | Windows UNC paths (`\\server\share`) misclassified as project-root on Unix. |

### Worth fixing now
- **#7 (Mid)**: Add a negative test case verifying `glob_scan_max_depth` is absent when project-root globs contain only single-level wildcards (`src/*`).

### Worth fixing eventually
- **S-001**: Add a guard in `addFilesystemRule` or after the loop to warn if `:project_roots` is already set as a string key before overwriting.
- **#6**: Add a brief comment in the source or docs linking the `8` value to Codex CLI defaults.
- **S-002**: Add validation to reject or warn on empty string patterns.

### Nice to have
- **#2, S-005**: Stricter guards around `glob_scan_max_depth` key handling in nested contexts.
- **S-006**: Align TOML output sanitization with the pattern used in geminicli-permissions.
- **#8**: Include issue link in future PR descriptions per repo guidelines.

dyoshikawa-claw:
## Summary
- Add negative test verifying `glob_scan_max_depth` is not set when project-root globs use only single-level wildcards (e.g., `src/*`)
- Guard against `:project_roots` key collision: warn when a direct string access rule would be overwritten by the project-root table
- Reject empty string patterns with a warning in `addFilesystemRule`
- Add comment linking `glob_scan_max_depth = 8` value to Codex CLI defaults

Closes #1619

### DIFF
--- a/src/features/permissions/codexcli-permissions.test.ts
+++ b/src/features/permissions/codexcli-permissions.test.ts
@@ -179,6 +179,61 @@ glob_scan_max_depth = 8
     expect(json.permission.edit?.["docs/**"]).toBe("allow");
   });
 
+  it("should warn when :project_roots is set as a direct string access rule", async () => {
+    const logger = createMockLogger();
+    const rulesyncPermissions = new RulesyncPermissions({
+      outputRoot: testDir,
+      relativeDirPath: ".rulesync",
+      relativeFilePath: "permissions.json",
+      fileContent: JSON.stringify({
+        permission: {
+          read: {
+            ":project_roots": "deny",
+            "src/**": "allow",
+          },
+        },
+      }),
+    });
+
+    await CodexcliPermissions.fromRulesyncPermissions({
+      outputRoot: testDir,
+      rulesyncPermissions,
+      logger,
+    });
+
+    expect(logger.warn).toHaveBeenCalledWith(
+      expect.stringContaining('":project_roots" is set as a direct filesystem access rule'),
+    );
+  });
+
+  it("should skip empty string patterns with a warning", async () => {
+    const logger = createMockLogger();
+    const rulesyncPermissions = new RulesyncPermissions({
+      outputRoot: testDir,
+      relativeDirPath: ".rulesync",
+      relativeFilePath: "permissions.json",
+      fileContent: JSON.stringify({
+        permission: {
+          read: {
+            "": "allow",
+            "src/**": "allow",
+          },
+        },
+      }),
+    });
+
+    const codexPermissions = await CodexcliPermissions.fromRulesyncPermissions({
+      outputRoot: testDir,
+      rulesyncPermissions,
+      logger,
+    });
+
+    expect(logger.warn).toHaveBeenCalledWith("Skipping empty pattern in filesystem permissions.");
+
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa-claw Thank you!

PR #1622, opened 2026-05-11
fix(codexcli-mcp): strip empty env tables from generated config

sirmacik:
## Summary

The codex generator currently emits an empty `[mcp_servers.X.env]` header for any server whose `env` is `{}`. Codex CLI 0.130+ rejects this on remote (sse/http/streamable_http) transports with:

```
Error loading config.toml: env is not supported for streamable_http
in `mcp_servers.X`
```

This is a hard startup failure for any source `mcp.json` that has a remote MCP server with an empty `env` (intentionally, or left over from another tool's schema shape).

## Repro

```jsonc
// .rulesync/mcp.json
{
  "mcpServers": {
    "aws-knowledge": {
      "type": "sse",
      "url": "https://knowledge-mcp.global.api.aws",
      "env": {}
    }
  }
}
```

After `rulesync generate -g`:

```toml
[mcp_servers.aws-knowledge]
type = "sse"
url = "https://knowledge-mcp.global.api.aws"

[mcp_servers.aws-knowledge.env]   ← codex rejects this
```

## Fix

Make `removeEmptyEntries` in `codexcli-mcp.ts` recursive so empty nested objects (`env: {}`) are stripped at any depth, not just at the top level. Arrays are preserved verbatim because an empty array can be a meaningful explicit override.

Side effect: stdio servers also get cleaner output (no more inert empty `[mcp_servers.X.env]` headers). Codex tolerates those, so this is purely cosmetic for stdio.

## Tests added

- `should strip empty env object from remote (sse) server` — repro test
- `should strip empty env object from stdio server (clean output)`
- `should preserve populated env table on stdio server`

All 54 codex MCP tests pass. `pnpm cicheck` is clean.

## Risk

Minimal — recursing into nested objects is a generalization of the existing behaviour (which already strips empty top-level server configs). No existing test relied on empty `env: {}` being preserved inside a server config.

### DIFF
--- a/src/features/mcp/codexcli-mcp.test.ts
+++ b/src/features/mcp/codexcli-mcp.test.ts
@@ -456,6 +456,99 @@ fontSize = 14
       expect(codexcliMcp.getToml().mcp_servers).toEqual({});
     });
 
+    it("should strip empty env object from remote (sse) server", async () => {
+      // codex CLI rejects empty `[mcp_servers.X.env]` for streamable_http
+      // transports (sse/http) with: "env is not supported for streamable_http".
+      const jsonData = {
+        mcpServers: {
+          "aws-knowledge": {
+            type: "sse",
+            url: "https://knowledge-mcp.global.api.aws",
+            env: {},
+          },
+        },
+      };
+      const rulesyncMcp = new RulesyncMcp({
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@sirmacik Thank you!

PR #1623, opened 2026-05-11
feat(codexcli-mcp): support codex-specific env_vars passthrough

sirmacik:
## Summary

Codex CLI supports a per-server array of shell env var names to inherit when launching the MCP server process. This is **distinct from `env`**: `env` is a literal `name → value` map; `envVars` is a list of names whose values come from the user's environment at runtime.

Use case: secrets like `OPENAI_API_KEY`, `JIRA_PERSONAL_TOKEN`, etc. that should NOT be literal-encoded into a committed `mcp.json`.

Before this PR, rulesync had no way to express this in the source schema. Users either lost their entries on every `rulesync generate` regen, or added them out-of-band, polluting other tools' configs.

### Source field: `envVars` (camelCase)

Matches the MCP schema's existing convention (`enabledTools`, `disabledTools`, `kiroAutoApprove`, `httpUrl`, `networkTimeout`, `alwaysAllow`, `kiroAutoBlock`). The codex generator renames it to snake_case `env_vars` at emit time, mirroring the established `enabledTools` → `enabled_tools` pattern (`convertToCodexFormat`).

### Strip is centralised in `getMcpServers()`

The field is stripped via `RulesyncMcp.getMcpServers()` so non-codex generators that go through that accessor never see it. The codex generator reads from `getJson()` (unfiltered) and is the only consumer.

### Generator audit + leak closure

Six generators were bypassing `getMcpServers()` and reading server configs directly from `getJson()` or `getFileContent()`. Without the migration in this PR they would have leaked `envVars` (and the pre-existing rulesync-only fields `targets`/`description`/`exposed`) into their tool outputs:

- `cline-mcp.ts` — was using `getFileContent()`
- `cursor-mcp.ts` — was using `getJson()`
- `factorydroid-mcp.ts` — was using `getJson()`
- `geminicli-mcp.ts` — was using `getJson()`
- `junie-mcp.ts` — was using `getFileContent()`
- `rovodev-mcp.ts` — was using `getJson()`

Each is migrated to `getMcpServers()`. `cline` and `junie` rebuild the file content from `getJson()` top-level + `getMcpServers()` to preserve `$schema` and other top-level fields. The cursor "complex configuration" test was updated — it had been asserting the buggy leaked-`targets` output.

`getMcpServers()` is also hardened to tolerate missing `mcpServers` (returns `{}` instead of throwing on `Object.entries(undefined)`) so callers no longer need the prior `isMcpServers` guard.

## Source example

```json
{
  "mcpServers": {
    "pal": {
      "type": "stdio",
      "command": "uvx",
      "args": ["--from", "git+https://github.com/BeehiveInnovations/pal-mcp-server.git", "pal-mcp-server"],
      "envVars": ["OPENAI_API_KEY", "OPENROUTER_API_KEY", "GEMINI_API_KEY"]
    }
  }
}
```

Generated `~/.codex/config.toml`:

```toml
[mcp_servers.pal]
type = "stdio"
command = "uvx"
args = ["--from", "git+https://github.com/BeehiveInnovations/pal-mcp-server.git", "pal-mcp-server"]
env_vars = ["OPENAI_API_KEY", "OPENROUTER_API_KEY", "GEMINI_API_KEY"]
```

Other tools' outputs (`~/.claude.json`, `~/.config/opencode/opencode.json`, `~/.config/kilo/kilo.json`, `~/.gemini/settings.json`, `.cursor/mcp.json`, `.cline/...`, `.junie/...`, `.droid/...`, `~/.rovodev/mcp.json`): `envVars` stripped, never written.

## Tests

- `codexcli`: source camelCase `envVars` → output snake_case `env_vars`; source key absent; defensive assertions on surviving fields
- `codexcli`: coexistence of `envVars` + `env` on the same server
- `codexcli`: round-trip codex `env_vars` → `toRulesyncMcp` → `envVars`
- `rulesync-mcp`: `envVars` is stripped from `getMcpServers()` output
- `rulesync-mcp`: `envVars` survives via `getJson()` for the codex generator
- `geminicli`: representative non-codex leak-prevention test — `envVars` is absent from gemini output

All 656 MCP tests pass. `pnpm cicheck` clean.

## Docs

Added new section to `docs/reference/file-formats.md` (synced into `skills/rulesync/file-formats.md`) covering the source vs codex naming, the `env` vs `envVars` distinction, the full enumeration of non-codex tools where it's stripped, and the precedence-deferral note.

## Risk

Low — additive change at the schema level. The leak-closure migrations are mechanical (one accessor swap per file). The cursor complex-config test update is the only test change that reflects a behaviour change (correctly stripping `targets`); all other changes are additive or strictly cleaner.

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -361,6 +361,31 @@ You can control which individual tools from an MCP server are enabled or disable
 - `enabledTools`: An array of tool names that should be explicitly enabled for this server.
 - `disabledTools`: An array of tool names that should be explicitly disabled for this server.
 
+### Codex-specific: pass shell env vars to MCP servers (`env_vars`)
+
+Codex CLI supports a per-server `env_vars` array that names shell environment variables to inherit when launching the MCP server process. This is distinct from `env` (which is a literal `{name: value}` map) — `env_vars` is a list of names whose values come from the user's environment.
+
+```json
+{
+  "mcpServers": {
+    "pal": {
+      "type": "stdio",
+      "command": "uvx",
+      "args": [
+        "--from",
+        "git+https://github.com/BeehiveInnovations/pal-mcp-server.git",
+        "pal-mcp-server"
+      ],
+      "env_vars": ["OPENAI_API_KEY", "OPENROUTER_API_KEY", "GEMINI_API_KEY"]
+    }
+  }
+}
+```
+
+- Emitted only into the codex CLI output (`~/.codex/config.toml` or project `.codex/config.toml`).
+- Stripped from `getMcpServers()` so it does not leak into other tools' generated configs (Claude Code, Kilo, OpenCode, Gemini CLI, etc. each ignore it).
+- Use this for secrets and API keys you do not want to literal-encode into a committed `mcp.json`.
+
 ## `.rulesync/.aiignore` or `.rulesyncignore`
 
 Rulesync supports a single ignore list that can live in either location below:
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -361,9 +361,11 @@ You can control which individual tools from an MCP server are enabled or disable
 - `enabledTools`: An array of tool names that should be explicitly enabled for this server.
 - `disabledTools`: An array of tool names that should be explicitly disabled for this server.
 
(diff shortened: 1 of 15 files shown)


dyoshikawa:
@sirmacik Thank you!

PR #1630, opened 2026-05-12
fix: harden codexcli-mcp converters against prototype pollution and add regression tests

### LINKED ISSUE #1626: Harden codexcli-mcp removeEmptyEntries: prototype pollution, recursion depth, and test coverage gaps
## Background

PR #1622 (https://github.com/dyoshikawa/rulesync/pull/1622) fixed a real bug in `src/features/mcp/codexcli-mcp.ts` by making `removeEmptyEntries` recursive so that nested empty objects (e.g. `env: {}`) are stripped before TOML serialization. This unblocks Codex CLI 0.130+ which rejects empty `[mcp_servers.X.env]` headers on remote (sse/http/streamable_http) transports with `env is not supported for streamable_http`.

The fix is correct and ships safely, but the review surfaced several follow-up items that are worth tracking separately so the codexcli-mcp generator stays robust as input shapes evolve and as Codex CLI's TOML schema changes.

## Details

Findings from the PR review (severity in parentheses):

### Mid

1. **Arrays-of-objects are not recursed into** (`src/features/mcp/codexcli-mcp.ts:191-197`)
   `Array.isArray(value)` deliberately preserves arrays verbatim, but if an array element is a plain object containing `env: {}`, the empty inline table will still appear in TOML. Codex's current `mcp_servers.*` schema is a map (not an array), so there is no live regression — but the boundary is implicit and undocumented in code, and would silently re-introduce the bug if Codex ever adds an array-shaped field.
   - Suggested action: add a regression test using e.g. `args: [{}]` to pin the documented behavior, and tighten the inline comment to spell out "array elements themselves are not recursed into".

2. **Tests assert on the parsed object only, not on the serialized TOML string** (`src/features/mcp/codexcli-mcp.test.ts:459-566`)
   The actual failure mode is smol-toml emitting a `[mcp_servers.X.env]` header that Codex rejects. The new tests verify the in-memory `getToml().mcp_servers` shape, which is a proxy. Asserting that the serialized output (e.g. `getFileContent()`) does not contain the substring `.env]` would map directly to the Codex 0.130+ rejection condition.

3. **Prototype pollution surface widened by the recursion** (`src/features/mcp/codexcli-mcp.ts:184-189`)
   `JSON.parse` preserves `__proto__` as an own property, and `Object.entries` enumerates it. The line `filtered[key] = cleaned` then assigns through the bracket setter, which mutates the prototype of `filtered` when `key === "__proto__"`. The same sink already exists in `convertToCodexFormat` (lines 47-70) and pre-dates this PR, so the recursion does not introduce a new vulnerability — but it does add another sink at every nesting level.
   - Threat model: rulesync is a local CLI reading the user's own `.rulesync/mcp.json`, so direct external attack is limited. The realistic vector is `git clone` + `rulesync generate` against an untrusted repo.
   - Suggested action: harden `removeEmptyEntries`, `convertToCodexFormat`, and `convertFromCodexFormat` by either using `Object.create(null)` for accumulator objects or skipping `__proto__` / `constructor` / `prototype` keys explicitly.

### Low

4. **Date / Map / Set / class instances misclassified as empty plain objects** (`codexcli-mcp.ts:193-197`)
   `typeof value === 'object' && !Array.isArray(value)` matches non-plain objects too, and `Object.entries(new Date())` returns `[]`, so they would be stripped. Inputs are always JSON-derived plain objects today, so no live impact — but an `isPlainObject` guard (`Object.getPrototypeOf(value) === Object.prototype || === null`) would future-proof this.

5. **`as Record<string, unknown>` cast and `eslint-disable`** (`codexcli-mcp.ts:194`)
   An `isRecord` helper would let the recursion narrow the type without disabling the lint rule. Surrounding code uses similar casts, so this is a stylistic improvement rather than a fix.

6. **Optional chaining weakens test assertions** (`codexcli-mcp.test.ts:494, 528, 562`)
   `expect(server?.env).toBeUndefined()` passes even when `server` itself is undefined. The tests do `expect(server).toBeDefined()` first so there is no real gap, but `expect(server!.env).toBeUndefined()` makes the intent clearer.

7. **Unbounded recursion can blow the stack on deeply nested input** (`codexcli-mcp.ts:177-193`)
   Self-DoS only — the user controls their own `mcp.json`. A depth cap (e.g. 32) or an iterative implementation would eliminate the risk.

8. **"Config-only" servers silently disappear** (`codexcli-mcp.ts:184-190`)
   A server entry like `{ disabled: false }` becomes an empty object after `convertToCodexFormat` drops `disabled: false`, and the recursive strip then removes the entire server from `mcp_servers`. The result is never a security boundary loss (such configs are non-functional), but the silent drop may surprise users.
   - Suggested action: emit a debug/log warning when a server entry collapses to empty.

## Solution / Next Steps

Recommended follow-up tasks, in priority order:

- [ ] Harden `removeEmptyEntries` and the surrounding converters in `src/features/mcp/codexcli-mcp.ts` against `__proto__` / `constructor` / `prototype` keys (finding #3). Apply the same fix to `convertToCodexFormat` and `convertFromCodexFormat` since the sink pre-exists there.
- [ ] Add a TOML-string-level regression test asserting that `getFileContent()` never contains `.env]` for a server with `env: {}` (finding #2).
- [ ] Add an `args: [{}]` style regression test and tighten the inline comment about "arrays are not recursed into" (finding #1).
- [ ] Replace `typeof value === 'object' && !Array.isArray(value)` with an `isPlainObject` guard (finding #4).
- [ ] Optional polish: introduce an `isRecord` type guard helper (finding #5), tighten test assertions to drop the optional chaining (finding #6), add a recursion depth cap (finding #7), and warn when a server entry collapses to empty (finding #8).

Findings #1–#3 are the highest leverage; #4–#8 can be bundled into a single follow-up PR if desired.

dyoshikawa-claw:
## Summary

Closes #1626

### Security hardening
- Skip `__proto__` / `constructor` / `prototype` keys in `removeEmptyEntries`, `convertToCodexFormat`, and `convertFromCodexFormat` to prevent prototype pollution
- Add `isPlainObject` guard to replace broad `typeof === 'object'` check, avoiding misclassification of Date/Map/Set as empty plain objects
- Add `isRecord` type guard helper to replace `as`-casts and `eslint-disable` comments

### Robustness
- Add recursion depth cap (`MAX_REMOVE_EMPTY_ENTRIES_DEPTH = 32`) to prevent stack overflow on deeply nested input
- Warn when a server entry collapses to empty after `removeEmptyEntries`

### Test coverage
- Add TOML-string-level regression test asserting `getFileContent()` never contains `.env]` for server with `env: {}`
- Add regression test for arrays-of-objects not recursed into (`args: [{}]`)
- Tighten test assertions: replace optional chaining with non-null assertion after `expect().toBeDefined()`

### Related
- PR #1622 (previous fix that made `removeEmptyEntries` recursive)

### DIFF
--- a/src/features/mcp/codexcli-mcp.test.ts
+++ b/src/features/mcp/codexcli-mcp.test.ts
@@ -488,10 +488,10 @@ fontSize = 14
           codexcliMcp.getToml().mcp_servers as Record<string, Record<string, unknown>>
         )?.["aws-knowledge"];
         expect(server).toBeDefined();
-        expect(server?.env).toBeUndefined();
+        expect(server!.env).toBeUndefined();
         // Defensive: assert non-env fields survive the strip.
-        expect(server?.type).toBe(transport);
-        expect(server?.url).toBe("https://knowledge-mcp.global.api.aws");
+        expect(server!.type).toBe(transport);
+        expect(server!.url).toBe("https://knowledge-mcp.global.api.aws");
       },
     );
 
@@ -522,10 +522,10 @@ fontSize = 14
         codexcliMcp.getToml().mcp_servers as Record<string, Record<string, unknown>>
       )?.["local"];
       expect(server).toBeDefined();
-      expect(server?.env).toBeUndefined();
+      expect(server!.env).toBeUndefined();
       // Defensive: assert non-env fields survive the strip.
-      expect(server?.command).toBe("node");
-      expect(server?.args).toEqual(["server.js"]);
+      expect(server!.command).toBe("node");
+      expect(server!.args).toEqual(["server.js"]);
     });
 
     it("should preserve populated env table on stdio server", async () => {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa-claw Thank you!

PR #1632, opened 2026-05-12
fix(kilo-subagent): default mode to 'all' to match Kilo CLI docs

sirmacik:
## Summary

Kilo CLI's documented default for user-defined agents is `all` (available both as a top-level pick AND as a subagent). See https://kilocode.ai/docs/customize/custom-modes — the `mode` reference states:

> `all` — Available both as a top-level pick and as a subagent (default for user-defined agents).

Rulesync previously defaulted to `subagent`, which hid every generated agent from Kilo's agent picker. Users had to either set `kilo: { mode: all }` explicitly in every source subagent, or post-process the output with sed.

After this PR, omitting `kilo.mode` writes `mode: all` — matching Kilo's own default for user-defined agents. The explicit override still wins, so users wanting subagent-only behaviour can opt in with `kilo: { mode: subagent }` in source frontmatter.

## Changes

- `src/features/subagents/kilo-subagent.ts`: change default from `"subagent"` → `"all"` in `fromRulesyncSubagent`. Added an inline comment linking to the Kilo docs reference.
- `src/features/subagents/kilo-subagent.test.ts`: updated the existing default-mode test to expect `mode: all`. Added regression test: explicit `kilo: { mode: subagent }` override is preserved.

## Tests

All 10 kilo-subagent tests pass. `pnpm cicheck` clean.

## Risk

Behaviour change for users who relied on the previous default. Opt-out path is a one-line frontmatter addition (`kilo: { mode: subagent }`). The new default matches Kilo's own documented default, so most users will see no behaviour change in practice — the generated agents simply become visible in the picker as Kilo's docs already say they should be.

### DIFF
--- a/src/e2e/e2e-subagents.spec.ts
+++ b/src/e2e/e2e-subagents.spec.ts
@@ -140,6 +140,61 @@ You are a primary agent. You appear in the Tab rotation.
     expect(generatedContent).toContain("A primary mode agent");
   });
 
+  it("should default kilo.mode to 'all' when omitted in source", async () => {
+    const testDir = getTestDir();
+
+    // Kilo's documented default for user-defined agents is `all`
+    // (https://kilocode.ai/docs/customize/custom-modes). Rulesync must
+    // emit `mode: all` when source frontmatter has no `kilo.mode`,
+    // otherwise generated agents are hidden from Kilo's agent picker.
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@sirmacik Thank you!

PR #1633, opened 2026-05-12
fix(claudecode-mcp): write global MCP to documented ~/.claude.json path

### LINKED ISSUE #1387: claudecode global MCP writes to wrong path (~/.claude/.claude.json vs ~/.claude.json)
## Bug

`rulesync generate --global --targets claudecode --features mcp` writes MCP config to `~/.claude/.claude.json`, but Claude Code reads global/user-scoped MCP servers from `~/.claude.json` (at HOME root, not inside `.claude/`).

## Reproduction

```bash
rulesync generate --global --targets claudecode --features mcp --dry-run --verbose
```

Output:
```
Generating files...
Base directories: /Users/tim
Generating MCP files...
[DRY RUN] Would write: /Users/tim/.claude/.claude.json
Successfully loaded 0 claudecode MCP files
[DRY RUN] Would write 1 MCP file
    .claude/.claude.json
```

## Expected

Output path should be `~/.claude.json`, not `~/.claude/.claude.json`.

## Reference

Claude Code MCP documentation confirms the correct global MCP path:

> User-scoped servers are stored in `~/.claude.json` and provide cross-project accessibility.

The settings hierarchy for Claude Code:

| Type | User/Global | Project |
|------|-------------|---------|
| MCP servers | `~/.claude.json` | `.mcp.json` |
| Settings | `~/.claude/settings.json` | `.claude/settings.json` |

## Secondary issue: --global source resolution

`--global` also changes the **source** directory to `$HOME`, looking for `$HOME/.rulesync/mcp.json` instead of the project directory's `.rulesync/mcp.json`. This means `--global` fails to find MCP source files unless they're manually placed in `$HOME/.rulesync/`.

## Versions

- rulesync: 7.20.0 (also reproduced on 7.23.0)
- Claude Code: latest
- macOS: Darwin 25.2.0

## Workaround

After `--global` generates, merge mcpServers from the project's `.rulesync/mcp.json` directly into `~/.claude.json` via a post-processing script.

sirmacik:
## Summary

This fixes Claude Code global MCP output to use the documented user-scope path: `~/.claude.json`.

Rulesync currently writes Claude Code global MCP to `~/.claude/.claude.json`. That path worked in some environments, but it does not match Claude Code's documented MCP location and makes the generated config harder to audit, automate against, or reason about.

Closes #1387.

## Why this needs care

Claude Code is widely used, including in enterprise setups where config files may be managed, audited, backed up, or inspected by automation. We should fix the path bug without introducing a destructive migration.

Earlier drafts considered deleting or archiving the legacy file to prevent possible stale/ghost MCP entries. I backed away from that after checking rulesync's own precedent: PR #333 introduced legacy path handling for `.rulesync/.mcp.json` by **reading the legacy path with a deprecation warning, but never modifying it**.

This PR follows that same pattern.

## Behavior after this change

- `fromRulesyncMcp({ global: true })` writes to the documented path:
  - `~/.claude.json`
- `fromFile({ global: true })` reads from:
  1. `~/.claude.json` first
  2. then falls back to `~/.claude/.claude.json` if the recommended path is absent
- When the legacy path is used, rulesync emits a deprecation warning.
- The legacy file is **never deleted, renamed, archived, merged, or modified**.

This keeps the migration conservative: rulesync fixes its output path while leaving any user-owned legacy file untouched.

## Notes for upgrading users

If you previously generated Claude Code global MCP with rulesync, you may have this legacy file:

```bash
~/.claude/.claude.json
```

After upgrading, rulesync writes to:

```bash
~/.claude.json
```

If you want to fully avoid any risk of Claude Code reading stale MCP entries from the legacy path, manually inspect and remove the old file after confirming your new config is correct:

```bash
rm ~/.claude/.claude.json
```

Rulesync intentionally does not do this automatically.

## Changes

- Update `ClaudecodeMcp.getSettablePaths({ global: true })` to return:
  - `relativeDirPath: "."`
  - `relativeFilePath: ".claude.json"`
- Add legacy fallback in `ClaudecodeMcp.fromFile`.
- Keep `fromRulesyncMcp` write behavior focused on the recommended path only.
- Add tests for:
  - recommended path priority
  - legacy fallback
  - deprecation warning
  - local mode not using legacy
  - malformed recommended / legacy JSON behavior
  - byte-for-byte preservation of the legacy file
  - read-modify-write preservation of unrelated `~/.claude.json` keys
- Update E2E coverage for the global Claude Code MCP output path.

## Verification

- `pnpm test src/features/mcp/claudecode-mcp.test.ts`
- `pnpm test:e2e src/e2e/e2e-mcp.spec.ts`
- `pnpm cicheck`

All pass locally.

## Risk

Low-to-medium.

The path change is straightforward and matches Claude Code documentation. The main risk is upgrade behavior for users who already have the legacy file. To minimize that risk, this PR deliberately avoids destructive actions and follows the existing rulesync legacy-path precedent from PR #333.

### DIFF
--- a/src/e2e/e2e-mcp.spec.ts
+++ b/src/e2e/e2e-mcp.spec.ts
@@ -5,7 +5,7 @@ import { setTimeout } from "node:timers/promises";
 import { describe, expect, it } from "vitest";
 
 import { RULESYNC_MCP_RELATIVE_FILE_PATH } from "../constants/rulesync-paths.js";
-import { readFileContent, writeFileContent } from "../utils/file.js";
+import { fileExists, readFileContent, writeFileContent } from "../utils/file.js";
 import {
   runGenerate,
   runImport,
@@ -221,7 +221,7 @@ describe("E2E: mcp (global mode)", () => {
   const { getProjectDir, getHomeDir } = useGlobalTestDirectories();
 
   it.each([
-    { target: "claudecode", outputPath: join(".claude", ".claude.json") },
+    { target: "claudecode", outputPath: ".claude.json" },
     { target: "cursor", outputPath: join(".cursor", "mcp.json") },
     { target: "geminicli", outputPath: join(".gemini", "settings.json") },
     { target: "opencode", outputPath: join(".config", "opencode", "opencode.jsonc") },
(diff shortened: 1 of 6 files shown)


sirmacik:
I consider this PR ready from my side. I’m keeping it in draft only to respect the current two-PR contributor limit while #1631 and #1632 are still in flight. Once one of those lands and a review slot opens, I’ll mark this ready for review.

dyoshikawa:
@sirmacik Thank you!

PR #1635, opened 2026-05-13
fix(kilo-subagent): align read defaults

### LINKED ISSUE #1634: PR #1632 follow-ups: kilo subagent default mode change
## Background

PR #1632 (https://github.com/dyoshikawa/rulesync/pull/1632) changes the Kilo subagent default `mode` from `"subagent"` to `"all"` in `src/features/subagents/kilo-subagent.ts` to match Kilo CLI's documented default for user-defined agents (https://kilocode.ai/docs/customize/custom-modes). The PR is small and CI is green, but the review surfaced several follow-up items worth tracking separately so the PR itself can ship without scope creep.

This issue consolidates those follow-ups.

### 1. (High) Asymmetry between write-default (`"all"`) and read-default (`"subagent"`)
The PR updates the write-time default in `KiloSubagent.fromRulesyncSubagent`, but the shared schema in `src/features/subagents/opencode-style-subagent.ts:14` still has:

```ts
mode: z._default(z.string(), "subagent"),
```

This schema is re-used by `KiloSubagentFrontmatterSchema` (`src/features/subagents/kilo-subagent.ts:22`). Consequences:

- `KiloSubagent.fromFile` reading an existing `.kilo/agent/*.md` that omits `mode` will coerce it to `"subagent"` (existing test `should apply default mode 'subagent' when mode is omitted` at `kilo-subagent.test.ts:207-225` encodes this).
- `KiloSubagent.forDeletion` (`kilo-subagent.ts:117`) still hard-codes `mode: "subagent"`. Currently harmless because the value is unused for deletion, but it is an orphan reference to the old default.
- An `import` → re-export round trip for a file without `mode` would flip back to `subagent` on import, then to `all` on the next regenerate.

### 2. (Mid) Symmetry test in `kilo-subagent.test.ts:165-191` is largely vacuous
The new test `should honour explicit kilo.mode = 'all' override (matches default)` cannot fail under the current implementation: the `typeof kiloSection.mode === "string" ? kiloSection.mode : "all"` ternary returns `"all"` whether the user wrote `"all"` or omitted it. The intent (guard against future regression where the override is dropped) is reasonable, but as written it would also pass against a literal `mode: "all"` hardcode.

### 3. (Mid) Breaking change is under-signalled
The commit prefix is `fix(...)` but for users who already rely on the previous default, this is a behaviour change in generated output. Existing Kilo subagents will appear in the agent picker on the next `rulesync generate`.

- Consider a `BREAKING CHANGE:` footer or `fix!:`/`feat!:` per Conventional Commits so it surfaces in release notes / changelog automation.
- `docs/reference/file-formats.md:228` (subagent frontmatter example block) currently only shows `opencode:`. Adding a one-liner that Kilo defaults to `all` and how to opt back into `subagent` would prevent surprise.

### 4. (Low) URL host inconsistency: `kilocode.ai` vs `kilo.ai`
PR #1632 comments and tests use `https://kilocode.ai/docs/customize/custom-modes`, while the in-repo reference file `.rulesync/skills/rulesync-feature-research/references/kilo.md` uses `https://kilo.ai/docs/...` everywhere. Both resolve, but mixing hosts is a grep/maintenance footgun. Pick one — the existing project convention is `kilo.ai` — and standardise.

### 5. (Low) Comment duplication between source and tests
The same rationale is repeated in three places:
- `src/features/subagents/kilo-subagent.ts:53-61`
- `src/features/subagents/kilo-subagent.test.ts:100-105`
- `src/e2e/e2e-subagents.spec.ts:143-147`

Keep the canonical explanation in the source file and replace the test-side blurbs with a one-line pointer to `kilo-subagent.ts`.

## Solution / Next Steps

- **Finding #1**: Either split the schema default per subclass (Kilo defaults `"all"`, OpenCode keeps `"subagent"`), or at minimum align `forDeletion` and add an explanatory comment in `opencode-style-subagent.ts` so the next reader does not unify the schema default to `"all"` for both tools by mistake.
- **Finding #2**: Replace the symmetry test with one that exercises a non-default value (e.g. assert that `kilo.temperature` or another `kilo.*` field round-trips alongside `mode`), or drop it as documentation.
- **Finding #3**: Add a note to `docs/reference/file-formats.md` documenting Kilo's per-tool default and the opt-out frontmatter. Decide whether the release including PR #1632 should be tagged as a breaking change.
- **Finding #4**: Replace `kilocode.ai` references in the codebase with `kilo.ai` to match the existing convention.
- **Finding #5**: Move the canonical rationale into `kilo-subagent.ts` and reduce test-side comments to a single-line `// see kilo-subagent.ts` pointer.

## References
- PR: https://github.com/dyoshikawa/rulesync/pull/1632
- Kilo CLI mode docs: https://kilo.ai/docs/customize/custom-modes

sirmacik:
## Summary

Draft follow-up for #1634. This addresses the Kilo subagent default-mode cleanup from the maintainer-scrap issue.

## Changes

- Give Kilo its own subagent frontmatter schema with `mode` defaulting to `all`.
- Route generated Kilo frontmatter through that schema so the default is centralized.
- Align `fromFile` and deletion placeholders with the same default.
- Strengthen the explicit `kilo.mode: all` test so it also proves other Kilo-specific fields survive.
- Document Kilo's default and the `kilo.mode: subagent` opt-out in file-format docs and synced skill docs.
- Update Kilo docs links from `kilocode.ai` to `kilo.ai`.

Fixes #1634.

## Verification

- `pnpm test src/features/subagents/kilo-subagent.test.ts`
- `pnpm test:e2e src/e2e/e2e-subagents.spec.ts`
- `pnpm cicheck`

## Review notes

Opened as draft to signal ownership of the maintainer-scrap follow-up while keeping review pressure low. Local regression-gate review and PAL review were run; the only follow-up identified is a broader future cleanup to separate per-target subagent schemas from the shared OpenCode-style base constructor.

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -235,6 +235,8 @@ opencode: # for OpenCode-specific parameters
   permission:
     bash:
       "git diff": allow
+kilo: # for Kilo-specific parameters
+  mode: all # (optional, defaults to "all") use "subagent" for hidden/subagent-only agents
 takt: # takt specific parameters (optional; emitted under .takt/facets/personas/)
   name: "renamed-stem" # (optional) override the emitted filename stem (no path separators or "..")
 ---
@@ -247,6 +249,8 @@ Attention, again, you are just the planner, so though you can read any files and
 ```
 
 > **Gemini CLI note (as of 2026-04-01):** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
+
+> **Kilo note (as of 2026-05-13):** Kilo's documented default for user-defined agents is `mode: all`, which makes the agent available both as a top-level pick and as a subagent. Set `kilo.mode: subagent` to opt into hidden/subagent-only behavior.
 
 ## `.rulesync/skills/*/SKILL.md`
 
--- a/skills/rulesync/file-formats.md
+++ b/skills/rulesync/file-formats.md
@@ -235,6 +235,8 @@ opencode: # for OpenCode-specific parameters
   permission:
     bash:
       "git diff": allow
+kilo: # for Kilo-specific parameters
+  mode: all # (optional, defaults to "all") use "subagent" for hidden/subagent-only agents
 takt: # takt specific parameters (optional; emitted under .takt/facets/personas/)
   name: "renamed-stem" # (optional) override the emitted filename stem (no path separators or "..")
 ---
@@ -247,6 +249,8 @@ Attention, again, you are just the planner, so though you can read any files and
 ```
 
 > **Gemini CLI note (as of 2026-04-01):** Subagents are generated to `.gemini/agents/`. To enable the agents feature, set `"experimental": { "enableAgents": true }` in your `.gemini/settings.json`.
+
+> **Kilo note (as of 2026-05-13):** Kilo's documented default for user-defined agents is `mode: all`, which makes the agent available both as a top-level pick and as a subagent. Set `kilo.mode: subagent` to opt into hidden/subagent-only behavior.
 
 ## `.rulesync/skills/*/SKILL.md`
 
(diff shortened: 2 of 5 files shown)


dyoshikawa:
@sirmacik Thank you!

PR #1637, opened 2026-05-13
fix(codexcli-mcp): tighten envVars handling

### LINKED ISSUE #1625: Follow-up: harden codex MCP generator and tighten envVars validation (from PR #1623 review)
## Background

PR [#1623](https://github.com/dyoshikawa/rulesync/pull/1623) added Codex CLI-specific `envVars` (camelCase) support to the MCP server schema. The codex generator renames it to `env_vars` (snake_case) in the emitted `config.toml`. Six other generators (cline / cursor / factorydroid / geminicli / junie / rovodev) were migrated from `getJson()`/`getFileContent()` to `getMcpServers()` to close leaks of rulesync-only fields (`targets`/`description`/`exposed`) and the new codex-only `envVars`.

A combined code review and security review was run on the PR. No high/critical issues were found and the PR was merged-quality, but several low/mid follow-ups surfaced that are out of scope for the original PR yet worth tracking separately.

### Code-quality follow-ups

1. **[mid] Codex generator still leaks `targets`/`description`/`exposed` into its own output.** The codex generator deliberately bypasses the `getMcpServers()` strip in order to preserve `envVars`. As a side effect, the other rulesync-only fields (`targets`, `description`, `exposed`) — which are stripped from every other tool's output — still flow into `~/.codex/config.toml`. Pre-PR these leaked everywhere; post-PR they leak only in codex, but they should leak nowhere. Suggested approach: extend `getMcpServers()` with an optional `exclude`/`keep` parameter so the codex generator can request "strip everything except `envVars`", or have the codex generator read the stripped result and re-merge `envVars` per-server from `getJson()`.
   - File: `src/features/mcp/codexcli-mcp.ts` (the `rulesyncMcp.getJson().mcpServers` path inside `fromRulesyncMcp`).

2. **[mid] `convertFromCodexFormat` / `convertToCodexFormat` if-else chain is getting long.** Both functions now contain three parallel rename branches (`enabledTools`/`enabled_tools`, `disabledTools`/`disabled_tools`, `envVars`/`env_vars`) plus the `disabled`/`enabled` boolean inversion. Adding a fourth field would require edits in two places. Suggested approach: extract a single key-mapping table such as `const CODEX_FIELD_MAP: Record<string,string> = { enabledTools: "enabled_tools", disabledTools: "disabled_tools", envVars: "env_vars" }` and iterate; keep the `disabled`↔`enabled` inversion as a named special case.
   - File: `src/features/mcp/codexcli-mcp.ts:18-89`.

3. **[low] Cursor test does not assert that `targets` is preserved at the source level.** The updated `cursor-mcp.test.ts` correctly asserts `targets` is absent from cursor output, but does not document that `targets` is still in the rulesync source — making it ambiguous whether the field is stripped at emit-time or removed entirely. A one-line `expect((rulesyncMcp.getJson() as any).mcpServers["complex-server"].targets).toBeDefined()` would make the intent explicit.
   - File: `src/features/mcp/cursor-mcp.test.ts:954+`.

4. **[low] Schema field ordering for `envVars`.** In `McpServerSchema`, `envVars` is currently inserted between `kiroAutoBlock` and `headers`, which puts it adjacent to kiro-specific fields rather than near `env` (its semantic sibling) or `enabledTools`/`disabledTools` (its naming-convention precedent referenced in the field comment). Pure cosmetic.
   - File: `src/types/mcp.ts:19-29`.

5. **[low] Precedence note in docs is vague.** `docs/reference/file-formats.md` ends with "the codex CLI behavior is upstream-defined — see codex documentation for the exact resolution rule" with no link or last-checked-date.
   - Files: `docs/reference/file-formats.md:399`, `skills/rulesync/file-formats.md` (kept in sync).

### Security-hardening follow-ups

6. **[low] No contract test enforces that every non-codex generator routes through `getMcpServers()`.** PR #1623 migrated six generators and the others were spot-checked at review time as already using `getMcpServers()`, but there is nothing preventing a future generator from regressing and leaking env var names. Impact is informational disclosure only (names, not values). Suggested approach: a single parametric test that, for every registered tool MCP class except `CodexcliMcp`, asserts `envVars`/`targets`/`description`/`exposed` are absent in the emitted output for a sample input.
   - File: `src/features/mcp/rulesync-mcp.ts:156` (the strip list).

7. **[low] codex TOML → rulesync round-trip lacks element-type validation.** `convertFromCodexFormat` copies `env_vars` straight into `converted["envVars"] = value` without verifying it is `string[]`. A malformed `config.toml` (e.g., `env_vars = [1, 2, 3]`) would propagate through `z.looseObject` and end up serialized into other tools' configs on the next `generate`. No code execution, but config-corruption potential. Same lax pattern exists for `enabled_tools`/`disabled_tools` and predates this PR.
   - File: `src/features/mcp/codexcli-mcp.ts:21-44`.

8. **[low] No character-set restriction on env var names.** `envVars: z.optional(z.array(z.string()))` accepts arbitrary strings. TOML injection is **not** exploitable (`smol-toml`'s `stringify` quotes/escapes correctly — verified during review). Tightening to `z.array(z.string().regex(/^[A-Za-z_][A-Za-z0-9_]*$/))` would document intent and harden against accidental misuse where the name is interpreted by a downstream shell.
   - File: `src/types/mcp.ts:27`.

## Solution / Next Steps

Pick up these items independently — they do not block any other work:

- [ ] **#1 (mid)**: Re-route the codex generator's server fetch through `getMcpServers()` so `targets`/`description`/`exposed` no longer leak into `~/.codex/config.toml`. Decide between (a) adding an `exclude`/`keep` parameter on `getMcpServers()` or (b) merging `envVars` back from `getJson()` per-server. Add a regression test asserting these fields are absent in codex output.
- [ ] **#2 (mid)**: Refactor `convertFromCodexFormat` / `convertToCodexFormat` to use a shared field-mapping table; keep the `disabled`↔`enabled` inversion as a named exception.
- [ ] **#3 (low)**: Add the source-side `targets` preservation assertion to `cursor-mcp.test.ts`.
- [ ] **#4 (low)**: Move `envVars` in `McpServerSchema` next to `env` or `enabledTools` for discoverability.
- [ ] **#5 (low)**: Link to the codex CLI documentation page (or note a verified codex version) for the env/envVars precedence rule in `docs/reference/file-formats.md` and re-sync `skills/rulesync/file-formats.md`.
- [ ] **#6 (low)**: Add a parametric contract test ensuring every non-codex generator strips `envVars`/`targets`/`description`/`exposed`.
- [ ] **#7 (low)**: Validate element types in `convertFromCodexFormat` for `enabled_tools`/`disabled_tools`/`env_vars` to prevent malformed TOML round-trips.
- [ ] **#8 (low)**: Tighten the `envVars` schema with a POSIX env-var-name regex.

These can be bundled or split into smaller PRs as convenient. #1 and #2 are the highest-leverage items; the rest are quality-of-life improvements.

sirmacik:
## Summary

Draft follow-up for #1625. This tightens Codex MCP `envVars` handling and prevents rulesync-only source fields from leaking into Codex output.

## Changes

- Generate Codex MCP from `RulesyncMcp.getMcpServers()` stripped output, then re-merge only valid codex-specific `envVars` from the raw source.
- Replace repeated field rename branches with shared mapping tables for `enabledTools` / `disabledTools` / `envVars`.
- Ignore malformed `env_vars`, `enabled_tools`, and `disabled_tools` arrays when importing Codex TOML.
- Validate `envVars` entries as environment variable names in the Rulesync MCP schema.
- Add regression tests proving Codex output keeps `env_vars` but strips `targets`, `description`, and `exposed`.
- Link docs to the Codex configuration reference with a last-checked date and sync `skills/rulesync` docs.

Fixes #1625.

## Verification

- `pnpm test src/features/mcp/codexcli-mcp.test.ts src/types/mcp.test.ts`
- `pnpm cicheck`

## Deferred

Two cross-generator follow-ups from #1625 are intentionally deferred to keep this PR focused:

- Cursor source-preservation assertion for `targets`.
- Parametric contract test across every non-Codex MCP generator for `envVars` / `targets` / `description` / `exposed` stripping.

Those are valuable, but they touch broader generator behavior and should stay separate from the Codex-specific cleanup.

## Review notes

Opened as draft to signal ownership and keep the maintainer-scrap follow-up visible while avoiding immediate review pressure.

### DIFF
--- a/src/features/mcp/codexcli-mcp.ts
+++ b/src/features/mcp/codexcli-mcp.ts
@@ -52,10 +52,7 @@ function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
           if (isStringArray(value)) {
             converted[mappedKey] = value;
           } else {
-            warnWithFallback(
-              undefined,
-              `Ignored malformed array for ${key} in MCP server ${name}`,
-            );
+            warnWithFallback(undefined, `Ignored malformed array for ${key} in MCP server ${name}`);
           }
         }
       } else {
--- a/src/features/mcp/codexcli-mcp.ts
+++ b/src/features/mcp/codexcli-mcp.ts
@@ -17,6 +17,18 @@ import {
   ToolMcpSettablePaths,
 } from "./tool-mcp.js";
 
+const CODEX_TO_RULESYNC_FIELD_MAP: Record<string, string> = {
+  enabled_tools: "enabledTools",
+  disabled_tools: "disabledTools",
+  env_vars: "envVars",
+};
+
+const RULESYNC_TO_CODEX_FIELD_MAP: Record<string, string> = {
+  enabledTools: "enabled_tools",
+  disabledTools: "disabled_tools",
+  envVars: "env_vars",
+};
+
 const MAX_REMOVE_EMPTY_ENTRIES_DEPTH = 32;
 
 const PROTOTYPE_POLLUTION_KEYS = new Set(["__proto__", "constructor", "prototype"]);
@@ -27,12 +39,15 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
   return proto === null || proto === Object.prototype;
 }
 
+function isStringArray(value: unknown): value is string[] {
+  return Array.isArray(value) && value.every((item) => typeof item === "string");
+}
+
 function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
   const result: McpServers = {};
 
   for (const [name, config] of Object.entries(codexMcp)) {
-    if (PROTOTYPE_POLLUTION_KEYS.has(name)) continue;
-    if (!isRecord(config)) continue;
+    if (PROTOTYPE_POLLUTION_KEYS.has(name) || !isRecord(config)) continue;
 
     const converted: Record<string, unknown> = {};
     for (const [key, value] of Object.entries(config)) {
@@ -41,15 +56,11 @@ function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
         if (value === false) {
           converted["disabled"] = true;
         }
-      } else if (key === "enabled_tools") {
-        converted["enabledTools"] = value;
-      } else if (key === "disabled_tools") {
-        converted["disabledTools"] = value;
-      } else if (key === "env_vars") {
-        // codex stores env-var passthrough names in snake_case (`env_vars`);
-        // the rulesync source schema uses camelCase (`envVars`) for
-        // consistency with `enabledTools`/`disabledTools`/etc.
-        converted["envVars"] = value;
+      } else if (key in CODEX_TO_RULESYNC_FIELD_MAP) {
+        const mappedKey = CODEX_TO_RULESYNC_FIELD_MAP[key];
+        if (mappedKey && isStringArray(value)) {
+          converted[mappedKey] = value;
+        }
       } else {
         converted[key] = value;
       }
@@ -74,16 +85,11 @@ function convertToCodexFormat(mcpServers: McpServers): Record<string, unknown> {
         if (value === true) {
           converted["enabled"] = false;
         }
-      } else if (key === "enabledTools") {
-        converted["enabled_tools"] = value;
-      } else if (key === "disabledTools") {
-        converted["disabled_tools"] = value;
-      } else if (key === "envVars") {
-        // Rename camelCase source `envVars` → snake_case `env_vars`
-        // for codex's native config.toml format. See `enabledTools`
-        // precedent above. `envVars` itself is stripped from
-        // getMcpServers() so non-codex tools never receive it.
-        converted["env_vars"] = value;
+      } else if (key in RULESYNC_TO_CODEX_FIELD_MAP) {
+        const mappedKey = RULESYNC_TO_CODEX_FIELD_MAP[key];
+        if (mappedKey && isStringArray(value)) {
+          converted[mappedKey] = value;
+        }
       } else {
         converted[key] = value;
       }
@@ -170,8 +176,23 @@ export class CodexcliMcp extends ToolMcp {
 
     const configToml = smolToml.parse(configTomlFileContent);
 
-    const mcpServers = rulesyncMcp.getJson().mcpServers;
-    const converted = convertToCodexFormat(mcpServers);
+    const strippedMcpServers = rulesyncMcp.getMcpServers();
+    const rawMcpServers = rulesyncMcp.getJson().mcpServers;
+    const mcpServersWithCodexFields = Object.fromEntries(
+      Object.entries(strippedMcpServers).map(([serverName, serverConfig]) => {
+        const rawServer = isRecord(rawMcpServers) ? rawMcpServers[serverName] : undefined;
+        return [
+          serverName,
+          {
+            ...serverConfig,
+            ...(isRecord(rawServer) && isStringArray(rawServer.envVars)
+              ? { envVars: rawServer.envVars }
+              : {}),
+          },
+        ];
+      }),
+    );
+    const converted = convertToCodexFormat(mcpServersWithCodexFields);
     const filteredMcpServers = this.removeEmptyEntries(converted);
 
     for (const name of Object.keys(converted)) {
--- a/src/features/mcp/codexcli-mcp.ts
+++ b/src/features/mcp/codexcli-mcp.ts
@@ -46,10 +46,17 @@ function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
         if (value === false) {
           converted["disabled"] = true;
         }
-      } else if (key in CODEX_TO_RULESYNC_FIELD_MAP) {
+      } else if (Object.hasOwn(CODEX_TO_RULESYNC_FIELD_MAP, key)) {
         const mappedKey = CODEX_TO_RULESYNC_FIELD_MAP[key];
-        if (mappedKey && isStringArray(value)) {
-          converted[mappedKey] = value;
+        if (mappedKey) {
+          if (isStringArray(value)) {
+            converted[mappedKey] = value;
+          } else {
+            warnWithFallback(
+              undefined,
+              `Ignored malformed array for ${key} in MCP server ${name}`,
+            );
+          }
         }
       } else {
         converted[key] = value;
@@ -78,7 +85,7 @@ function convertToCodexFormat(
         if (value === true) {
           converted["enabled"] = false;
         }
-      } else if (key in RULESYNC_TO_CODEX_FIELD_MAP) {
+      } else if (Object.hasOwn(RULESYNC_TO_CODEX_FIELD_MAP, key)) {
         const mappedKey = RULESYNC_TO_CODEX_FIELD_MAP[key];
         if (mappedKey) {
           if (isStringArray(value)) {
--- a/src/types/mcp.test.ts
+++ b/src/types/mcp.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 
-import { isMcpServers } from "./mcp.js";
+import { McpServerSchema, isMcpServers } from "./mcp.js";
 
 describe("isMcpServers", () => {
   it("should return true for a plain object", () => {
@@ -19,5 +19,25 @@ describe("isMcpServers", () => {
     expect(isMcpServers("x")).toBe(false);
     expect(isMcpServers(1)).toBe(false);
     expect(isMcpServers(true)).toBe(false);
+  });
+});
+
+describe("McpServerSchema", () => {
+  it("should accept valid envVars names", () => {
+    const result = McpServerSchema.safeParse({
+      command: "node",
+      envVars: ["OPENAI_API_KEY", "_PRIVATE", "KEY_123"],
+    });
+
+    expect(result.success).toBe(true);
+  });
+
+  it("should reject invalid envVars names", () => {
+    const result = McpServerSchema.safeParse({
+      command: "node",
+      envVars: ["1BAD", "HAS-DASH", ""],
+    });
+
+    expect(result.success).toBe(false);
   });
 });
--- a/src/types/mcp.ts
+++ b/src/types/mcp.ts
@@ -1,4 +1,13 @@
-import { z } from "zod/mini";
+import { refine, z } from "zod/mini";
+
+const EnvVarNameSchema = z
+  .string()
+  .check(
+    refine(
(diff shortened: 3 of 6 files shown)


dyoshikawa (commented):
Approach looks right — strip everything via `getMcpServers()`, then re-introduce only the codex-specific allow-listed field. The regression tests cover the leak scenarios well, and the schema validation for env var names is reasonable.

Main thing to address is the rebase: `isPlainObject` was just consolidated to `src/utils/type-guards.ts` on main (#1643), so the local copy here will need to go. A few smaller nits inline.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -27,12 +39,15 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
   return proto === null || proto === Object.prototype;
 }
 
+function isStringArray(value: unknown): value is string[] {
```
dyoshikawa: Heads up: `main` recently consolidated `isPlainObject` into `src/utils/type-guards.ts` (commit 09db6b2b / #1643). After rebasing, the local copy just above should be removed and imported from there. Worth hoisting `isStringArray` into the same module while you're at it — it's a general-purpose guard that other generators are likely to want.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -74,16 +85,11 @@ function convertToCodexFormat(mcpServers: McpServers): Record<string, unknown> {
...
-        converted["disabled_tools"] = value;
-      } else if (key === "envVars") {
-        // Rename camelCase source `envVars` → snake_case `env_vars`
-        // for codex's native config.toml format. See `enabledTools`
-        // precedent above. `envVars` itself is stripped from
-        // getMcpServers() so non-codex tools never receive it.
-        converted["env_vars"] = value;
+      } else if (key in RULESYNC_TO_CODEX_FIELD_MAP) {
+        const mappedKey = RULESYNC_TO_CODEX_FIELD_MAP[key];
+        if (mappedKey && isStringArray(value)) {
```
dyoshikawa: On the import path (TOML → rulesync) silently dropping non-string-array values is defensible — malformed external TOML shouldn't break things. But this code path also runs on the *generate* side via `convertToCodexFormat`, where the source is the user's own rulesync file and Zod has already validated the shape. If a wrong-shape value somehow slipped through, the field will just vanish from the generated `config.toml` with no signal. Consider a `logger.warn` for the unexpected-shape branch, or splitting validation between the two directions.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -170,8 +176,23 @@ export class CodexcliMcp extends ToolMcp {
...
-    const converted = convertToCodexFormat(mcpServers);
+    const strippedMcpServers = rulesyncMcp.getMcpServers();
+    const rawMcpServers = rulesyncMcp.getJson().mcpServers;
+    const mcpServersWithCodexFields = Object.fromEntries(
+      Object.entries(strippedMcpServers).map(([serverName, serverConfig]) => {
+        const rawServer = isRecord(rawMcpServers) ? rawMcpServers[serverName] : undefined;
+        return [
+          serverName,
+          {
+            ...serverConfig,
```
dyoshikawa, line 191: A short comment here would save future readers some time: only `envVars` is re-merged because everything else codex-specific is already preserved in `strippedMcpServers`, and adding more keys here would re-introduce the very leak this PR fixes. Without that hint someone is going to look at this pattern and think the same approach should apply to other rulesync-only fields like `targets`.

src/types/mcp.test.ts
```diff
@@ -21,3 +21,23 @@ describe("isMcpServers", () => {
...
+      envVars: ["OPENAI_API_KEY", "_PRIVATE", "KEY_123"],
+    });
+
+    expect(result.success).toBe(true);
+  });
+
+  it("should reject invalid envVars names", () => {
+    const result = McpServerSchema.safeParse({
+      command: "node",
+      envVars: ["1BAD", "HAS-DASH", ""],
```
dyoshikawa: Bundling three failure modes into one `safeParse` works, but if any single one regresses (say someone relaxes the leading-digit check), the test still fails on the others and the signal is lost. Worth splitting into one assertion per case. Also consider adding a happy-path for lowercase names (`openai_api_key`) since the regex allows them — locks in the POSIX-compliance choice explicitly.

sirmacik:
Thanks for the review! I've resolved the merge conflict and addressed all of your feedback:

- Removed the local `isPlainObject` and imported it from `src/utils/type-guards.ts`.
- Hoisted `isStringArray` into `src/utils/type-guards.ts` as well.
- Added a `logger.warn` (via `warnWithFallback`) in `convertToCodexFormat` if an unexpected non-string-array slips through.
- Added an inline comment explaining why only `envVars` needs manual re-merging (to prevent future readers from leaking other fields).
- Split the `McpServerSchema` tests into individual assertions and added a happy-path test for lowercase POSIX names.

Everything should be green now.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -55,7 +66,10 @@ function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
   return result;
 }
 
-function convertToCodexFormat(mcpServers: McpServers): Record<string, unknown> {
+function convertToCodexFormat(
```
dyoshikawa, line 69: Could you use a static `import type { Logger } from "../../utils/logger.js"` at the top of the file instead of an inline `import()` here? The sibling files (`tool-mcp.ts`, `rulesync-mcp.ts`, `mcp-processor.ts`) all use a static import for `Logger`, and the project coding guidelines favor static imports.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -164,8 +180,25 @@ export class CodexcliMcp extends ToolMcp {
...
+            ...serverConfig,
+            // Only envVars needs manual re-merging here. Other codex-specific fields
+            // (like disabledTools) are preserved by RulesyncMcp's filtering natively.
+            ...(isRecord(rawServer) && isStringArray(rawServer.envVars)
+              ? { envVars: rawServer.envVars }
+              : {}),
+          },
+        ];
+      }),
+    );
```
dyoshikawa, line 200: The `logger` argument here is always `undefined`, and `convertFromCodexFormat` was not given a logger parameter even though it now also warns via `warnWithFallback(undefined, ...)`. As-is, the parameter adds API surface without any behavioral effect, and the test "should ignore malformed codex array fields when importing" will spam `console.warn` lines on every run because `warnWithFallback` falls back to the global console. Either plumb a real logger through both functions (from `fromRulesyncMcp`/`toRulesyncMcp`) or drop the parameter entirely.

src/features/mcp/codexcli-mcp.ts
```diff
@@ -35,15 +46,15 @@ function convertFromCodexFormat(codexMcp: Record<string, unknown>): McpServers {
...
-        converted["disabledTools"] = value;
-      } else if (key === "env_vars") {
-        // codex stores env-var passthrough names in snake_case (`env_vars`);
-        // the rulesync source schema uses camelCase (`envVars`) for
-        // consistency with `enabledTools`/`disabledTools`/etc.
-        converted["envVars"] = value;
+      } else if (Object.hasOwn(CODEX_TO_RULESYNC_FIELD_MAP, key)) {
+        const mappedKey = CODEX_TO_RULESYNC_FIELD_MAP[key];
+        if (mappedKey) {
+          if (isStringArray(value)) {
```
dyoshikawa, line 52: Minor: `Object.hasOwn(CODEX_TO_RULESYNC_FIELD_MAP, key)` already guarantees the lookup yields a string (the map is typed `Record<string, string>`), so the inner `if (mappedKey)` is unreachable defensive code. Same pattern appears in `convertToCodexFormat` around line 86. Not a bug — just dead branches you could drop for clarity.

src/types/mcp.ts
```diff
@@ -1,4 +1,13 @@
-import { z } from "zod/mini";
+import { refine, z } from "zod/mini";
+
+const EnvVarNameSchema = z
+  .string()
+  .check(
+    refine(
```
dyoshikawa, line 6: This tightening is a good idea, but it is also a backward-incompatible change: existing user configs with `envVars` entries containing dashes, dots, or other non-identifier characters will now fail to parse. Could you (a) call this out as a possible breaking change in the PR description and (b) document the constraint in the Codex-specific section of `docs/reference/file-formats.md` and `skills/rulesync/file-formats.md`? The `.claude/rules/feature-change-guidelines.md` doc asks for docs and README to stay in sync with implemented behavior.

src/types/mcp.test.ts
```diff
@@ -21,3 +21,45 @@ describe("isMcpServers", () => {
...
 });
+
+describe("McpServerSchema", () => {
+  it("should accept valid uppercase envVars names", () => {
+    const result = McpServerSchema.safeParse({
+      command: "node",
+      envVars: ["OPENAI_API_KEY", "_PRIVATE", "KEY_123"],
+    });
+    expect(result.success).toBe(true);
+  });
```
dyoshikawa, line 32: Tiny nit: POSIX `name = [a-zA-Z_][a-zA-Z0-9_]*` is the identifier rule and strictly reserves the all-uppercase subset for environment variables, so calling lowercase names "POSIX compliant" is a little misleading. Consider renaming this test to something like "should accept lowercase envVars names" to avoid that implication. (Allowing lowercase is still a reasonable choice — only the label is off.)

dyoshikawa (commented):
Nice cleanup. The shared field maps remove a lot of duplication, the `isStringArray` guard makes the malformed-input handling explicit, and `Object.hasOwn` guarding the map lookup is a thoughtful touch against prototype pollution. The strip-then-remerge pattern in `fromRulesyncMcp` is sound, and the regression test asserting that pal-server only exposes `env_vars` (not `targets`/`description`/`exposed`) is exactly what was missing.

A few things worth tightening before merge: the new `envVars` regex makes the schema stricter than before, so it's worth calling out as a possible breaking change for users with non-conforming names and documenting the constraint in `file-formats.md`. The `logger?` parameter added to `convertToCodexFormat` is always called with `undefined`, while `convertFromCodexFormat` was given no logger at all even though it now warns too — the asymmetry suggests an in-progress refactor. Also note that CI is red purely because of `oxfmt --check` on `codexcli-mcp.ts`; a `pnpm fmt` run should clear that. Inline comments cover the smaller suggestions.

dyoshikawa-claw (approved):
All CI checks passed, code & security reviews completed. LGTM.

Issue #1603, opened 2026-05-07
Refactor: Reuse toPosixPath in CopilotRule path normalization and improve sameRelativePath API


dyoshikawa:
## Background

PR #1584 introduced `normalizeRelativePath` and `sameRelativePath` helper functions in `src/features/rules/copilot-rule.ts` to fix Windows/WSL path separator issues in Copilot root path detection. While the fix is functionally correct, a code review identified several improvement opportunities — most notably a DRY violation against the existing `toPosixPath` utility.

## Details

The following findings were identified during review of PR #1584:

### 1. DRY violation: `normalizeRelativePath` duplicates `toPosixPath` (mid)

`src/utils/file.ts` already exports `toPosixPath` which performs the same `p.replace(/\\/g, "/")` transformation. The new `normalizeRelativePath` in `copilot-rule.ts` reimplements this and adds `replace(/\/+/g, "/")` for consecutive slash collapsing. The project's coding guidelines explicitly state to use `toPosixPath` for non-filesystem path normalization.

### 2. `sameRelativePath` takes 4 positional string parameters (low)

The function signature `(leftDir, leftFile, rightDir, rightFile)` is error-prone — callers can easily swap left/right arguments. An object-based API would improve readability and safety.

### 3. Parameter name `path` shadows `node:path` module scope (low)

The parameter name `path` in `normalizeRelativePath` can be confused with the `node:path` import in the same file.

### 4. Test coverage gaps (low)

Only `.github\\` (trailing backslash) is tested. Mid-path backslashes (`.github\\subfolder`) and mixed separators are not covered.

### 5. Consecutive slash normalization necessity is unclear (low)

`join()` already normalizes consecutive slashes in its output. The extra `replace(/\/+/g, "/")` may be unnecessary defensive code — if it is needed, it should be justified with a concrete test case.

## Solution / Next Steps

1. **Replace `normalizeRelativePath` with `toPosixPath`**: Import `toPosixPath` from `../../utils/file.js` and compose it:
   ```typescript
   import { toPosixPath } from "../../utils/file.js";
   const normalizeRelativePath = (p: string): string =>
     toPosixPath(p).replace(/\/+/g, "/");
   ```
   Or, if the consecutive-slash handling proves unnecessary, just use `toPosixPath` directly.

2. **Consider refactoring `sameRelativePath` to use object parameters**:
   ```typescript
   const sameRelativePath = (
     left: { dir: string; file: string },
     right: { dir: string; file: string },
[...]

PR #1644, opened 2026-05-15
refactor(copilot-rule): reuse toPosixPath and switch sameRelativePath to object params

Chen17-sq:
## Summary

Addresses the refactor checklist in #1603 for `src/features/rules/copilot-rule.ts`:

- **DRY** — Replace the local `path.replace(/\\/g, "/")` with the shared `toPosixPath` utility from `src/utils/file.ts`, in line with the coding guideline that points to `toPosixPath` for non-filesystem path normalization.
- **Keep the slash collapse, but justify it** — The follow-up `replace(/\/+/g, "/")` is still needed: on POSIX, `node:path.join(".github\\", "x.md")` keeps the trailing backslash literal, so after `toPosixPath` runs we get `.github//x.md`. A short comment now spells this out so future readers don't think the collapse is dead code.
- **Avoid the `path` parameter name** — Rename `path` → `p` so it no longer shadows the `node:path` import in the same file.
- **Object-based `sameRelativePath`** — Switch the four positional strings to `{dir, file}` objects per the project's coding guidelines, and update both call sites (`fromFile`, `forDeletion`).

## Test plan

- [x] `pnpm cicheck` (fmt + oxlint + eslint + tsgo + vitest 5552 tests + cspell + secretlint) passes locally on Node 22.
- [x] New `forDeletion` tests:
  - Root deletion target is detected when `relativeDirPath` is `".github\\"`.
  - Non-root paths with a mid-path backslash (`".github\\instructions"`) stay non-root.
- [x] Existing root-detection test for `.github\\` continues to pass — behavior is preserved, only the helper implementation changes.

Refs #1603.

### DIFF
--- a/src/features/rules/copilot-rule.test.ts
+++ b/src/features/rules/copilot-rule.test.ts
@@ -620,11 +620,11 @@ This should be treated as a non-root rule.`;
       expect(copilotRule.getBody()).toBe("This should be treated as a non-root rule.");
     });
 
-    it("should normalize separators when detecting explicit root paths", async () => {
+    it("should normalize a trailing-backslash relativeDirPath when detecting explicit root paths", async () => {
       const githubDir = join(testDir, ".github");
       await ensureDir(githubDir);
 
-      const rootContent = "Root detected with mixed separators.";
+      const rootContent = "Root detected with a trailing backslash on relativeDirPath.";
       await writeFileContent(join(githubDir, "copilot-instructions.md"), rootContent);
 
       const copilotRule = await CopilotRule.fromFile({
@@ -636,6 +636,40 @@ This should be treated as a non-root rule.`;
 
       expect(copilotRule.isRoot()).toBe(true);
       expect(copilotRule.getBody()).toBe(rootContent);
+    });
+
+    it("should treat non-root paths as non-root when relativeDirPath has a mid-path backslash", async () => {
+      // Symmetric coverage to the forDeletion equivalent — exercises the
+      // sameRelativePath check on the fromFile call site so the original
+      // coverage gap flagged in #1603 is closed on both call sites.
+      //
+      // Note on the fixture path: fromFile reads the file from the *raw*
+      // relativeDirPath after the root check (only the comparison goes
+      // through normalizeRelativePath). On POSIX `\` is a regular filename
+      // character, so we materialize the directory at the literal
+      // ".github\instructions" name so the read inside fromFile succeeds.
+      const literalDir = join(testDir, ".github\\instructions");
+      await ensureDir(literalDir);
+      const fileContent = `---
+description: "Mid-path backslash on relativeDirPath"
+applyTo: "**/*.ts"
+---
+
+This should resolve to a non-root rule.`;
+      await writeFileContent(join(literalDir, "feature.instructions.md"), fileContent);
+
+      const copilotRule = await CopilotRule.fromFile({
+        outputRoot: testDir,
+        // Mid-path backslash (".github\\instructions" → ".github\instructions").
+        // sameRelativePath normalizes both sides before comparing against
+        // paths.root, so the comparison correctly returns false here.
+        relativeDirPath: ".github\\instructions",
+        relativeFilePath: "feature.instructions.md",
+        validate: true,
+      });
+
+      expect(copilotRule.isRoot()).toBe(false);
+      expect(copilotRule.getBody()).toBe("This should resolve to a non-root rule.");
     });
 
     it("should detect root only when both relativeDirPath and filename match", async () => {
@@ -797,7 +831,7 @@ description: "Test trimming"
   });
 
   describe("forDeletion", () => {
-    it("should mark root deletion target when separators are mixed", () => {
+    it("should mark root deletion target when relativeDirPath has a trailing backslash", () => {
       const copilotRule = CopilotRule.forDeletion({
         outputRoot: testDir,
         relativeDirPath: ".github\\",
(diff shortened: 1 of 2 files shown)


src/features/rules/copilot-rule.ts
```diff
@@ -48,12 +48,18 @@ export type CopilotRuleSettablePathsGlobal = ToolRuleSettablePathsGlobal & {
   };
 };
 
-const normalizeRelativePath = (path: string): string =>
-  path.replace(/\\/g, "/").replace(/\/+/g, "/");
+// toPosixPath converts backslashes to forward slashes so paths compare equally
+// on Windows and POSIX. The extra slash collapse covers a WSL/mixed-separator
+// edge case where `node:path/posix.join` keeps a literal backslash inside an
```
dyoshikawa: Minor: this references `node:path/posix.join`, but the actual import at the top of the file is `node:path` (`import { join } from "node:path"`). On POSIX, `node:path.join(".github\\", "x.md")` keeps the backslash literal, which is what the slash-collapse handles. The PR description has the accurate phrasing — consider updating the comment to say `node:path.join on POSIX` so future readers do not get confused.

src/features/rules/copilot-rule.ts
```diff
@@ -48,12 +48,18 @@ export type CopilotRuleSettablePathsGlobal = ToolRuleSettablePathsGlobal & {
...
+// input segment (e.g. ".github\\") and produces ".github//instructions/x.md"
+// after the backslash is rewritten.
+const normalizeRelativePath = (p: string): string => toPosixPath(p).replace(/\/+/g, "/");
```
dyoshikawa, line 60: Nit: `p` works for avoiding the `node:path` shadow, but `relativePath` would communicate intent a bit better. Feel free to ignore — `toPosixPath` in `utils/file.ts` already uses `p`, so this is at least consistent with the existing helper.

dyoshikawa (commented):
Nice, focused refactor — it matches the coding guidelines (object params, shared toPosixPath) cleanly, and the new forDeletion tests cover the mixed-separator edge case well. One small thing about the comment accuracy flagged inline.

src/features/rules/copilot-rule.ts
dyoshikawa: The comment refers to `node:path/posix.join`, but the import at the top of this file is `node:path` (the platform-aware variant), not `node:path/posix`. The behavior it describes — `node:path.join(".github\\", "x.md")` leaving the backslash literal — happens on POSIX where `\` isn't a separator, and that's actually what we rely on here. Also the example output mentions `.github//instructions/x.md`, but the join at the call sites in this file produces `.github//copilot-instructions.md` for the root case. Mind tightening the comment so it matches the real import and the real string under test? Otherwise the next reader may go hunting for a `path/posix` import that doesn't exist.

src/features/rules/copilot-rule.test.ts
```diff
@@ -796,6 +796,28 @@ description: "Test trimming"
     });
   });
 
+  describe("forDeletion", () => {
+    it("should mark root deletion target when separators are mixed", () => {
```
dyoshikawa: Small naming nit: both new cases say "separators are mixed" / "mixed separators", but neither input actually mixes `/` and `\` in the same string — they use backslashes only (trailing here, mid-path below). Maybe rename to something like "should mark root deletion target when the dir uses a trailing backslash" and "should treat non-root paths with backslash separators as non-root"? Or, if you'd like to keep the phrasing, add one truly mixed case like `relativeDirPath: ".github\\instructions/sub"` — that matches what #1603 explicitly asked for.

src/features/rules/copilot-rule.test.ts
```diff
@@ -796,6 +796,28 @@ description: "Test trimming"
     });
   });
 
+  describe("forDeletion", () => {
```
dyoshikawa, line 833: Nice that `forDeletion` now has coverage for the backslash path. A symmetric test for `fromFile` with `relativeDirPath: ".github\\instructions"` would close the other half of the issue's coverage ask, since `fromFile` also routes through `sameRelativePath`. Not blocking — the helper is shared — but adding it would make the safety net explicit at both call sites.

dyoshikawa (commented):
Clean little refactor. The DRY consolidation onto `toPosixPath`, the `path` → `p` rename, and the `{dir, file}` object form on `sameRelativePath` all match the project's coding guidelines, and the slash-collapse stays put with a justification — that part is exactly right. CI is fully green across all four platforms.

Two things worth a second look before merge. The new comment block in `copilot-rule.ts` describes `node:path/posix.join`, but the file actually imports from `node:path`, and the example path it shows doesn't match the string that's actually joined at the call sites. And the new `forDeletion` tests are labeled "mixed separators" but don't actually mix `/` and `\`. Both are non-blocking nits, but worth tightening for future readers. Optionally, a symmetric `fromFile` case for the mid-path backslash would close the coverage gap that #1603 explicitly called out.

Chen17-sq:
Thanks for the thorough read — pushed [`69cde2f`](https://github.com/dyoshikawa/rulesync/pull/1644/commits/69cde2f) addressing all three:

- **Comment accuracy.** Rewrote the `normalizeRelativePath` block so the module reference matches the actual `node:path` import on line 1, and the example matches the actual call shape (`.github\\` joined with `copilot-instructions.md`) instead of the multi-segment path the previous version implied.
- **Test labeling.** Renamed both `forDeletion` tests and the existing `fromFile` "normalize separators" test to spell out what they actually exercise — "trailing backslash" vs "mid-path backslash" — since none of them really mix `/` and `\\`.
- **Symmetric `fromFile` mid-path test.** Added the optional one. Since `fromFile` reads from the raw (un-normalized) `relativeDirPath` after the root check, the fixture materializes the directory at the literal `.github\\instructions` name on disk so the read succeeds — that path comment is inline on the test for future readers.

CI green; 1062 → 1063 tests on the copilot-rule file (65 → 66 in the suite).

dyoshikawa:
@Chen17-sq Thank you!

PR #1653, opened 2026-05-18
feat(mcp): add Amp CLI target support

anandghegde:
## Summary
- Add an Amp MCP adapter that writes `amp.mcpServers` into project `.amp/settings.json` and global `.config/amp/settings.json` while preserving other Amp settings.
- Register `amp` as an MCP-capable target and include optional gitignore coverage for project settings.
- Document Amp support and add unit/E2E coverage for generate, import, global mode, and non-deletable settings behavior.

## Test Plan
- `pnpm cicheck`
- `pnpm test:e2e src/e2e/e2e-mcp.spec.ts`

No linked issue; this adds support requested for Amp CLI MCP settings.


src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  static async fromFile({
+    outputRoot = process.cwd(),
+    validate = true,
+    global = false,
+  }: ToolMcpFromFileParams): Promise<AmpMcp> {
+    const paths = this.getSettablePaths({ global });
+    const fileContent =
+      (await readFileContentOrNull(
+        join(outputRoot, paths.relativeDirPath, paths.relativeFilePath),
```
dyoshikawa: Amp also supports `settings.jsonc` for both project and global configs. Right now only `settings.json` is checked, so a user whose Amp config is in `.jsonc` will get a brand-new `.json` written alongside it and their actual settings won't be touched. Worth mirroring the `OpencodeMcp` pattern: try `.jsonc` first via `readFileContentOrNull`, fall back to `.json`, parse with `jsonc-parser`, and emit back to whichever filename was read.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,122 @@
...
+export class AmpMcp extends ToolMcp {
+  private readonly json: Record<string, unknown>;
+
+  constructor(params: ToolMcpParams) {
+    super(params);
+    this.json = JSON.parse(this.fileContent || "{}");
+  }
+
+  getJson(): Record<string, unknown> {
+    return this.json;
```
dyoshikawa: This hands back the internal mutable object — a caller can do `ampMcp.getJson()["amp.mcpServers"] = ...` and silently desync from `fileContent`. The `readonly` modifier on the field doesn't help because it only freezes the reference. Consider returning `structuredClone(this.json)` or `Object.freeze`-ing it in the constructor. (Same latent issue lives in `GeminiCliMcp`, but no need to spread it further.)

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  toRulesyncMcp(): RulesyncMcp {
+    return this.toRulesyncMcpDefault({
+      fileContent: JSON.stringify({ mcpServers: this.json[AMP_MCP_SERVERS_KEY] ?? {} }, null, 2),
+    });
+  }
+
+  validate(): ValidationResult {
+    return { success: true, error: null };
```
dyoshikawa, line 232: `validate()` is a no-op, which means a malformed `settings.json` (non-object root, `amp.mcpServers` set to a string, `__proto__`/`constructor` keys, etc.) is accepted silently and round-tripped back to disk. `CodexcliMcp` and `OpencodeMcp` both validate with a Zod schema and explicitly skip `PROTOTYPE_POLLUTION_KEYS`. Even a minimal `isRecord` check plus the pollution-key guard would meaningfully harden this against user-supplied input.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,122 @@
...
+    outputRoot = process.cwd(),
+    relativeDirPath,
+    relativeFilePath,
+    global = false,
+  }: ToolMcpForDeletionParams): AmpMcp {
+    return new AmpMcp({
+      outputRoot,
+      relativeDirPath,
+      relativeFilePath,
+      fileContent: "{}",
```
dyoshikawa, line 256: `forDeletion` is required by the abstract interface but here it constructs an instance whose content is literally `{}` (no `amp.mcpServers` key). It's safe today only because `isDeletable()` short-circuits before this content is ever written — if a future refactor forgets that guard, this would happily overwrite the user's full Amp settings. A short `// SAFETY:` comment explaining the contract (or throwing here since it's a logical no-op) would prevent that footgun.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,122 @@
...
+  ToolMcpForDeletionParams,
+  ToolMcpFromFileParams,
+  ToolMcpFromRulesyncMcpParams,
+  ToolMcpParams,
+  ToolMcpSettablePaths,
+} from "./tool-mcp.js";
+
+const AMP_MCP_SERVERS_KEY = "amp.mcpServers";
+
+export class AmpMcp extends ToolMcp {
```
dyoshikawa, line 58: Roughly 95% of this class is a clone of `GeminiCliMcp` with the key name and paths swapped out. Not blocking — fine to ship as-is — but if a fourth `settings.json`-wrapper target appears, it's probably time to pull out a `MergedSettingsToolMcp` base that takes `{ mcpKey, projectPaths, globalPaths }` and folds in load/merge/serialize. A follow-up issue would be enough for now.

dyoshikawa (commented):
Thanks for putting this together — the adapter mirrors `GeminiCliMcp` closely and the docs/e2e/gitignore wiring is all coherent. CI is green and there's nothing concerning from a security standpoint.

A few things stood out while reading. The biggest one is that Amp's manual lets users author config in either `settings.json` or `settings.jsonc`, and right now only `.json` is read/written, so a `.jsonc` user would end up with a fresh sibling `.json` and a split configuration. `OpencodeMcp` already has the fallback pattern with `jsonc-parser` if you want a template.

The other thing worth tightening before merge is `validate()` — it's currently a no-op, so malformed input or prototype-pollution keys (`__proto__`, etc.) slip through. `CodexcliMcp` and `OpencodeMcp` both run Zod schemas plus an explicit pollution-key guard; mirroring that here would be cheap and defensive.

Also — this is now the third class doing the same "load JSON file, merge our key, preserve the rest" dance (Gemini, Codex, now Amp). I'm not asking you to refactor in this PR, but a follow-up to extract a shared base would pay off if more `settings.json`-style targets land.

The rest are smaller polish items inline. Nice work overall.

anandghegde:
Pushed suggested changes

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,275 @@
...
+  ToolMcpSettablePaths,
+} from "./tool-mcp.js";
+
+const AMP_MCP_SERVERS_KEY = "amp.mcpServers";
+
+// Block prototype pollution keys
+const PROTOTYPE_POLLUTION_KEYS = new Set(["__proto__", "constructor", "prototype"]);
+
+// Valid transport types for Amp MCP servers
+const VALID_TRANSPORT_TYPES = new Set(["stdio", "sse", "http", "streamable_http"]);
```
dyoshikawa: This transport allowlist is the only one of its kind across the MCP adapters — Cursor, Claudecode, Cline, Roo, Codexcli etc. all accept unknown fields. `.claude/rules/coding-guidelines.md` says to use `z.looseObject()` for the same reason: AI tools evolve quickly and configs gain fields. As written, the next transport Amp ships (e.g. a new variant) will be rejected here. Consider dropping the allowlist or moving validation into a `zod/mini` loose schema.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,275 @@
...
+  ToolMcpFromFileParams,
+  ToolMcpFromRulesyncMcpParams,
+  ToolMcpParams,
+  ToolMcpSettablePaths,
+} from "./tool-mcp.js";
+
+const AMP_MCP_SERVERS_KEY = "amp.mcpServers";
+
+// Block prototype pollution keys
+const PROTOTYPE_POLLUTION_KEYS = new Set(["__proto__", "constructor", "prototype"]);
```
dyoshikawa: This is the same `Set` literal as `PROTOTYPE_POLLUTION_KEYS` in `src/features/mcp/codexcli-mcp.ts`. Worth extracting to a small shared module (e.g. `src/utils/prototype-pollution.ts`) so future adapters can reuse it.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,275 @@
...
+      fileContent = await readFileContentOrNull(jsonPath);
+      if (fileContent) {
+        relativeFilePath = "settings.json";
+      } else {
+        // Neither file exists, default to jsonc
+        relativeFilePath = "settings.jsonc";
+      }
+    } else {
+      relativeFilePath = "settings.jsonc";
+    }
```
dyoshikawa: This ~15-line JSONC/JSON probing block is duplicated almost verbatim in `fromRulesyncMcp` below. A private static helper like `resolveSettingsFile(jsonDir)` returning `{ fileContent, relativeFilePath }` would keep them in sync. Also note that several `else` branches reassign `relativeFilePath` to the same `"settings.jsonc"` it was initialized to — those are no-ops and can be removed.

src/cli/commands/gitignore-entries.ts
```diff
@@ -203,6 +203,10 @@ export const GITIGNORE_ENTRY_REGISTRY: ReadonlyArray<GitignoreEntryTag> = [
   // Keep this after ignore entries like "**/.aiignore" so the exception remains effective.
   { target: "common", feature: "general", entry: "!.rulesync/.aiignore" },
 
+  // AMP
+  { target: "amp", feature: "mcp", entry: "**/.amp/settings.json" },
+  { target: "amp", feature: "mcp", entry: "**/.amp/settings.jsonc" },
```
dyoshikawa: These entries diverge from the closest precedent. `KiloMcp` and `OpencodeMcp` are also `isDeletable=false` and also preserve other user settings on round-trip, and they intentionally have **no** gitignore entry for their settings file — Kilo even has an explicit comment explaining that adding one would be too aggressive because users want to check the file in. Since `amp/settings.json` holds things like `amp.dangerouslyAllowAll`, the same reasoning applies. I'd drop both lines.

src/e2e/e2e-mcp.spec.ts
```diff
@@ -19,6 +19,7 @@ describe("E2E: mcp", () => {
   const { getTestDir } = useTestDirectory();
 
   it.each([
+    { target: "amp", outputPath: join(".amp", "settings.jsonc") },
     { target: "claudecode", outputPath: ".mcp.json" },
```
dyoshikawa, line 23: Only the basic generate matrix is updated, but the same file has three more `it.each` matrices that should include Amp to preserve the Tool×Feature happy-path coverage CLAUDE.md asks for: (1) `E2E: mcp (import)`, (2) `E2E: mcp (global mode)` — Amp supports global, (3) the non-deletable check-mode matrix where Amp matches geminicli/codexcli/opencode/kilo. The unit tests in `amp-mcp.test.ts` cover these paths but the E2E matrix is the contract.

dyoshikawa (commented):
Thanks for adding this — the adapter is well-tested and the layered prototype-pollution defense is nice to see. A few things stood out though.

The `VALID_TRANSPORT_TYPES` allowlist will break whenever Amp adds a new transport. Other MCP adapters in this repo accept unknown fields on purpose (see `.claude/rules/coding-guidelines.md` on `z.looseObject()`); I'd drop this check or switch to a loose schema. The JSONC/JSON probing block is also duplicated verbatim between `fromFile` and `fromRulesyncMcp` — a tiny private static helper would prevent drift.

A couple of consistency items worth fixing before merge: (1) The `**/.amp/settings.json{,c}` gitignore entries contradict the precedent in `KiloMcp`/`OpencodeMcp`, where `isDeletable=false` files that hold user-owned settings are intentionally **not** gitignored (Kilo even has an explicit comment about this). Since `amp/settings.json` holds things like `amp.dangerouslyAllowAll` that the user authors and wants in VCS, this looks too aggressive. (2) The README "Supported Tools and Features" table and `docs/reference/file-formats.md` aren't updated for Amp — `.claude/rules/feature-change-guidelines.md` requires this. (3) Only the basic generate matrix in `e2e-mcp.spec.ts` was extended. The `import`, `global-mode`, and non-deletable check-mode matrices in the same file are missing Amp entries, and CLAUDE.md explicitly calls out preserving the Tool×Feature happy-path E2E coverage.

No security concerns — `PROTOTYPE_POLLUTION_KEYS` filtering is applied in three places, no new deps, no exec/network/telemetry. Just a small DRY note that the constant duplicates the one in `codexcli-mcp.ts` and could move to a shared util.

anandghegde:
Hi @dyoshikawa — I've merged the latest `main` into this branch and resolved the conflict in `src/features/mcp/codexcli-mcp.ts` (combined the `PROTOTYPE_POLLUTION_KEYS` import from this branch with the `isStringArray` addition from main; both symbols are used in the file). `pnpm cicheck:code` passes locally (format, lint, typecheck, 5614 tests). Could you please take another look when you have a chance? Thanks!

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,245 @@
...
+    rulesyncMcp,
+    validate = true,
+    global = false,
+  }: ToolMcpFromRulesyncMcpParams): Promise<AmpMcp> {
+    const basePaths = this.getSettablePaths({ global });
+    const jsonDir = join(outputRoot, basePaths.relativeDirPath);
+    const { fileContent, relativeFilePath } = await this.resolveSettingsFile(jsonDir);
+
+    // If neither exists, use default config
+    const parsed = fileContent ? parseJsonc(fileContent) : { [AMP_MCP_SERVERS_KEY]: {} };
```
dyoshikawa: `parseJsonc` is lenient — on malformed input it recovers and returns `{}` instead of throwing (e.g. `parseJsonc("{ not json")` → `{}`). Since this path then writes `{ ...json, "amp.mcpServers": ... }` back to disk, a corrupted or truncated existing settings file is parsed to empty and silently overwritten, destroying the user's other Amp settings — the opposite of this adapter's purpose. cursor-mcp.ts / claudecode-mcp.ts throw on parse failure so generation fails loudly instead. Consider passing the `errors: ParseError[]` out-param to `parse` and throwing a descriptive error when the existing file fails to parse. Same concern applies to `fromFile` at line 77.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,245 @@
...
+    // eslint-disable-next-line no-type-assertion/no-type-assertion
+    const json = parsed as Record<string, unknown>;
+
+    // Filter out prototype pollution keys from MCP servers
+    const mcpServers = rulesyncMcp.getMcpServers();
+    const filteredMcpServers: Record<string, unknown> = {};
+
+    for (const [name, config] of Object.entries(mcpServers)) {
```
dyoshikawa: This "skip pollution keys at the server-name level, then again at the config-key level" loop is repeated almost verbatim in `toRulesyncMcp` (~line 146) and `validate` (~line 188). codexcli-mcp.ts factors this into shared `convertTo/FromCodexFormat` helpers — extracting a single `filterMcpServers()` here would remove the triplication. Minor: `!config || typeof config !== "object"` doesn't exclude arrays; `isRecord` / `isPlainObject` from src/utils/type-guards.ts would be tighter.

src/utils/prototype-pollution.ts
```diff
@@ -0,0 +1,16 @@
...
+ * that copies arbitrary user-supplied keys into a fresh object — frontmatter
+ * parsing, MCP config conversion, settings round-trip — should skip these.
+ */
+export const PROTOTYPE_POLLUTION_KEYS: ReadonlySet<string> = new Set([
+  "__proto__",
+  "constructor",
+  "prototype",
+]);
+
+export function isPrototypePollutionKey(key: string): boolean {
```
dyoshikawa, line 14: `isPrototypePollutionKey` isn't imported anywhere — only `PROTOTYPE_POLLUTION_KEYS` is consumed (here and in codexcli-mcp.ts). Either use it in the filter callbacks or drop it so we don't ship unused API surface.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,245 @@
...
+} from "./tool-mcp.js";
+
+const AMP_MCP_SERVERS_KEY = "amp.mcpServers";
+
+export class AmpMcp extends ToolMcp {
+  private readonly json: Record<string, unknown>;
+
+  constructor(params: ToolMcpParams) {
+    super(params);
+    // Use parseJsonc to preserve __proto__ and other special keys
```
dyoshikawa: This comment isn't quite accurate: jsonc-parser actually *drops* `__proto__` entirely during parsing (it never becomes an own key), so it isn't "preserved". What survives as own enumerable keys are `constructor`/`prototype`, which is what the validate/filter guards are really catching. Worth rewording — the `__proto__` branches in `validate()` are effectively dead for jsonc-parsed input.

src/features/mcp/amp-mcp.ts
```diff
@@ -0,0 +1,245 @@
...
+          }
+        }
+      }
+    }
+
+    return { success: true, error: null };
+  }
+
+  /**
+   * settings.json may contain other Amp settings, so it should not be deleted.
```
dyoshikawa, line 236: Small wording nit: the adapter primarily writes `settings.jsonc`, so "settings.json may contain other Amp settings" reads as if only `.json` is in play. `settings.json/.jsonc` would be clearer. The always-false logic itself is correct and matches codexcli.

dyoshikawa (commented):
Nice addition — the adapter tracks the existing MCP pattern closely, the README/docs/skills tables stay in sync, and the E2E matrix coverage (generate / import / global / non-deletable check) is solid. Security also looks clean: I verified the prototype-pollution guard holds (jsonc-parser drops `__proto__` and the validate/filter loops reject `constructor`/`prototype`), the paths are static so there's no traversal vector, and `isDeletable() === false` keeps the orchestrator from ever writing the `{}` deletion content over the real settings file.

The one thing I'd want addressed before merge is the malformed-JSON case. `parseJsonc` recovers silently and returns `{}` on invalid input, so a broken or truncated existing `settings.json[c]` gets parsed to empty and then overwritten — wiping the user's other Amp settings, which is the opposite of what this adapter is meant to do. The other adapters (cursor/claudecode) throw on parse errors instead. Everything else is cleanup: some duplicated filter logic, an unused export, and a couple of stale comments. Details inline.

anandghegde:
Addressed the current review feedback in `3c4f7be2`:

- Hardened `AmpMcp` JSONC parsing so malformed existing `settings.json/settings.jsonc` now fails loudly instead of being recovered as `{}` and overwritten.
- Made `getJson()` return a defensive clone so callers cannot mutate the adapter's internal parsed settings state.
- Tightened validation for Amp settings shape: non-object roots, non-object `amp.mcpServers`, non-object server configs, and prototype-pollution keys are rejected.
- Extracted the repeated MCP server filtering into a shared `filterMcpServers()` helper and switched Amp to use `isPrototypePollutionKey()`.
- Reworded the JSONC/prototype-pollution comment to reflect that `jsonc-parser` drops `__proto__` while `constructor`/`prototype` can remain own keys.
- Updated the non-deletable MCP E2E matrix to include `opencode` and `kilo`; Amp/import/global coverage was already present from the previous review commit.
- Confirmed the `.amp/settings.{json,jsonc}` gitignore entries were already removed and Amp is listed with the other intentionally user-owned settings files.
- Left the shared base-class abstraction as a future follow-up since the review called it non-blocking unless another similar settings-wrapper target appears.

Verification:

- `pnpm cicheck`
- `pnpm run test` (219 files, 5619 tests)
- `pnpm vitest run src/features/mcp/amp-mcp.test.ts --silent=false` (23 tests)
- `pnpm vitest run --config vitest.e2e.config.ts src/e2e/e2e-mcp.spec.ts --silent=false` (54 tests)

dyoshikawa:
@anandghegde Thank you!

PR #1655, opened 2026-05-18
feat(kilo): explicitly declare supported Kilo agent frontmatter fields

### LINKED ISSUE #1639: Follow-ups from PR #1635 (Kilo subagent schema / OpenCodeSubagent.forDeletion global)
## Background

PR #1635 (`fix(kilo-subagent): align read defaults`) gave `KiloSubagent` its own frontmatter schema, decoupling it from `OpenCodeStyleSubagentFrontmatterSchema`. During review, two follow-up items were identified that are out of scope for that PR but should be tracked.

### 1. `KiloSubagent.validate()` still uses parent class schema

`KiloSubagent` does not override `validate()`, so it inherits `OpenCodeStyleSubagent.validate()`, which validates against `OpenCodeStyleSubagentFrontmatterSchema` (default `mode: "subagent"`). While both schemas are currently structurally identical (`z.looseObject({ description, mode, name })`), if Kilo's schema diverges in the future (e.g., new Kilo-specific fields), validation would produce false negatives.

**Affected code**: `src/features/subagents/opencode-style-subagent.ts` `validate()` method (inherited by `KiloSubagent`)

### 2. `OpenCodeSubagent.forDeletion` silently drops `global` parameter

PR #1635 added `global` passthrough to `KiloSubagent.forDeletion`, but `OpenCodeSubagent.forDeletion` has the same gap. When `subagents-processor.ts` calls `forDeletion` with `global: this.global`, the `OpenCodeSubagent` variant ignores it, defaulting to `global: false`. Not functionally critical for deletion placeholders, but inconsistent.

**Affected code**: `src/features/subagents/opencode-subagent.ts` `forDeletion()` static method

## Solution / Next Steps

1. Consider having `KiloSubagent` override `validate()` to use `KiloSubagentFrontmatterSchema`, or refactor `OpenCodeStyleSubagent.validate()` to be overridable with a dynamic schema reference.
2. Add `global` parameter passthrough to `OpenCodeSubagent.forDeletion` (mirroring the `KiloSubagent` fix in PR #1635).

sirmacik:
## Summary

- Replaces the bare `OpenCodeStyleSubagentFrontmatterSchema` alias in `KiloSubagentFrontmatterSchema` with a dedicated schema that explicitly declares all Kilo-runtime-supported agent frontmatter fields: `displayName`, `deprecated`, `native`, `hidden`, `top_p`, `temperature`, `color`, `permission`, `model`, `variant`, `prompt`, `options`, `steps`, `disable`
- **Enforces Kilo schema validation at runtime** by adding a constructor and overriding `validate()` to use `KiloSubagentFrontmatterSchema` rather than falling back to the parent class.
- **Strictly parses rulesync config** in `fromRulesyncSubagent` so invalid types (e.g. `temperature: "hot"`) are caught as Zod errors.
- **Restores `global` passthrough** inside `OpenCodeSubagent.forDeletion()`.
- Keeps `z.looseObject()` so unknown future Kilo fields still pass through (per project guidelines)

Fixes #1639.

## Backward compatibility

All existing tests pass. No fields are removed. `looseObject` ensures any Kilo fields added in future still pass through without a schema update.

## Fields not included

`skills` preload, per-agent `mcpServers`, and persistent `memory` are intentionally omitted — Kilo runtime does not support them.


src/features/subagents/kilo-subagent.ts
```diff
@@ -53,6 +88,7 @@ export class KiloSubagent extends OpenCodeStyleSubagent {
     const kiloFrontmatter: KiloSubagentFrontmatter = KiloSubagentFrontmatterSchema.parse({
       ...kiloSection,
       description: rulesyncFrontmatter.description,
+      mode: typeof kiloSection.mode === "string" ? kiloSection.mode : "all",
```
dyoshikawa: This line looks redundant with the schema itself: `KiloSubagentFrontmatterSchema` already declares `mode: z._default(z.string(), "all")`, so `.parse()` would supply `"all"` whenever `mode` is missing. The only thing this explicit line adds is silently coercing a non-string `mode` to `"all"` instead of letting Zod raise — which is a bit inconsistent with the rest of the PR's message of "invalid types like `temperature: 'hot'` should throw". If the soft fallback is intentional for `mode` specifically, a one-line comment explaining why would help; otherwise dropping this line and trusting the schema default would be cleaner.

dyoshikawa (commented):
Solid tightening of the Kilo frontmatter surface — the dedicated schema plus the `validate()` override genuinely strengthens runtime checking, the type narrowing via `declare protected readonly` is the right idiom, and the existing happy-path matrix is preserved. CI is green.

A few things worth a look before merging:

The PR description says `forDeletion()` restores the `global` passthrough from #1635, but `git diff origin/main...HEAD` shows no change to that function — the passthrough was already on `main`. Probably worth trimming that bullet from the description so reviewers aren't hunting for a non-change.

Per `.claude/rules/feature-change-guidelines.md`, the `Each File Format` section should track implemented functionality. `docs/reference/file-formats.md` (around the Kilo section) currently only mentions `kilo.mode`. Now that fields like `displayName`, `temperature`, `top_p`, `model`, `permission`, `prompt`, etc. are officially declared, it would be good to enumerate them there so users can discover the supported surface.

Optional follow-up: the same loose-alias pattern still exists in `OpenCodeSubagent` (and the sibling tools that extend the same parent). Worth an issue to track applying this pattern across the family for consistency.

src/features/subagents/kilo-subagent.ts
```diff
@@ -18,19 +19,53 @@ import {
...
+
   protected getToolTarget(): Extract<ToolTarget, "opencode" | "kilo"> {
     return "kilo";
   }
 
+  getFrontmatter(): KiloSubagentFrontmatter {
+    return this.frontmatter;
+  }
+
+  validate(): ValidationResult {
```
dyoshikawa, line 65: One subtlety worth flagging: this override is not called from the parent constructor. `OpenCodeStyleSubagent`'s constructor only runs the *parent*'s `OpenCodeStyleSubagentFrontmatterSchema.safeParse(...)` when `validate: true`, so `new KiloSubagent({ validate: true, frontmatter: { temperature: "hot" } })` does not throw. The strict Kilo schema only kicks in via `KiloSubagent.fromFile`, the explicit `.parse()` in `fromRulesyncSubagent`, or a manual `subagent.validate()` call. That's arguably fine and matches existing conventions, but the PR description reads as if construction-time validation is strict — worth either documenting the actual surface in a class-level JSDoc or having the child constructor re-validate against the Kilo schema.

src/features/subagents/kilo-subagent.test.ts
```diff
@@ -308,4 +308,129 @@ Body content`,
 
     expect(subagent.getFrontmatter().mode).toBe("all");
   });
+
+  it("should accept and expose all explicit Kilo frontmatter fields", async () => {
+    const dirPath = join(testDir, ".kilo", "agent");
```
dyoshikawa, line 340: All three new tests are happy-path. The headline behavior of the PR — that bad types now raise Zod errors — isn't covered, so a regression that, say, drops the `validate()` override or loosens `z.number()` back to `z.unknown()` would pass CI. A single negative test (e.g. `KiloSubagentFrontmatterSchema.safeParse({ mode: "subagent", temperature: "hot" })` asserting `result.success === false`, and/or `KiloSubagent.fromFile` rejecting a file with an invalid type) would lock in the strict behavior this PR is introducing.

sirmacik (commented):
Thanks for the review! I've pushed an update addressing all your feedback:
- Dropped the redundant `mode` fallback coercion in `fromRulesyncSubagent` to let Zod raise errors on bad types.
- Added a constructor to `KiloSubagent` that calls `this.validate()` when `validate: true`, ensuring strict validation kicks in during base initialization.
- Added a negative test in `kilo-subagent.test.ts` to assert that invalid schema types (like `temperature: "hot"`) correctly fail validation.
- Enumerated the supported Kilo fields in `docs/reference/file-formats.md` so the surface is discoverable.
- Trimmed the `forDeletion()` bullet from the PR description.
- Created issue #1658 to track applying this strict schema pattern to the rest of the OpenCodeStyle subagents.

Everything is green now.

dyoshikawa:
@sirmacik Thank you!

PR #1676, opened 2026-05-25
refactor: delegate draft-release workflow prompt to skill

dyoshikawa:
## Summary
- Shrink the inline prompt in `.github/workflows/draft-release.yml` down to a single sentence that delegates to the existing `draft-release` skill.
- The 10-step release procedure already lives in `.rulesync/skills/draft-release/SKILL.md` (and is mirrored to `.opencode/skill/draft-release/SKILL.md` by `pnpm generate`), so duplicating it in the workflow was redundant.
- The skill receives the workflow input via the existing `INPUT_NEW_VERSION` env var as its `$ARGUMENTS`; when empty, the skill falls back to the `release-dry-run` skill for automatic version determination.

## Test plan
- [ ] Trigger the `Draft Release` workflow on this branch via `workflow_dispatch` (with and without `new_version`) before merging to verify OpenCode still produces the expected draft release PR and GitHub draft release

### DIFF
--- a/.github/workflows/draft-release.yml
+++ b/.github/workflows/draft-release.yml
@@ -64,27 +64,4 @@ jobs:
           use_github_token: "true"
           share: false
           prompt: |
-            Let's work on the following steps.
-
-            1. Confirm that you are currently on the main branch and pull the latest changes. If not on main branch, switch to main branch.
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1680, opened 2026-05-26
feat(commands): add research-tool-updates and understand-scrap-issues commands

dyoshikawa:
## Summary

Add two new project commands (source files under `.rulesync/commands/`, generated into the configured tool outputs):

- **`research-tool-updates`** — Researches each rulesync target tool's recent upstream releases (official release notes / GitHub releases preferred) via one `general-purpose` subagent per tool, compares them against rulesync's current implementation, and files one GitHub issue per tool for any upstream capability rulesync has not yet followed.
  - Reads the `Supported Tools and Features` matrix in `README.md` each run as the authoritative tool/feature list (no hardcoding).
  - Accepts an optional `$ARGUMENTS` to scope the investigation to a single tool (display name or `--targets` id); empty means all tools.
  - Subagents start from the `rulesync-feature-research` skill / `references/<tool>.md`, ground claims in `src/**` and `--dry-run` output, and report only material capability gaps.
  - Caps subagent concurrency (~5). Issue content is always written in English.
  - **Mandatory duplicate check before filing**: searches open and closed issues by both the display name and the `--targets` id, and reads candidates before deciding. When a matching issue already exists, it **supplements that issue with a comment** carrying only the newly found releases/gaps instead of filing a duplicate; if there is nothing new to add, it is skipped.
  - `targets: [claudecode]` because the workflow depends on subagent orchestration (mirrors the existing `agent-team` command).
- **`understand-scrap-issues`** — Fetches the most recent `maintainer-scrap` issues (up to 3) and summarizes each (topic / background / details / next steps). `targets: ["*"]`.

## Test plan

- `pnpm cicheck` passes (5613 tests, lint/format/typecheck, cspell, secretlint, skill-docs sync).
- `pnpm run dev generate` produces the corresponding generated outputs (gitignored) and the new commands appear as available slash commands.

### DIFF
--- a/.rulesync/commands/research-tool-updates.md
+++ b/.rulesync/commands/research-tool-updates.md
@@ -1,0 +1,156 @@
+---
+targets:
+  - claudecode
+description: >-
+  Research recent upstream releases of every rulesync target tool, detect
+  capabilities rulesync has not yet followed, and file one GitHub issue per
+  tool for the gaps.
+---
+
+# Research Tool Updates
+
+TARGET = $ARGUMENTS
+
+Purpose: for every target tool rulesync supports, investigate the tool's recent
+releases (official release notes / GitHub releases preferred), compare them
+against rulesync's current implementation, and open a per-tool GitHub issue for
+any upstream capability rulesync has not yet caught up with.
+
+## Step 0: Determine Scope
+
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1730, opened 2026-06-02
fix(mcp): allow trailing commas in Amp JSONC parsing

### LINKED ISSUE #1728: AmpMcp rejects valid JSONC with trailing commas (missing allowTrailingComma)
## Background

PR #1653 (`feat(mcp): add Amp CLI target support`) introduces `AmpMcp`, an adapter that reads and writes Amp MCP configuration in `.amp/settings.json` / `.amp/settings.jsonc`. During review, a read-time correctness issue was found in the JSONC parsing helper.

Amp's advertised, primary config format is **`settings.jsonc`** — a format whose whole point is to tolerate JSON-with-comments **and trailing commas**. The current parser rejects perfectly legal JSONC that uses trailing commas, which is exactly the kind of file a human would hand-author.

## Details

In `src/features/mcp/amp-mcp.ts`, `parseAmpSettingsJsonc()` calls `jsonc-parser`'s `parse` without options:

```ts
function parseAmpSettingsJsonc(fileContent: string): Record<string, unknown> {
  const errors: ParseError[] = [];
  const parsed: unknown = parseJsonc(fileContent || "{}", errors);

  if (errors.length > 0) {
    const details = errors
      .map((error) => `${printParseErrorCode(error.error)} at offset ${error.offset}`)
      .join(", ");
    throw new Error(`Failed to parse Amp settings: ${details}`);
  }
  // ...
}
```

`jsonc-parser` defaults to `allowTrailingComma: false`. A trailing comma therefore produces entries in `errors` (`PropertyNameExpected` / `ValueExpected`), and because the helper throws whenever `errors.length > 0`, a valid `.jsonc` file with a trailing comma is rejected with `Failed to parse Amp settings` — even though `jsonc-parser` still returns the correct value.

Observations:

- **Read-time only.** The generate/write path emits `JSON.stringify` output (no trailing commas), so it is unaffected. The bug bites only when reading a pre-existing hand-authored `.amp/settings.jsonc` (e.g. `fromFile`, `fromRulesyncMcp`, and `validate`).
- **Inconsistent with sibling adapters.** `src/features/permissions/kilo-permissions.ts` uses the exact same strict `errors[]` + `printParseErrorCode` pattern but deliberately passes `{ allowTrailingComma: true }`. `opencode-mcp.ts` / `kilo-mcp.ts` call `parseJsonc` without an errors array (best-effort), so they silently accept trailing commas. Amp is currently the **only** adapter that throws on a legal JSONC file.
- **Untested gap.** The existing JSONC test in `src/features/mcp/amp-mcp.test.ts` only covers `//` comments (accepted by default). There is no trailing-comma fixture, so this regression is not caught.
- The malformed-input protection added earlier in the PR (throwing instead of silently overwriting user settings with `{}`) must be **preserved** — the fix below keeps that behavior; only trailing commas become tolerated, while genuinely malformed input still throws.

## Solution / Next Steps

1. Pass `{ allowTrailingComma: true }` to the `parseJsonc` call in `parseAmpSettingsJsonc()`:

   ```ts
   const parsed: unknown = parseJsonc(fileContent || "{}", errors, {
     allowTrailingComma: true,
   });
   ```

   Mirror the precedent in `src/features/permissions/kilo-permissions.ts`.

2. Add a test fixture in `src/features/mcp/amp-mcp.test.ts` that reads a `settings.jsonc` containing a trailing comma and asserts it parses successfully (alongside the existing `//`-comment case).

3. Verify the malformed-input guard still throws: a separate negative test (e.g. `"{ not json"`) should continue to raise `Failed to parse Amp settings`. (The existing "reject malformed existing settings" test already covers this — confirm it still passes.)

4. Run `pnpm cicheck` to confirm.

## References

- PR: #1653 (`feat(mcp): add Amp CLI target support`)
- File: `src/features/mcp/amp-mcp.ts` (`parseAmpSettingsJsonc`)
- Precedent: `src/features/permissions/kilo-permissions.ts` (`allowTrailingComma: true`)
- Test: `src/features/mcp/amp-mcp.test.ts`

axelray-dev:
Fixes #1728

## What changed

Pass `{ allowTrailingComma: true }` as the third argument to `jsonc-parser`'s `parse()` in `parseAmpSettingsJsonc()`, so that valid JSONC files with trailing commas are accepted instead of rejected.

Amp's primary config format is `settings.jsonc` -- a format whose whole point is to tolerate JSON-with-comments and trailing commas. The current parser rejected perfectly legal JSONC that uses trailing commas, which is exactly the kind of file a human would hand-author.

**Files changed:**
- `src/features/mcp/amp-mcp.ts` -- one-line fix: add `{ allowTrailingComma: true }` option to `parseJsonc()` call
- `src/features/mcp/amp-mcp.test.ts` -- add regression test for JSONC with trailing commas in the existing `JSONC support` describe block

## Pattern reference

This matches the existing pattern in `src/features/permissions/kilo-permissions.ts` line 43:
```ts
const parsed = parseJsonc(content, errors, { allowTrailingComma: true });
```

## Verification

```
pnpm test     # 233 files, 5840 tests passed
pnpm check    # format, oxlint, eslint, typecheck all passed
```

### DIFF
--- a/src/features/mcp/amp-mcp.test.ts
+++ b/src/features/mcp/amp-mcp.test.ts
@@ -420,6 +420,31 @@ describe("AmpMcp", () => {
       expect(ampMcp.getFilePath()).toBe(join(testDir, ".amp", "settings.jsonc"));
     });
 
+    it("should parse settings.jsonc with trailing commas", async () => {
+      const jsoncContent = `{
+  // Amp settings with trailing commas
+  "amp.mcpServers": {
+    "context7": {
+      "type": "http",
+      "url": "https://mcp.context7.com/mcp",
+    },
+  },
+}`;
+      await ensureDir(join(testDir, ".amp"));
+      await writeFileContent(join(testDir, ".amp", "settings.jsonc"), jsoncContent);
+
+      const ampMcp = await AmpMcp.fromFile({ outputRoot: testDir });
+
+      expect(ampMcp.getJson()).toEqual({
+        "amp.mcpServers": {
+          context7: {
+            type: "http",
+            url: "https://mcp.context7.com/mcp",
+          },
+        },
+      });
+    });
+
     it("should fall back to settings.json if settings.jsonc does not exist", async () => {
       const jsonData = {
         "amp.mcpServers": {
--- a/src/features/mcp/amp-mcp.ts
+++ b/src/features/mcp/amp-mcp.ts
@@ -20,7 +20,7 @@ const AMP_MCP_SERVERS_KEY = "amp.mcpServers";
 
 function parseAmpSettingsJsonc(fileContent: string): Record<string, unknown> {
   const errors: ParseError[] = [];
-  const parsed: unknown = parseJsonc(fileContent || "{}", errors);
+  const parsed: unknown = parseJsonc(fileContent || "{}", errors, { allowTrailingComma: true });
 
   if (errors.length > 0) {
     const details = errors


dyoshikawa:
@axelray-dev Thank you!

PR #1731, opened 2026-06-02
fix(hooks): wrap matcher handlers in block scope to avoid duplicate const __re

### LINKED ISSUE #1725: [Bug]: opencode hooks plugin emits duplicate `const __re` declarations for multiple matcher handlers on the same event, causing SyntaxError

## Problem

When `opencode` is included as a target and a single hook event (e.g.
`postToolUse` → `tool.execute.after`) contains **two or more handlers with a
`matcher`**, the generated `.opencode/plugins/rulesync-hooks.js` declares
`const __re` multiple times within the same function scope. The resulting
plugin is not valid JavaScript and fails to load with:

> SyntaxError: Identifier '__re' has already been declared

Because the plugin fails to load entirely, **all** hooks for the opencode
target stop working — not just the colliding ones.

## Evidence

**1. Generation logic — `opencode-style-generator.ts`**

In `generateOpencodeStylePluginCode`, the matcher branch pushes a
`const __re` declaration once per handler, directly at the top level of the
event-handler function body:

```typescript
for (const handler of handlers) {
  if (handler.matcher) {
    const safeMatcher = validateAndSanitizeMatcher(handler.matcher);
    lines.push(`      const __re = new RegExp("${safeMatcher}");`);  // ← re-declared per handler
    lines.push(`      if (__re.test(input.tool)) {`);
    lines.push(`        await $\`${escapedCommand}\`;`);
    lines.push("      }");
  } else {
    lines.push(`      await $\`${escapedCommand}\`;`);
  }
}
```

When `handlers` has more than one matcher entry for the same event, each
iteration emits another `const __re` into the same scope.

**2. Generated output**

```js
"tool.execute.after": async (input) => {
  const __re = new RegExp(".*");
  if (__re.test(input.tool)) { await $`echo a`; }
  const __re = new RegExp("Edit|Write");   // ← duplicate const in same scope
  if (__re.test(input.tool)) { await $`echo b`; }
  const __re = new RegExp("Edit|Write");   // ← duplicate const in same scope
  if (__re.test(input.tool)) { await $`echo c`; }
},
```

**3. The syntax error**

The generated plugin is an ES module (`export const RulesyncHooksPlugin`),
so it must be evaluated as one. Checking it as an ES module confirms the
failure:

```
node --input-type=module --check < .opencode/plugins/rulesync-hooks.js
SyntaxError: Identifier '__re' has already been declared
```

Note that `kilo` shares `opencode-style-generator.ts`, so it is affected in
the same way.

This is distinct from #1652 — that issue concerned `*` being rejected at
validation time, and the `*` → `.*` normalization is already in effect here
(the generated code shows `new RegExp(".*")`). The duplicate-declaration
problem surfaces only after that fix, once two matcher handlers coexist on
one event.

## Expected Behavior

The generated plugin must be valid JavaScript regardless of how many matcher
handlers an event has. Either:

1. Wrap each matcher handler in its own block scope, so each `__re` lives in
   a separate scope:

   ```js
   {
     const __re = new RegExp(".*");
     if (__re.test(input.tool)) { await $`echo a`; }
   }
   {
     const __re = new RegExp("Edit|Write");
     if (__re.test(input.tool)) { await $`echo b`; }
   }
   ```

2. Or give each declaration a unique identifier (`__re0`, `__re1`, …).

## Reproduction

`.rulesync/hooks.json`:

```json
{
  "version": 1,
  "hooks": {
    "postToolUse": [
      { "type": "command", "command": "echo a", "matcher": "*" },
      { "type": "command", "command": "echo b", "matcher": "Edit|Write" },
      { "type": "command", "command": "echo c", "matcher": "Edit|Write" }
    ]
  }
}
```

Then run:

```
rulesync generate --targets opencode --features hooks
```

and check the generated plugin as an ES module:

```
# The plugin is an ES module, so check it as one:
node --input-type=module --check < .opencode/plugins/rulesync-hooks.js
# → SyntaxError: Identifier '__re' has already been declared
```

(`node --check` without `--input-type=module` may pass silently when the
working directory has no `package.json` with `"type": "module"`, which can
mask the error.)

Tested with rulesync 8.23.0.

axelray-dev:
Fixes #1725

## What changed

The `generateOpencodeStylePluginCode` function in `opencode-style-generator.ts` emits `const __re = new RegExp(...)` once per handler in the same function scope. When multiple handlers with matchers exist for the same event (e.g., `postToolUse` with 3 matchers), the generated code has duplicate `const __re` declarations in the same scope, causing `SyntaxError: Identifier '__re' has already been declared`.

The fix wraps each matcher handler in its own block scope (`{ ... }`), so each `const __re` declaration lives in a separate scope. This affects both opencode and kilo targets since they share the same generator.

**Files changed:**
- `src/features/hooks/opencode-style-generator.ts` -- wrap matcher handlers in block scope
- `src/features/hooks/opencode-hooks.test.ts` -- add regression test with 3 matchers and `node --input-type=module --check`
- `src/features/hooks/kilo-hooks.test.ts` -- same regression test for kilo

## Verification

```
pnpm cicheck
```

All checks passed: format, oxlint, eslint, typecheck, 233 test files (5842 tests), cspell, secretlint.

DCO sign-off included.

### DIFF
--- a/src/features/hooks/kilo-hooks.test.ts
+++ b/src/features/hooks/kilo-hooks.test.ts
@@ -1,3 +1,4 @@
+import { execFileSync } from "node:child_process";
 import { join } from "node:path";
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@axelray-dev Thank you!

Issue #1669, opened 2026-05-22
fix(permissions): Cursor cli.json missing required permissions.allow and editor.vimMode on deny-only configs


jitendravyas:
## Summary

When generating project-level `.cursor/cli.json` from a deny-only `.rulesync/permissions.json`, rulesync produces a file that fails **Cursor CLI schema validation**. Cursor requires `permissions.allow` (array, may be empty) and `editor.vimMode` (boolean). Rulesync omits both on fresh generation when there are no `allow` rules in source.

This breaks Cursor CLI in repos that use rulesync's Cursor `permissions` feature with a deny-list-only policy (common for security guardrails).

## Environment

- **rulesync:** 8.19.0 (`npx rulesync@8.19.0 generate`)
- **Cursor CLI:** project config at `.cursor/cli.json`
- **OS:** macOS
- **rulesync.jsonc target:** `"cursor": [..., "permissions"]`

## Steps to reproduce

1. Create `.rulesync/permissions.json` with **only deny rules** (no `"allow"` entries):

```json
{
  "$schema": "https://github.com/dyoshikawa/rulesync/releases/latest/download/permissions-schema.json",
  "permission": {
    "bash": {
      "git push --force*": "deny",
      "npm install*": "deny"
    },
    "write": {
      ".env": "deny"
    }
  }
}
```

2. Ensure `.cursor/cli.json` is fresh (or `{}`).

3. Run:

```bash
npx rulesync@8.19.0 generate
```

4. Inspect `.cursor/cli.json` and run Cursor CLI in the project directory.

## Actual output (rulesync 8.19.0)

```json
{
  "version": 1,
  "permissions": {
    "deny": [
      "Shell(git push --force*)",
      "Shell(npm install*)",
      "Write(.env)"
    ]
  }
}
```

## Expected output (per Cursor CLI schema)

Cursor documents these as **required fields** for both project `.cursor/cli.json` and global `cli-config.json`:

- https://cursor.com/docs/cli/reference/configuration

Minimal valid example with deny-only policy:

```json
{
  "version": 1,
  "editor": { "vimMode": false },
  "permissions": {
    "allow": [],
    "deny": [
      "Shell(git push --force*)",
      "Shell(npm install*)",
      "Write(.env)"
    ]
  }
}
```

## Cursor CLI error

```
Invalid project config at <project>/.cursor/cli.json: schema validation failed.
[
  {
    "code": "invalid_type",
    "expected": "array",
    "received": "undefined",
[...]

PR #1732, opened 2026-06-03
fix(cursor): emit schema-required cli config fields

zerone0x:
## Summary\n- default-stamp Cursor CLI `editor.vimMode` when missing\n- always emit `permissions.allow` as an array, even for deny-only configs\n- add regression coverage for deny-only generation and preserving existing editor settings\n\n## Testing\n- corepack pnpm exec vitest run src/features/permissions/cursor-permissions.test.ts --silent=true\n- corepack pnpm exec eslint src/features/permissions/cursor-permissions.ts src/features/permissions/cursor-permissions.test.ts\n\nCloses #1669

### DIFF
--- a/src/features/permissions/cursor-permissions.test.ts
+++ b/src/features/permissions/cursor-permissions.test.ts
@@ -190,6 +190,61 @@ describe("CursorPermissions", () => {
       expect(parsed.version).toBe(1);
     });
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@zerone0x Thank you!

PR #1738, opened 2026-06-03
fix(mcp): clarify generate no-op result so zero count is not read as failure

dyoshikawa:
## Summary

A user reported that `generate` in the rulesync MCP server "wasn't working." Investigation showed that `generate` **does work** — it is idempotent. `writeAiFiles` (`src/types/feature-processor.ts`) only writes files whose content actually changed, so re-running it when outputs are already up to date returns `success: true` with `totalCount: 0`. Callers (especially LLM agents) misread that zero count as a silent failure.

This PR makes the no-op outcome explicit so a zero count is no longer mistaken for a broken generate.

## Changes

- **`src/mcp/generate.ts`**
  - Add a `message` field to `McpGenerateResult`.
  - `buildGenerateMessage` returns `Generated N file(s) ...` when files were written, and an explicit "already up to date — this is a successful no-op, not a failure" message when `totalCount` is 0.
  - Document the idempotent behavior in the `executeGenerate` tool description so agents interpret a zero count correctly.
- **`src/mcp/generate.test.ts`** — Unit tests for both the "generated" and "no-op" messages.
- **`src/e2e/e2e-mcp-server.spec.ts`** — Add daemon-level e2e coverage for `generate` (previously only `permissions`/`hooks` were covered): `rule put` → `generate` verifies the output file is written, and a second `generate` verifies the successful no-op.

## Test Plan

- [x] `pnpm cicheck` passes (fmt, oxlint, eslint, typecheck, 5846 unit tests, cspell, secretlint)
- [x] `pnpm test:e2e src/e2e/e2e-mcp-server.spec.ts` passes (4 tests, incl. new generate round-trip)
- [x] Verified the live MCP daemon (`pnpm dev mcp`) generates files for fresh targets and returns the new no-op message on re-run

## Note

No behavior change to the generated files themselves — this is purely a clarity/observability improvement to the MCP `generate` result.

### DIFF
--- a/src/e2e/e2e-mcp-server.spec.ts
+++ b/src/e2e/e2e-mcp-server.spec.ts
@@ -211,6 +211,64 @@ describe("E2E: mcp server (daemon over stdio JSON-RPC)", () => {
   });
 
   it(
+    "should generate output files through the daemon and report a no-op on re-run",
+    { timeout: 30_000 },
+    async () => {
+      const testDir = getTestDir();
+
+      const connection = await connectRulesyncMcpServer(testDir);
+      close = connection.close;
+      const { client } = connection;
+
+      // Seed a rule via the MCP tool so generate has something to work with.
+      const putRule = await client.callTool({
+        name: "rulesyncTool",
+        arguments: {
+          feature: "rule",
+          operation: "put",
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #1707, opened 2026-05-26
Support symlinked shared skills in multi-directory repositories


mocyuto:
When using Rulesync in multiple directories within the same repository, it is common to reuse the same skills across those directories.

To avoid duplicating identical skill directories, I tried centralizing shared skills in one location and referencing them from each directory via symbolic links. However, Rulesync does not detect those symlinked skills.

## Expected behavior

Rulesync should detect and process skills that are provided via symbolic links, so shared skills can be centralized and reused across multiple directories without duplication.

## Actual behavior

Symlinked skill directories are not detected, so the shared skills are effectively ignored.

## Use case

In a repository with multiple subprojects or working directories, the same skill may need to be available in more than one place. Managing duplicated copies of the same skill is error-prone and makes maintenance harder.

Supporting symlinked skills would make it easier to:

    centralize shared skill definitions
    avoid duplication across directories
    keep updates consistent across all consumers of the same skill

## Notes

It looks like Rulesync currently does not follow symbolic links in at least some file discovery paths, which may be the reason the skills are not detected.

If full symlink support is not desirable by default for safety reasons, it may still be useful to support one of these options:

    follow symlinked skill directories by default
    add an opt-in flag or config option to allow symlink traversal
    document symlink behavior explicitly and recommend an alternative pattern for shared skills

## Possible alternative

The --input-root workflow helps for centralized rule sources, but it does not fully solve the case where shared skills are reused across multiple directories in the same repository structure.

saitota:
I hit the same issue today.

In my case, a symlinked skill directory was partially exported: `SKILL.md` was exported, but other files under the skill directory were not.

After a quick look at the implementation, this may not be limited to skills. `findFilesByGlobs` uses `followSymbolicLinks: false`, so other features that rely on glob-based discovery, such as `commands` ,  `rules` ,and tool-side `subagents` , may hit a similar issue.

Skills may just be where this is more likely to surface because they often use subdirectories such as `docs/` or `scripts/`.

saitota:
Following up on my comment — I realized afterward that my case is actually the reverse direction of the original issue.

The original report is about symlinking **skill → skill** (sharing in skill dir), where the link structure should be preserved on export.

Mine is the opposite: the real files live **outside** the skills dir (a dotfiles repo), symlinked *into* the skill. There I'd want Rulesync to **follow the symlink and export the resolved file as a regular file**, not keep it as a link.

[...]

PR #1802, opened 2026-06-09
fix(file): follow symbolic links in findFilesByGlobs

saitota:
Refs #1707

## Summary

- Change `followSymbolicLinks: false` to `true` in `findFilesByGlobs`

## Scope

`findFilesByGlobs` is the single discovery function used by **all features** — rules, commands, subagents, hooks, and skills. This one-line fix covers every feature, not just skills.

Two distinct symlink scenarios are resolved:

1. A symlinked skill/rule/command **directory** under `.rulesync/` was not detected at all (original report by @mocyuto)
2. A file **inside** a skill directory that is a symlink (e.g. pulled in from a dotfiles repo) was not exported (reported in comments by @saitota)

## Design note

The alternative — preserving the symlink itself in the output — was considered but does not fit the architecture. `findFilesByGlobs` feeds into `AiDir.collectOtherFiles`, which reads every file as a `Buffer` and re-writes it as a plain file via `writeFileContent`. There is no layer in the pipeline that carries symlink metadata, so the output is always a regular file regardless. Following and copying is the natural behavior here.

## Test plan

- [x] `pnpm vitest run src/utils/file.test.ts` — added two tests: symlinked file is found, symlinked directory and files inside it are found


dyoshikawa:
@saitota Thank you!

PR #1801, opened 2026-06-09
fix(codexcli): restore sandbox baseline fields lost on permissions round-trip

saitota:
Fix #1733 — Codex CLI permission profiles lose sandbox baseline fields on round-trip.

## Background

Codex CLI v0.131.0 renamed `:project_roots` → `:workspace_roots` and removed legacy parsing ([#22624](https://github.com/openai/codex/pull/22624)). v0.133.0 added `extends` inheritance resolution ([#22270](https://github.com/openai/codex/pull/22270)) and made `deny` the canonical filesystem value ([#23493](https://github.com/openai/codex/pull/23493)). v0.134.0 removed legacy profile v1 resolution entirely ([#24051](https://github.com/openai/codex/pull/24051), [#24055](https://github.com/openai/codex/pull/24055)).

After these changes, a functional Codex profile requires:

- **`extends = ":workspace"`** — inherits workspace-write baseline protections; without it, writes inside workspace roots are not covered ([Codex permissions docs](https://developers.openai.com/codex/permissions))
- **`network.enabled = true`** — defaults to `false`; `network.domains` entries are silently ignored unless this is set ([Codex permissions docs](https://developers.openai.com/codex/permissions#network-permissions))
- **`filesystem.":workspace_roots"."." = "write"`** — scopes filesystem write rules to workspace roots

Rulesync modelled `network` as `domains` only and had no concept of `extends` or `network.enabled`, so these fields were lost on every import/generate cycle regardless of the source tool.

## Changes

- Emit `extends = ":workspace"` when any `edit`/`write` rule is present
- Emit `network.enabled = true` alongside `network.domains` when any `webfetch` rule is present
- Import `network.enabled = true` (no domains) as `webfetch: { "*": "allow" }`
- Preserve `description`, `network.mode`, `network.unix_sockets` from existing config on regeneration

The converter infers these values from existing rule semantics rather than extending `permissions.json` with Codex-specific fields. This keeps the schema tool-agnostic and guarantees idempotency: any round-trip (Claude Code → Rulesync → Codex, Cursor → Rulesync → Codex, etc.) produces the same result.

Issue #1733, opened 2026-06-03
Codex CLI permissions export can lose network and filesystem sandbox baselines


saitota:
## Background

Codex CLI permission profiles have sandbox-level fields such as `extends`, `filesystem`, and `network`:
https://github.com/openai/codex/blob/main/codex-rs/config/src/permissions_toml.rs

For example, `NetworkToml` has both `enabled` and `domains`, and `enabled` is applied separately from domain rules:
https://github.com/openai/codex/blob/main/codex-rs/config/src/permissions_toml.rs

```toml
[permissions.rulesync]
extends = ":workspace"

[permissions.rulesync.filesystem.":workspace_roots"]
"." = "write"

[permissions.rulesync.network]
enabled = true
```

This became more visible after Codex CLI moved further toward permission profiles: 0.133.0 strengthened permission profiles, and 0.134.0 removed legacy profile v1 resolution/write paths.
https://github.com/openai/codex/releases/tag/rust-v0.133.0
https://github.com/openai/codex/releases/tag/rust-v0.134.0

## Concern

Rulesync's canonical permissions model only represents tool/category rules with `allow | ask | deny` actions:
https://github.com/dyoshikawa/rulesync/blob/main/src/types/permissions.ts

```json
{
  "permission": {
    "edit": { "*": "allow" },
    "write": { "*": "allow" },
    "webfetch": { "*": "allow" }
  }
}
```

The Codex CLI converter currently models `network` as `domains` only, so `network.enabled` cannot be preserved:
https://github.com/dyoshikawa/rulesync/blob/main/src/features/permissions/codexcli-permissions.ts

As a result, Codex-specific sandbox baseline settings can be lost during import/export. This is not source-tool specific: a Codex CLI → Rulesync → Codex CLI round trip can also drop `network.enabled = true`.

## Additional Note

Rulesync may also emit invalid Codex filesystem values like:

```toml
[permissions.rulesync.filesystem.":workspace_roots"]
"*" = "allow"
```

Codex filesystem entries are parsed as `FileSystemAccessMode`, not Rulesync permission actions:
https://github.com/openai/codex/blob/main/codex-rs/config/src/permissions_toml.rs

Observed error:

```text
Error loading config.toml: data did not match any variant of untagged enum FilesystemPermissionToml
in `permissions`
```

Related Issues
https://github.com/dyoshikawa/rulesync/issues/1716
https://github.com/dyoshikawa/rulesync/issues/1716

PR #1810, opened 2026-06-10
fix(codexcli): scope :workspace extends to workspace-wide writes and import extends-only profiles

### LINKED ISSUE #1809: Review findings for PR #1801 (codexcli permissions round-trip): extends over-grant, extends-only import loss, and related cleanups
# Background

PR #1801 (`fix(codexcli): restore sandbox baseline fields lost on permissions round-trip`, branch `i1733`, addressing #1733) was reviewed with parallel code-quality and security review passes. The reviews cross-checked the implementation against the upstream Codex CLI source (openai/codex), notably `PermissionProfile::workspace_write()` and the network sandbox config loader. No critical/high findings; CI is fully green (Code Quality & Tests + 3-OS E2E). However, several mid/low findings are worth tracking so they are not lost after the PR merges. All findings concern `src/features/permissions/codexcli-permissions.ts` (line numbers refer to the PR head).

# Details

## Mid severity

1. **`extends = ":workspace"` is emitted too broadly, widening permissions.** The `:workspace` baseline grants write to the entire workspace root plus `/tmp` (verified in Codex's `PermissionProfile::workspace_write()`), not just `.git`/`.codex` read-only protections. Because `hasWriteRules` (PR lines ~259-264) triggers on any write rule — including narrowly scoped ones like `edit: { "src/**": "allow" }` or even workspace-external ones like `edit: { "~/notes/**": "allow" }` — the generated profile grants workspace-wide write the user never asked for. The PR description's rationale ("without it, writes inside workspace roots are not covered") does not match Codex's implementation: `filesystem` entries grant access without `extends`.

2. **Profiles granting write solely via `extends = ":workspace"` are still lost on import.** `convertCodexProfileToRulesync` (PR lines ~270-304) never converts `extends` into rulesync rules. Importing a profile that has `extends = ":workspace"` but no `filesystem` table yields no write rules, so regeneration drops `extends` too and a working profile degrades to default-deny (only a warn is logged). This is a remaining form of the round-trip loss reported in #1733. Suggested fix: import `extends = ":workspace"` as the equivalent of `edit: { ".": "allow" }` so regeneration converges to the `extends` + `":workspace_roots"."." = "write"` shape.

3. **Internal flag `domainsHadUnknown` lives inside the serialized `CodexNetwork` type** (PR lines ~29-37, 321, 339-340). The TOML-serialized type carries a "do not serialize" flag; the destructuring removal in `mergeWithExistingProfile` is the only guard (and is effectively dead code today since no path sets the flag on `newProfile.network`). If `toCodexProfile`'s return value is ever serialized directly, `domainsHadUnknown = true` leaks into `config.toml`. Suggested fix: return metadata out-of-band, e.g. `{ profile, hadUnknownDomains }`, and drop the flag from the type. This also removes the codebase's first `eslint-disable` comment (the project uses oxlint).

## Low severity

4. **Asymmetric import behavior for missing `network.enabled`** (PR line ~290, `enabled !== false`): in Codex, a missing `enabled` means restricted (same as `false`, domains inactive), but the implementation imports domains only when the key is absent. Likely intentional rescue for older rulesync output — document the rationale in a comment.
5. **Unreachable branch in `mergeWithExistingProfile` description merge** (PR lines ~354-358): `convertRulesyncToCodexProfile` never sets `description`, so simplify to `newProfile.description ?? existingProfile.description`.
6. **Coding guideline: multiple positional args** — `mergeWithExistingProfile(newProfile, existingProfile)` (PR lines ~333-336) should take an object argument per `coding-guidelines.md`.
7. **`unix_sockets` preservation silently filters unknown values and warns on every generate** (PR lines ~314, 100-104): values outside allow/deny are dropped without warning (unlike domains), and the preservation warn fires on each generate run, creating noise. If the field is pass-through-only, preserving the raw table is more faithful.
8. **Deny-only webfetch rules enable the network** (PR line ~266): a rule set like `webfetch: { "evil.com": "deny" }` emits `network.enabled = true`, and since unlisted domains default to allow in Codex, the result is "everything except the listed domains is reachable" — wider than the previous behavior (network fully disabled). Consider emitting `enabled = true` only when at least one allow rule exists, or warning on deny-only sets.
9. **Existing `network.mode` is preserved without a warning** (PR lines ~342-343): `unix_sockets` preservation logs a warning, but `mode` (e.g. `mode = "full"`, broader than domain restrictions) is silently carried over on every regeneration. Emit the same kind of warning.

## Adjacent pre-existing bug (not introduced by the PR)

10. **`webfetch: { "*": "deny" }` generates `domains."*" = "deny"`, which Codex rejects at config load time** — the generated `config.toml` makes Codex itself fail to start. Worth a guard (warn + skip) since this exact mapping area is being touched.

# Solution / Next Steps

- Fix findings 1-3 in PR #1801 before merge, or record an explicit design decision (especially for 1 and 2, which affect permission widening and round-trip completeness).
- For finding 1, restrict `hasWriteRules` to workspace-root writes and document in `docs/reference/file-formats.md` (and the synced `skills/rulesync/file-formats.md`) that `:workspace` grants workspace-wide + `/tmp` write, not just baseline protections.
- For finding 2, add an import path for `extends = ":workspace"` → `edit: { ".": "allow" }` with a round-trip test.
- Address the low-severity items (4-9) opportunistically in the same PR or a follow-up.
- File/handle the pre-existing `"*" = "deny"` startup-breaking bug (finding 10) as its own fix: warn and skip emitting a global deny wildcard.

Refs: PR #1801, Issue #1733.

dyoshikawa:
# Background

Issue #1809 collected the mid/low review findings from PR #1801 (codexcli permissions round-trip, #1733) so they would not be lost after merge. The findings were re-validated against the upstream Codex CLI source and docs before this fix:

- Codex's `:workspace` baseline grants read access to the whole filesystem and write access to the entire workspace root plus `/tmp`/`$TMPDIR`, and `filesystem` entries grant access on their own without `extends` (verified against `openai/codex` `PermissionProfile::workspace_write()` / `FileSystemSandboxPolicy::workspace_write`). Emitting `extends = ":workspace"` for any write rule therefore over-grants.
- Profiles granting write solely via `extends = ":workspace"` were dropped on import, degrading to default-deny on regeneration (remaining form of #1733).
- The global `*` wildcard in `network.domains` is rejected at Codex config load time in both allow and deny positions ([network-proxy README](https://github.com/openai/codex/blob/main/codex-rs/network-proxy/README.md)), so the previous round-trip shape `enabled = true` → `webfetch: { "*": "allow" }` → `domains."*" = "allow"` generated a config Codex refuses to start with.
- One premise in finding 8 of the issue turned out to be incorrect: unlisted domains are denied (allowlist-first), not allowed. The fix still emits `enabled = true` only when an allow rule exists, which is behavior-preserving and keeps deny-only sets round-trippable.

# Changes

- Emit `extends = ":workspace"` only for workspace-wide write rules (`.`, `./`, `**`, `./**`); narrow or workspace-external writes are expressed purely as `filesystem` entries (finding 1)
- Import `extends = ":workspace"` as `edit: { ".": "allow" }` with a round-trip test so extends-only profiles converge instead of being lost (finding 2)
- Return `domainsHadUnknown` out-of-band from profile parsing instead of carrying it inside the serialized `CodexNetwork` type; removes the `eslint-disable` workaround (finding 3)
- Document the `enabled !== false` import rationale (finding 4)
- Simplify the description merge and use an object argument for `mergeWithExistingProfile` (findings 5, 6)
- Preserve `unix_sockets` values verbatim instead of silently filtering unknown values (finding 7)
- Emit `network.enabled = true` only when at least one allow rule exists; deny-only domain sets are emitted without `enabled` and still re-import (finding 8, with corrected premise)
- Warn when preserving an existing `network.mode` (finding 9)
- Guard global wildcard domain entries: `webfetch: { "*": "allow" }` becomes `enabled = true` with no domain entry (lossless round-trip with the existing import path), `webfetch: { "*": "deny" }` is skipped with a warning (finding 10)
- Sync `docs/reference/file-formats.md` and `skills/rulesync/file-formats.md` with the new behavior

`pnpm cicheck` passes locally (the content check passes once the synced skills file is committed).

Closes #1809


dyoshikawa:
@dyoshikawa Thank you!

PR #1811, opened 2026-06-10
fix(file): dedup symlink-cycle glob results and clarify symlink trust model

### LINKED ISSUE #1808: Follow-ups from PR #1802 review: symlink following in findFilesByGlobs (cycle dedup, comment fix, boundary check, tests, docs)
## Background

PR #1802 (`fix(file): follow symbolic links in findFilesByGlobs`) resolves issue #1707 by flipping `followSymbolicLinks` from `false` to `true` in `findFilesByGlobs` (`src/utils/file.ts`). A code review and a security review of the PR surfaced several follow-up items that are not blockers for the PR itself but should be addressed afterward. This issue consolidates those findings so they are not lost.

All CI checks on the PR were green (Build binaries, Code Quality & Tests, E2E on ubuntu/macos/windows). No high/critical findings; the items below are mid/low severity.

### #1 [mid] Circular symlinks silently produce massively duplicated output

- Location: `src/utils/file.ts` (`followSymbolicLinks: true`), affecting `src/types/ai-dir.ts` (`join(dirPath, "**", "*")`)
- Verified empirically with the repo's globby version: if the input tree contains a directory symlink cycle (e.g. a link inside a skill directory pointing to an ancestor), `**` traversal follows it up to the kernel ELOOP limit (~40 levels) and silently returns ~40x duplicated paths. `AiDir.collectOtherFiles` then reads and writes all of them, producing a deeply nested duplicated tree in the generated output. No crash, no warning. This is a regression relative to the old `followSymbolicLinks: false` behavior. The security review also flagged the same spot as a low-severity resource-exhaustion (DoS) vector, mitigated by the fact that the symlink must live inside the user-controlled input tree.

### #2 [mid] New comment misstates why git-client.ts skips symlinks

- Location: `src/utils/file.ts` (comment added in PR #1802)
- The comment claims git-client.ts skips symlinks because "unresolved symlinks are meaningless in a bare clone". This is factually wrong: `src/lib/git-client.ts` performs a non-bare clone (`--no-checkout --filter=blob:none` followed by `git checkout`), so a real working tree is materialized and symlinks are resolvable. The actual reason for skipping is a security hardening measure introduced in commit `51bf0443` ("address security review feedback for git transport") to avoid following symlinks from untrusted remote repositories. Leaving the wrong rationale in place risks someone later removing the skip "for consistency" and silently dropping a security control.

### #3 [mid] Symlink targets outside the input root are copied into generated output; `checkPathTraversal` does not catch it

- Location: `src/utils/file.ts`
- A repository that commits a symlink like `.rulesync/skills/x/creds.md -> ~/.aws/credentials` causes the link target's content to be copied verbatim into generated output when the user runs `rulesync generate`. `checkPathTraversal` validates only the symlink's own path (which is inside the root), so it does not detect that the dereferenced content originates outside the boundary. Note: the remote-fetch path (`git-client.ts` `walkDirectory`) skips symlinks and writes plain files, so untrusted remote content is NOT an attack vector here; the behavior is a documented, intentional trade-off (security review rated it low/informational). However, issue #1707's primary use case is sharing skills within the same repository, which a `realpath`-based containment check against the repository root would still satisfy.

### #4 [low] No processor/E2E-level test for the issue #1707 scenario

- Location: `src/utils/file.test.ts` (new tests)
- The two added tests pin the glob behavior at the unit level, but there is no end-to-end test covering the reported scenario (a symlinked skill directory under `.rulesync/skills/` being emitted by `generate`). A `skills-processor.test.ts` case loading a symlinked skill directory would make this more robust.

### #5 [low] New tests may fail on local Windows environments

- Location: `src/utils/file.test.ts` (`symlink()` calls)
- `fs.symlink` with the file type requires administrator privileges or Developer Mode on Windows. CI unit tests run on ubuntu only, so CI passes, but Windows contributors running `vitest` locally may hit `EPERM`.

### #6 [low] Symlink behavior documented only in the `--input-root` guide

- Location: `docs/guide/separate-input-root.md`, `skills/rulesync/separate-input-root.md`
- The behavior change applies to all features and to regular `.rulesync/` usage, and issue #1707's main use case does not involve `--input-root`. Users who never read that guide cannot discover the new behavior.

## Solution / Next Steps

1. (#1) Deduplicate glob results by `realpath` (keep the first path per real file), or pass a `deep` bound to globby for recursive `**` callers plus a warning log. Add a test with a circular symlink.
2. (#2) Rewrite the comment in `src/utils/file.ts` to state that git-client.ts skips symlinks as a security hardening for untrusted remote content, while `findFilesByGlobs` operates on user-specified local trees inside the trust boundary.
3. (#3) Consider a `realpath`-based containment check that resolved symlink targets must stay within the repository/working-tree root; at minimum record the accept/reject decision for this trade-off.
4. (#4) Add a `skills-processor.test.ts` case for a symlinked skill directory.
5. (#5) Guard the symlink tests with `it.skipIf(process.platform === "win32")` or use the `junction` type for directory symlinks.
6. (#6) Add a short "Symlinks" note to `docs/reference/file-formats.md` (or FAQ), keeping `docs/` and `skills/rulesync/` in sync.

Ref: PR #1802, issue #1707.

dyoshikawa:
# Background

Issue #1808 consolidated the follow-up findings from the code and security reviews of PR #1802 (`fix(file): follow symbolic links in findFilesByGlobs`, resolving #1707). None blocked that PR, but they were tracked so they would not be lost. Each was re-validated against the current code before this fix.

# Changes

- **#1 (mid) — circular symlink duplication.** `findFilesByGlobs` now deduplicates results by real path (`realpath`), keeping the first path per real file in sorted order. A directory symlink cycle (which globby follows up to the kernel ELOOP limit, ~40 levels) no longer yields ~40× duplicated entries that `AiDir.collectOtherFiles` would read and re-emit. Added a circular-symlink test and a link-and-target-collapse test.
- **#2 (mid) — wrong comment rationale.** Confirmed `src/lib/git-client.ts` does a **non-bare** clone (`--no-checkout --filter=blob:none` then `git checkout`), so symlinks are resolvable — the old comment ("unresolved symlinks in a bare clone are meaningless") was factually wrong. Rewrote it to state the real reason: `walkDirectory` skips symlinks as a **security hardening** for untrusted remote content (commit `51bf0443`), a distinct code path from the local, trusted `findFilesByGlobs`.
- **#3 (mid) — realpath containment.** Recorded an explicit design decision instead of adding a containment check: enforcing `realpath` containment against a single root would break issue #1707's use case of sharing files that live elsewhere in the same repository (and `--input-root` trees legitimately point outside themselves). The remote-fetch path already skips symlinks, so untrusted content is not an attack vector. Documented in the code comment and the new docs section.
- **#4 (low) — E2E-level coverage.** Added a `skills-processor.test.ts` case that loads a skill directory which is a symlink to a real directory elsewhere.
- **#5 (low) — Windows.** Guarded the symlink unit tests with `describe.skipIf(process.platform === "win32")` / `it.skipIf(...)` since `fs.symlink` needs admin or Developer Mode on Windows.
- **#6 (low) — discoverability.** Added a feature-wide **Symlinks** section to `docs/reference/file-formats.md` (covering all features and plain `.rulesync/` usage, not just `--input-root`), cross-linked it from the input-root guide, and synced `skills/rulesync/`.

`pnpm cicheck` passes locally (6194 tests; the content sync-check passes once the synced skills files are committed).

Closes #1808


dyoshikawa:
@dyoshikawa Thank you!

Issue #1783, opened 2026-06-09
Follow up Kiro upstream updates: IDE Markdown subagents, `.kiro.hook` hooks, steering `inclusion`, and global scope


dyoshikawa:
## Summary

rulesync's `kiro` adapters target a mix of the Kiro **CLI** (JSON agent config) and **IDE** surfaces, and several recent Kiro IDE capabilities are not followed: IDE subagents/hooks use Markdown / `.kiro.hook` files (rulesync emits CLI JSON instead), steering files never emit `inclusion` frontmatter, and global (`~/.kiro/`) scope is unsupported across the board. A first question for the maintainer is whether `kiro` should represent the IDE or the CLI.

## Recent Releases

- Kiro IDE v0.9 (2026-02-05), https://kiro.dev/changelog/ide/0-9/ — Markdown subagents and new hook tool-use triggers.
- Kiro CLI v2.3.0 (2026-05-12) — `KIRO_HOME` env var for relocating global agents/prompts/skills/steering/settings, https://kiro.dev/changelog/
- Steering docs (updated 2026-02-18): https://kiro.dev/docs/steering/
- Subagents: https://kiro.dev/docs/chat/subagents/ · Hooks: https://kiro.dev/docs/hooks/ and https://kiro.dev/docs/hooks/types/

## Gaps

- **subagents — IDE uses Markdown, rulesync emits CLI JSON (`project`).** Kiro IDE custom subagents are `.kiro/agents/*.md` (YAML frontmatter: `name` required; optional `description`, `tools`, `model`, `includeMcpJson`, `includePowers`). `src/features/subagents/kiro-subagent.ts` emits `.kiro/agents/<name>.json` in the CLI agent-config schema, which the IDE does not read as agents.
[...]

PR #1841, opened 2026-06-11
feat(kiro): map rule globs to steering inclusion frontmatter

dyoshikawa:
## Summary

Maps rulesync rule frontmatter onto Kiro's steering `inclusion` modes so `.kiro/steering/*.md` files are no longer all implicitly always-on.

Kiro reads an `inclusion` frontmatter block at the top of each steering file (`always`, `fileMatch` with a `fileMatchPattern`, or `manual`) to decide when it is loaded. Previously rulesync emitted every steering file without that block, so each was treated as `always`.

This is the self-contained slice of #1783 that does **not** depend on the IDE-vs-CLI framing decision. The other items (IDE Markdown subagents, `.kiro.hook` hooks, global scope, and the Powers bundle) remain a maintainer call and are intentionally **not** addressed here.

## Mapping

For non-root steering files (`deriveKiroInclusion`):
1. An explicit `kiro.inclusion` block round-trips as-is (with `fileMatchPattern` from the block, or derived from `globs` for `fileMatch`).
2. Otherwise specific (non-wildcard) `globs` map to `inclusion: fileMatch` with a comma-joined `fileMatchPattern`, scoping the rule to matching files.
3. Otherwise the file stays `always` — written plain (no block), matching Kiro's no-frontmatter default.

The root overview index stays plain so Kiro always loads it.

## Changes

- `deriveKiroInclusion` + reworked `KiroRule.fromRulesyncRule` / `toRulesyncRule` (round-trips inclusion into the rulesync `kiro` block + `globs`).
- New `kiro` frontmatter block on the rulesync rule schema (`inclusion`, `fileMatchPattern`).
- Unit tests for `deriveKiroInclusion` and fromRulesyncRule/toRulesyncRule round-trip; docs updated (skill docs synced).

## Verification

- `pnpm cicheck:code` passes (251 files, 6259 tests).
- Manual generate confirmed: glob-scoped rules get `inclusion: fileMatch`; globless rules stay plain.

Part of #1783

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -40,7 +40,7 @@ augmentcode: # augmentcode specific parameters
   description: "When to apply this rule" # (optional) used with "agent_requested" type
 kiro: # kiro specific parameters (steering inclusion)
   inclusion: "fileMatch" # always, fileMatch, or manual
-  fileMatchPattern: "src/components/**/*.tsx" # (optional) single glob used when inclusion is "fileMatch"
+  fileMatchPattern: ["src/components/**/*.tsx"] # (optional) glob string or array of globs, used when inclusion is "fileMatch"
 takt: # takt specific parameters (optional; emitted under .takt/facets/policies/ — frontmatter is dropped on emit)
   name: "renamed-stem" # (optional) override the emitted filename stem (no path separators or "..")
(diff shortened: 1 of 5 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1856, opened 2026-06-13
refactor: centralize per-tool path conventions into constants modules

saitota:
## Background

Each external tool's configuration conventions — directory names, file names, scope-specific paths — were independently hardcoded as string literals across up to 9 feature files per tool (`{tool}-rule.ts`, `{tool}-command.ts`, `{tool}-skill.ts`, `{tool}-mcp.ts`, etc.). There was no programmatic connection between these scattered literals; they were implicit duplicates of the same knowledge.

This created a concrete maintenance burden: when a tool renames its config directory (e.g. the Gemini CLI → Antigravity migration that produced `antigravity-rule.ts` / `antigravity-cli-rule.ts` / `antigravity-ide-rule.ts` and legacy aliases), up to 9 files required changes with no compile-time safety net. A missed file produces a silent mismatch in generated output — no type error, no test failure.

## Changes

- Add `src/constants/{tool}-paths.ts` for every supported tool (27 new files), each exporting the tool's directory names, file names, and scope-specific paths as named constants.
- Update all `src/features/**/{tool}-*.ts` files to import path constants from the corresponding module instead of repeating string literals.

No runtime behavior changes. Path values are identical — this is a pure refactor.

## Why this structure

The feature-first directory layout (`src/features/rules/`, `src/features/commands/`, …) is preserved. Only the volatile knowledge — where a tool stores its files — is extracted to a single authoritative location per tool. A tool path change now requires editing one file; all dependents get a compile-time error if they break.

## Test plan

- [x] `pnpm build` — clean
- [x] `pnpm typecheck` — clean
- [x] `pnpm test` — 6301/6302 pass (1 pre-existing failure in `warp-permissions.test.ts` unrelated to this change)
- [x] `pnpm oxlint` — clean
- [x] `pnpm fmt` — applied


dyoshikawa:
@saitota Thank you!

PR #1858, opened 2026-06-13
test(generate): add regression tests for execution-order constraints on shared output files

saitota:
## Summary

- Add two regression tests to `generate.test.ts` that pin the implicit execution-order constraints documented in `generate.ts:188-204`
- `IgnoreProcessor` must run before `PermissionsProcessor` — both write to `.claude/settings.json` and reversing the order silently drops ignore-generated `deny` entries
- `McpProcessor` must run before `RulesProcessor` — both write to `kilo.jsonc` and reversing the order silently drops the `mcp`/`tools` block
- Each test asserts that both `writeAiFiles` mocks are actually invoked (`toHaveBeenCalled`), then compares `invocationCallOrder` to catch any reordering or parallelization regression

## References

- `src/lib/generate.ts` lines 188–204 — existing NOTEs that describe the constraint in prose

## Test plan

- [x] `pnpm test` — 53 tests pass (2 new)

### DIFF
--- a/src/lib/generate.test.ts
+++ b/src/lib/generate.test.ts
@@ -608,6 +608,72 @@ describe("generate", () => {
     });
   });
 
+  describe("execution order constraints", () => {
+    it("should run ignore before permissions to prevent shared settings.json data loss", async () => {
+      mockConfig.getFeatures.mockReturnValue(["ignore", "permissions"]);
+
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@saitota Thank you!

PR #1859, opened 2026-06-13
test: add consistency checks for processor tool target schemas

saitota:
## Background

The knowledge of which tools are supported by which processor feature is scattered across multiple places: each processor's hardcoded tool list (e.g. \`rulesProcessorToolTargets\` in \`rules-processor.ts\`), each processor's factory \`Map\`, and \`ALL_TOOL_TARGETS\` in \`src/types/tool-targets.ts\`. There was no runtime or test-time check that these lists stay in sync.

A concrete failure mode: adding a new tool to \`ALL_TOOL_TARGETS\` without updating a processor's schema list produces no type error and no test failure — the tool silently generates nothing for that feature.

## Changes

- Add a \`"processor tool target consistency"\` describe block to \`src/types/tool-targets.test.ts\` that verifies all 8 processor \`*ProcessorToolTargetSchema.options\` are subsets of \`ALL_TOOL_TARGETS\`
- Export \`toolRuleFactories\` from \`rules-processor.ts\` and add a parity test that \`RulesProcessorToolTargetSchema.options\` and \`toolRuleFactories\` keys are equal sets — catching drift between the hardcoded enum array and the factory Map at runtime

No production behavior changes. Test-only addition (plus one \`export\` keyword on an existing constant).

## Test plan

- [x] \`pnpm test\` — 9 new tests pass
- [x] \`pnpm oxlint\` — clean
- [x] \`pnpm fmt\` — applied
- [x] \`pnpm typecheck\` — clean (pre-existing errors in \`scripts/\` unrelated to this change)

PR #1867, opened 2026-06-15
test: extend processor schema-vs-factory parity checks to all processors

### LINKED ISSUE #1866: Follow-ups from PR #1859: extend processor schema parity checks and clarify subset guarantee
## Background

PR #1859 ("test: add consistency checks for processor tool target schemas") was merged. It adds consistency-check tests in `src/types/tool-targets.test.ts` and exports `toolRuleFactories` from `rules-processor.ts` so the test can compare the schema enum against the factory map keys. Security review found nothing, CI is green, and the only production change is widening visibility of an existing const, so it was merged. This scrap records low-severity clarity/coverage follow-ups.

## Details

**1. (low) Subset test does not catch the failure mode named in the PR description.**
The PR body cites the motivating failure as "adding a new tool to `ALL_TOOL_TARGETS` without updating a processor's schema list." The subset tests assert the opposite direction — `schema.options ⊆ ALL_TOOL_TARGETS` (catches a target in a processor missing from `ALL_TOOL_TARGETS`). The reverse (`ALL_TOOL_TARGETS ⊆ schema.options`) is intentionally not asserted because not every tool supports every feature — a correct design choice, but it means the suite does not guard the headline scenario. Clarify the description.

**2. (low) Only RulesProcessor gets the schema-vs-factory parity check.**
The strong guarantee (`schema.options` set-equals the factory `Map` keys) is applied only to `RulesProcessor`; the other 7 processors get the weaker subset check. If other processors use analogous factory maps keyed by tool target, equivalent drift there would go undetected. Extend the parity pattern, or document why only rules has it.

**3. (low) `.toSorted()` relies on a recent runtime (ES2023).**
Fine in practice given modern Node, but worth a glance for consistency with the rest of the codebase's sorting idioms.

**4. (low) The `processors` array is hand-maintained.**
The list of 8 processors is manual; a 9th added later won't be auto-included, reintroducing the "scattered list drift" the PR targets, at the test level. Acceptable without a central registry, but worth a comment noting the footgun.

## Solution / Next Steps

- (1) Soften the PR/test description to reflect the subset (not superset) guarantee, or add a documented allowlist-based superset check if desired.
- (2) Extend the schema-vs-factory parity check to the other processors that use a factory map, or add a comment explaining the asymmetry.
- (4) Add a comment by the `processors` array noting it must be kept in sync, or derive it from a central source if one becomes available.

dyoshikawa:
## Background

Follow-ups from PR #1859, which added consistency checks for processor tool target schemas. That PR applied the strong schema-vs-factory parity check to `RulesProcessor` only and left low-severity clarity/coverage items for later. This PR addresses them.

## Changes

- **Extend parity check to every processor (item 2).** Export the seven remaining `tool*Factories` maps (commands, hooks, ignore, mcp, permissions, skills, subagents) — matching the visibility already given to `toolRuleFactories` — and assert that each processor's schema enum set-equals its factory `Map` keys. This catches drift in either direction (a target in the schema with no factory, or a factory for a target the schema rejects) uniformly across all 8 processors, not just rules.
- **Clarify the subset guarantee (item 1).** Document in comments that the subset check intentionally guards only the "a target declared by a processor must exist in `ALL_TOOL_TARGETS`" direction; the reverse is deliberately not asserted because not every tool supports every feature.
- **Note the hand-maintained list footgun (item 4).** Add a comment that the `processors` list must be kept in sync until a central processor registry exists.

Item 3 (`.toSorted()` requiring ES2023) needs no action: `.toSorted()` is already the dominant sorting idiom across `src` (45 occurrences), so this is consistent with the codebase.

The only production change is widening the visibility of existing `const` factory maps to `export`; there is no behavioral change. All checks pass via `pnpm cicheck`.

Closes #1866

### DIFF
--- a/src/features/commands/commands-processor.ts
+++ b/src/features/commands/commands-processor.ts
@@ -99,7 +99,7 @@ export const CommandsProcessorToolTargetSchema = z.enum(commandsProcessorToolTar
  * Factory Map mapping tool targets to their command factories.
  * Using Map to preserve insertion order for consistent iteration.
  */
-const toolCommandFactories = new Map<CommandsProcessorToolTarget, ToolCommandFactory>([
+export const toolCommandFactories = new Map<CommandsProcessorToolTarget, ToolCommandFactory>([
   [
     "agentsmd",
     {
--- a/src/features/hooks/hooks-processor.ts
+++ b/src/features/hooks/hooks-processor.ts
@@ -101,7 +101,7 @@ type ToolHooksFactory = {
   supportsMatcher: boolean;
 };
 
-const toolHooksFactories = new Map<HooksProcessorToolTarget, ToolHooksFactory>([
+export const toolHooksFactories = new Map<HooksProcessorToolTarget, ToolHooksFactory>([
   [
     "antigravity-cli",
     {
--- a/src/features/ignore/ignore-processor.ts
+++ b/src/features/ignore/ignore-processor.ts
@@ -65,7 +65,7 @@ type ToolIgnoreFactory = {
   };
 };
 
-const toolIgnoreFactories = new Map<IgnoreProcessorToolTarget, ToolIgnoreFactory>([
+export const toolIgnoreFactories = new Map<IgnoreProcessorToolTarget, ToolIgnoreFactory>([
   ["antigravity-cli", { class: AntigravityCliIgnore }],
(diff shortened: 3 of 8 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1868, opened 2026-06-15
test(generate): add real data-loss regression tests for shared output files

### LINKED ISSUE #1865: Follow-ups from PR #1858: order tests don't cover the data-loss regression they describe
## Background

PR #1858 ("test(generate): add regression tests for execution-order constraints on shared output files") was merged. It adds two regression tests to `src/lib/generate.test.ts` pinning the execution-order constraints documented in `generate.ts` (ignore-before-permissions for `.claude/settings.json`, mcp-before-rules for `kilo.jsonc`). Security review found nothing and CI is green, so it was merged. This scrap records non-blocking quality follow-ups from review.

## Details

**1. (mid) Tests verify call order, not the data-loss behavior the description claims.**
The PR description frames the tests as guarding against "silently dropping deny entries" / "dropping the mcp/tools block." But the assertions only check `invocationCallOrder` of a mocked `writeAiFiles`. Because the processors are fully mocked, these tests still pass even if the real read-modify-write merge/preserve logic against `kilo.jsonc` / `settings.json` were broken. They are pure orchestration-order tests, acceptable at the `generate.ts` unit level, but they do not cover the underlying data-loss regression. A genuine data-loss regression test would need an integration test exercising the real processors against a temp dir.

**2. (low) The order constraint is minimally pinned.**
For the ignore/permissions pair the test only asserts `ignoreCall < permissionsCall`. That is the right minimal invariant, so no change is strictly needed — noted for completeness.

**3. (low) Inconsistent non-null assertions.**
`expect(ignoreCall).toBeLessThan(permissionsCall!)` uses `!` on `permissionsCall` but not on `ignoreCall`, though both are `number | undefined`. Style inconsistency only; the preceding `toHaveBeenCalled()` makes both safe.

**4. (low) Heavy mock duplication (DRY).**
Each test re-declares the full processor mock object inline, duplicating the existing `createMockProcessor()` helper (generate.test.ts:127-136). A helper like `createMockProcessorWith(writeAiFilesMock)` would cut ~30 lines and ease maintenance. Note one test sets `convertRulesyncFilesToToolFiles` to return `[]` (vs one file in the shared helper); this is safe because `writeAiFiles` is called unconditionally in `processFeatureGeneration` (generate.ts:60), but the divergence is unexplained.

## Solution / Next Steps

- (1) Consider adding an integration test that exercises the real Ignore/Permissions and Mcp/Rules processors against a temp dir to actually assert the deny-entry / mcp-block preservation, matching what the PR description claims. Alternatively, soften the description to "orchestration order" framing.
- (3) Make the non-null assertions consistent (use `!` on both, or rely on the `toHaveBeenCalled()` guard).
- (4) Extract a shared mock-processor helper to remove the inline duplication, and document why the empty-array override is safe.

dyoshikawa:
## Background

Follow-ups from PR #1858, which added execution-order regression tests for the shared output files `.claude/settings.json` (ignore before permissions) and `kilo.jsonc` (mcp before rules).

As noted in the review, those tests mock every processor and only assert `invocationCallOrder` of a mocked `writeAiFiles`. Because the processors are fully mocked, the tests still pass even if the real read-modify-write merge/preserve logic were broken — they are orchestration-order tests, not the data-loss guards the PR description framed them as.

## Changes

- **Add genuine data-loss regression tests** (`src/lib/shared-file-data-loss.test.ts`). These exercise the *real* tool-file writers against a temp dir, in the documented order, and assert both features' data survives the shared file:
  - **ignore → permissions**: `ClaudecodeIgnore` writes `Read(...)` deny entries, then `ClaudecodePermissions` merges in its `Bash` entries. Asserts the ignore-written Read entries are preserved alongside the new Bash entry (the deny-entry data-loss the description claims to guard).
  - **mcp → rules**: `KiloMcp.fromRulesyncMcp` writes the `mcp` block, then `KiloMcp.fromInstructions` (the rules feature's kilo.jsonc registration) adds `instructions`. Asserts the mcp block is preserved alongside the instructions (the mcp-block data-loss).
- **Remove the duplicated inline processor mocks (item 4 / DRY).** Extract a module-level `createMockProcessorWith(writeAiFiles)` helper in `generate.test.ts`, used by both order tests, and document why returning an empty `convertRulesyncFilesToToolFiles` is safe (`writeAiFiles` runs unconditionally in `processFeatureGeneration`).
- **Consistent assertions (item 3).** Replace the lone non-null assertion (`permissionsCall!`) with the `?? 0` idiom already used elsewhere in the file.

Item 2 (the order constraint being minimally pinned) needs no change — the minimal invariant is the correct one, as the review itself noted.

All checks pass via `pnpm cicheck`.

Closes #1865

### DIFF
--- a/src/lib/generate.test.ts
+++ b/src/lib/generate.test.ts
@@ -74,6 +74,22 @@ const createMockAiFile = (filePath: string, content: string) => ({
   getFileContent: () => content,
 });
 
+/**
+ * Build a minimal processor mock whose `writeAiFiles` is the supplied spy, used
+ * by the execution-order tests to observe invocation order on a shared output
+ * file. `convertRulesyncFilesToToolFiles` returns an empty array because these
+ * tests only assert call order, not generated content; `writeAiFiles` is invoked
+ * unconditionally in `processFeatureGeneration` (see generate.ts), so the empty
+ * array does not skip the write.
+ */
+const createMockProcessorWith = (writeAiFiles: ReturnType<typeof vi.fn>) => ({
+  loadToolFiles: vi.fn().mockResolvedValue([]),
+  removeOrphanAiFiles: vi.fn().mockResolvedValue(0),
+  loadRulesyncFiles: vi.fn().mockResolvedValue([{ file: "test" }]),
+  convertRulesyncFilesToToolFiles: vi.fn().mockResolvedValue([]),
+  writeAiFiles,
+});
+
 describe("generate", () => {
   let mockConfig: {
     getVerbose: ReturnType<typeof vi.fn>;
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1869, opened 2026-06-15
refactor(constants): dedupe and prune centralized path constants

### LINKED ISSUE #1864: Follow-ups from PR #1856: dedupe and prune centralized path constants
## Background

PR #1856 ("refactor: centralize per-tool path conventions into constants modules") was merged. It centralizes per-tool path string literals into dedicated `src/constants/*-paths.ts` modules. Value-equivalence was verified during review (every replaced literal resolves byte-for-byte to its prior value) and runtime behavior is preserved, so the PR was merged. This scrap captures the non-blocking DRY / dead-code / completeness follow-ups surfaced during that review.

## Details

**1. (mid) Duplicate constant definitions in the antigravity modules**
`src/constants/antigravity-cli-paths.ts` independently defines `ANTIGRAVITY_CLI_PERMISSIONS_DIR_PATH` (= `.gemini/antigravity-cli`) and `ANTIGRAVITY_CLI_PERMISSIONS_FILE_NAME` (= `settings.json`), while `src/constants/antigravity-paths.ts` also defines identically-named constants. The consumer (`src/features/permissions/antigravity-cli-permissions.ts`) imports the cli-paths copy, leaving the antigravity-paths copy as an unused duplicate. This is exactly the duplication the PR set out to eliminate.

**2. (mid) 30+ exported constants are never imported by any consumer**
Some are legitimately internal building blocks (`PI_AGENT_DIR`, `TAKT_FACETS_DIR_PATH`, `WINDSURF_SUBDIR`, `KILO_RULES_DIR_NAME`, etc.), but genuinely dead exports include, in `src/constants/antigravity-paths.ts`: `ANTIGRAVITY_MCP_PATH`, `ANTIGRAVITY_HOOKS_PATH`, `ANTIGRAVITY_GLOBAL_MCP_PATH`, `ANTIGRAVITY_GLOBAL_HOOKS_PATH`, `ANTIGRAVITY_GLOBAL_SKILLS_IDE_PATH`, `ANTIGRAVITY_GLOBAL_SKILLS_CLI_PATH`, `ANTIGRAVITY_GLOBAL_WORKFLOWS_IDE_PATH`, `ANTIGRAVITY_RULE_FILE_CLI`, `ANTIGRAVITY_RULE_FILE_IDE`, `ANTIGRAVITY_COMMANDS_DIR_PATH`. `knip` is in `package.json` but is not wired into `cicheck`/`check`, so these do not fail CI but would surface under `pnpm knip`.

**3. (low) Incomplete centralization for rule/steering directories**
Constants like `CURSOR_RULES_DIR_PATH`, `KILO_RULES_DIR_PATH`, `ROO_RULES_DIR_PATH`, `AUGMENTCODE_RULES_DIR_PATH`, `KIRO_STEERING_DIR_PATH` were added, but the matching feature files still hardcode the leaf segment (e.g. `cursor-rule.ts:53` uses `buildToolPath(CURSOR_DIR, "rules", ...)`). The reason is real (`buildToolPath` supports an `excludeToolDir` toggle a static constant cannot express), but the added `*_RULES_DIR_PATH` constants end up unused and misleading.

**4. (low) No tests added/changed**
Acceptable for a pure refactor, but there is no new test asserting the centralized constants equal their historical literals, so a future accidental edit to a constants file is only caught if an existing path-assertion test happens to cover it.

## Solution / Next Steps

- (1) Have `antigravity-cli-paths.ts` re-export the permissions dir/file constants from `antigravity-paths.ts` instead of redefining them (single source of truth).
- (2) Remove the genuinely dead exported constants, or wire up the intended consumers. Optionally consider adding `knip` to the CI pipeline to prevent regressions.
- (3) Either drop the unused `*_RULES_DIR_PATH` / `*_STEERING_DIR_PATH` constants, or refactor `buildToolPath` to consume them so the leaf segment is no longer hardcoded.
- (4) Optionally add a lightweight test asserting key centralized constants resolve to their expected path strings.

dyoshikawa:
## Background

Follow-ups from PR #1856, which centralized per-tool path literals into `src/constants/*-paths.ts` modules. Review surfaced non-blocking DRY / dead-code / completeness items, captured in #1864. This PR resolves them.

## Changes

- **Dedupe the Antigravity CLI permissions constants (item 1).** `antigravity-paths.ts` is now the single source of truth for `ANTIGRAVITY_CLI_PERMISSIONS_SUBDIR` / `_DIR_PATH` / `_FILE_NAME`. `antigravity-cli-paths.ts` re-exports them (the subdir under the `ANTIGRAVITY_CLI_GLOBAL_SUBDIR` name its skill consumer uses) instead of redefining identical literals.
- **Remove genuinely dead exported path constants (items 2 & 3).** Verified with `knip` + `grep` that no consumer imports them: the unused `ANTIGRAVITY_*_PATH` / `ANTIGRAVITY_RULE_FILE_*` / `ANTIGRAVITY_COMMANDS_DIR_PATH` family, the dead `antigravity-ide` config/skills/mcp/hooks constants, and the leftover `*_RULES_DIR_PATH` / `*_STEERING_DIR_PATH` / `*_MEMORIES_DIR_PATH` / `*_INSTRUCTIONS_DIR_PATH` constants (cursor, kilo, roo, augmentcode, kiro, copilot, etc.). After this change, `pnpm knip` reports **zero** dead exports in `src/constants/*-paths.ts`.
- **Un-export internal building blocks.** `WINDSURF_SUBDIR`, `PI_AGENT_DIR`, and `TAKT_FACETS_DIR_PATH` are used only to build other constants in their own file, so they drop `export` and stay module-private rather than appearing as unused public exports.
- **Add a regression test (item 4).** `src/constants/paths.test.ts` pins key centralized constants to their expected path strings and asserts the deduped Antigravity CLI constants resolve identically through both modules.

This is a pure refactor: `tsgo --noEmit` is clean and the full suite passes via `pnpm cicheck`. No runtime behavior changes (the removed constants had no consumers; the deduped ones keep byte-for-byte values, locked in by the new test). `knip` is intentionally **not** wired into CI here — the repo still has many intentional unused exports (frontmatter schemas, processor target lists) that would need triage first; that remains an optional future step noted in the issue.

Closes #1864


dyoshikawa:
@dyoshikawa Thank you!

PR #1920, opened 2026-06-18
fix(cursor): escape rule description in generated .mdc frontmatter

### LINKED ISSUE #1919: Cursor `.mdc` generation produces invalid YAML when a rule `description` contains a colon (breaks `import` round-trip)
## Summary

`rulesync generate -t cursor` writes the rule `description` into the `.mdc` YAML frontmatter **without quoting**. If the description contains a YAML indicator — most commonly a `": "` sequence (e.g. `"Use when: ..."`, `"Note: ..."`), but also a leading `#` or bare values like `true`/`123` — the generated frontmatter is **invalid YAML**.

rulesync's own parser then can't read it back, so `rulesync import -t cursor` throws a `YAMLException` and **silently drops the rule**. Cursor itself (which parses the `.mdc` frontmatter as YAML) is affected the same way.

## Reproduction

`.rulesync/rules/strict.md`:

```md
---
root: false
targets: ["*"]
description: "Use this rule: enforce strict TypeScript settings"
globs: ["*.ts"]
---
Always enable strict mode.
```

Generate the Cursor rule:

```bash
rulesync generate -t cursor -f rules
```

Generated `.cursor/rules/strict.mdc` — note the bare `: ` inside the value:

```md
---
description: Use this rule: enforce strict TypeScript settings
globs: *.ts
---

Always enable strict mode.
```

Round-trip it back:

```bash
rulesync import -t cursor -f rules
```

```
Failed to load tool files for cursor: Error: Failed to parse frontmatter in .cursor/rules/strict.mdc: YAMLException: incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 2, column 27:
    description: Use this rule: enforce strict TypeScript settings
                              ^
No rule files found for cursor. Skipping import.
No files imported for enabled features: rules
```

The rule is silently dropped — no non-zero exit, just lost data.

## Expected

A `description` containing colons or other YAML-special characters should be emitted as a valid, quoted YAML scalar and round-trip cleanly through `generate` → `import`.

## Root cause

`src/features/rules/cursor-rule.ts` → `stringifyCursorFrontmatter` interpolates the value raw:

```ts
if (frontmatter.description) {
  lines.push(`description: ${frontmatter.description}`);
}
```

The parse side (`parseCursorFrontmatter`) only quote-fixes the `globs` line, never `description`, so a `": "` in the value reads as a second mapping indicator and the YAML parser rejects the line.

## Suggested fix

Serialize the description as a proper YAML scalar (e.g. `js-yaml`'s `dump`, already a dependency, with `lineWidth: -1`). js-yaml quotes only when required, so plain descriptions stay unquoted — matching the existing "no unnecessary quotes" MDC convention used for `globs` — while `": "`, leading `#`, `true`/`123`, etc. are emitted safely.

I have a fix + regression test ready and will open a PR.

**Environment:** rulesync 8.30.1, Node 22+.

glitch-ux:
Closes #1919.

## Problem

`stringifyCursorFrontmatter` interpolated the rule `description` directly into the `.mdc` YAML frontmatter without quoting. Any description containing a YAML indicator — most commonly a `": "` sequence (`"Use when: ..."`, `"Note: ..."`), but also a leading `#` or bare values like `true`/`123` — produced **invalid YAML**. The generated file then failed to parse with rulesync's own parser, so `rulesync import -t cursor` threw a `YAMLException` and **silently dropped the rule**. Cursor (which reads the `.mdc` frontmatter as YAML) is affected the same way.

## Fix

Serialize the description with `js-yaml`'s `dump` (already a dependency, with `lineWidth: -1`), which quotes only when required:

- plain descriptions stay unquoted (`description: Test rule`) — output is unchanged, matching the existing "no unnecessary quotes" MDC convention already used for `globs`;
- `": "`, leading `#`, `true`/`123`, etc. are emitted as valid quoted scalars and round-trip cleanly.

`globs` handling is intentionally left untouched — Cursor requires it unquoted, and `parseCursorFrontmatter` already preprocesses that line on the way back in.

## Tests

- Added a round-trip regression test in `cursor-rule.test.ts` (`generate` → `fromFile` parse) for a description containing a colon. It **fails on `main`** with the `YAMLException` and **passes** with this change.
- `pnpm cicheck` is green locally: `fmt:check`, `oxlint`, `typecheck`, full test suite (6560 tests), `cspell`, `secretlint`.

Diff is ~38 lines across 2 files (1 implementation line + comment, plus the test).

### DIFF
--- a/src/features/rules/cursor-rule.test.ts
+++ b/src/features/rules/cursor-rule.test.ts
@@ -410,6 +410,34 @@ This is the rule content
       expect(rule.getRelativeFilePath()).toBe("test.mdc");
     });
 
+    it("should round-trip a description containing YAML-special characters", async () => {
+      // A description with a ": " sequence (and other YAML indicators) must be emitted
+      // as a quoted scalar; otherwise the generated frontmatter is invalid YAML and
+      // fromFile throws when reading it back. See the constructor stringify path.
+      const description = "Use this rule: enforce strict TypeScript settings";
+      const generated = new CursorRule({
+        frontmatter: { description, globs: "*.ts", alwaysApply: false },
+        body: "Always enable strict mode.",
+        relativeDirPath: ".cursor/rules",
+        relativeFilePath: "strict.mdc",
+      });
+
+      const filePath = join(testDir, ".cursor/rules", "strict.mdc");
+      await writeFileContent(filePath, generated.getFileContent());
+
+      const parsed = await CursorRule.fromFile({
+        outputRoot: testDir,
+        relativeFilePath: "strict.mdc",
+      });
+
+      expect(parsed.getFrontmatter()).toEqual({
+        description,
+        globs: "*.ts",
+        alwaysApply: false,
+      });
+      expect(parsed.getBody()).toBe("Always enable strict mode.");
+    });
+
     it("should handle file with no frontmatter", async () => {
       const filePath = join(testDir, ".cursor/rules", "simple.mdc");
       const content = "Just plain content without frontmatter";
--- a/src/features/rules/cursor-rule.ts
+++ b/src/features/rules/cursor-rule.ts
@@ -1,5 +1,6 @@
 import { join } from "node:path";
 
+import { dump } from "js-yaml";
 import { z } from "zod/mini";
 
 import { CURSOR_DIR } from "../../constants/cursor-paths.js";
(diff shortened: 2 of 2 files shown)


dyoshikawa:
@glitch-ux Thank you!

PR #1927, opened 2026-06-18
fix: use .kiroignore for Kiro ignore files

miloquinn:
## Summary
- Use `.kiroignore` for Kiro ignore generation/import instead of `.aiignore`
- Add Kiro's `.kiroignore` entry to generated gitignore blocks
- Keep Junie's `.aiignore` gitignore coverage explicit

## Why
The Kiro feature map already documents `.kiroignore` as the Kiro ignore surface, but the implementation still generated/imported `.aiignore`. That collides with other tools that also use `.aiignore` and makes the Kiro target write the wrong ignore file.

## Testing
- `pnpm exec vitest run --silent=true src/e2e/e2e-ignore.spec.ts src/cli/commands/gitignore.test.ts`
- `pnpm exec tsgo --noEmit`
- `pnpm exec vitest run --config vitest.e2e.config.ts --silent=false src/e2e/e2e-ignore.spec.ts`
- `pnpm exec oxfmt --check src/constants/kiro-paths.ts src/cli/commands/gitignore-entries.ts src/e2e/e2e-ignore.spec.ts src/cli/commands/gitignore.test.ts`
- `git diff --check`

### DIFF
--- a/src/cli/commands/gitignore-entries.ts
+++ b/src/cli/commands/gitignore-entries.ts
@@ -11,6 +11,8 @@ import {
   CLAUDECODE_SETTINGS_LOCAL_FILE_NAME,
   CLAUDECODE_SKILLS_DIR_PATH,
 } from "../../constants/claudecode-paths.js";
+import { JUNIE_IGNORE_FILE_NAME } from "../../constants/junie-paths.js";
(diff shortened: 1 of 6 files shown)


dyoshikawa:
@miloquinn Thank you!

PR #1928, opened 2026-06-18
feat(skills): support root-level disable-model-invocation with per-target overrides

sybers:
Users can now set `disable-model-invocation` once at the rulesync skill frontmatter root and have it applied to every supported tool (`claudecode`, `cursor`, `zed`, `pi`, `qwencode`, `factorydroid`).
A value set inside a tool section still wins, so single-target opt-outs remain possible.


dyoshikawa:
@sybers Thank you!

Issue #1830, opened 2026-06-10
Follow-up: doc/gitignore sync gaps from PR #1766 (Codex .agents/skills path & hook events)


dyoshikawa:
## Background

PR #1766 (https://github.com/dyoshikawa/rulesync/pull/1766, fixes #1685) updates the Codex CLI skills output path from `.codex/skills/` to `.agents/skills/` and wires three previously-missing hook events (`subagentStart`, `subagentStop`, `preCompact`) to their Codex PascalCase equivalents.

The core logic (`hooks.ts` map additions, `codexcli-skill.ts` / `gh-paths.ts` path changes) is correct, CI is fully green, and a security review found no issues. However, the review surfaced several **documentation / generated-artifact synchronization gaps** that violate the explicitly-documented requirements in `.claude/rules/feature-change-guidelines.md`. A contributing factor is that the PR branch is ~106 commits behind `main`, so some current-main content is not yet reflected.

This scrap consolidates those follow-up items so they are not lost.

## Details

Findings from the PR #1766 review:

1. **[high] Generated `.gitignore` not regenerated.** `src/cli/commands/gitignore-entries.ts:113` changed the registry entry `**/.codex/skills/` → `**/.agents/skills/`, but the generated `/workspace/.gitignore` was not regenerated. The PR branch still carries an orphaned `**/.codex/skills/` line (`.agents/skills/` already exists via another tool). `feature-change-guidelines.md` requires running `pnpm dev gitignore`.

2. **[high] Hook event × tool matrix doc not updated.** In `docs/reference/file-formats.md` (the "Hook event × tool matrix"), the Codex CLI column for `subagentStart` / `subagentStop` / `preCompact` still shows `—` (unsupported). Since this PR adds support for those events, the cells must become `✅`. Violates the "keep README/docs synchronized with implementation" rule.

3. **[mid] Sidecar description will become stale after rebase.** Current `main`'s `docs/reference/file-formats.md` contains "Rulesync emits `.codex/skills/<name>/agents/openai.yaml`". The PR branch is too old to include this line yet, but after rebasing onto `main` the path becomes inconsistent with the new `.agents/skills/` output. Needs updating to `.agents/skills/<name>/agents/openai.yaml`.

[...]

Issue #1207, opened 2026-02-27
fix: generated GitHub Copilot hooks are not ignored by rulesync gitignore


flanny7:
Hello,
I always appreciate rulesync.

I found a minor issue regarding the newly supported GitHub Copilot hooks, and I'd like to share it along with a proposed fix.

## Issue Description

While GitHub Copilot's hooks feature is now supported, it seems that the generated files under `.github/hooks/` are not currently targeted by the `.gitignore` generated by rulesync. 
Because of this, when running `rulesync gitignore`, these generated files might accidentally be tracked by Git.

## Proposed Solution

I propose updating the `rulesync gitignore` command output to include `**/.github/hooks/`.

I have already forked the repository and implemented this fix locally. I plan to submit a Pull Request shortly. 

There is no rush at all. It would be wonderful if you could review it when your current tasks settle down.

Your consideration would be appreciated.

Issue #619, opened 2025-12-01
gitignore command should respect targets and features configuration


mkanakana:
## Summary

The `rulesync gitignore` command adds all possible output paths to `.gitignore` regardless of the `targets` and `features` configuration in `rulesync.jsonc`. This results in unnecessary entries being added.

## Current Behavior

When running `rulesync gitignore`, all tool output paths are added to `.gitignore`, including paths for features that are not configured:

For example, if `rulesync.jsonc` has:
```jsonc
{
  "targets": ["claudecode"],
  "features": ["rules"]
}
```

The command still adds entries like:
- `**/.claude/commands/`
- `**/.claude/agents/`
- `**/.claude/skills/`
- `**/.github/prompts/`
- `**/.cursor/`
- etc.

## Expected Behavior

`rulesync gitignore` should read `rulesync.jsonc` and only add `.gitignore` entries for:
1. Tools specified in `targets` (or all if `"*"`)
2. Features specified in `features` (or all if `"*"`)

For example, with the configuration above, only `**/CLAUDE.md` and `**/.claude/memories/` should be added.

## Related Code

The current implementation in `src/cli/commands/gitignore.ts` uses a hardcoded list without checking the configuration.

## Benefits

- Cleaner `.gitignore` files
- Avoids confusion about which tools are actually being used
- Consistent behavior with other rulesync commands that respect configuration

cm-dyoshikawa:
@mkanakana Thank you. 

Certainly your suggestion is stricter however current implementation is not critical for this cli users. The richer the specification becomes, the higher the maintenance cost.

At this moment, I think I'd like to focus on another point than `gitignore` command, such as plugin system.

dyoshikawa:
This is now implemented. The `gitignore` command resolves targets from `rulesync.jsonc`/`rulesync.local.jsonc` via `resolveGitignoreTargets` (honoring `gitignoreTargetsOnly`, default true) and filters entries by target/feature in `resolveGitignoreEntries` (`src/cli/commands/gitignore-entries.ts`, `resolve-gitignore-targets.ts`, wired in `src/cli/index.ts`). It landed in commit a760d790 (2026-04-14), after the earlier deferral. The hardcoded-list premise no longer holds, so closing as resolved.

[...]

PR #1929, opened 2026-06-19
fix(gitignore): correct 7 drifted tool output entries and add getSettablePaths coverage test

saitota:
## Background

`gitignore-entries.ts` hand-maintains a `{ target, feature, entry }` registry of where every tool writes output, duplicating each tool class's `getSettablePaths` (the real source of truth). Drift produces **no type error and no test failure** — it only surfaces when users find rulesync-generated files committed to their repo.

Auditing the registry against `getSettablePaths` found **7 already-shipping drifts**:

| Tool / feature | Was | Should be |
|---|---|---|
| kilo / commands | `**/.kilo/workflows/` | `**/.kilo/commands/` |
| kilo / subagents | (missing) | `**/.kilo/agents/` |
| kilo / hooks | (missing) | `**/.kilo/plugins/` |
| junie / commands | (missing) | `**/.junie/commands/` |
| roo / commands | (missing) | `**/.roo/commands/` |
| cline / skills | (missing) | `**/.cline/skills/` |
| geminicli / permissions | (missing) | `**/.gemini/policies/` |

Each confirmed by running `generate` and observing un-gitignored artifacts.

## Changes

- Fix the 7 entries and regenerate `.gitignore`.
- Add a `getSettablePaths coverage` test asserting every tool's project-scope output is covered by the registry, or in an explicit exception set (user-managed `settings.json` files, global-only paths, deprecated `antigravity` alias). Turns the implicit coupling into a CI-enforced invariant.

This is the recommended first step, not the full derivation (deliberately-broad entries like `**/.cursor/` make naive derivation behavior-changing — left as follow-up).

## Related

- #619 — same hardcoded list in `gitignore-entries.ts`; that issue asks to filter it by config, this PR guards it against drift. Complementary, not overlapping.
- #1207, #1830 (closed) — prior instances of the exact same drift (a tool's output not gitignored), motivating the CI guard added here.

## Test plan

- [x] `pnpm fmt:check` / `oxlint` / `typecheck` / `cspell` / `secretlint` — clean
- [x] `gitignore-entries.test.ts` — 166 pass (verified to fail when an entry is reverted)
- [ ] `pnpm test` (full) — `EPERM` in sandbox (pre-existing, unrelated); run before merge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -256,6 +256,7 @@ rulesync.local.jsonc
 **/.claude/*.lock
 **/.clinerules/
 **/.clinerules/workflows/
+**/.cline/skills/
 **/.clineignore
 **/.cline/agents/
 **/.cline/command-permissions.json
@@ -282,6 +283,7 @@ rulesync.local.jsonc
 **/.gemini/agents/
 **/.gemini/skills/
 **/.gemini/memories/
+**/.gemini/policies/
 **/.goosehints
 **/.goose/
 **/.gooseignore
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@saitota Thank you!

PR #1945, opened 2026-06-19
fix(cursor): flatten multi-line .mdc description to avoid YAML block scalar

### LINKED ISSUE #1933: Follow-ups from PR #1920: Cursor .mdc multiline description block-scalar gap and shared-serializer DRY
## Background

PR #1920 (`fix(cursor): escape rule description in generated .mdc frontmatter`, by @glitch-ux) was merged after batch review. It replaces raw string interpolation of the rule `description` in Cursor's `.mdc` YAML frontmatter with `js-yaml`'s `dump()`, which quotes/escapes only when required. This **closes a real YAML/frontmatter-injection vector**: previously a description containing `": "`, a leading `#`, a bare scalar, or fence/key break-out payloads could corrupt the file or inject arbitrary frontmatter keys. Security review verified the fix neutralizes those payloads (the `---`/`alwaysApply:` stay safely indented inside the value), round-trips correctly via rulesync's parser, and over-escaping is avoided (plain descriptions remain unquoted). CI was green.

The findings below are **non-blocking** follow-ups; the merged fix is already a net improvement. Captured here for later.

### Finding 1 (mid) — Multiline descriptions still emit a YAML block scalar Cursor's simplified parser may not read
The inline comment claims `lineWidth: -1` "prevents js-yaml from wrapping long descriptions into block scalars," which is misleading. `lineWidth: -1` only disables *folding of long single-line strings*. A description containing genuine newlines still serializes to a literal block scalar (`description: |-` ...). `description` is typed as a plain optional string with no newline restriction and is passed through untouched, so this is reachable from user YAML. The codebase already has an `avoidBlockScalars` option (`src/utils/frontmatter.ts`) precisely because Cursor's simplified MDC parser misinterprets `|-`/`>-` indicators. This fix does not protect that case. rulesync's own parser round-trips fine, but Cursor's native reader may not. Not a security gap (injection stays neutralized); a compatibility gap. No multiline-description test exists.

### Finding 2 (mid) — Hand-rolled serialization instead of the existing shared Cursor serializer (DRY)
The codebase already has a purpose-built serializer: `stringifyFrontmatter(body, frontmatter, { avoidBlockScalars: true })`, used by sibling Cursor features (`cursor-command.ts`, `cursor-subagent.ts`). That path flattens newlines (avoiding Finding 1) and quotes only when required. The PR instead hand-rolls `dump({ description }, { lineWidth: -1 }).trimEnd()` inline in `stringifyCursorFrontmatter` (`src/features/rules/cursor-rule.ts`). Note `CursorRule` keeps a custom serializer because Cursor requires `globs` unquoted (the shared path would quote glob patterns), so a full switch isn't trivial — but the `description` line specifically should at minimum flatten newlines (e.g. `description.replace(/\n+/g, " ").trim()`) before dumping, to match `avoidBlockScalars` semantics.

### Finding 3 (low) — Comment accuracy
The block comment at `src/features/rules/cursor-rule.ts` overstates `lineWidth: -1`. Correct it to reflect that it only prevents folding of long single-line values, not block scalars for genuinely multiline strings.

## Solution / Next Steps

1. (Findings 1 & 2, primary) Flatten newlines in the description before `dump()` (or route the description through the shared `avoidBlockScalars` serializer) so Cursor itself — not just rulesync's parser — can read multi-line descriptions.
2. (Finding 3) Fix the misleading comment about `lineWidth: -1`.
3. Add a `description: "Line one\nLine two"` round-trip test to lock in the multiline behavior.

dyoshikawa:
## Summary

Follow-ups from PR #1920 (Cursor `.mdc` description escaping). Addresses the multiline block-scalar compatibility gap.

- **Findings 1 & 2 (mid):** A genuinely multi-line `description` still serialized to a YAML block scalar (`description: |-`), which Cursor's simplified MDC frontmatter parser does not read. We now flatten newlines to spaces (`description.replace(/\n+/g, " ").trim()`) before `dump()`, matching the `avoidBlockScalars` semantics used by the sibling Cursor features, so Cursor itself — not just rulesync's parser — can read the description. Non-string values on the `validate: false` path pass through unchanged.
- **Finding 3 (low):** Corrected the misleading comment about `lineWidth: -1` (it only prevents folding of long single-line values, not block scalars for multi-line strings).
- Added a multi-line `description` round-trip test asserting no block-scalar indicator is emitted.

Note: `CursorRule` keeps its custom serializer because Cursor requires `globs` unquoted, so this targets the `description` line specifically rather than switching the whole path to the shared serializer.

## Verification

- `pnpm cicheck:code` passes (lint, types, 6728 unit tests incl. the new multi-line case).

Closes #1933

Ref: #1920

### DIFF
--- a/src/features/rules/cursor-rule.test.ts
+++ b/src/features/rules/cursor-rule.test.ts
@@ -438,6 +438,36 @@ This is the rule content
       expect(parsed.getBody()).toBe("Always enable strict mode.");
     });
 
+    it("should flatten a multi-line description into a single-line scalar Cursor can read", async () => {
+      // A genuinely multi-line description must NOT serialize to a YAML block scalar
+      // (`description: |-`): Cursor's simplified MDC parser does not read block-scalar
+      // indicators. The newlines are flattened to spaces before serialization.
+      const generated = new CursorRule({
+        frontmatter: { description: "Line one\nLine two", globs: "*.ts", alwaysApply: false },
+        body: "Body.",
+        relativeDirPath: ".cursor/rules",
+        relativeFilePath: "multiline.mdc",
+      });
+
+      const content = generated.getFileContent();
+      // No block-scalar indicator, and the value is on a single line.
+      expect(content).not.toContain("description: |");
+      expect(content).not.toContain("description: >");
+      expect(content).toContain("description: Line one Line two");
+
+      const filePath = join(testDir, ".cursor/rules", "multiline.mdc");
+      await writeFileContent(filePath, content);
+      const parsed = await CursorRule.fromFile({
+        outputRoot: testDir,
+        relativeFilePath: "multiline.mdc",
+      });
+      expect(parsed.getFrontmatter()).toEqual({
+        description: "Line one Line two",
+        globs: "*.ts",
+        alwaysApply: false,
+      });
+    });
+
     it("should handle file with no frontmatter", async () => {
       const filePath = join(testDir, ".cursor/rules", "simple.mdc");
       const content = "Just plain content without frontmatter";
--- a/src/features/rules/cursor-rule.test.ts
+++ b/src/features/rules/cursor-rule.test.ts
@@ -468,6 +468,54 @@ This is the rule content
       });
     });
 
+    it("should flatten and still quote a multi-line description containing YAML indicators", async () => {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1948, opened 2026-06-19
fix(skills): make factorydroid disable-model-invocation consistent with other tools

### LINKED ISSUE #1931: Follow-ups from PR #1928: factorydroid disable-model-invocation override gap and skills resolve-helper DRY
## Background

PR #1928 (`feat(skills): support root-level disable-model-invocation with per-target overrides`, by @sybers) was merged after batch review. It adds a root-level `disable-model-invocation` boolean to the rulesync skill frontmatter and wires it into six tools (claudecode, cursor, zed, pi, qwencode, factorydroid) as a shared default that each tool's own section value overrides. The core merge logic (`section?.[key] ?? root[key]`) is correct — including the important edge case where a per-target `false` correctly overrides a root `true` — and security review found no issues. CI was green.

The findings below are **non-blocking** follow-ups surfaced during review; none blocked the merge. Captured here for later.

### Finding 1 (mid) — factorydroid cannot actually be overridden per-target; docs/PR text are misleading
`RulesyncSkillFrontmatterSchemaInternal` has no `factorydroid:` section, so a user has no way to set a per-target `disable-model-invocation` for factorydroid. `factorydroid-skill.ts` only reads the root value:
```ts
const resolvedDisableModelInvocation = rulesyncFrontmatter["disable-model-invocation"];
```
Yet the PR description ("A value set inside a tool section still wins, so single-target opt-outs remain possible") and the new doc comment ("Any of those tool sections can override it") claim per-target override works for all six tools including factorydroid. This is a correctness/consistency gap between documented and actual behavior.

**Suggestion:** Either add a `factorydroid` section schema with the key (plus the standard `??` merge + round-trip), or qualify the docs/PR text to state factorydroid supports the root default only.

### Finding 2 (low) — factorydroid does not round-trip the value on import (`toRulesyncSkill`)
`factorydroid-skill.ts` builds the rulesync frontmatter with only `name`/`description`/`targets`, dropping `disable-model-invocation`. The other five tools round-trip the value into their section, so `generate` then `import` is asymmetric for factorydroid (value emitted but lost on re-import). Related to Finding 1.

### Finding 3 (low) — DRY: the resolve pattern is duplicated across six files
The identical merge expression `section?.["disable-model-invocation"] ?? rulesyncFrontmatter["disable-model-invocation"]` is copy-pasted in `claudecode-skill.ts`, `cursor-skill.ts`, `zed-skill.ts`, `pi-skill.ts`, `qwencode-skill.ts`, and `factorydroid-skill.ts`. A small shared helper (e.g. `resolveDisableModelInvocation(rootFrontmatter, section)`) would make the "root default + per-target override" contract a single source of truth as the supported-tool list grows.

### Finding 4 (low) — pi/zed rely on subtle spread-then-reapply ordering
In `pi-skill.ts` and `zed-skill.ts` the section is spread first (`...piSection` / `...zedSection`) then the resolved value is conditionally re-applied afterward. This is correct (the `??` already prefers the section value), but the ordering dependency is easy to break in future edits. A brief comment or the helper from Finding 3 would make the intent explicit.

## Solution / Next Steps

1. (Finding 1, primary) Align factorydroid's documented vs. actual behavior — add the missing `factorydroid` section schema, or qualify the docs/PR claim to "root default only".
2. (Finding 2) If a factorydroid section is added, round-trip the value in `toRulesyncSkill` too.
3. (Findings 3 & 4) Extract a shared `resolveDisableModelInvocation` helper to deduplicate and make the override contract explicit.

dyoshikawa:
## Summary

Closes #1931
Ref: #1928
Issue: https://github.com/dyoshikawa/rulesync/issues/1931

PR #1928 added a root-level `disable-model-invocation` default to the rulesync skill frontmatter and wired it into six tools (`claudecode`, `cursor`, `zed`, `pi`, `qwencode`, `factorydroid`), where each tool's own section value overrides the shared default. Five of those tools have a per-target section schema and round-trip the value, but **factorydroid did not**: it had no `factorydroid:` section in the rulesync skill schema, only read the root value, and dropped the value on import. This contradicted the documented "any tool section can override" claim.

## Changes

- **factorydroid consistency**
  - Added a `factorydroid` section (with the `disable-model-invocation` key) to `RulesyncSkillFrontmatterSchemaInternal` and its input type in `src/features/skills/rulesync-skill.ts`.
  - `FactorydroidSkill.fromRulesyncSkill` now resolves the value via the shared helper (root default + per-target override).
  - `FactorydroidSkill.toRulesyncSkill` now round-trips `disable-model-invocation` back into the `factorydroid` section, matching the other five tools. The documented per-target override claim is now true for factorydroid too.
- **DRY helper**
  - Extracted `resolveDisableModelInvocation({ rootFrontmatter, section })` into `src/features/skills/skills-utils.ts` and replaced the copy-pasted `section?.["disable-model-invocation"] ?? rootFrontmatter["disable-model-invocation"]` expression in all six tool skill files. For `pi`/`zed` (which spread the section then re-apply the resolved value), behavior is identical and a clarifying comment now makes the spread-then-reapply intent explicit. No observable output change for the five already-correct tools.
- **Docs**
  - Documented the `factorydroid:` skill section in `docs/reference/file-formats.md` and re-synced `skills/rulesync/file-formats.md` via `scripts/sync-skill-docs.ts`.
- **Tests**
  - Added a direct unit test for `resolveDisableModelInvocation`.
  - Added factorydroid tests for the per-target override (including the `false`-overrides-`true` edge case) and the import round-trip.

## Verification

`pnpm cicheck` (code + content) passes: format, oxlint, typecheck, 6738 tests, skill-doc-sync, cspell, secretlint all green.

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -478,6 +478,8 @@ cursor: # for Cursor-specific parameters (optional)
   disable-model-invocation: true # (optional) only include the skill when invoked via /skill-name
   metadata: # (optional) free-form metadata
     author: rulesync
+factorydroid: # for Factory Droid-specific parameters (optional)
+  disable-model-invocation: true # (optional) prevent the model from auto-invoking this skill
 takt: # takt specific parameters (optional; emitted under .takt/facets/knowledge/ — frontmatter is dropped on emit)
   name: "renamed-stem" # (optional) override the emitted filename stem (no path separators or "..")
   extends: "base" # (optional) emit a leading `{extends:<parent>}` facet-inheritance directive (Takt 0.39.0+)
--- a/skills/rulesync/file-formats.md
+++ b/skills/rulesync/file-formats.md
@@ -478,6 +478,8 @@ cursor: # for Cursor-specific parameters (optional)
   disable-model-invocation: true # (optional) only include the skill when invoked via /skill-name
   metadata: # (optional) free-form metadata
     author: rulesync
+factorydroid: # for Factory Droid-specific parameters (optional)
+  disable-model-invocation: true # (optional) prevent the model from auto-invoking this skill
 takt: # takt specific parameters (optional; emitted under .takt/facets/knowledge/ — frontmatter is dropped on emit)
   name: "renamed-stem" # (optional) override the emitted filename stem (no path separators or "..")
   extends: "base" # (optional) emit a leading `{extends:<parent>}` facet-inheritance directive (Takt 0.39.0+)
--- a/src/features/skills/claudecode-skill.ts
+++ b/src/features/skills/claudecode-skill.ts
@@ -11,6 +11,7 @@ import { RULESYNC_SKILLS_RELATIVE_DIR_PATH } from "../../constants/rulesync-path
 import { ValidationResult } from "../../types/ai-dir.js";
 import { formatError } from "../../utils/error.js";
 import { RulesyncSkill, RulesyncSkillFrontmatterInput, SkillFile } from "./rulesync-skill.js";
+import { resolveDisableModelInvocation } from "./skills-utils.js";
 import {
   ToolSkill,
   ToolSkillForDeletionParams,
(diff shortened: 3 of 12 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1949, opened 2026-06-19
test(gitignore): make global-only skip symmetric and remove silent catch-swallow

### LINKED ISSUE #1930: Follow-ups from PR #1929: gitignore coverage-test polish and .gemini/policies precision
## Background

PR #1929 (`fix(gitignore): correct 7 drifted tool output entries and add getSettablePaths coverage test`, by @saitota) was merged after batch review. It fixed 7 genuine drifts in the hand-maintained gitignore registry (`src/cli/commands/gitignore-entries.ts`) and added a coverage test (`gitignore-entries.test.ts`) that couples the registry to each tool's `getSettablePaths` (the real source of truth). All 7 drift fixes were verified correct against the actual tool path constants, and CI was green.

The findings below are **non-blocking maintainability polish** surfaced during review; none were severe enough to block the merge. They are captured here for later follow-up.

### Finding 1 (mid) — Asymmetric handling of global-only tools between the two test branches
In `gitignore-entries.test.ts`, the `fileFeatures` loop cleanly skips global-only tools via `meta.supportsProject === false`, but the `dirFeatures` loop (commands/skills/subagents) does not. Instead it relies on `getSettablePaths({ global: false })` *throwing* for global-only tools (e.g. `CodexcliCommand` throws), which a broad `try/catch` silently swallows. The `dirFeatures` factories also expose `meta.supportsProject`, so the same clean guard can be applied. As written, if a global-only dir tool is ever refactored to return a path instead of throwing, the test would falsely require that global path to be gitignored.

**Suggestion:** Make both branches symmetric by checking `meta.supportsProject === false` in the `dirFeatures` branch too.

### Finding 2 (low) — Silent-pass risk from early returns and broad try/catch
Both test loops contain multiple early `return;` statements and a broad `try/catch`. A Vitest `it` that returns without asserting *passes*, so a future refactor that makes `getSettablePaths` throw unexpectedly or return an empty path would turn a real coverage gap into a silently-green test.

**Suggestion:** Assert the skip reason explicitly (e.g. `expect.assertions`, or distinguish intentionally-skipped from unexpectedly-empty), or narrow the `catch` to the expected global-only error.

### Finding 3 (low) — Directory-level gitignore of `**/.gemini/policies/` may capture user-authored files
rulesync writes only `rulesync.toml` into `.gemini/policies/`, and `geminicli-permissions.ts` explicitly anticipates hand-authored sibling `.toml` files in that directory. A user's hand-authored `.gemini/policies/my-rules.toml` would also be gitignored. This is consistent with the directory-level convention used elsewhere (`.kilo/commands/`, `.cline/skills/`), but unlike those, this directory is documented to also hold user files.

**Suggestion:** If precision is preferred, a file-scoped `**/.gemini/policies/rulesync.toml` entry would be safer.

## Solution / Next Steps

1. (Finding 1) Add a `meta.supportsProject === false` guard to the `dirFeatures` branch of `gitignore-entries.test.ts` to match the `fileFeatures` branch.
2. (Finding 2) Tighten the test so intentional skips are distinguished from accidental no-assertion passes.
3. (Finding 3) Decide whether to narrow the `.gemini/policies/` entry to the specific `rulesync.toml` file, weighing precision vs. the existing directory-level convention.

dyoshikawa:
## Summary

Follow-ups from PR #1929 (gitignore coverage test maintainability polish).

- **Finding 1 (mid):** The `dirFeatures` branch of `gitignore-entries.test.ts` relied on `getSettablePaths({ global: false })` *throwing* for global-only tools, with a broad `catch` swallowing it — asymmetric with the `fileFeatures` branch, which cleanly skips via `meta.supportsProject === false`. If a global-only dir tool were ever refactored to return a path instead of throwing, the test would falsely require that global path to be gitignored. Now both branches use the same explicit `meta.supportsProject === false` guard.
- **Finding 2 (low):** Removed the broad `try/catch` in both branches. With global-only tools skipped explicitly, a project-supporting tool must resolve a path without throwing, so an unexpected throw now fails the test instead of silently passing (a returning Vitest `it` with no assertion passes).

**Finding 3 (`.gemini/policies/` directory- vs file-level precision) is intentionally NOT changed.** Narrowing the registry entry to `**/.gemini/policies/rulesync.toml` is technically more precise, but it diverges from the established directory-level convention used elsewhere (`.kilo/commands/`, `.cline/skills/`) and, because this project dogfoods rulesync, regeneration leaves the old broad directory entry preserved as an unmanaged line — defeating the precision goal without manual cleanup. Keeping the directory-level convention is the cleaner decision; this can be revisited if precise per-file ignoring becomes a priority.

## Verification

- `pnpm cicheck` passes (code + content; 6737 unit tests incl. the 166 gitignore coverage tests, cspell, secretlint, skill-doc sync).

Closes #1930

Ref: #1929

### DIFF
--- a/src/cli/commands/gitignore-entries.test.ts
+++ b/src/cli/commands/gitignore-entries.test.ts
@@ -113,13 +113,17 @@ describe("getSettablePaths coverage", () => {
   for (const [feature, factories] of dirFeatures) {
     for (const [target, factory] of factories) {
       if (TARGETS_WITHOUT_GITIGNORE_ENTRIES.has(target)) continue;
+      // Skip global-only tools explicitly (same guard as the fileFeatures branch
+      // below) rather than relying on `getSettablePaths({ global: false })`
+      // throwing and a broad catch swallowing it — otherwise a global-only tool
+      // refactored to return a project path would silently change coverage.
+      const meta = factory.meta as { supportsProject?: boolean } | undefined;
+      if (meta && meta.supportsProject === false) continue;
       it(`covers ${feature} output for ${target}`, () => {
-        let paths: { relativeDirPath?: string };
-        try {
-          paths = factory.class.getSettablePaths({ global: false });
-        } catch {
-          return;
-        }
+        // No try/catch: a project-supporting tool must resolve a project path
+        // without throwing, so an unexpected throw fails the test instead of
+        // passing silently.
+        const paths = factory.class.getSettablePaths({ global: false });
         const dir = paths.relativeDirPath;
         if (!dir || dir === ".") return;
         const glob = dirToGlob(dir);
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #1958, opened 2026-06-19
fix: ignore shared AGENTS root in rules check

### LINKED ISSUE #1894: generate --check for codexcli rules fails when another target owns AGENTS.md
## Problem

When two targets both emit the root `AGENTS.md` rule file, `rulesync generate --check` for one target can fail even when that target's **non-root** outputs (for example `.codex/memories/*.md`) are already up to date.

We use **target order** so one IDE target (last in config) owns `AGENTS.md`. Codex topic rules live under `.codex/memories/`. Full `rulesync generate` works. A split verify for Codex rules alone does not.

## Example setup

```jsonc
{
    "targets": {
        "claudecode": ["rules", "ignore", "mcp", "skills", "subagents", "permissions"],
        "cursor": ["rules", "ignore", "mcp", "skills", "subagents"],
        "agentsskills": ["skills"],
        "codexcli": ["rules", "mcp", "subagents"],
        "devin": ["rules", "ignore", "mcp", "skills"],
        "antigravity-ide": ["rules", "mcp"]
    },
    "delete": true,
    "gitignoreTargetsOnly": true
}
```

- Root `overview.md` has `root: true`, `targets: ["*"]`.
- `antigravity-ide` is **last** so it wins `AGENTS.md` (per target-order docs).
- Codex non-root rules go to `.codex/memories/`.

## Reproduction

1. Run full generate (succeeds):

   ```bash
   rulesync generate
   ```

2. Confirm `.codex/memories/*.md` exists and looks correct.

3. Run check for Codex rules only (fails):

   ```bash
   rulesync generate --check -t codexcli -f rules
   ```

**Actual output (example):**

```text
[DRY RUN] Would write: AGENTS.md
[DRY RUN] Would write 1 rule
    AGENTS.md
Files are not up to date. Run 'rulesync generate' to update.
```

Exit code: `1`

**Expected (one of):**

- Check passes when `.codex/memories/` is up to date, even if on-disk `AGENTS.md` matches another target's root output, **or**
- A flag to verify **non-root outputs only** for a target, **or**
- A built-in verify command that understands shared files and target order.

## Why this hurts

We split verify to avoid other target conflicts:

```bash
rulesync generate --check -t claudecode,cursor,devin,antigravity-ide \
  && rulesync generate --check -t agentsskills -f skills \
  && rulesync generate --check -t codexcli -f mcp,subagents
```

We **cannot** add `rules` to the `codexcli` check without it failing on `AGENTS.md`, so `.codex/memories/` drift is not caught by rulesync's built-in check.

## Current workaround

A small custom script that:

1. Runs `rulesync generate -o <temp-dir> -t codexcli -f rules`
2. Compares only `<temp-dir>/.codex/memories/*.md` to the local `.codex/memories/`
3. Ignores `AGENTS.md`

This works but is easy to get wrong and should not be required in every multi-target repo.

## Suggested solutions

Any of these would help:

1. **`--check` scope** — verify only non-root rule outputs for a target.
2. **Shared-file-aware verify** — when multiple targets write the same path, respect target order / "owner" and do not fail other targets' checks.
3. **`rulesync verify`** — documented or built-in multi-target verify for repos using `agentsskills`, `antigravity-ide`, `codexcli`, etc.
4. **Docs** — official recipe for "one target owns `AGENTS.md`, Codex memories verified separately".

## Environment

- **rulesync:** 8.29.0
- **Node:** 24.x
- **OS:** macOS
- **Repo:** multi-target setup; generated AI tool outputs gitignored (`gitignoreTargetsOnly: true`)

ProfTrader:
## Summary
- Skip project-root `AGENTS.md` during rules `--check` for modular rule targets when non-root rule outputs are present.
- Keep normal generation unchanged, and keep Codex memory files checked.
- Add unit and e2e coverage for the shared-root `AGENTS.md` case.

Fixes #1894

## Verification
- `pnpm exec vitest run src/features/rules/rules-processor.test.ts -t "ignore shared root AGENTS.md" --silent=true`
- `pnpm exec oxfmt --check src/features/rules/rules-processor.test.ts src/features/rules/rules-processor.ts src/lib/generate.ts src/e2e/e2e-rules.spec.ts`
- `pnpm exec oxlint src/features/rules/rules-processor.test.ts src/features/rules/rules-processor.ts src/lib/generate.ts src/e2e/e2e-rules.spec.ts --max-warnings 0`
- `pnpm run typecheck`

## Notes
- Full `src/e2e/e2e-rules.spec.ts` is blocked locally on Windows because the e2e helper spawns `node_modules/.bin/tsx` without the `.cmd` shim, so every e2e case fails before reaching this change.
- The full `rules-processor.test.ts` file currently has an unrelated failure on this branch: `should generate copilot global non-root files via round-trip`.

### DIFF
--- a/src/e2e/e2e-rules.spec.ts
+++ b/src/e2e/e2e-rules.spec.ts
@@ -166,6 +166,69 @@ This is a test rule for E2E testing.
     expect(stdout).not.toContain("All files are up to date (rules)");
   });
 
+  it("should check codexcli non-root rules when another target owns project AGENTS.md", async () => {
+    const testDir = getTestDir();
+
+    const codexRootRuleContent = `---
+root: true
+targets: ["codexcli"]
+description: "Codex root rule"
+globs: ["**/*"]
+---
+
+# Codex Root Rule
+`;
+    const opencodeRootRuleContent = `---
+root: true
+targets: ["opencode"]
+description: "OpenCode root rule"
+globs: ["**/*"]
+---
+
+# OpenCode Root Rule
+`;
+    const codexMemoryRuleContent = `---
+targets: ["codexcli"]
+description: "Codex memory"
+globs: ["src/**/*"]
+---
+
+# Codex Memory Rule
+`;
+
+    await writeFileContent(
+      join(testDir, RULESYNC_RULES_RELATIVE_DIR_PATH, "codex-root.md"),
+      codexRootRuleContent,
+    );
+    await writeFileContent(
+      join(testDir, RULESYNC_RULES_RELATIVE_DIR_PATH, "opencode-root.md"),
+      opencodeRootRuleContent,
+    );
+    await writeFileContent(
+      join(testDir, RULESYNC_RULES_RELATIVE_DIR_PATH, "codex-memory.md"),
+      codexMemoryRuleContent,
+    );
+
+    await runGenerate({ target: "codexcli,opencode", features: "rules", deleteFiles: true });
+
+    expect(await readFileContent(join(testDir, "AGENTS.md"))).toContain("OpenCode Root Rule");
+    expect(await readFileContent(join(testDir, ".codex", "memories", "codex-memory.md"))).toContain(
+      "Codex Memory Rule",
+    );
+
+    const { stdout, stderr } = await runGenerate({
+      target: "codexcli",
+      features: "rules",
+      deleteFiles: true,
+      check: true,
+      env: { NODE_ENV: "e2e" },
+    });
+
+    expect(stderr).toBe("");
+    expect(stdout).toContain("All files are up to date.");
+    expect(stdout).not.toContain("Would write: AGENTS.md");
+  });
+
   it("should write BOTH instructions (rules) and mcp into a single kilo.jsonc when generating rules+mcp together", async () => {
     const testDir = getTestDir();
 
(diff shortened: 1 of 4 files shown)


dyoshikawa-claw (approved):
## Review Summary

**Severity: none** — No merge-blocker findings.

The fix correctly skips project-root `AGENTS.md` in `--check` mode for modular rule targets (like codexcli) when non-root rule outputs are present. This prevents false-positive staleness reports when another target owns the shared root file. The skip logic is applied only in check mode and doesn't affect normal generation. Both unit and e2e test coverage are included.

Note: this PR has a merge conflict with main that needs resolution before merging.

✅ Ready to merge after conflict resolution.

dyoshikawa:
Thank you for this, @ProfTrader — the approach was correct and the review was positive.

Unfortunately this got overtaken: #1894 has just been fixed and closed by #1978, which lands an equivalent fix via root-file ownership computation in `src/lib/generate.ts` (rather than the `check`-flag skip inside `rules-processor.ts` taken here). Since the two are competing implementations of the same fix and #1894 is now resolved, this PR is superseded and would only add a redundant second fix on top of `main` (which is also why it currently conflicts).

Closing as superseded by #1978. Really appreciate the contribution — the diagnosis and the shared-root `AGENTS.md` test coverage were spot on, and they helped confirm the fix direction.

PR #1959, opened 2026-06-19
fix: override gray-matter js-yaml advisory

ProfTrader:
## Summary

- add a pnpm workspace override so `gray-matter` resolves its transitive `js-yaml` dependency to `4.2.0`
- add a tiny pnpm dependency patch for `gray-matter@4.0.3` so its default YAML engine uses `js-yaml` 4's `load` / `dump` APIs, with fallback support for the old names
- remove the vulnerable `js-yaml@3.14.2` / `argparse@1` / `sprintf-js` dependency path from the lockfile

## Verification

- `npx pnpm@11.5.2 install --ignore-scripts --config.confirmModulesPurge=false`
- `npx pnpm@11.5.2 audit --prod --json` reports 0 vulnerabilities
- `npx pnpm@11.5.2 exec vitest run src/utils/frontmatter.test.ts src/utils/content-equivalence.test.ts --silent=true`
- `npx pnpm@11.5.2 run typecheck`
- `git diff --check`


dyoshikawa:
Reviewed (security): **safe to merge.** This is a minimal, correctly-scoped supply-chain hardening — the patch binds gray-matter onto js-yaml 4's safe-by-default `load`/`dump` (removing the vulnerable js-yaml 3.14.2 / argparse@1 / sprintf-js path), the patch content is benign and exactly the described engine-binding change, the lockfile adds no new packages and the js-yaml@4.2.0 integrity hash matches the npm registry, and no unsafe loader is introduced. Two low/informational notes only: the `safeLoad`/`safeDump` fallback is dead-but-harmless under js-yaml 4, and js-yaml 4 is slightly stricter (rejects duplicate mapping keys) — a safety improvement.

Holding off on auto-merge: this PR edits dependency manifests (`pnpm-lock.yaml`, `pnpm-workspace.yaml`) and adds a dependency patch, which our autonomous review flow intentionally does **not** admin-merge. Leaving it for a maintainer to merge manually.

dyoshikawa:
@ProfTrader Thank you!

PR #1961, opened 2026-06-19
feat(skills): support user-invocable on claudecode and at the rulesync root with per-target overrides

sybers:
Claude Code's skill spec documents a `user-invocable: false` flag that hides a skill from the `/` menu while keeping it model-invokable (https://code.claude.com/docs/en/skills). rulesync already routed this field through `qwencode`, `vibe`, and `factorydroid`, but it was missing from `claudecode` itself — and from the rulesync root.

Users can now set `user-invocable` once at the rulesync skill frontmatter root and have it applied to every supported tool (`claudecode`, `qwencode`, `vibe`, `factorydroid`). A value set inside a tool section still wins, so single-target overrides remain possible — same pattern as `disable-model-invocation`.

Also fixes a small `factorydroid` bug where `user-invocable` was accepted in the SKILL.md schema but silently dropped on conversion to/from rulesync.

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -385,6 +385,10 @@ targets: ["*"] # * = all, or specific tools
 # zed, pi, qwencode, and factorydroid. Any of those tool sections can override it
 # by setting their own `disable-model-invocation` value below.
 disable-model-invocation: true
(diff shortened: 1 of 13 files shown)


dyoshikawa:
Merged. Thanks! Filed #1967 to track the minor docs follow-ups (vibe example, factorydroid verification).

PR #1962, opened 2026-06-19
refactor(rules): decouple RulesProcessor from other features' internals

saitota:
## Background

`RulesProcessor` had become a god class. While the other 7 feature processors keep 0–1 tool branches and import no other features, `rules-processor.ts` imported 10 other-feature classes and held **11 `this.toolTarget === "..."` branches**. It reached into the MCP feature's internals (`KiloMcp` / `OpencodeMcp.fromInstructions`) and hardcoded the simulate-section wording for commands/subagents/skills.

So `convertRulesyncFilesToToolFiles` forced a reader to hold many unrelated tools' behavior in their head at once, and adding a Kilo/OpenCode-style tool meant adding another branch here instead of in the owning feature.

## Changes

- **MCP integration inversion** — non-root rules are registered via a declarative `meta.mcpInstructionsRegistrar` contract instead of importing and calling specific MCP classes inline. The rules feature only knows it has a list of non-root paths; the MCP feature owns the file format.
- **Simulate extraction** — convention-section wording moved into each feature processor's `getSimulatedConventionSection`; the rules feature only decides which features contribute and concatenates them.
- **Declarative flags** — all 11 `toolTarget` string branches replaced with `meta` flags (`foldsNonRootIntoRoot`, `mcpInstructionsRegistrar`, `localRootMode` / `localRootFileName`, `mirrorsRootToAgentsMd`). `this.toolTarget === "..."` count is now **11 → 0**.

No runtime behavior changes — generated output is identical. This is a pure refactor.

## Test plan

- [x] `pnpm fmt:check` / `oxlint` / `typecheck` — clean
- [x] `pnpm test` — 6800/6801 pass (1 pre-existing failure in `warp-permissions.test.ts`, present on `main` and unrelated to this change)

### DIFF
--- a/src/features/commands/commands-processor.ts
+++ b/src/features/commands/commands-processor.ts
@@ -2,6 +2,7 @@ import { basename, join, relative } from "node:path";
 
 import { z } from "zod/mini";
 
+import { RULESYNC_COMMANDS_RELATIVE_DIR_PATH } from "../../constants/rulesync-paths.js";
 import { FeatureProcessor } from "../../types/feature-processor.js";
 import { RulesyncFile } from "../../types/rulesync-file.js";
 import { ToolFile } from "../../types/tool-file.js";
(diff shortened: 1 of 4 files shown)


dyoshikawa:
Merged. Thanks!

PR #1964, opened 2026-06-19
test(gitignore): close getSettablePaths coverage gaps and fix 3 drifted entries

saitota:
## Background

The `getSettablePaths coverage` test from #1929 couples each tool’s `getSettablePaths` (the real output location) to the hand-maintained `GITIGNORE_ENTRY_REGISTRY`. Drift there produces no type error and no test failure — it surfaces only as generated files committed to a user’s repo.

The guard had gaps: it skipped `rules` and `ignore`, matched directories by prefix (a renamed dir could pass by matching a parent), and never checked for ghost entries no tool emits. Closing them surfaced **3 already-shipping drifts**.

## Changes

Coverage gaps closed:
- Added `ignore` and `rules` to the checks (`rules` gets its own loop for its `root`/`alternativeRoots`/`nonRoot` shape).
- Directory matching is now exact, not prefix, so `.augment/commands/` → `.augment/cmds/` fails instead of matching a parent. Deliberate subtree roots (`.cursor`, `.agents`, `.goose`, `.rovodev/.rulesync`) are listed explicitly and asserted to exist.
- Reverse check: every non-`general` entry must map to a real emitted output, catching ghosts.

Drifts fixed (were shipping un-gitignored / stale):
- junie root rule → `**/.junie/AGENTS.md` (`guidelines.md` kept as legacy fallback)
- junie non-root rules → `**/.junie/memories/`
- removed stale `**/.kilo/mcp.json` ghost (MCP output moved to root `kilo.json`)

`.gitignore` regenerated accordingly.

## Related

- #1929 — adds the original coverage test; this closes its remaining gaps and fixes the drifts they surface.
- #1930 — follow-up findings from #1929. Finding 2 (early-return / try-catch silent-pass risk) is reduced here: the reverse check no longer relies on broad try/catch.

## Test plan

- [x] `pnpm cicheck:code` — clean; `gitignore-entries.test.ts` 218 pass (fails when a drift is reintroduced). One pre-existing unrelated failure (`warp-permissions`, also on `main`).
- [x] `pnpm cicheck:content` — clean
- [x] `pnpm dev gitignore` — regenerated, idempotent

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -303,7 +303,9 @@ rulesync.local.jsonc
 **/.copilot/mcp-config.json
 **/.copilot/agents/
 **/.copilot/hooks/
+**/.junie/AGENTS.md
 **/.junie/guidelines.md
+**/.junie/memories/
 **/.junie/commands/
 **/.junie/mcp/mcp.json
 **/.junie/skills/
@@ -313,7 +315,6 @@ rulesync.local.jsonc
 **/.kilo/skills/
 **/.kilo/commands/
 **/.kilo/agents/
-**/.kilo/mcp.json
 **/.kilo/plugins/
 **/.kilocodeignore
 **/.kiro/steering/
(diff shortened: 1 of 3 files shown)


dyoshikawa:
Merged. Thanks!

PR #1914, opened 2026-06-17
fix: omit unsupported codex subagent short-description

ProfTrader:
## Summary
- Prevent codexcli subagent generation from emitting the unsupported `short-description` TOML field.
- Preserve supported Codex fields such as `model` while filtering this compatibility-only frontmatter key.
- Add a regression test for the Codex parser failure described in #1714.

## Validation
- `pnpm vitest run src/features/subagents/codexcli-subagent.test.ts --silent=true`

Closes #1714.

PR #1968, opened 2026-06-20
fix(codexcli): drop short-description on subagent import for symmetry

### LINKED ISSUE #1915: Follow-ups from PR #1914: codex subagent short-description low-severity notes
## Background

PR #1914 fixed a bug where the Codex CLI subagent generator emitted a `short-description` field that Codex does not support, causing Codex to reject the entire agent file as a malformed agent role (related to #1714). The fix adds `short-description` to the `excludeFields` list passed to `filterToolSpecificSection` in `codexcli-subagent.ts`, matching the existing pattern used by other subagent implementations. CI was green and the PR had no mid-or-above review findings, so it was merged.

During review, two low-severity, non-blocking observations were surfaced and are recorded here for potential future follow-up. Neither blocks anything today.

## Details

1. **Import/export asymmetry.** The generation path (`fromRulesyncSubagent` in `src/features/subagents/codexcli-subagent.ts`) now excludes `short-description`, but the import path (`toRulesyncSubagent`) passes unknown TOML fields through to the `codexcli` section because `CodexCliSubagentTomlSchema` is a `z.looseObject`. Impact is limited: Codex itself rejects files containing `short-description`, so a valid Codex file can never contain that field in practice.

2. **Test assertion robustness.** In `src/features/subagents/codexcli-subagent.test.ts`, the assertion `expect(...).not.toContain("short-description")` is a substring match. It works correctly today, but asserting the TOML-key form (e.g. `'short-description ='`) would be slightly more precise. Not required.

## Solution / Next Steps

- These are optional improvements only; no action is strictly required.
- If revisited: consider whether the import path should also strip `short-description` for symmetry, and optionally tighten the test assertion to match the TOML key form.

dyoshikawa:
## Summary

Follow-ups from PR #1914 (Codex CLI subagent `short-description`).

- **Finding 1 (low):** `toRulesyncSubagent` (import) spread all unknown TOML fields into the rulesync `codexcli` section, including `short-description` — asymmetric with the generation side, which already excludes it (`short-description` is the field Codex rejects). Now dropped on import too, so a stray value never round-trips. Added an import regression test asserting the field does not reach the `codexcli` section.
- **Finding 2 (low):** Hardened the "does not emit short-description" assertion to match the TOML-key assignment form (`^\s*"?short-description"?\s*=`) instead of a bare substring, so it cannot be satisfied incidentally by a value.

## Verification

- `pnpm cicheck` passes (code + content; 40 codexcli-subagent tests incl. the new import case).

Closes #1915

Ref: #1914

### DIFF
--- a/src/features/subagents/codexcli-subagent.test.ts
+++ b/src/features/subagents/codexcli-subagent.test.ts
@@ -323,7 +323,34 @@ describe("CodexCliSubagent", () => {
       }) as CodexCliSubagent;
 
       expect(codexcliSubagent.getBody()).toContain('model = "gpt-5"');
-      expect(codexcliSubagent.getBody()).not.toContain("short-description");
+      // Assert the TOML-key assignment form is absent rather than a bare
+      // substring, so the check cannot be satisfied incidentally by the value.
+      expect(codexcliSubagent.getBody()).not.toMatch(/^\s*"?short-description"?\s*=/m);
+    });
+
+    it("should drop short-description on import so it does not round-trip", () => {
+      // A (hypothetical) Codex TOML carrying `short-description` must not carry
+      // that field into the rulesync `codexcli` section — import/export symmetry.
+      const toml = [
+        'name = "reviewer"',
+        'description = "Code reviewer"',
+        '"short-description" = "Review source changes"',
+        'model = "gpt-5"',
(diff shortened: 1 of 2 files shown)


dyoshikawa:
Merged. Thanks!

PR #1960, opened 2026-06-19
feat(codex): emit and preserve :minimal filesystem baseline

saitota:
## Background

Codex CLI `v0.112.0-alpha.8` introduced `:minimal` as a `FileSystemSpecialPath` that enables `include_platform_defaults()`, providing the platform/runtime read access needed for sandboxed command execution on macOS, Linux, and Windows. Without `:minimal = "read"` in the profile, commands fail inside the sandbox on those platforms.

Rulesync's existing model had no concept of these Codex-native sandbox baseline paths (`:minimal`, `:root`, `:tmpdir`, `:slash_tmp`), so they were silently dropped on import and never emitted on generate.

## Changes

- Always emit `":minimal" = "read"` in the generated `permissions.<profile>.filesystem` table
- Skip `:minimal`, `:root`, `:tmpdir`, `:slash_tmp` on import so they never pollute rulesync's permission model (they're Codex-native baselines, not user rules)
- Preserve existing `:root`, `:tmpdir`, `:slash_tmp` values verbatim on regeneration via `mergeWithExistingProfile`
- Remove dead `Object.keys(filesystem).length > 0` guard (filesystem is always non-empty after `:minimal` init)
- Add `:minimal = "read"` assertion to E2E test per Tool × Feature matrix requirement
- Document the new behavior in `docs/reference/file-formats.md`

Closes https://github.com/dyoshikawa/rulesync/issues/1876

## References

- openai/codex#13434 — `FileSystemSpecialPath::Minimal` and `include_platform_defaults()`
- #1733 — prior issue on Codex sandbox baseline loss (network)

## Test plan

- [x] `pnpm fmt:check` / `oxlint` / `typecheck` — clean
- [x] `cspell` / `secretlint` — clean
- [x] `codexcli-permissions.test.ts` — 39 pass (6 new cases covering `:minimal` emit, import skip, round-trip, `:root`/`:tmpdir`/`:slash_tmp` pass-through)
- [x] E2E `codexcli` permissions case updated to assert `filesystem[":minimal"] === "read"`

PR #1970, opened 2026-06-20
fix(codexcli): round-trip :root/:tmpdir/:slash_tmp through the rulesync model

### LINKED ISSUE #1965: Follow-ups from PR #1960: codex :minimal import drops user :root/:tmpdir/:slash_tmp on fresh regenerate
## Background

PR #1960 (`feat(codex): emit and preserve :minimal filesystem baseline`, by @saitota) was merged after batch review. It makes the Codex CLI permissions adapter always emit `[":minimal" = "read"]` in the `filesystem` table (which enables Codex's `include_platform_defaults()` in a Restricted sandbox) and preserve user-customized `:root`/`:tmpdir`/`:slash_tmp` baseline keys by overlaying them from an existing `.codex/config.toml`. Correctness of the `:minimal` semantics was verified against the upstream Codex source (`codex-rs/protocol/src/permissions.rs`, `core/src/config/permissions.rs`), code review was clean, and CI was green.

The findings below are **non-blocking** follow-ups surfaced during review; none was severity high/critical, so they did not block the merge.

### Finding 1 (mid) — Baseline keys skipped on import can silently drop a user's restrictive `:root`/`:tmpdir`/`:slash_tmp` in the fresh-clone regenerate path
`src/features/permissions/codexcli-permissions.ts` — `convertCodexProfileToRulesync` now skips `CODEX_FILESYSTEM_BASELINE_KEYS` (`:minimal`, `:root`, `:tmpdir`, `:slash_tmp`) on import. Before this PR, importing a hand-authored Codex config with e.g. `":root" = "deny"` populated the rulesync model and round-tripped on every regenerate. After this PR, those keys never enter `.rulesync/permissions.json` (the committed source of truth); their preservation depends entirely on `mergeWithExistingProfile` reading them back from an existing `.codex/config.toml`. But `.codex/config.toml` is gitignored generated output, so the realistic workflow — commit `.rulesync/permissions.json`, fresh clone, `rulesync generate` with no pre-existing config — does NOT preserve the value, and a user's deliberate `:root = "deny"` tightening is silently lost.

This is a (narrow) weakening of permission posture: it only affects users who imported a hand-authored Codex config that set these special baseline keys restrictively, and the keys are platform-path toggles rather than arbitrary files, hence mid rather than high. The PR's "preserve :root/:tmpdir" test passes only because it keeps the existing config file in place; it does not cover the model-is-source-of-truth fresh-clone case.

**Suggested fix (reviewer):** either (a) continue importing `:root`/`:tmpdir`/`:slash_tmp` into the rulesync model (as before) so they live in the committed source of truth — only `:minimal` needs the always-emit treatment — or (b) emit a `logger.warn` when these baseline keys are encountered on import and dropped, so the user knows the value persists only if the generated config file is retained.

### Finding 2 (low) — `:minimal` is forced to `"read"` and cannot be overridden
`:minimal` is always emitted as `"read"` and the merge never lets an existing config's `:minimal` value win; it is also skipped on import, so a user cannot express any other value through rulesync. This is acceptable because `:minimal` is a read-only platform-defaults toggle (it does not grant write) and matches Codex's intended design, but unlike the `extends`/`network.mode`/`unix_sockets` overrides it overwrites a hand-set `:minimal` without a warning. Low impact; arguably by design.

## Solution / Next Steps

1. (Finding 1, primary) Stop dropping `:root`/`:tmpdir`/`:slash_tmp` on import (keep them in the rulesync model), or warn when they are dropped — so an imported restrictive baseline is not silently lost on a fresh-clone regenerate.
2. (Finding 2) Optionally warn when overwriting a hand-set `:minimal`.

## References

- PR #1960 — Codex `:minimal` filesystem baseline.
- Upstream: https://github.com/openai/codex/blob/main/codex-rs/protocol/src/permissions.rs (FileSystemSpecialPath, include_platform_defaults), https://github.com/openai/codex/blob/main/codex-rs/core/src/config/permissions.rs (parse_special_path).

dyoshikawa:
## Summary

Fixes a permission-posture weakening introduced by PR #1960 in `src/features/permissions/codexcli-permissions.ts`.

PR #1960 added `:root`, `:tmpdir`, and `:slash_tmp` to `CODEX_FILESYSTEM_BASELINE_KEYS` and **skipped all of them on import**, re-emitting them only by overlaying values from an **existing** `.codex/config.toml` in `mergeWithExistingProfile`. Because `.codex/config.toml` is gitignored generated output, the realistic workflow — commit `.rulesync/permissions.json`, fresh clone, `rulesync generate` with no pre-existing config — **silently dropped** a user's restrictive `:root = "deny"`.

## Changes

- **Import**: only `:minimal` is skipped on import now (it is the always-emitted, non-user-managed fixed baseline). `:root`, `:tmpdir`, and `:slash_tmp` flow through `addRulesyncFilesystemRule` like any ordinary filesystem pattern, so they land in the rulesync model (`:root = "deny"` → read/edit `deny`; `:tmpdir = "write"` → edit `allow`; etc.).
- **Export**: unchanged — `addFilesystemRule` already routes any `:`-prefixed pattern into `filesystem[pattern] = access`, so a model containing `:root = "deny"` re-emits `filesystem[":root"] = "deny"`. Round-trips cleanly with no existing config.
- **Merge**: removed the now-redundant baseline-overlay loop in `mergeWithExistingProfile`. It could re-introduce a **stale** value the user intentionally removed. `:minimal` is always set by `newProfile`, so it was never actually overlaid; `:workspace` / `:workspace_roots` / `extends` / `network` handling is untouched, and the always-emit `:minimal = "read"` behavior is preserved.
- Removed the `CODEX_FILESYSTEM_BASELINE_KEYS` constant and updated comments so the code accurately reflects that only `:minimal` is a non-user-managed baseline.
- Updated `docs/reference/file-formats.md` (and synced `skills/rulesync/`) to describe the new round-trip behavior.

## Tests

- Replaced the old "preserve user-customized `:root`/`:tmpdir`" test (which kept an existing config file in place) with one that asserts the behavior **through the model**: import → model contains `:root`/`:tmpdir` → regenerate from a **fresh** dir with **no** existing config → the values still appear.
- Flipped "should not import `:root` or `:tmpdir`" to assert they now **are** imported (and `:minimal` still is not).
- Added the regression test PR #1960 missed: import `:root = "deny"` + `:tmpdir`, assert they land in the model, regenerate into a fresh/empty output dir, and assert `:root = "deny"` / `:tmpdir` are **not lost** while `:minimal = "read"` is still always emitted.
- Existing codexcli-permissions tests and the Tool×Feature happy-path e2e (`e2e-permissions.spec.ts`) all pass.

Full `pnpm cicheck` (code + content): all 6892 tests pass, format/lint/typecheck/cspell/secretlint clean.

Closes #1965
https://github.com/dyoshikawa/rulesync/issues/1965
Ref: #1960


dyoshikawa:
Merged. Thanks!

PR #1978, opened 2026-06-22
fix: skip non-owned root files in generate --check for shared AGENTS.md

### LINKED ISSUE #1894: generate --check for codexcli rules fails when another target owns AGENTS.md

dyoshikawa-claw:
## Summary

When multiple targets share the root `AGENTS.md` file (e.g., `codexcli` and `antigravity-ide`), `rulesync generate --check` for a single target would fail because the target's expected `AGENTS.md` content differs from the on-disk content written by the last target in config order (different TOON reference paths like `@.codex/memories/*` vs `@.agents/rules/*`).

## Fix

Compute root file ownership from the config file's full target list (before `--targets` CLI override). In check mode, skip root files the target doesn't own, so only the owning target's root file is checked. Non-root files (e.g., `.codex/memories/*.md`) are always checked regardless.

### Changes

- Add `Config.getConfigFileTargets()` returning config file targets before CLI `--targets` override
- Add `computeRootFileOwnership` helper in `generate.ts` that maps root file paths to their owner (last target in config order)
- Skip non-owned root files in `processFeatureGeneration` (write check only; orphan detection still sees the full tool file list so shared root files are not treated as orphans)
- Add e2e test: check codexcli passes when antigravity-ide owns `AGENTS.md`

Closes #1894

### DIFF
--- a/src/cli/commands/generate.test.ts
+++ b/src/cli/commands/generate.test.ts
@@ -46,6 +46,7 @@ describe("generateCommand", () => {
       getSilent: vi.fn().mockReturnValue(false),
       getOutputRoots: vi.fn().mockReturnValue(["."]),
       getTargets: vi.fn().mockReturnValue(["claudecode"]),
+      getConfigFileTargets: vi.fn().mockReturnValue(["claudecode"]),
       getFeatures: vi.fn().mockReturnValue(["rules", "ignore", "mcp", "commands", "subagents"]),
       getFeatureOptions: vi.fn().mockReturnValue(undefined),
       getDelete: vi.fn().mockReturnValue(false),
--- a/src/config/config-resolver.ts
+++ b/src/config/config-resolver.ts
@@ -7,6 +7,12 @@ import {
   RULESYNC_LOCAL_CONFIG_RELATIVE_FILE_PATH,
 } from "../constants/rulesync-paths.js";
 import {
+  ALL_TOOL_TARGETS,
+  type ToolTarget,
+  isRulesyncConfigTargetsObject,
+  type RulesyncConfigTargets,
+} from "../types/tool-targets.js";
+import {
   fileExists,
   getHomeDirectory,
   readFileContent,
@@ -310,6 +316,7 @@ export class ConfigResolver {
       // captured `cwd` so the value is still deterministic.
       inputRoot: resolvedInputRoot !== undefined ? resolve(resolvedInputRoot) : cwd,
       sources: configByFile.sources ?? getDefaults().sources,
+      configFileTargets: extractConfigFileTargets(configByFile.targets),
     };
     const config = new Config(configParams);
     // The legacy `antigravity` target is never produced by wildcard expansion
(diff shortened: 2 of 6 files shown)


dyoshikawa:
@dyoshikawa-claw Thank you!

PR #1979, opened 2026-06-22
fix(codexcli): fold non-root rules into root AGENTS.md instead of inert .codex/memories

### LINKED ISSUE #1765: codexcli non-root rules written to .codex/memories/ which Codex CLI does not read
## Summary

Non-root rules generated for `codexcli` are written to `.codex/memories/*.md`. Codex CLI does not load instruction files from this directory — neither project-level `.codex/memories/` nor global `~/.codex/memories/`. The files are silently ignored.

## Reproduction

```bash
# 1. Initialize rulesync with a non-root rule
mkdir /tmp/test && cd /tmp/test && git init
npx rulesync init
# Edit .rulesync/rules/overview.md — ensure root: false (or remove root: true)

# 2. Generate for codexcli
npx rulesync generate -t codexcli -f rules -V

# 3. Observe output goes to .codex/memories/overview.md

# 4. Verify Codex does NOT load it
codex debug prompt-input
# The memories file content does not appear in the prompt input
```

## Root Cause

`CodexcliRule.getSettablePaths()` in `src/features/rules/codexcli-rule.ts:54` sets the non-root path to `.codex/memories`. This directory is Codex CLI's own auto-memory system (SQLite-backed persistent session memory at `~/.codex/memories_1.sqlite`), not a location for user-provided instruction files.

The path was modeled after the `agentsmd` target's `.agents/memories/` convention, but Codex CLI does not implement this convention.

## Codex CLI's actual instruction loading

Per the [official docs](https://developers.openai.com/codex/guides/agents-md) and source verification:

1. **`AGENTS.md`** at project root (and `AGENTS.override.md`)
2. **Hierarchical walking** from git root to cwd, loading `AGENTS.md` in each directory
3. **`config.toml`** `instructions` / `developer_instructions` fields
4. **Skills** (`~/.codex/skills/*/SKILL.md`, `.agents/skills/*/SKILL.md`)

There is no "directory of loose rule files" discovery mechanism.

## TOON workaround is unreliable

The `ruleDiscoveryMode: "toon"` setting prepends a TOON-formatted reference section to AGENTS.md asking the LLM to read the non-root rule files. This is prompt engineering, not native file loading — the model may or may not follow the instruction. Verified that `codex debug prompt-input` does not show the referenced files in the loaded context.

## Impact

- **Root rules work correctly** — `root: true` generates `AGENTS.md`, which Codex reads.
- **All non-root rules are silently dropped** — they go to `.codex/memories/` where Codex ignores them.
- **Single-rule setups are unaffected** — if you have one rule with `root: true`, everything lands in AGENTS.md.
- **Multi-rule setups lose all non-root rules** without warning.

## Suggested fix

Since Codex CLI has no directory-based rule discovery, the options are:

1. **New `ruleDiscoveryMode: "inline"`** — concatenate non-root rule content directly into the root `AGENTS.md` under section headers. This is what actually works with Codex.
2. **Maintain the current TOON approach** but document that it is best-effort and relies on LLM compliance, not native loading.

Option 1 would be a correct fix. Option 2 would at minimum need a warning in the docs/README.

## Related

- #1685 — similar path mismatch for skills (`.codex/skills/` vs canonical `.agents/skills/`)

## Environment

- rulesync: 8.24.1
- Codex CLI: 0.137.0
- Verified with `codex debug prompt-input` and `codex exec` testing

dyoshikawa-claw:
## Summary

Codex CLI loads instructions only from the `AGENTS.md` family (root + hierarchical directory walk per the [official docs](https://developers.openai.com/codex/guides/agents-md)). It does **NOT** scan `.codex/memories/` for instruction files — that directory is Codex's separate SQLite-backed auto-memory system. Non-root rules written there were silently ignored.

## Fix

Add `foldsNonRootIntoRoot: true` to the codexcli rule factory entry and remove the `nonRoot` path from `CodexcliRule.getSettablePaths()`, so non-root rule bodies are concatenated into the single root `AGENTS.md` by the RulesProcessor — same pattern as **grokcli**, **warp**, and **deepagents**.

### Changes

- Rewrite `codexcli-rule.ts` following the grokcli/warp pattern (root-only settable paths, both root and non-root rules target `./AGENTS.md`)
- Change `ruleDiscoveryMode` from `'toon'` to `'auto'` (no inert TOON references)
- Update `codexcli-rule.test.ts`: 41 tests updated for folding behavior
- Update gitignore tests: `.codex/memories/` no longer generated
- Regenerate `.gitignore` (drops `**/.codex/memories/`)

### Precedent

This mirrors commit `3a57ea7a" which fixed the identical bug for warp (`.warp/memories/` was inert).

Closes #1765


dyoshikawa:
@dyoshikawa-claw Thank you!

PR #1999, opened 2026-06-23
fix(devcontainer): run pnpm install through mise exec in init script

dyoshikawa:
## Summary

- Wrap the dependency install in `.devcontainer/init.sh` with `mise exec` so `pnpm i` runs with the toolchain versions pinned in `mise.toml` rather than whatever `pnpm` happens to be on `PATH` during container setup.
- This keeps devcontainer setup consistent with the runtimes/package managers managed by mise.

## Test plan

- `pnpm cicheck` passes (code style, type safety, tests, content checks).
- Change is limited to the devcontainer init script; no application code is affected.


dyoshikawa:
@dyoshikawa Thank you!

PR #2008, opened 2026-06-23
fix(generate): correct root-file ownership for wildcard configs and mirrored roots

### LINKED ISSUE #1981: Follow-up: non-blocking findings from PR #1978 (skip non-owned root files in generate --check)
## Background

PR #1978 ("fix: skip non-owned root files in generate --check for shared AGENTS.md") fixes issue #1894: when multiple targets emit the shared root `./AGENTS.md` with differing content, `generate --check` for a single non-owning target failed. The fix computes root-file ownership from the config file's full target list (last target in config order wins) and, in check mode, skips checking root files a target doesn't own. Orphan detection is correctly left intact (full `toolFiles` is still passed to `removeOrphanAiFiles`). CI was fully green and the security review was clean, so it was merged during batch triage.

These are the non-blocking (mid/low) findings surfaced during that review, captured here for follow-up.

### #1 (mid) — Wildcard config `targets: ["*"]` silently bypasses the fix
Files: `src/config/config-resolver.ts:359-360` (`extractConfigFileTargets`) and `src/config/config.ts:436`.

For `targets: ["*"]`, `extractConfigFileTargets` filters out `*` and returns `[]` (empty), not `undefined`. In `config.ts:436`, `return this.configFileTargets ?? this.getTargets();` — `[]` is not nullish, so the fallback to `getTargets()` is NOT triggered and `getConfigFileTargets()` returns `[]`. `computeRootFileOwnership` then receives an empty target list, the ownership map is empty, no root file is skipped, and the #1894 bug reproduces for any repo using the very common `targets: ["*"]` form (e.g. `rulesync generate --check -t codexcli -f rules` still fails on `AGENTS.md`). This is an incomplete-fix gap rather than a new regression (wildcard configs stay at status quo).

### #2 (mid) — rovodev's mirrored `./AGENTS.md` is invisible to ownership computation
Files: `src/lib/generate.ts:252-265` (`computeRootFileOwnership`) vs `src/features/rules/rules-processor.ts` (mirror logic).

`computeRootFileOwnership` derives the owned root path solely from `factory.class.getSettablePaths({ global }).root`. For rovodev that is `.rovodev/AGENTS.md`, but at generation time rovodev additionally pushes a mirrored project-root `./AGENTS.md` (`mirrorsRootToAgentsMd: true`). That mirror is exactly the shared-collision path this PR targets, yet it is never entered into `ownerByPath`. In a config like `{ rovodev: [...], codexcli: [...] }` where rovodev is last and actually owns on-disk `./AGENTS.md` via mirror, the map instead assigns `./AGENTS.md` to codexcli, inverting ownership/skip decisions. The same blind spot applies to any target using `alternativeRoots` / `mirrorsRootToAgentsMd`.

### #3 (low) — Ownership map ignores `outputRoot`
`src/lib/generate.ts:285-329`.

Keys are built from cwd-relative paths without `outputRoot`, so a single ownership decision applies uniformly across all output roots. Unlikely to bite real users, but the model conflates per-output-root state.

### #4 (low) — Ownership determinism depends on config object-key order
`src/lib/generate.ts:254-260`.

"Last target in config order wins" relies on `Object.keys()` insertion order. This is correct and consistent with the existing target-order semantics, but the behavior is load-bearing and undocumented at this call site.

### #5 (low) — Test coverage gaps
`src/e2e/e2e-rules.spec.ts:264-331`.

The added e2e covers the object-form happy path. Missing coverage that would catch #1/#2: a `targets: ["*"]` variant, the reverse direction (owning target passes while a non-owning target is present), a unit test on `getConfigFileTargets()` for `["*"]`/empty inputs, and a `computeRootFileOwnership` unit test for owner selection + reorder.

## Solution / Next Steps

- **#1**: In `extractConfigFileTargets`, expand `*` to the full non-legacy target set (mirroring `getTargets()`), and/or return `undefined` when the filtered result is empty so the `?? getTargets()` fallback engages. Add a `["*"]` regression test.
- **#2**: Source the owned root paths from the same mechanism that determines actual emitted root files (mirror + alternative roots), not just `getSettablePaths().root`, so mirrored/aliased root files are correctly attributed.
- **#3**: Document the single-decision-across-output-roots assumption, or key ownership per output root if multi-root `--check` correctness is desired later.
- **#4**: Add a one-line comment noting that the last writer in config order is intentionally the owner (consistent with generation write order), plus a reorder test asserting the owner flips when the two targets are swapped.
- **#5**: Backfill the missing e2e/unit tests listed above.

dyoshikawa:
## Summary

Follow-up fixes for the #1978 root-file-ownership work, addressing the non-blocking findings in #1981.

### #1 (mid) — wildcard `targets: ["*"]` silently bypassed the fix
`extractConfigFileTargets` filtered out `*` and returned `[]` (not `undefined`), so `getConfigFileTargets()` did not fall back and root-file ownership was computed against an empty target list — reproducing #1894 for the very common `targets: ["*"]` config (e.g. `generate --check -t codexcli -f rules` failing on `AGENTS.md`). Now `*` is expanded to the full non-legacy target set, mirroring `Config.getTargets()`.

> The `?? getTargets()` fallback is intentionally **not** relied on here: `getTargets()` is CLI `-t` filtered, whereas ownership needs the full *config-file* target list (CLI-independent).

### #2 (mid) — rovodev's mirrored `./AGENTS.md` was invisible to ownership
`computeRootFileOwnership` derived owned roots only from `getSettablePaths().root`, so rovodev's generation-time mirror to the project-root `./AGENTS.md` (`mirrorsRootToAgentsMd`) and any target's `alternativeRoots` were never attributed — inverting ownership in configs like `{ codexcli, rovodev }`. Both are now registered.

### #3 / #4 (low) — documentation
Documented the single-ownership-decision-across-output-roots assumption and the load-bearing "last target in config order wins" semantics at the call site.

### #5 (low) — tests
- config-resolver regression test: `["*"]` expands to the full non-legacy list (and stays full even when CLI `-t` selects one target).
- E2E: a non-owning target (`codexcli`) passes `--check` when rovodev owns the mirrored `./AGENTS.md`.

## Verification

- `pnpm cicheck` passes (fmt, oxlint, typecheck, tests, content).
- E2E smoke: `generate --check -t codexcli` passes when rovodev (last in config) owns the mirrored `AGENTS.md`.

Closes #1981

### DIFF
--- a/src/config/config-resolver.test.ts
+++ b/src/config/config-resolver.test.ts
@@ -158,6 +158,44 @@ describe("config-resolver", () => {
     });
   });
 
+  describe("config file targets (getConfigFileTargets)", () => {
+    it("expands wildcard targets ['*'] to the full non-legacy target list", async () => {
+      // Regression for #1981 / #1894: a `targets: ["*"]` config must not collapse
+      // to an empty config-file target list, otherwise root-file ownership in
+      // `generate --check` is computed against no targets and the bug reproduces.
+      const configContent = JSON.stringify({ outputRoots: ["./"], targets: ["*"] });
+      await writeFileContent(join(testDir, "rulesync.jsonc"), configContent);
+
+      const config = await ConfigResolver.resolve({
+        configPath: join(testDir, "rulesync.jsonc"),
+      });
+
+      const configFileTargets = config.getConfigFileTargets();
+      expect(configFileTargets.length).toBeGreaterThan(1);
+      expect(configFileTargets).toContain("claudecode");
+      expect(configFileTargets).toContain("codexcli");
+      // Legacy targets are excluded from wildcard expansion (must be explicit).
+      expect(configFileTargets).not.toContain("claudecode-legacy");
+      expect(configFileTargets).not.toContain("antigravity");
+    });
+
+    it("keeps the full config-file target list even when CLI -t selects one target", async () => {
+      const configContent = JSON.stringify({ outputRoots: ["./"], targets: ["*"] });
+      await writeFileContent(join(testDir, "rulesync.jsonc"), configContent);
+
+      const config = await ConfigResolver.resolve({
+        configPath: join(testDir, "rulesync.jsonc"),
+        targets: ["codexcli"],
+      });
+
+      // CLI -t narrows the generated targets, but config-file ownership must
+      // still see every target the config lists.
+      expect(config.getTargets()).toEqual(["codexcli"]);
+      expect(config.getConfigFileTargets().length).toBeGreaterThan(1);
+      expect(config.getConfigFileTargets()).toContain("claudecode");
+    });
+  });
+
   describe("base directory resolution", () => {
     it("should load configured outputRoots from file", async () => {
       const configContent = JSON.stringify({
--- a/src/config/config-resolver.ts
+++ b/src/config/config-resolver.ts
@@ -26,6 +26,7 @@ import {
   ConfigFile,
   ConfigFileSchema,
   ConfigParams,
+  LEGACY_TARGETS,
   PartialConfigParams,
   RequiredConfigParams,
 } from "./config.js";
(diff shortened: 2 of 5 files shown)


dyoshikawa:
Thanks — merging. Fixes the #1978 root-file-ownership follow-up findings (wildcard `["*"]` configs and rovodev's mirrored `./AGENTS.md`), closing #1981.

PR #2010, opened 2026-06-23
test(mcp): add cross-generator field-stripping contract tests

### LINKED ISSUE #1659: Test: add cross-generator contract tests for MCP field stripping
## Background

PR #1637 cleaned up `envVars` handling for Codex CLI MCP configurations. During the review of the broader changes, we identified two cross-generator testing gaps that are worth closing in a dedicated PR.

## Details

1. **Cursor Source-Preservation Assertion**
   We need to explicitly test and assert that fields like `targets` are preserved in the Rulesync source but stripped accurately at emit time across the Cursor generator.
2. **Parametric Non-Codex MCP Generator Leak Contract**
   We should implement a parameterized contract test verifying that *every* non-Codex MCP generator correctly strips `envVars`, `targets`, `description`, and `exposed` fields, ensuring they do not leak into output configs (like they historically did).

## Next Steps

Add the cross-generator contract tests outlined above to harden the MCP export paths universally.

dyoshikawa:
## Summary

Adds the cross-generator MCP contract tests requested in #1659, hardening the MCP export paths universally.

`RulesyncMcp.getMcpServers()` omits the rulesync-source-only fields (`targets`, `description`, `exposed`) plus the codex-only `envVars` field, and every non-codex tool MCP generator builds its output from `getMcpServers()`. These tests pin that contract so a future generator that bypasses the canonical stripping is caught.

## Changes

- **Parametric non-codex leak contract**: `it.each(toolMcpFactories)` (excluding `codexcli`) builds a `RulesyncMcp` whose server carries `targets`/`description`/`exposed`/`envVars` and asserts none of those field names or canary values leak into the emitted `getFileContent()` — format-agnostic, covering all 28 non-codex generators (global-only tools like augmentcode are driven in global mode).
- **Cursor source-preservation assertion**: verifies the rulesync source retains all four fields verbatim while the Cursor emit strips them, keeping only the real MCP server config.

Codex is the deliberate exception (it reads `envVars` directly to emit `[mcp_servers.<name>.env]`).

## Verification

- `pnpm cicheck` passes; the new file adds 29 passing tests. Test-only — no production changes were needed (all generators already strip correctly).

Closes #1659

### DIFF
--- a/src/features/mcp/mcp-field-stripping.test.ts
+++ b/src/features/mcp/mcp-field-stripping.test.ts
@@ -1,6 +1,7 @@
-import { describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { RULESYNC_RELATIVE_DIR_PATH } from "../../constants/rulesync-paths.js";
+import { setupTestDirectory } from "../../test-utils/test-directories.js";
 import { CursorMcp } from "./cursor-mcp.js";
 import { toolMcpFactories } from "./mcp-processor.js";
 import { RulesyncMcp } from "./rulesync-mcp.js";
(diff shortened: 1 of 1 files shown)


dyoshikawa:
Thanks — merging. Adds cross-generator MCP field-stripping contract tests (parametric non-codex leak contract + Cursor source-preservation), closing #1659.

PR #2012, opened 2026-06-23
fix(codexcli): preserve per-tool MCP approval_mode on regenerate

### LINKED ISSUE #1709: Codex MCP approvals
Hey.

This could've been reported before, but I didn't find relevant issues. The problem currently is some MCPs in Codex require explicit approvals stored in Codex config, even when running in "Full access" mode. One example is Playwright - whenever you Codex attempts to use Playwright tools, it explicitly asks user permissions. When you click "Approve", it writes these `approval_mode` lines to `.codex/config.toml`:

```toml
sandbox_mode = "danger-full-access"

[mcp_servers.playwright]
type = "stdio"
command = "npx"
args = [ "@playwright/mcp@latest" ]

[mcp_servers.playwright.tools.browser_navigate]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_click]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_type]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_evaluate]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_run_code_unsafe]
approval_mode = "approve"

[mcp_servers.playwright.tools.browser_press_key]
approval_mode = "approve"
```

For Rulesync, these approvals are just keys nested under `mcp_servers.playwright.tools`. So when you run Rulesync after, it generates a new presentation for `mcp_servers.playwright` key, without the `mcp_servers.playwright.tools` key in it, and overwrites the approvals in `.codex/config.toml`. 

Would it make sense here that we always do recursive deep merges instead of overwrites, and use `undefined` as value for keys that we want to explicitly unset, like so?

```typescript
// config.toml part generated by Rulesync, to be recursively merged with current config
const newConfigPart = {
  mcp_servers: {
    playwright: {
      type: 'stdio',
      command: 'npx',
      args: ['@playwright/mcp@latest'],
      // If those are set in the current `.codex/config.toml` - unset/remove those
      enabled: undefined,
      cwd: undefined,
    }
  }
};
```

I can provide a fix like that for either just Codex or all of Rulesync if you feel like this makes sense :)

dyoshikawa:
## Summary

Fixes #1709. When a user approves an MCP tool in Codex, the CLI saves that decision as a nested table in `.codex/config.toml`:

```toml
[mcp_servers.playwright.tools.browser_navigate]
approval_mode = "approve"
```

rulesync's `CodexcliMcp.fromRulesyncMcp` rebuilt the entire `mcp_servers` table from its own model (command/args/env/env_vars/enabled_tools/…), which does **not** include this per-tool `tools` table, and assigned it wholesale — so every `rulesync generate` silently **wiped the user's saved approvals**, re-introducing approval prompts.

## Fix

Re-merge each existing server's `tools` approval table onto the regenerated server config (only when rulesync doesn't itself emit a `tools` key, which it never does for codex). rulesync still fully owns every field it emits; only the unmodeled approval-policy state is preserved.

## Verification

- New regression test asserts that pre-existing `tools.<tool>.approval_mode` entries survive a regenerate while the rulesync-owned `command`/`args` are still written. (Fails without the fix.)
- `pnpm cicheck` passes (68 codex MCP tests + full suite).

Closes #1709

### DIFF
--- a/src/features/mcp/codexcli-mcp.test.ts
+++ b/src/features/mcp/codexcli-mcp.test.ts
@@ -411,6 +411,40 @@ approval_mode = "approve"
       });
     });
 
+    it("does not clobber a rulesync-emitted tools value with the preserved approval table", async () => {
+      // When rulesync itself supplies a `tools` value, rulesync owns it: the
+      // preserved approval table must NOT override it.
+      const existingToml = `[mcp_servers.srv]
+command = "node"
+
+[mcp_servers.srv.tools.some_tool]
+approval_mode = "approve"
+`;
+      await ensureDir(join(testDir, ".codex"));
+      await writeFileContent(join(testDir, ".codex/config.toml"), existingToml);
+
+      const jsonData = {
+        mcpServers: {
+          srv: { command: "node", tools: ["explicit_tool"] },
+        },
+      };
+      const rulesyncMcp = new RulesyncMcp({
+        relativeDirPath: RULESYNC_RELATIVE_DIR_PATH,
+        relativeFilePath: ".mcp.json",
+        fileContent: JSON.stringify(jsonData),
+      });
+
+      const codexcliMcp = await CodexcliMcp.fromRulesyncMcp({
+        outputRoot: testDir,
+        rulesyncMcp,
+        global: true,
+      });
+
+      const servers = codexcliMcp.getToml().mcp_servers as Record<string, Record<string, unknown>>;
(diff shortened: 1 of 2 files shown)


dyoshikawa:
Thanks — merging. Preserves per-tool MCP `approval_mode` state in `.codex/config.toml` across regenerates, closing #1709.

autaut03:
Thank you @dyoshikawa :)

PR #2014, opened 2026-06-23
refactor(opencode): apply strict frontmatter schema to OpenCodeSubagent

### LINKED ISSUE #1658: Refactor: apply strict frontmatter schema pattern across all OpenCodeStyle subagents
## Background

During the review of PR #1655 (which explicitly typed the `KiloSubagent` frontmatter schema), we noticed that `OpenCodeSubagent` (and sibling tools extending the same parent) still use the loose-alias pattern.

## Details

The pattern where a subagent merely aliases `OpenCodeStyleSubagentFrontmatterSchema` rather than declaring its own explicit schema (like we did for Kilo) exists across the family of OpenCode-style tools. 

## Next Steps

Applying the pattern introduced in #1655 across the family for consistency would tighten validation and make the supported surface area clear for each specific tool.

dyoshikawa:
## Summary

Closes #1658. `OpenCodeSubagent` merely aliased the shared `OpenCodeStyleSubagent` base frontmatter schema, whereas `KiloSubagent` (the other tool extending the same parent) declares an explicit schema since #1655. This applies the same strict-schema pattern across the OpenCode-style family.

## Changes

- Declare an explicit `OpenCodeSubagentFrontmatterSchema` (`z.looseObject`) enumerating the documented OpenCode agent fields: `description`, `mode` (default `subagent`), `name`, `model`, `temperature`, `top_p`, `prompt`, `disable`, `tools`, `permission`.
- Mirror the Kilo pattern: normalize the `mode` default in the constructor, `declare` the narrowed `frontmatter` type, validate through the schema in `validate()`, and `safeParse` in `fromRulesyncSubagent()` / `fromFile()`.
- The schema stays `looseObject`, so forward-compatible/unknown fields (e.g. `hidden`) still round-trip — behavior is unchanged for valid configs; validation is merely tightened for the declared field types.

## Verification

- `pnpm cicheck` passes (typecheck + 12 OpenCodeSubagent tests + subagents processor + full suite). New test asserts the declared field types validate, a wrong type is rejected, and unknown fields still round-trip.

Closes #1658

### DIFF
--- a/src/features/subagents/opencode-subagent.test.ts
+++ b/src/features/subagents/opencode-subagent.test.ts
@@ -204,6 +204,35 @@ Assist with any tasks`,
     expect(result.success).toBe(true);
   });
 
+  it("should validate the documented field types but keep unknown fields (looseObject)", () => {
+    // Known fields are now typed explicitly (strict-schema pattern, #1658)...
+    expect(
+      OpenCodeSubagentFrontmatterSchema.safeParse({
+        description: "Agent",
+        temperature: 0.4,
+        top_p: 0.9,
+        model: "some-model",
+        tools: { read: true, write: false },
+      }).success,
+    ).toBe(true);
+    // ...a wrong type for a declared field is rejected...
+    expect(
+      OpenCodeSubagentFrontmatterSchema.safeParse({
+        description: "Agent",
+        temperature: "hot",
+      }).success,
+    ).toBe(false);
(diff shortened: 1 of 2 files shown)


dyoshikawa:
Thanks — merging. Applies the explicit strict-frontmatter-schema pattern to OpenCodeSubagent (matching KiloSubagent), closing #1658.

PR #2019, opened 2026-06-23
feat(hermesagent): add Hermes Agent support for rules and MCP

dyoshikawa:
## Summary

Adds a new tool target `hermesagent` for [Hermes Agent](https://hermes-agent.nousresearch.com) by Nous Research.

- **Rules** → project-root `.hermes.md` (project scope only). Hermes auto-injects `.hermes.md` from the project root into its system prompt.
- **MCP** → `~/.hermes/config.yaml` under the top-level `mcp_servers` key (global only). Merges into the shared config (preserving unrelated keys like `model`/`terminal`), mirroring the existing Goose `config.yaml` pattern.

Skills, commands, subagents, hooks, ignore, and permissions are intentionally out of scope for this first iteration (those Hermes surfaces are global-only/agent-managed) and therefore appear as unsupported in the generated supported-tools matrix.

## Changes
- New: `hermesagent-rule.ts`, `hermesagent-mcp.ts`, `hermesagent-paths.ts` (+ unit tests)
- Registered the target in tuples, display names, rules/mcp processors, and the relevant e2e matrices
- Regenerated supported-tools tables and `.gitignore`

## Verification
`pnpm cicheck` passes (lint, types, 6817 tests, content/spelling/secrets).

Closes #2018

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2018, opened 2026-06-23
Add Hermes Agent support


rudironsoni:
## Summary

Add Hermes Agent as a supported Rulesync target.

Homepage: https://hermes-agent.nousresearch.com

Hermes Agent is a Nous Research agent with desktop and CLI install surfaces. Its site currently describes it as open source under the MIT License and shows Hermes Agent v0.17.0.

## Why

Rulesync already supports several AI coding and agent tools. Hermes Agent looks like a natural addition if it has stable local files for instructions, commands, skills, MCP config, hooks, or subagents.

## Scope to investigate

- Add a new tool target id, likely `hermes` or `hermesagent` depending on project naming conventions.
- Identify Hermes Agent's local config paths for global and project-level instructions.
- Check whether Hermes supports commands, skills, MCP config, hooks, or subagents in a form Rulesync can generate.
- Wire the target through the target schema, supported-tools docs, README/generated docs, feature processors, and tests as applicable.
- Add conversion support only if Hermes has a compatible source or destination format.

## Acceptance criteria

- `rulesync --targets hermes` or the chosen target id is accepted.
- Generated files land in the correct Hermes Agent locations.
- Unsupported Hermes features are documented clearly rather than silently ignored.
- Unit and e2e coverage are added for the supported feature set.
- `docs/reference/supported-tools.md` and related generated docs list Hermes Agent with accurate feature support.

PR #2023, opened 2026-06-23
fix(hermesagent): carry MCP server timeout through conversion

dyoshikawa:
## Summary

Follow-up to #2019 (Hermes Agent support). During review of #2019, a finding was raised that the canonical `timeout` field (and the `networkTimeout` alias) was silently dropped in both MCP conversion directions. #2019 was merged before this fix landed, so this PR completes it.

Hermes `mcp_servers` entries support a `timeout` key, so the conversion now maps it like the Goose adapter does:
- Export: `timeout` / `networkTimeout` (canonical) → `timeout` (Hermes)
- Import: `timeout` (Hermes) → `timeout` (canonical)

Adds tests for both directions.

Refs #2018

### DIFF
--- a/src/features/mcp/hermesagent-mcp.test.ts
+++ b/src/features/mcp/hermesagent-mcp.test.ts
@@ -173,6 +173,25 @@ describe("HermesagentMcp", () => {
       expect(server?.args).toEqual(["-y", "server", "--flag"]);
     });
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
Merged — completes the Hermes Agent MCP timeout handling from #2018.

PR #2026, opened 2026-06-23
feat: enhance Hermes Agent support

rudironsoni:
## Summary
- Add Hermes Agent support across rules, skills, MCP, commands, hooks, permissions, and subagents.
- Generate native Hermes subagent plugin files and JSON specs using delegate_task dispatch.
- Preserve shared .hermes/config.yaml content when multiple Hermes features contribute config.
- Update supported-tool docs and feature matrix tests.

## Verification
- pnpm cicheck:code
- pnpm cicheck:content
- Hermes Agent all-feature E2E generation into temp HOME and HERMES_HOME
- Verified generated .hermes/config.yaml retains hooks and rulesync-subagents
- Verified generated plugin uses delegate_task dispatch


.github/copilot-instructions.md
```diff
@@ -1,18 +1,3 @@
...
-  - path: @.github/instructions/feature-change-guidelines.instructions.md
-    description: "When you add or change features, must follow these guidelines."
-  - path: @.github/instructions/github-actions-security.instructions.md
-    description: Guidelines to avoid GitHub Actions script injection vulnerabilities.
-    applyTo[1]: .github/workflows/*.yml
-  - path: @.github/instructions/testing-guidelines.instructions.md
-    description: "When you write tests, must follow these guidelines."
-    applyTo[2]: **/*.test.ts,src/e2e/**/*.spec.ts
-
 # Rulesync Project Overview
```
dyoshikawa, line 16: This drops the whole TOON rules-reference header (including the `github-actions-security.instructions.md` pointer). This file is a generated artifact and the non-root rules still exist, so a full regeneration should keep this header. Looks like an unintended side effect — please restore it.

src/features/permissions/hermesagent-permissions.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  static fromRulesyncPermissions({
+    outputRoot,
+    rulesyncPermissions,
+  }: ToolPermissionsFromRulesyncPermissionsParams): HermesagentPermissions {
+    const permissions = rulesyncPermissions.getJson();
+    const commandAllowlist = Object.entries(permissions.permission ?? {})
+      .filter(([, value]) => value.allow?.length)
+      .flatMap(([command]) => command);
```
dyoshikawa: This allowlist extraction is broken in two ways. `permissions.permission` is a `Record<glob, "allow"|"ask"|"deny">`, so `value.allow` is almost always `undefined` and the filter matches nothing. Even if it did match, `.flatMap(([command]) => command)` spreads the command string into individual characters (e.g. `"bash"` → `["b","a","s","h"]`). You probably want to keep entries whose action value is `"allow"`. A unit test here would catch this.

src/features/rules/rules-processor.ts
```diff
@@ -475,33 +475,27 @@ export const toolRuleFactories = new Map<RulesProcessorToolTarget, ToolRuleFacto
...
-      // `.grok/memories/` directory, so non-root rules are folded into the
-      // single root AGENTS.md below (same handling as warp / deepagents).
-      class: GrokcliRule,
+      class: HermesagentRule,
       meta: {
         extension: "md",
         supportsGlobal: true,
-        ruleDiscoveryMode: "auto",
-        foldsNonRootIntoRoot: true,
+        ruleDiscoveryMode: "toon",
```
dyoshikawa: This new `meta` (`supportsGlobal: true`, `ruleDiscoveryMode: "toon"`, dropping `foldsNonRootIntoRoot`) doesn't match what `HermesagentRule` actually does: its `getSettablePaths()` returns only the project-scope root `.hermes.md` with no non-root path, and the class docstring still says rules are project-scope only and folded into the single root file. With `toon` and no folding, multiple non-root rules all resolve to `.hermes.md` (duplicate ToolRules / dangling TOON references), and `supportsGlobal: true` has no global file backing it. Either update the class to truly support global + toon discovery, or keep the previous meta.

dyoshikawa (commented):
Thanks for the broad push to bring Hermes up to feature parity across rules, skills, commands, hooks, permissions, subagents and MCP. The MCP `networkTimeout` round-trip and its tests look solid. That said, I think a few things need to be sorted out before this can merge.

The biggest one is the permissions allowlist (commented inline) — as written it never produces a usable `command_allowlist`, and there are no unit tests for the new permissions/hooks/command/skill classes, so this and similar issues slip through. Could you add unit tests for the new Hermes classes? The rules-processor `meta` change also seems to contradict what `HermesagentRule` actually emits (inline).

A couple of cross-cutting points worth addressing too:

- The new Hermes features (commands/skills/subagents/hooks/permissions) aren't added to the e2e Tool×Feature happy-path specs (`src/e2e/e2e-*.spec.ts`). The project guidelines ask us to always keep those matrix cases in place, so please add `hermesagent` cases there.
- `src/types/feature-processor.ts` hardcodes `.hermes/config.yaml` in the generic base class and matches it with `endsWith(".hermes/config.yaml")`. Since `getFilePath()` builds paths with `path.join`, this won't match on Windows (`.hermes\config.yaml`), and tool-specific strings in the base class break the abstraction. Better to model "is this file mergeable" as a hook on the tool class.
- `HermesagentSubagent` extends `AiFile` directly rather than `ToolSubagent`, which forces the double `as unknown as` cast in the factory registration. Worth aligning with the other subagent classes or widening the factory type properly.
- `.github/copilot-instructions.md` lost its TOON rules-reference header (inline) — looks like an unintended side effect of local regeneration.

Smaller stuff: `hermesagent-command.ts` hardcodes the SKILL.md description instead of using the rulesync command's `description`, and its `toRulesyncCommand()` embeds the generated frontmatter into the command body which breaks the import round-trip; and the `import { join }` at the end of `hermesagent-paths.ts` should move to the top. Happy to dig into any of these further.

.github/copilot-instructions.md
rudironsoni: Restored in 359cb6d6. I pulled the generated TOON header back from upstream/main, including the github-actions-security.instructions.md pointer. `check:sync-skill-docs` and `pnpm cicheck` both pass locally now.

src/features/permissions/hermesagent-permissions.ts
rudironsoni: Fixed in 359cb6d6. The allowlist now reads the canonical permission shape by action value, so `bash: { "git *": "allow" }` becomes `command_allowlist: ["git *"]` without splitting strings. I also added focused coverage for that path and for preserving existing `.hermes/config.yaml` content while writing permissions.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in 359cb6d6. Hermes rules are back to project scope with `ruleDiscoveryMode: "auto"` and `foldsNonRootIntoRoot: true`, which matches the `.hermes.md` adapter behavior. I updated the tests and regenerated the support tables.

src/types/feature-processor.ts
```diff
@@ -61,8 +61,19 @@ export abstract class FeatureProcessor {
     const changedPaths: string[] = [];
     for (const aiFile of aiFiles) {
       const filePath = aiFile.getFilePath();
+      const existingFileContent = await readFileContentOrNull(filePath);
+
+      if (
+        existingFileContent !== null &&
+        filePath.endsWith(".hermes/config.yaml") &&
```
dyoshikawa: `getFilePath()` builds the path with `path.join`, so on Windows this is `.hermes\config.yaml` and `endsWith(".hermes/config.yaml")` never matches — the config-preservation merge silently stops working there. Also, hardcoding a tool-specific path in this generic base class leaks the abstraction. Better to model "is this file mergeable" as a hook on the tool class rather than a string check here.

src/features/commands/hermesagent-command.ts
```diff
@@ -0,0 +1,84 @@
...
+function commandSlug(relativeFilePath: string): string {
+  return basename(relativeFilePath, ".md").replace(/[^a-zA-Z0-9_-]/g, "-");
+}
+
+function commandSkillContent(rulesyncCommand: RulesyncCommand): string {
+  const slug = commandSlug(rulesyncCommand.getRelativeFilePath());
+  return [
+    "---",
+    `name: ${slug}`,
+    `description: ${slug} command`,
```
dyoshikawa: The description is hardcoded to `${slug} command` instead of using the rulesync command's `description` frontmatter. Also `rulesyncCommand.getFileContent()` below includes the rulesync frontmatter in the body — you likely want `getBody()` so the generated SKILL.md doesn't embed the source frontmatter.

src/features/commands/hermesagent-command.ts
```diff
@@ -0,0 +1,84 @@
...
+    return { success: true, error: null };
+  }
+
+  toRulesyncCommand(): RulesyncCommand {
+    const slug = basename(dirname(this.getRelativePathFromCwd()));
+    return new RulesyncCommand({
+      relativeDirPath: RULESYNC_COMMANDS_RELATIVE_DIR_PATH,
+      relativeFilePath: `${slug}.md`,
+      frontmatter: {},
+      body: this.getFileContent(),
```
dyoshikawa: `fromRulesyncCommand` stores content that already starts with a `---` name/description frontmatter block, and here `toRulesyncCommand()` puts that whole `getFileContent()` straight into `body` with `frontmatter: {}`. Importing the generated file back will leave a literal `---` block inside the command body, breaking the round-trip.

src/features/subagents/hermesagent-subagent.ts
```diff
@@ -0,0 +1,310 @@
...
+    prompt: rulesyncSubagent.getBody(),
+    toolsets: ["terminal", "file", "web"],
+    hermes: {
+      command: hermesCommandName(slug),
+      dispatch: "delegate_task",
+    },
+  };
+}
+
+export class HermesagentSubagent extends AiFile {
```
dyoshikawa: This extends `AiFile` directly rather than `ToolSubagent` like the other subagent classes, which is what forces the `as unknown as ...` double cast at factory registration (and similar casts ripple into the commands/hooks registrations). Aligning with `ToolSubagent` — or widening the factory type properly — would remove those casts.

src/lib/generate.ts
```diff
@@ -105,6 +105,45 @@ async function processDirFeatureGeneration(params: {
   return { count: totalCount, paths: allPaths, hasDiff };
 }
 
+function mergeDuplicateAiFiles(toolFiles: AiFile[]): AiFile[] {
```
dyoshikawa: This adds a generic cross-tool merge stage to `processFeatureWithRulesyncFiles` to solve a Hermes-specific problem (multiple features writing the same `config.yaml`). The two branches are asymmetric (one mutates in place + `continue`, the other does last-write-wins via `merged.set`), and non-mergeable files at the same path are silently overwritten. This is the same abstraction leak as the `feature-processor` path check — worth keeping the Hermes concern out of the shared path, and it has no unit test.

src/features/hermes-config.ts
```diff
@@ -0,0 +1,25 @@
+import { load, dump } from "js-yaml";
+
+type HermesConfig = Record<string, unknown>;
+
+export function parseHermesConfig(fileContent: string): HermesConfig {
+  if (!fileContent.trim()) {
+    return {};
+  }
+  const parsed = load(fileContent);
+  if (parsed && typeof parsed === "object" && !Array.isArray(parsed)) {
```
dyoshikawa: Minor / defense-in-depth: the MCP adapter (`hermesagent-mcp.ts`) and the recent goose hardening reject non-plain objects with `isPlainObject` and strip `__proto__`/`constructor` for prototype-pollution safety. This shared parser only checks `typeof === "object"`. I confirmed it doesn't poison the global prototype today, but for parity it'd be good to reuse `isPlainObject` + `omitPrototypePollutionKeys` here so a malicious imported `config.yaml` can't round-trip a literal `__proto__` key.

dyoshikawa (commented):
Thanks for the follow-up round — the core blockers from the last review are resolved. The permissions allowlist now extracts correctly and has unit tests, the rules-processor `meta` matches what `HermesagentRule` emits, and there are new tests for permissions/hooks/mcp/subagents. CI is fully green too.

A couple of things still need attention before this can merge, though.

The big one is the e2e Tool×Feature happy-path matrix. The new Hermes commands/skills/hooks/permissions/subagents features aren't represented in `src/e2e/e2e-*.spec.ts` (only rules and mcp have `hermesagent` cases). The project guidelines require always keeping those matrix cases in place, so please add `hermesagent` cases for the new features.

The Windows path check in `feature-processor.ts` is a real cross-platform bug (inline). Beyond that there are a few mid-level items on the command import round-trip and the subagent inheritance that I left inline, plus the generic merge stage added to `generate.ts` for what is really a Hermes-specific concern. Happy to talk through any of these.

src/types/feature-processor.ts
rudironsoni: Fixed in 01026e4b. FeatureProcessor now uses a generic shouldMergeExistingFileContent hook instead of checking a Hermes path string, so Windows path separators are no longer part of the behavior. The Hermes config writers opt in at the class level.

src/features/commands/hermesagent-command.ts
rudironsoni: Fixed in 01026e4b. Hermes command generation now uses the Rulesync command description and body, so the generated SKILL.md does not embed the source frontmatter. I added focused coverage for this path.

src/features/commands/hermesagent-command.ts
rudironsoni: Fixed in 01026e4b. Import now parses the Hermes skill frontmatter and maps only the description plus body back into RulesyncCommand, so the round trip no longer leaves a literal frontmatter block in the command body.

src/features/subagents/hermesagent-subagent.ts
rudironsoni: Fixed in 01026e4b. HermesagentSubagent now extends ToolSubagent, and the processor registration no longer needs the double cast.

src/lib/generate.ts
rudironsoni: Fixed in 01026e4b. I removed the generic duplicate-file merge from generate.ts. Existing Hermes config preservation now lives behind the tool-level merge hook and each Hermes config writer owns its merge behavior.

src/features/hermes-config.ts
rudironsoni: Fixed in 01026e4b. The shared Hermes config parser now requires a plain-object root and strips prototype-pollution keys recursively before round-tripping. I added parser coverage for that case.

src/features/mcp/hermesagent-mcp.ts
```diff
@@ -185,6 +200,22 @@ export class HermesagentMcp extends ToolMcp {
...
+  override setFileContent(fileContent: string): void {
+    const config = parseHermesConfig(fileContent, this.relativeDirPath, this.relativeFilePath);
+    const mcpServers = isRecord(this.config.mcp_servers) ? this.config.mcp_servers : {};
+    const merged = mergeHermesMcpServers(
+      config,
+      mcpServers as Record<string, Record<string, unknown>>,
+    );
+
+    this.config = merged;
+    super.setFileContent(dump(merged));
```
dyoshikawa, line 216: This serializes with raw `dump(merged)`, while the shared `hermes-config.ts` (used by hooks/permissions/subagents) goes through `stringifyHermesConfig` with `{ sortKeys: false }` + `.trimEnd() + "\n"`. Since all of these features write to the same `~/.hermes/config.yaml`, the differing formatting can cause non-idempotent diff churn. The MCP adapter also keeps its own private `parseHermesConfig`. Worth consolidating onto `hermes-config.ts` so the serialization matches and the parser logic isn't duplicated.

src/lib/generate.ts
```diff
@@ -145,7 +145,7 @@ async function processFeatureWithRulesyncFiles(params: {
   if (rulesyncFiles.length === 0) {
     return processEmptyFeatureGeneration({ config, processor, skipFilePaths });
   }
-  const toolFiles = await processor.convertRulesyncFilesToToolFiles(rulesyncFiles);
+  let toolFiles = await processor.convertRulesyncFilesToToolFiles(rulesyncFiles);
```
dyoshikawa, line 148: `toolFiles` is no longer reassigned now that the merge stage is gone — this can go back to `const`.

src/types/feature-processor.ts
```diff
@@ -61,8 +61,19 @@ export abstract class FeatureProcessor {
     const changedPaths: string[] = [];
     for (const aiFile of aiFiles) {
       const filePath = aiFile.getFilePath();
+      const existingFileContent = await readFileContentOrNull(filePath);
+
+      if (
+        existingFileContent !== null &&
+        "shouldMergeExistingFileContent" in aiFile &&
```
dyoshikawa, line 68: `shouldMergeExistingFileContent` is declared on the `AiFile` base class now, so the `"..." in aiFile && typeof ... === "function"` guard is redundant — `aiFile.shouldMergeExistingFileContent()` alone is enough and keeps the type contract explicit rather than duck-typed.

src/constants/hermesagent-paths.ts
```diff
@@ -21,3 +21,30 @@ export const HERMESAGENT_GLOBAL_DIR = ".hermes";
 
 /** MCP servers and other settings live in `config.yaml` under `~/.hermes/`. */
 export const HERMESAGENT_MCP_FILE_NAME = "config.yaml";
+export const HERMESAGENT_SOUL_FILE_NAME = "SOUL.md";
+export const HERMESAGENT_CONFIG_FILE_NAME = "config.yaml";
```
dyoshikawa, line 25: `HERMESAGENT_CONFIG_FILE_NAME` and `HERMESAGENT_MCP_FILE_NAME` are both `"config.yaml"`. Two constants for the same value invites drift — consider collapsing to one.

src/features/commands/hermesagent-command.ts
```diff
@@ -0,0 +1,86 @@
...
+  static getSettablePaths({ slug = "command" }: { slug?: string; global?: boolean } = {}) {
+    return {
+      relativeDirPath: join(HERMESAGENT_SKILLS_DIR_PATH, slug),
+      relativeFilePath: SKILL_FILE_NAME,
+    };
+  }
+
+  constructor({ slug, ...params }: HermesagentCommandParams) {
+    const resolvedSlug =
+      slug ?? basename(dirname(params.relativeDirPath)) ?? commandSlug(params.relativeFilePath);
```
dyoshikawa, line 46: `basename(dirname(...))` never returns nullish, so the `?? commandSlug(...)` fallback is dead code. Minor, but worth trimming for clarity.

dyoshikawa (commented):
Nice turnaround — most of the last round is sorted. The `feature-processor` path check is now a polymorphic `shouldMergeExistingFileContent()` hook (no more Windows-breaking `endsWith` or hardcoded tool string), the command class uses the real `description` frontmatter and `getBody()` with a proper round-trip, `HermesagentSubagent` extends `ToolSubagent` so the double casts are gone, and the generic merge stage in `generate.ts` was removed. The shared `hermes-config.ts` parser also got the prototype-pollution hardening (isPlainObject + recursive `__proto__`/`constructor` stripping), which clears the security note from before.

One blocker still stands: the e2e Tool×Feature happy-path matrix. The five new features (commands, skills, hooks, permissions, subagents) still have no `hermesagent` case — only `e2e-rules.spec.ts` and `e2e-mcp.spec.ts` cover it. The project guidelines treat keeping those matrix cases in place as mandatory, so please add `hermesagent` rows to `src/e2e/e2e-commands.spec.ts`, `e2e-skills.spec.ts`, `e2e-hooks.spec.ts`, `e2e-permissions.spec.ts`, and `e2e-subagents.spec.ts` (each is a one-line addition to the existing `it.each` table). That's the one thing holding up the merge.

The rest are small cleanups I left inline — the MCP/shared-config DRY divergence is the most worth doing, the others are minor. Thanks for sticking with this.

dyoshikawa:
@rudironsoni Thank you!

PR #2040, opened 2026-06-24
feat(factorydroid): collapse commandBlocklist onto deny on import

### LINKED ISSUE #1883: Follow up Factory Droid upstream updates: commandBlocklist hard-block permission tier
## Summary

Factory Droid's `settings.json` defines **three** command permission tiers, but rulesync models only two (`commandAllowlist`/`commandDenylist`, tracked previously). The hard-block tier `commandBlocklist` has no representation.

## Recent Releases

- Settings reference (current as of 2026-06-15): https://docs.factory.ai/cli/configuration/settings

## Gaps

- **permissions — `commandBlocklist` (hard-block tier) not modeled.** Factory Droid `settings.json` has `commandAllowlist`, `commandDenylist`, **and** `commandBlocklist`. The blocklist is a distinct, stronger tier: blocked commands "can never run… no prompt and no way to approve them: the block applies even under full autonomy, auto-run, or `--skip-permissions-unsafe`," and the resolver inspects the actual program invoked to prevent wrapper bypasses. The denylist only forces confirmation and can be approved. `src/features/permissions/factorydroid-permissions.ts` models only `commandAllowlist`/`commandDenylist`; `commandBlocklist` appears nowhere in `src/`. The canonical action model in `src/types/permissions.ts` is `allow | ask | deny` only — there is no "block" action, so the never-run-even-under-autonomy semantic cannot be expressed or round-tripped, and an existing `commandBlocklist` is silently ignored on import (preserved only as a non-managed key).

## Proposed Follow-up

- Decide how to represent Factory's hard-block tier: either (a) introduce a canonical "block"/stronger-than-deny action mapped to `commandBlocklist`, or (b) at minimum read/preserve `commandBlocklist` explicitly and document that rulesync does not manage it.

## References

- https://docs.factory.ai/cli/configuration/settings — `commandAllowlist` / `commandDenylist` / `commandBlocklist` semantics.

dyoshikawa:
## Background

Follow-up to #1883. Factory Droid's `settings.json` defines a third, stronger command-permission tier — `commandBlocklist` — whose commands can never run, not even under full autonomy / auto-run. rulesync's canonical action model is `allow | ask | deny` only, with no never-runs equivalent, so the import path previously ignored `commandBlocklist` and silently dropped it from the canonical view.

Per the maintainer decision, this resolves the gap with the **lossy collapse** option.

## What changed

- Add `commandBlocklist?: string[]` to the modeled settings shape.
- On **import** (`toRulesyncPermissions`), map `commandBlocklist` entries onto canonical `deny` under the `bash` category, applied last so a hard-block command outranks an allowlist entry. Lossy: the never-runs guarantee weakens to a deny the user can still approve, but the rule is no longer dropped.
- On **export** (`fromRulesyncPermissions`), behavior is unchanged: with no canonical `block` action rulesync never emits `commandBlocklist`, and an existing one on disk stays preserved verbatim as an unmanaged key (via the existing `{...settings}` merge).

## Tests

- New cases: `commandBlocklist` collapses onto `deny`; a hard-block command outranks an allowlist entry; an existing `commandBlocklist` is preserved verbatim on export.
- `pnpm cicheck` green.

## Out of scope

The 2026-06-22 re-check comment on #1883 surfaced two additional, separate Factory Droid gaps — `settings.local.json` override layer and global `~/.factory/rules/` non-root rules. Those are distinct features and are not addressed here; I will file a fresh scrap issue to track them so they are not lost.

Closes #1883


dyoshikawa:
@dyoshikawa Thank you!

PR #2049, opened 2026-06-25
fix(goose): fold non-root rules into root .goosehints; document skills path decision

dyoshikawa:
## Summary

Resolves the two Goose surface divergences reported in #2038. Each was verified against the current code and the cited upstream docs before changing behavior.

## Finding 1 (primary, fixed): non-root rules written to an inert directory

rulesync routed detail/scoped (non-root) Goose rules to `.goose/memories/*.md`. Per Goose's hints doc, session context is loaded only from the `.goosehints` / `AGENTS.md` family (the configured `CONTEXT_FILE_NAMES`) — discovered from the working directory up to the repo root plus nested directories Goose touches. `.goose/memories/` is the separate Memory extension's storage and is **not** auto-loaded as context, so those non-root rules were effectively invisible to Goose.

**Fix:** Switch the goose rules target to the fold-into-root pattern already used by `grokcli` / `warp` / `deepagents`. The `nonRoot` settable path is removed, and the convention now uses `ruleDiscoveryMode: "auto"` + `foldsNonRootIntoRoot: true`, so every non-root rule body is merged into the single root `.goosehints` that Goose actually reads. Root-rule behavior is unchanged.

I chose folding over emitting nested `.goosehints` files because rulesync's topic-based non-root rules have no project subdirectory to map onto (they are not subproject rules), so there is no nested directory for Goose's hierarchical loader to pick them up from. Folding guarantees the content is loaded, mirrors the existing handling for analogous AGENTS.md-family tools, and avoids a large architectural change. This matches the established precedent comment on the grokcli target.

- `src/features/rules/goose-rule.ts`: `getSettablePaths` is now root-only (`nonRoot: undefined`); `fromFile` / `fromRulesyncRule` / `forDeletion` updated accordingly, with a doc comment citing the loader behavior.
- `src/features/rules/rules-processor.ts`: goose meta switched to `auto` + `foldsNonRootIntoRoot`.
- Updated `goose-rule.test.ts` and added a goose fold happy-path case to `src/e2e/e2e-rules.spec.ts`.
- Regenerated `.gitignore` (drops the now-unemitted `**/.goose/memories/`).

## Finding 2 (low severity): skills emit the legacy `.goose/skills/` path

Upstream now recommends `.agents/skills/` (project) and `~/.agents/skills/` (global), discovering `.goose/skills/` only for backward compatibility. rulesync's `goose` skill target writes the legacy `.goose/skills/` and throws on `--global` (delegating global to `agentsskills`).

[...]

PR #2054, opened 2026-06-25
refactor(rules): tidy GooseRule paths type and drop dead excludeToolDir

### LINKED ISSUE #2051: Follow-ups from PR #2049 review: goose fold-into-root cleanup and stale .goose/memories migration
## Background

PR #2049 switched the goose rules target to the fold-into-root pattern (`foldsNonRootIntoRoot: true`), mirroring the existing `grokcli` / `warp` / `deepagents` targets. The PR was reviewed and merged with **no merge-blocker findings**. During review, three low-severity (non-blocking) findings were surfaced and captured here for later instead of blocking the merge.

### Finding 1 (low) — Stale `.goose/memories/*.md` files are not cleaned up on regeneration
After this change, `goose-rule.ts` `getSettablePaths` sets `nonRoot: undefined`, so the non-root deletion branch in `rules-processor.ts` returns `[]` for goose. Users upgrading from a prior rulesync version that emitted `.goose/memories/detail.md` will keep those orphaned files on disk after regeneration. This is identical to the existing `grokcli` / `warp` / `deepagents` behavior (none of them clean their legacy memories dirs), so it is consistent with precedent and acceptable — but it is a real migration footgun worth a note in the changelog/docs or a dedicated cleanup mechanism.

### Finding 2 (low) — Type definition diverges from cited grokcli precedent
`goose-rule.ts` defines `GooseRuleSettablePaths` as a fresh `{ root: {...}; nonRoot?: undefined }` object type, whereas `grokcli` (`grokcli-rule.ts`) uses `Pick<ToolRuleSettablePaths, "root"> & { root: {...}; nonRoot?: undefined }`. Both compile and are functionally equivalent, but since the doc comment states it mirrors `grokcli` / `warp` / `deepagents`, matching the `Pick<>` form keeps the pattern uniform and makes future refactors of `ToolRuleSettablePaths` propagate consistently. Cosmetic.

### Finding 3 (low) — Dead `excludeToolDir` parameter
`goose-rule.ts` `getSettablePaths` still declares `excludeToolDir?: boolean` in its options type but no longer uses it (the `buildToolPath(GOOSE_DIR, "memories", excludeToolDir)` call was removed). `grokcli` keeps the same vestigial parameter too, so it is consistent with precedent, but it is genuinely unused now and could be dropped for clarity. Cosmetic.

## Solution / Next Steps

- (a) Consider documenting and/or handling the stale `.goose/memories` migration footgun — ideally across all fold-into-root tools (`grokcli` / `warp` / `deepagents` / `goose`), not just goose, since they share the same gap.
- (b) A small cleanup pass to align the `GooseRuleSettablePaths` type definition with the `Pick<>` form and drop the unused `excludeToolDir` parameter.

_Source: review of PR #2049._

dyoshikawa:
## Background

Follow-up to the PR #2049 review, which switched the goose rules target to the fold-into-root pattern. Three low-severity, non-blocking findings were captured in #2051. This PR resolves the two concrete refactoring findings (Findings 2 and 3).

## Changes

- **Finding 2** — Define `GooseRuleSettablePaths` via `Pick<ToolRuleSettablePaths, "root">`, matching the `grokcli` precedent it documents itself as mirroring, so future refactors of `ToolRuleSettablePaths` propagate consistently.
- **Finding 3** — Drop the now-dead `excludeToolDir` option from `GooseRule.getSettablePaths`. Goose folds non-root rules into the root `.goosehints` and no longer calls `buildToolPath`, so the parameter was unused.

## On Finding 1 (stale `.goose/memories` migration footgun)

Intentionally not addressed here. The issue itself concluded this is "acceptable and consistent with precedent" — `grokcli`/`warp`/`deepagents` share the exact same behavior and none clean their legacy memories dirs. A cross-tool cleanup mechanism would be a separate design decision, out of scope for this cosmetic cleanup.

## Verification

`pnpm cicheck` passes (6546 tests, lint, typecheck, content checks).

Closes #2051

### DIFF
--- a/src/features/rules/goose-rule.ts
+++ b/src/features/rules/goose-rule.ts
@@ -9,12 +9,13 @@ import {
   ToolRuleFromFileParams,
   ToolRuleFromRulesyncRuleParams,
   ToolRuleParams,
+  ToolRuleSettablePaths,
   ToolRuleSettablePathsGlobal,
 } from "./tool-rule.js";
 
 export type GooseRuleParams = ToolRuleParams;
 
-export type GooseRuleSettablePaths = {
+export type GooseRuleSettablePaths = Pick<ToolRuleSettablePaths, "root"> & {
   root: {
     relativeDirPath: string;
     relativeFilePath: string;
@@ -48,7 +49,6 @@ export class GooseRule extends ToolRule {
     global,
   }: {
     global?: boolean;
-    excludeToolDir?: boolean;
   } = {}): GooseRuleSettablePaths | GooseRuleSettablePathsGlobal {
     if (global) {
       return {


dyoshikawa:
@dyoshikawa Thank you!

PR #2058, opened 2026-06-26
chore: resolve all knip findings

dyoshikawa:
## Summary

Make `pnpm run knip` pass by cleaning up the ~120 findings it reported (unused exports, unused exported types, duplicate exports, and a config hint). This is a follow-up housekeeping pass; knip is not part of `cicheck` and had drifted out of a passing state.

## Changes

- **De-export internal-only symbols.** The bulk of findings were `export`ed symbols (zod frontmatter schemas, path/event constants, `*Params`/`*Result` types, and the `*ToolSchemas` objects in `src/mcp/**`) that are used only within their own declaring file. The `export` keyword was removed so they become file-local; the handful that were fully dead were deleted along with any imports they orphaned.
- **knip config.** Removed the redundant `src/index.ts` entry (auto-detected from package.json `exports`/`main`/`module`) per knips own hint, and left `includeEntryExports` at its default so the library public API (`Feature`, `ToolTarget`, etc. re-exported from `src/index.ts`) is no longer flagged as unused.
- **Intentional alias exports.** Disabled knips `duplicates` rule for `src/types/hooks.ts`, where Kilo intentionally aliases the OpenCode hook-event constants (`KILO_HOOK_EVENTS = OPENCODE_HOOK_EVENTS`, etc.). Each name is imported by its own tool integration; aliasing keeps the data DRY while letting the two diverge later. A comment documents the rationale.
- **Removed a contrived test.** Deleted the `AugmentcodeLegacyRuleParams` type alias and its dedicated export-assertion test, consistent with how the other `*Params` types were de-exported.

No runtime behavior changes — only `export` removals and dead-code deletion.

## Test plan

- `pnpm run knip` now exits 0
- `pnpm cicheck` passes (fmt, oxlint, `tsgo` typecheck, 6545 tests, and all content checks)
- `pnpm run build` succeeds and the published public exports are unchanged (`ALL_FEATURES`, `ALL_TOOL_TARGETS`, `convertFromTool`, `generate`, `importFromTool`)

### DIFF
--- a/knip.ts
+++ b/knip.ts
@@ -2,10 +2,10 @@ import { type KnipConfig } from "knip";
 
 const config: KnipConfig = {
   entry: [
-    // `src/cli/index.ts` is auto-detected from the package.json `bin` field, so it
-    // does not need to be listed here. The library entry and the test files are kept
-    // explicit because they are not covered by an enabled plugin's defaults.
-    "src/index.ts",
+    // `src/cli/index.ts` (package.json `bin`) and `src/index.ts` (package.json
+    // `exports`/`main`/`module`) are both auto-detected from package.json, so they
+    // do not need to be listed here. The test files are kept explicit because they
+    // are not covered by an enabled plugin's defaults.
     "src/**/*.test.ts",
     // Standalone task runners under scripts/ (executed via tsx) and their colocated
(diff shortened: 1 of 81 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2060, opened 2026-06-26
chore: switch security scan and asset publishing to repomix XML instead of TOON

dyoshikawa:
## Summary

Switches both the scheduled security scan and the release asset publishing to use repomix XML output instead of converting it to TOON first.

### Security scan
- Add `scripts/generate-repomix-xml.ts` that produces `repomix-output-*.xml` variants directly via repomix (`style: "xml"`), dropping the JSON-to-TOON conversion step.
- Rename `scripts/security-scan-toon.ts` to `scripts/security-scan-xml.ts`; it now reads `.xml` files and uses an XML-oriented input-format description in the prompt.
- Rename `getToonFiles` to `getXmlFiles` (now matches `.xml`) and the `runSecurityScan` `toonContent` parameter to `content` in `scripts/security-scan-lib.ts`.
- Update `.github/workflows/security-scan.yml` to run the new XML generation and scan scripts.

### Asset publishing
- Update `.github/workflows/publish-assets.yml` to generate repomix XML variants and publish them as `repomix-xml.zip` (previously `repomix-toon.zip`).
- Delete the now-unused `scripts/generate-repomix-toon.ts`.
- Update `.gitignore` accordingly (`repomix-xml.zip`, drop `repomix-output*.toon`/`repomix-toon.zip`).

The `@toon-format/toon` dependency is intentionally kept because the rules and skills processors (`src/features/rules/rules-processor.ts`, `src/features/skills/skills-processor.ts`) still use TOON for rule discovery output.

## Test plan

- `pnpm cicheck` passes (lint, typecheck, tests, content checks).


dyoshikawa:
@dyoshikawa Thank you!

PR #2062, opened 2026-06-26
fix: align security scan with @openrouter/sdk request shape

dyoshikawa:
## Summary

The scheduled Security Scan workflow ([failing run](https://github.com/dyoshikawa/rulesync/actions/runs/28215330012)) failed for every variant with:

```
SDKValidationError: Input validation failed: [
  { "expected": "object", "code": "invalid_type", "path": ["chatRequest"],
    "message": "Invalid input: expected object, received undefined" }
]
```

`@openrouter/sdk` (0.13.7) expects `chat.send` to receive the request body under `chatRequest` and the app display name under `appTitle`, but the code still passed the older `chatGenerationParams` / `xTitle` field names, so the SDK rejected every call and the scan aborted with "All scans failed".

## Changes
- Rename the `chat.send` request fields in `scripts/security-scan-lib.ts` to match the SDK: `chatGenerationParams` -> `chatRequest`, `xTitle` -> `appTitle` (both the scan and the overall-summary calls).
- Assert the request shape in `runSecurityScan` tests so a future field-name drift fails fast instead of only at runtime.

## Test plan
- `pnpm cicheck` passes (lint, typecheck, tests, content checks).

### DIFF
--- a/scripts/security-scan-lib.test.ts
+++ b/scripts/security-scan-lib.test.ts
@@ -460,6 +460,17 @@ describe("runSecurityScan", () => {
 
     expect(result).toEqual(scanResult);
     expect(mockClient.chat.send).toHaveBeenCalledOnce();
+    // Lock in the OpenRouter SDK request shape (chatRequest/appTitle), which
+    // the SDK validates at call time and silently breaks the scan if wrong.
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2081, opened 2026-06-29
refactor(generate): enforce shared-file write order via a declarative step graph

saitota:
## Description

The write order of features sharing one config file lived only in two prose comments in `generate()`. Reordering the `await generate*Core(...)` lines, or switching to `Promise.all`, would silently drop another feature's keys. PR #1858 pinned the two known pairs with tests; this makes the constraint explicit in data.

## Changes

- Each step declares `writesSharedFile` / `dependsOn`; `resolveExecutionOrder` topologically sorts them and throws on an unordered shared-file writer pair, an unknown dependency, or a cycle — so the array can be reordered freely and a silent data-loss trap becomes a startup error.
- Declares every multi-writer tool, not just the two example pairs: claudecode, kilo, opencode, zed, qwencode, augmentcode, hermesagent, kiro, amp, codexcli, grokcli, vibe.

## Test Plan

- Unit tests for `resolveExecutionOrder`; existing #1858 regression tests preserved.
- `pnpm cicheck` green (6553 tests).

Refs #1858


dyoshikawa:
@saitota Thank you!

PR #2087, opened 2026-06-30
ci(release): auto-open Homebrew/core bump PR after npm publish

### LINKED ISSUE #2074: Homebrew formula is behind latest GitHub release
## Description

The Homebrew formula for `rulesync` appears to be behind the latest GitHub release.

- Homebrew formula: https://formulae.brew.sh/formula/rulesync#default
- Latest GitHub release: https://github.com/dyoshikawa/rulesync/releases/tag/v9.0.1

At the time of checking, the latest GitHub release is `v9.0.1`, while the Homebrew formula is still reporting an older stable version. I originally saw Homebrew reported as `8.32.0`; on the formula page currently available to me it shows `stable 8.31.0`, but in either case it is behind `v9.0.1`.

## Expected behavior

`brew install rulesync` should install the latest stable release, currently `v9.0.1`, or the Homebrew formula should be updated soon after a new stable release is published.

## Actual behavior

Homebrew installs or advertises an older `8.x` version while GitHub Releases has `v9.0.1` marked as the latest release.

## Why this matters

Users installing via Homebrew may not get the latest fixes and behavior available in the GitHub release.

## Suggested fix

I think the durable fix should be release automation rather than a one-off formula bump.

Current state I verified:

- GitHub latest release is `v9.0.1`.
- npm latest is `rulesync@9.0.1`.
- Homebrew/core currently advertises an older `8.x` formula.
- BrewTestBot has already opened a Homebrew/core PR for `rulesync 9.0.0`, but that still misses the current `9.0.1` release.

Common Homebrew automation practices I found:

- For packages already in Homebrew/core, the normal community path is to rely on Homebrew's `livecheck` and BrewTestBot autobump system, but upstream does not control when the PR is opened, whether it races with a newer release, or when it is merged.
- For upstream-controlled immediacy, projects commonly maintain their own tap and update it from release automation. GoReleaser-style workflows use this model, publishing a formula or cask into a repository the project controls.
- For existing Homebrew/core formula updates, Homebrew's supported command is `brew bump-formula-pr`, ideally run with the exact version, source URL, and sha256 after the release artifact is available.

Given that contributors may not have access to create a repository like `dyoshikawa/homebrew-tap`, I would suggest this implementation path:

1. Keep Homebrew/core as the canonical public install path: `brew install rulesync`.
2. Add a release workflow step after npm publish and GitHub release publication that:
   - derives the version from the release tag,
   - waits until `npm view rulesync@<version>` exposes the tarball,
   - computes the npm tarball sha256,
   - checks whether Homebrew/core already has that exact version or an open PR for it,
   - opens a Homebrew/core PR with `brew bump-formula-pr rulesync --version <version> --url <npm tarball> --sha256 <sha256>` when needed.
3. Use a maintainer-owned GitHub token secret for the Homebrew/core PR step, because the default repository `GITHUB_TOKEN` is not enough for cross-repository PR creation.
4. Optionally, if the maintainer wants instant Homebrew availability independent of Homebrew/core merge latency, create an official tap later and wire the same release metadata into it. That should be a maintainer decision, not required for this issue.

For `v9.0.1`, the computed npm tarball target would be:

- URL: `https://registry.npmjs.org/rulesync/-/rulesync-9.0.1.tgz`
- sha256: `863ccfbf7f57ae04936afdc4250e5574d0f7fb2e2e6eb7fc3ac27333a2610ca3`

References:

- Homebrew Autobump: https://docs.brew.sh/Autobump
- `brew bump-formula-pr`: https://docs.brew.sh/Manpage#bump-formula-pr-options-formula
- Homebrew Action for bumping packages: https://github.com/Homebrew/actions/tree/main/bump-packages
- Formula Cookbook, npm formula support: https://docs.brew.sh/Formula-Cookbook#std_npm_args

dyoshikawa:
## Summary

Resolves the Homebrew formula lagging behind GitHub/npm releases (#2074) by automating the formula bump from the release pipeline.

Adds a `homebrew` job to `.github/workflows/publish.yml` that runs after the npm publish + GitHub release are finalized and:
1. Derives the version from the `release/v*` tag (validated).
2. Waits for the npm tarball (`https://registry.npmjs.org/rulesync/-/rulesync-<version>.tgz`) to propagate.
3. Computes its sha256.
4. Runs `brew bump-formula-pr` to open a Homebrew/core update PR (idempotent — refuses duplicates).

## ⚠️ Requires maintainer action — DO NOT auto-merge
- **Secret required:** the job is gated on a maintainer-provisioned `HOMEBREW_BUMP_TOKEN` secret — a PAT with `public_repo` scope (the default `GITHUB_TOKEN` cannot open cross-repository PRs against `Homebrew/homebrew-core`). When the secret is absent the job is a no-op warning, so merging is safe but inert until the secret is added.
- **Design decision:** this keeps Homebrew/core as the canonical install path (per the issue). If you would prefer an official `dyoshikawa/homebrew-tap` for instant availability instead, that is a separate maintainer choice.

## Security
Untrusted `workflow_run` data (`head_branch`) is passed via env vars and the tag is regex-validated before use, per `.claude/rules/github-actions-security.md`. `actionlint` passes locally.

## Notes
This is a GitHub Actions workflow change, so per this repo's autonomous-batch safety policy it is opened for **manual maintainer review and merge**, not auto-merged.

Closes #2074


.github/workflows/publish.yml
```diff
@@ -43,3 +43,60 @@ jobs:
...
+
+          SHA256="$(curl -sfL "$URL" | sha256sum | cut -d' ' -f1)"
+          if [ -z "${SHA256}" ]; then
+            echo "::error::Failed to compute sha256 for $URL"
+            exit 1
+          fi
+
+          # Skip if Homebrew/core already has this exact version or an open bump PR
+          # for it; brew bump-formula-pr is idempotent and will refuse duplicates.
+          brew bump-formula-pr rulesync \
```
dyoshikawa, line 108: GitHub-hosted `ubuntu-latest` runners have Homebrew preinstalled but it's not added to `PATH` by default (see the runner-images README: you need `eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"` or the `Homebrew/actions/setup-homebrew` action first). As written, this step will very likely fail with `brew: command not found`, meaning the whole point of this job wouldn't actually work on a real release run.

.github/workflows/publish.yml
```diff
@@ -43,3 +43,60 @@ jobs:
             exit 1
           fi
           gh release edit "$TAG" --draft=false --latest
+
+  homebrew:
+    name: Bump Homebrew/core formula
+    needs: publish
+    runs-on: ubuntu-latest
+    # Mirror the publish job's gating so the bump only runs for real releases.
+    if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' && startsWith(github.event.workflow_run.head_branch, 'release/v')
```
dyoshikawa: Heads up: once a job defines its own `if:` that doesn't reference `success()` or `needs.publish.result`, that condition replaces the implicit "only run if `publish` succeeded" gate that `needs: publish` would otherwise provide — it doesn't add to it. This `if:` only re-checks the upstream `workflow_run` trigger fields, not whether the sibling `publish` job actually succeeded. So if `publish` fails after `pnpm publish` already succeeded (e.g. the `gh release edit` step errors out), this job would still run and could open a public Homebrew/core PR pointing at a release GitHub still shows as an unpublished draft. Adding `success() &&` (or `needs.publish.result == 'success' &&`) to this condition would close that gap.

.github/workflows/publish.yml
```diff
@@ -43,3 +43,60 @@ jobs:
             exit 1
           fi
           gh release edit "$TAG" --draft=false --latest
+
+  homebrew:
```
dyoshikawa, line 47: Minor: this job doesn't check out code or touch `GITHUB_TOKEN`/OIDC (it uses its own `HOMEBREW_BUMP_TOKEN`), but it still inherits the workflow-level `permissions: id-token: write / contents: write`. Might be worth adding `permissions: {}` here for least privilege, similar to the per-job overrides already used in `draft-release.yml`.

.github/workflows/publish.yml
```diff
@@ -43,3 +43,60 @@ jobs:
...
+          # cross-repository PRs, so the step is skipped when this is absent.
+          HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.HOMEBREW_BUMP_TOKEN }}
+          HEAD_BRANCH: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          if [ -z "${HOMEBREW_GITHUB_API_TOKEN}" ]; then
+            echo "::warning::HOMEBREW_BUMP_TOKEN secret is not set; skipping Homebrew formula bump."
+            exit 0
+          fi
+
+          TAG="${HEAD_BRANCH#release/}"
```
dyoshikawa, line 74: Small DRY note: this `TAG="${HEAD_BRANCH#release/}"` + regex-validation block is now duplicated a third time (it already exists in the `publish` job above and in `publish-assets.yml`). Not blocking, but worth centralizing at some point so the three copies don't drift.

dyoshikawa (commented):
This is a solid idea for automating the Homebrew bump, and the script-injection handling (passing `HEAD_BRANCH` through `env:` and regex-validating it) is done right. Two things below look like they'd actually break or misbehave in practice, though, so I'd want those addressed before merging. Also worth noting: since this job only runs on `workflow_run` after a real release merge, none of these issues would show up in this PR's own CI checks — they'd only surface (or silently misbehave) during an actual release.

dyoshikawa:
Pushed 26336e1f to address the review findings:

- **Success gate**: added `success() &&` to the job's `if:` so it no longer replaces the implicit `needs: publish` gate — the bump won't run when `publish` failed after the tarball was already published.
- **brew on PATH**: added `eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"` before the `brew bump-formula-pr` call, since GitHub-hosted runners ship Homebrew preinstalled but not on PATH.
- **Least privilege**: scoped the job to `permissions: {}` (it only opens a cross-repo PR with its own `HOMEBREW_BUMP_TOKEN`).

Also allowlisted `linuxbrew`/`shellenv` in cspell. Ran `pnpm cicheck` locally (all green). Left the TAG-derivation DRY cleanup and the `HOMEBREW_BUMP_TOKEN` fine-grained-PAT suggestion out of scope for a follow-up.

rudironsoni:
Thanks, [dyoshikawa](https://github.com/dyoshikawa)

rudironsoni:
@dyoshikawa 

HOMEBREW_BUMP_TOKEN secret is not set; skipping Homebrew formula bump.


https://github.com/dyoshikawa/rulesync/actions/runs/28786400981

https://github.com/dyoshikawa/rulesync/actions/runs/28786400981/job/85354216126

PR #2104, opened 2026-07-01
fix(hooks): quote $CLAUDE_PROJECT_DIR-prefixed hook commands

### LINKED ISSUE #2097: Unquoted $CLAUDE_PROJECT_DIR in generated Claude hook commands breaks on project paths with spaces
## Summary

When generating `.claude/settings.json` for the Claude Code target, rulesync
rewrites a project-relative hook command (`./<path>`) into
`$CLAUDE_PROJECT_DIR/<path>` but does **not** quote it. Claude Code executes hook
commands via `/bin/sh -c "<command>"`, so if the project directory contains a
space, the unquoted variable expansion is word-split and the hook fails.

## Environment

- rulesync 8.23.0
- Target: `claude` (`.claude/settings.json` hooks)
- Project path containing a space, e.g. `/Volumes/Crucial X9/.../my-project`
- OS: macOS (any POSIX `/bin/sh`)

## Reproduction

1. Source `.rulesync/hooks.json`:
   ```json
   {
     "hooks": {
       "sessionStart": [
         { "type": "command", "command": "./.rulesync/hooks/session-start.sh" }
       ]
     }
   }
   ```
2. Run `rulesync generate` (Claude target) from a project whose absolute path
   contains a space.
3. Generated `.claude/settings.json`:
   ```json
   "command": "$CLAUDE_PROJECT_DIR/.rulesync/hooks/session-start.sh"
   ```
4. Claude Code fires the SessionStart hook and it errors:
   ```
   SessionStart:resume hook error
   Failed with non-blocking status code:
   /bin/sh: /Volumes/Crucial: No such file or directory
   ```

## Root cause

The `./` → `$CLAUDE_PROJECT_DIR/` rewrite emits an **unquoted** shell token.
`/bin/sh -c "$CLAUDE_PROJECT_DIR/.rulesync/hooks/session-start.sh"` with a spaced
expansion splits into `/Volumes/Crucial` (command) + `X9/.../session-start.sh`
(argument) → "No such file or directory".

## Expected

The generated command should quote the substituted path:

```json
"command": "\"$CLAUDE_PROJECT_DIR/.rulesync/hooks/session-start.sh\""
```

Note: rulesync's **plugin** output already quotes the analogous path
(`"${CLAUDE_PLUGIN_ROOT}/hooks/session-start.sh"`), so the Claude settings
generator should match that behaviour for consistency.

## Suggested fix

In the code that rewrites a relative hook `command` to
`$CLAUDE_PROJECT_DIR/...` for the Claude target, wrap the substituted path token
in double quotes (mirroring the `CLAUDE_PLUGIN_ROOT` generator). Quote only the
`$CLAUDE_PROJECT_DIR/<token>` path so any trailing arguments are preserved.

## Workaround (downstream, for reference)

Until a release, we post-process the generated file with a small script that
re-quotes each hook command:

```
// quote unquoted $CLAUDE_PROJECT_DIR/<token> in every
// .claude/settings.json hooks.*[].hooks[].command
command.replace(/(?<!["'])\$CLAUDE_PROJECT_DIR\/\S+/g, (m) => `"${m}"`)
```

This produces the expected quoted command and the SessionStart hook then runs
cleanly (exit 0) even with a spaced project path.

dyoshikawa:
## Summary
- Claude Code (and Factory Droid, which shares the same converter) execute hook commands via `/bin/sh -c "<command>"`. The `./` → `$CLAUDE_PROJECT_DIR/`-style rewrite emitted an unquoted substitution, so a project path containing a space gets word-split and the hook fails with `No such file or directory`.
- Wrap the generated `$CLAUDE_PROJECT_DIR/...` (and `$FACTORY_PROJECT_DIR/...`) substitution in double quotes at export time.
- Teach the import-side `stripCommandPrefix` to recognize both the new quoted form and the legacy unquoted form, so round-tripping existing generated files keeps working.
- Manually-authored override commands that already start with `$` are left untouched, as before.

## Test plan
- [x] `pnpm cicheck` (full lint/typecheck/test/docs/gitignore/schema/spelling/secrets suite) passes
- [x] Added a unit test asserting the exact quoted output for Claude Code
- [x] Added a unit test asserting the quoted form round-trips back to a `./`-relative command on import
- [x] Existing Claude Code / Factory Droid hook tests and the hooks E2E spec still pass

Closes #2097

### DIFF
--- a/src/features/hooks/tool-hooks-converter.ts
+++ b/src/features/hooks/tool-hooks-converter.ts
@@ -99,7 +99,7 @@ function applyCommandPrefix({
     (!converterConfig.prefixDotRelativeCommandsOnly || trimmedCommand.startsWith("."));
 
   return shouldPrefix && typeof trimmedCommand === "string"
-    ? `${converterConfig.projectDirVar}/${trimmedCommand.replace(/^\.\//, "")}`
+    ? `"${converterConfig.projectDirVar}/${trimmedCommand.replace(/^\.\//, "")}"`
     : def.command;
 }
 
@@ -206,15 +206,16 @@ function stripCommandPrefix({
   converterConfig: ToolHooksConverterConfig;
 }): string | undefined {
   const cmd = typeof command === "string" ? command : undefined;
-  if (
-    converterConfig.projectDirVar !== "" &&
-    typeof cmd === "string" &&
-    cmd.includes(`${converterConfig.projectDirVar}/`)
-  ) {
-    return cmd.replace(
-      new RegExp(`^${converterConfig.projectDirVar.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}\\/?`),
-      "./",
-    );
+  if (converterConfig.projectDirVar === "" || typeof cmd !== "string") {
+    return cmd;
+  }
+  const escapedVar = converterConfig.projectDirVar.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+  const quotedMatch = cmd.match(new RegExp(`^"${escapedVar}\\/(.*)"$`));
+  if (quotedMatch) {
+    return `./${quotedMatch[1]}`;
+  }
+  if (cmd.includes(`${converterConfig.projectDirVar}/`)) {
+    return cmd.replace(new RegExp(`^${escapedVar}\\/?`), "./");
   }
   return cmd;
 }
--- a/src/features/hooks/tool-hooks-converter.ts
+++ b/src/features/hooks/tool-hooks-converter.ts
@@ -98,8 +98,11 @@ function applyCommandPrefix({
     !trimmedCommand.startsWith("$") &&
     (!converterConfig.prefixDotRelativeCommandsOnly || trimmedCommand.startsWith("."));
 
+  // Only the variable itself is quoted (not the whole command) so a project path
+  // containing a space can't be word-split by the shell, while any trailing
+  // arguments after the script path stay outside the quotes and still split normally.
   return shouldPrefix && typeof trimmedCommand === "string"
-    ? `"${converterConfig.projectDirVar}/${trimmedCommand.replace(/^\.\//, "")}"`
+    ? `"${converterConfig.projectDirVar}"/${trimmedCommand.replace(/^\.\//, "")}`
     : def.command;
 }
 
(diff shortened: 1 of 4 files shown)


src/features/hooks/tool-hooks-converter.ts
```diff
@@ -98,8 +98,11 @@ function applyCommandPrefix({
     !trimmedCommand.startsWith("$") &&
     (!converterConfig.prefixDotRelativeCommandsOnly || trimmedCommand.startsWith("."));
 
+  // Only the variable itself is quoted (not the whole command) so a project path
+  // containing a space can't be word-split by the shell, while any trailing
```
dyoshikawa, line 102: This quotes the *entire* remainder of the command, not just the `projectDirVar`/path token. For a canonical hook command like `.rulesync/hooks/lint.sh --fix`, this now emits the single string `"$CLAUDE_PROJECT_DIR/.rulesync/hooks/lint.sh --fix"` (quote characters included). When the target tool runs that via `/bin/sh -c "<command>"`, the double quotes suppress word-splitting across the whole thing, so after `$CLAUDE_PROJECT_DIR` expansion the shell treats path *and* arguments as one un-splittable word — `argv[0]` becomes a literal path containing a space, which doesn't exist, and the hook fails with the same class of error this PR is fixing.

I checked this against a real shell: pre-PR (unquoted), `.rulesync/hooks/lint.sh --fix` splits correctly into `lint.sh` + `--fix` even with no space in the path. Post-PR (quoted), the same command fails with `not found`. So this trades a narrow bug (spaces in the project path, no-arg commands only) for a broader one (any dot-relative command with trailing arguments, regardless of path spaces) — and issue #2097's own suggested fix called this out explicitly: quote only the `$CLAUDE_PROJECT_DIR/<path>` token so trailing arguments are preserved outside the quotes.

Worth noting Factory Droid (`factorydroid-hooks.ts`) doesn't set `prefixDotRelativeCommandsOnly`, so there `shouldPrefix` is true for *any* command not starting with `$` — meaning this same failure mode hits a much wider set of commands there, not just dot-relative ones.

None of the new tests cover a command with trailing arguments (both only test bare, no-arg scripts), so this wouldn't be caught by `pnpm cicheck` as currently written. Adding a case like `.rulesync/hooks/lint.sh --fix` to the export test, and asserting the quotes wrap only the path (e.g. `"$CLAUDE_PROJECT_DIR/.rulesync/hooks/lint.sh" --fix`), would both prove the fix and guard against this regressing again.

dyoshikawa (commented):
The fix correctly resolves the exact reproduction in #2097 (a bare, no-argument dot-relative script with a space in the project path), and the `stripCommandPrefix` refactor (early-return guard, shared `escapedVar`) is a nice cleanup. But I think there's a real regression here that's worth fixing before merge: quoting the whole command string instead of just the path segment breaks any hook that has trailing arguments, independent of whether the project path has spaces at all. I verified this empirically with `/bin/sh -c` — details in the inline comment.

dyoshikawa:
@dyoshikawa Thank you!

PR #2105, opened 2026-07-01
refactor(generate): make the step graph directly testable and harden it

### LINKED ISSUE #2082: Harden the declarative generate() step graph: verify edge direction & writesSharedFile completeness (follow-up to #2081)
## Background

PR #2081 (`refactor(generate): enforce shared-file write order via a declarative step graph`, by external contributor @saitota) refactors the shared-file write ordering in `generate()` (`src/lib/generate.ts`) from a fixed sequence of awaited calls into a declarative step graph (`dependsOn` + `writesSharedFile`), topologically sorted by `resolveExecutionOrder()`. `assertSharedFilesOrdered()` checks at startup that any two steps writing the same shared file have *some* declared order between them.

The PR was reviewed and merged with no merge-blocker findings (CI green, no high/critical issues, regression-free against the original ordering). This issue captures the non-blocking follow-ups surfaced during review so they are not lost.

Reference PR: https://github.com/dyoshikawa/rulesync/pull/2081
Related files: `src/lib/generate.ts`, `src/lib/generate.test.ts`

### 1. [mid] The assertion proves "an order exists", not "the order is correct" or "declarations are complete"

`assertSharedFilesOrdered` only checks `dependsOnReachable(a, b) || dependsOnReachable(b, a)` — i.e. that *some* ordering exists between two writers of the same shared file. It does not catch:

- **Wrong direction**: a future `dependsOn` declared the wrong way around (e.g. `ignore` depending on `permissions`) still satisfies the assertion. The build stays green while keys in `.claude/settings.json` etc. are silently dropped.
- **Missing declaration**: the safety net assumes *every* shared-file read-modify-write is listed in `writesSharedFile`. There is no mechanism verifying that this enumeration is complete; a feature that RMWs a shared config file but forgets to list it reopens the silent-data-loss trap.

The original code carried prose comments explaining *why* each direction was required (e.g. "permissions must run after ignore because both write `.claude/settings.json`"). That rationale is now gone, and the PR's "the array can be reordered freely and stay safe" claim only holds while the graph is both complete and correctly directed.

Possible mitigations to evaluate: cross-check each feature's `getSettablePaths` against `writesSharedFile` to verify enumeration completeness; add tests that pin the required directions; re-add a short rationale comment for the non-obvious edges.

### 2. [low-mid] `mcp` execution position shifted implicitly (2nd → 5th)

The topological sort changes the runtime order from `ignore, mcp, commands, subagents, skills, hooks, permissions, rules` to `ignore, commands, subagents, skills, mcp, hooks, permissions, rules` (the dependency-free `commands`/`subagents`/`skills` precede `mcp` in the ready queue). This is currently safe because those three steps do not write shared config files, but there is no regression test pinning this order. If OpenCode (or another tool) later moves to inline writes into a shared file, this could regress silently. Consider adding a test that fixes the expected execution order.

### 3. [low] No direct test that the real `steps` graph in `generate()` is well-formed

`resolveExecutionOrder` has thorough unit tests, but the actual `steps` array in `generate()` is only validated indirectly (whether `generate()`/e2e throws at runtime). A dedicated test passing the real `steps` to `resolveExecutionOrder` and asserting it does not throw would make declaration mistakes easier to debug.

### 4. [low] Effectively unreachable defensive code

`rules`'s `skills: skillsResult?.skills` optional chaining, the `if (!skillsResult) throw`, and the `get()` helper's missing-result throw are effectively unreachable given the `dependsOn` graph guarantees `skillsResult` is set before `rules` runs. The module-scoped `let skillsResult` hand-off also slightly undermines the "side-effect-free step" cleanliness of the declarative graph.

### 5. [nit] `as never` cast in the test helper

`generate.test.ts`'s `executionStep` helper casts its result with `as never`. A partial of `GenerationStep` could give it proper typing.

## Solution / Next Steps

- Prioritize item 1: decide on a mechanism that guarantees both edge direction correctness and `writesSharedFile` completeness (completeness cross-check and/or direction-pinning tests), and re-add concise rationale comments for non-obvious edges.
- Add the regression/well-formedness tests from items 2 and 3.
- Optionally simplify the unreachable defensive code (item 4) and tighten the test helper typing (item 5) if touching this area.

dyoshikawa:
## Summary
- Follow-up to #2081 / PR #2081, which refactored `generate()`'s shared-file write ordering into a declarative step graph (`dependsOn` + `writesSharedFile`), topologically sorted and validated by `resolveExecutionOrder`/`assertSharedFilesOrdered`.
- Extracts the graph's static shape (`id`, `writesSharedFile`, `dependsOn`) into an exported `GENERATION_STEP_GRAPH`, separate from the `run` closures that need a live `config`/`logger`. This lets tests exercise `resolveExecutionOrder` against the *real* graph instead of a hand-copied stand-in.
- Restores short rationale comments on each non-obvious `dependsOn` edge, explaining which shared file forces the ordering (the original PR's prose comments were dropped when the graph became declarative).
- Adds a test that walks every `dependsOn` edge and asserts it corresponds to either an actual overlapping `writesSharedFile` token or the one documented value-dependency (`rules` reading the `skills` step's output) — this catches a stale or unjustified edge, and combined with the existing `assertSharedFilesOrdered` runtime check (which catches a *missing* edge for a real overlap), the two together keep the declared edges honest in both directions.
- Adds a test pinning the full execution order of the real graph, so the `mcp` step's position (2nd → 5th after the refactor) — or any future reordering that changes runtime behavior — is caught rather than silently regressing.

## Not addressed (left for a follow-up if needed)
- Fully automated verification that `writesSharedFile` enumerates *every* shared file across all tool targets (proving there's no un-declared overlap) would require cross-referencing each feature's `getSettablePaths()` per tool target, which is a materially larger undertaking than this hardening pass. The new edge-justification test narrows the risk (no stale/incorrect edges) but doesn't fully close this gap.
- The two "effectively unreachable" defensive checks flagged in the original issue (`if (!skillsResult) throw`, `get()`'s missing-result throw) were left as-is — they're a cheap safety net against future graph regressions and only rated `low`/`nit`.

## Test plan
- [x] `pnpm cicheck` passes
- [x] New tests: graph well-formedness, edge-justification, and full execution-order pinning, all run against the real `GENERATION_STEP_GRAPH`
- [x] Existing `generate.test.ts` and `resolveExecutionOrder` tests still pass unmodified

Closes #2082

### DIFF
--- a/src/lib/generate.test.ts
+++ b/src/lib/generate.test.ts
@@ -12,7 +12,12 @@ import { SkillsProcessor } from "../features/skills/skills-processor.js";
 import { SubagentsProcessor } from "../features/subagents/subagents-processor.js";
 import { createMockLogger } from "../test-utils/mock-logger.js";
 import { fileExists, readFileContentOrNull } from "../utils/file.js";
-import { checkRulesyncDirExists, generate, resolveExecutionOrder } from "./generate.js";
+import {
+  checkRulesyncDirExists,
+  generate,
+  GENERATION_STEP_GRAPH,
+  resolveExecutionOrder,
+} from "./generate.js";
 
 const logger = createMockLogger();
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2108, opened 2026-07-02
refactor(rules): move rovodev quirks from hub to spoke hooks

saitota:
## Description

`mirrorsRootToAgentsMd` was cited as a single-tool (rovodev) quirk whose interpretation lived in the hub (`rules-processor.ts`) rather than the spoke. The hub hardcoded `instanceof RovodevRule` / `=== RovodevRule` checks and rovodev path constants across the generate and delete paths, leaking tool-specific knowledge into code shared by all 34 tools.

## Changes

- Move the root-mirror quirk to optional spoke hooks on `RovodevRule` (`getRootMirrorFiles`, `getRootMirrorDeletionGlobs`), following the existing `getAuxiliaryFiles` / `fromRulesyncSubagents` hook pattern used by other features.
- Move the `separate-local-file` deletion glob to a `getLocalRootDeletionGlob` hook, so the hub no longer branches on `=== RovodevRule` for it.
- The `mirrorsRootToAgentsMd` meta flag stays as a pure dispatch declaration; rovodev-specific paths and `instanceof` checks now live entirely in `RovodevRule`.

## Test Plan

- [x] `pnpm run typecheck`, `pnpm run fmt:check`, `pnpm run oxlint` — clean.
- [x] `rules-processor.test.ts` / `rovodev-rule.test.ts` — 112 pass (root-mirror generate/delete and local-root deletion covered).

### DIFF
--- a/src/features/rules/rovodev-rule.ts
+++ b/src/features/rules/rovodev-rule.ts
@@ -376,4 +376,15 @@ export class RovodevRule extends ToolRule {
       mirrorGlob: join(outputRoot, ROVODEV_RULE_FILE_NAME),
     };
   }
+
+  /** Glob for the `separate-local-file` deletion; rovodev writes it at project root, not under `.rovodev/`. */
+  static getLocalRootDeletionGlob({
+    outputRoot,
+    fileName,
+  }: {
+    outputRoot: string;
+    fileName: string;
+  }): string {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@saitota Thank you!

PR #2109, opened 2026-07-02
test(generate): derive shared-file writers from the registry and enforce declarations

saitota:
## Background

Which config file each feature read-modify-writes is known in two places. The tool classes know their real write targets through `getSettablePaths()` and their implementations; `GENERATION_STEP_GRAPH` re-declares the same knowledge as opaque tokens (`"reasonix-config"`, ...). Nothing — no type, no test — connects the two: `generate.test.ts` only checks the graph's internal consistency (are `dependsOn` edges justified for duplicated tokens), never whether the token set matches reality.

So `assertSharedFilesOrdered` can only order writers that are declared. An undeclared second writer passes silently and looks safe only while the execution order happens to be correct. And a tool adder has no reason to open `generate.ts`: the code compiles, tests pass, and output is correct under the current order — the drift surfaces only in review (e.g. ab61e383).

## Changes

- Derive the multi-writer set from `PROCESSOR_REGISTRY` × `getSettablePaths` (both scopes) plus a `getExtraSharedWritePaths` contract for writes not visible as settable paths, and assert it matches the graph — a missing declaration now fails CI, not review.
- Replace opaque tokens with real `dir/file` keys so declarations and derivation share one namespace.
- Fixes surfaced by the derivation: register `.takt/config.yaml`; drop permissions from `kilo.json` (it writes `kilo.jsonc`); cover the global-only `.config/devin/config.json` writer.

No change to generated output.

## Test Plan

- `pnpm cicheck` green (6740 tests).

Refs #2081

### DIFF
--- a/src/features/hooks/codexcli-hooks.ts
+++ b/src/features/hooks/codexcli-hooks.ts
@@ -7,6 +7,7 @@ import {
   CODEXCLI_HOOKS_FILE_NAME,
   CODEXCLI_MCP_FILE_NAME,
 } from "../../constants/codexcli-paths.js";
+import type { SharedWritePath } from "../../lib/shared-file-derive.js";
 import type { AiFileParams, ValidationResult } from "../../types/ai-file.js";
 import {
   CODEXCLI_HOOK_EVENTS,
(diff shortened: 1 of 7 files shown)


dyoshikawa:
@saitota Thank you!

PR #2115, opened 2026-07-03
refactor(claudecode): centralize .claude/settings.json permissions ownership in a gateway

saitota:
## Background

Who owns which entries inside `.claude/settings.json` `permissions` lived in two places: `ignore` writes `Read(...)` into `permissions.deny`, `permissions` writes the whole `allow`/`ask`/`deny`, and both re-implemented the entry format and the preserve/overwrite rule — `permissions` even guessed about the ignore feature by name in a warning. The two could drift while each feature's own tests kept passing. Ordering is already guarded by `GENERATION_STEP_GRAPH` (#2081, #2109); this closes the remaining merge/ownership half.

## Changes

- New `claudecode-settings-gateway.ts` owns the `permissions` entry format, merge, and cross-feature ownership rule (permissions' `Read` rules win over ignore-derived `Read` denies).
- `ignore` and `permissions` now state intent only; the warning drops its `possibly from ignore feature` guess.

No change to generated output. Claude Code only, matching the staged-adoption note in #2081.

## Test Plan

- New gateway unit tests; existing claudecode regression tests preserved.
- `pnpm cicheck` green (6749 tests).

Refs #2081, #2109

### DIFF
--- a/src/features/claudecode-settings-gateway.test.ts
+++ b/src/features/claudecode-settings-gateway.test.ts
@@ -1,0 +1,138 @@
+import { describe, expect, it } from "vitest";
+
+import { createMockLogger } from "../test-utils/mock-logger.js";
+import type { ClaudeSettingsJson } from "../types/claude-settings.js";
+import {
+  applyIgnoreReadDenies,
+  applyPermissions,
+  buildReadDenyEntry,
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@saitota Thank you!

PR #2116, opened 2026-07-03
test(e2e): assert the generate matrix covers every declared tool target

saitota:
## Background

Which tools a feature supports, and where each writes output, lives in two places — the implementation (`getToolTargets()`) and the hardcoded `it.each` arrays in `src/e2e/*.spec.ts` — with nothing connecting them. CLAUDE.md having to *tell a human* "preserve the Tool × Feature matrix tests" is the evidence it was never mechanized. So adding a tool (32 in H1 2026) means reconstructing from memory which specs need new rows; a miss leaves CI green while coverage silently drifts.

## Changes

- Add `assertGenerateMatrixCoversTargets` (e2e-helper): derives the tool set from `getToolTargets()` and asserts declared targets partition exactly into the matrix's tested tools plus an explicit, reasoned `untested` list — a declared-but-absent tool or a stray entry now fails CI.
- Wire it into all 8 features (project + global where supported). `-legacy` aliases are excluded, as the supported-tools tables already do.

Same "derive from implementation, fail on drift" idiom as the TOOL_DISPLAY check and gitignore derivation. Per-tool expected paths stay hand-written; only the enumeration is derived.

## Test Plan

- `pnpm cicheck` green (6740 tests).
- Verified drift detection: removing a tool from a matrix array fails with "declared by the processor but missing from the e2e matrix".

### DIFF
--- a/src/e2e/e2e-commands.spec.ts
+++ b/src/e2e/e2e-commands.spec.ts
@@ -3,60 +3,110 @@ import { join } from "node:path";
 import { describe, expect, it } from "vitest";
 
 import { RULESYNC_COMMANDS_RELATIVE_DIR_PATH } from "../constants/rulesync-paths.js";
+import { CommandsProcessor } from "../features/commands/commands-processor.js";
 import { readFileContent, writeFileContent } from "../utils/file.js";
 import {
+  assertGenerateMatrixCoversTargets,
   runGenerate,
   runImport,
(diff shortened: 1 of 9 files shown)


saitota:
Trimmed comments: now +1,318 / -884.

dyoshikawa:
@saitota Thank you!

PR #2118, opened 2026-07-03
fix(sources): handle root skill with metadata dirs

### LINKED ISSUE #2117: Declarative sources miss root SKILL.md when repo has metadata directories
## Summary

Declarative source install fails for a single-skill repository whose `SKILL.md` is at the repository root when the repository also contains an unrelated top-level directory, such as plugin metadata.

This currently affects `blader/humanizer`, which has a root `SKILL.md` plus `.claude-plugin/`. Both documented root-source forms resolve the repository, but install zero skills.

## Reproduction

Using current local `main` source (`package.json` reports `9.1.1`) from `~/src/rudironsoni/rulesync`:

```bash
tmpdir=$(mktemp -d /private/tmp/rulesync-humanizer-repro.XXXXXX)
printf '{"sources":[{"source":"blader/humanizer:.","skills":["humanizer"]}]}' > "$tmpdir/rulesync.jsonc"
cd "$tmpdir"
GH_TOKEN="$(gh auth token)" /path/to/rulesync/node_modules/.bin/tsx /path/to/rulesync/src/cli/index.ts install --update
```

Observed:

```text
Fetched 0 skill(s) from blader/humanizer:.: (none)
All skills up to date (1 source(s) checked).
```

The resulting lockfile contains no skills:

```json
{
  "lockfileVersion": 1,
  "sources": {
    "blader/humanizer:.": {
      "requestedRef": "main",
      "resolvedRef": "1b48564898e999219882660237fde01bf4843a0f",
      "skills": {}
    }
  }
}
```

The git transport form also fails:

```jsonc
{
  "sources": [
    {
      "source": "https://github.com/blader/humanizer",
      "transport": "git",
      "path": ".",
      "skills": ["humanizer"]
    }
  ]
}
```

Observed:

```text
Fetched 0 skill(s) from https://github.com/blader/humanizer: (none)
```

For comparison, pinning to the commit before `.claude-plugin/` was added works:

```jsonc
{
  "source": "blader/humanizer@9600f2b7241cb4eed6ad803abee5ea01d67fe8e4:.",
  "skills": ["humanizer"]
}
```

Observed:

```text
Fetched 1 skill(s) from blader/humanizer@9600f2b7241cb4eed6ad803abee5ea01d67fe8e4:.: humanizer
```

## Expected

For an explicit single-skill source like:

```jsonc
{ "source": "blader/humanizer:.", "skills": ["humanizer"] }
```

Rulesync should install the root `SKILL.md` as `.rulesync/skills/.curated/humanizer/SKILL.md`, even if the repository root also contains unrelated directories such as `.claude-plugin/`.

This matches the documented intent in `docs/guide/declarative-sources.md`: repository-root paths are useful for single-skill repositories whose `SKILL.md` lives at `<repo>/SKILL.md` rather than under a `skills/` container.

## Root Cause

The root-level fallback is gated too narrowly.

For git transport, `groupRemoteFilesBySkillRoot` only maps root-level files to the requested skill when no top-level directories were grouped:

```ts
if (grouped.size === 0 && !isWildcard && skillFilter.length === 1) {
  const [singleSkillName] = skillFilter;
  if (singleSkillName !== undefined && rootLevelFiles.length > 0) {
    grouped.set(singleSkillName, rootLevelFiles);
  }
}
```

When a repository has root `SKILL.md` plus `.claude-plugin/plugin.json`, `grouped.size` is nonzero because `.claude-plugin` is grouped as a top-level directory. The fallback never fires. The explicit filter `skills: ["humanizer"]` then filters out `.claude-plugin`, leaving no fetched skills.

For GitHub transport, `discoverGithubSkillDirs` similarly only tries `fetchRootLevelFallbackSkill` when `remoteSkillDirs.length === 0`:

```ts
if (remoteSkillDirs.length === 0 && !isWildcard && skillFilter.length === 1) {
  const fallback = await fetchRootLevelFallbackSkill(...)
}
```

`blader/humanizer:.` has `.claude-plugin/`, so `remoteSkillDirs.length !== 0`; fallback is skipped before the explicit `humanizer` filter is applied.

## Existing Coverage Gap

`src/lib/sources.test.ts` currently covers:

- root `SKILL.md` when there are no top-level directories
- classic `<skill>/SKILL.md` layout
- wildcard multi-skill top-level directories
- GitHub root fallback when the selected path itself contains only files

It does not cover root `SKILL.md` plus an unrelated top-level directory.

I ran the current test file locally:

```text
pnpm exec vitest run src/lib/sources.test.ts --silent=true
Test Files  1 passed (1)
Tests  36 passed (36)
```

## Fix Proposal

Preserve wildcard behavior, but for a single explicit skill filter, allow the root fallback when the requested skill is not present as a top-level directory and root files include `SKILL.md`.

Suggested shape:

1. In `groupRemoteFilesBySkillRoot`, change the fallback condition from `grouped.size === 0` to roughly:

```ts
const [singleSkillName] = skillFilter;
const hasRootSkill = rootLevelFiles.some((file) => file.relativePath === SKILL_FILE_NAME);

if (!isWildcard && skillFilter.length === 1 && singleSkillName && hasRootSkill && !grouped.has(singleSkillName)) {
  grouped.set(singleSkillName, rootLevelFiles);
}
```

This avoids treating arbitrary root files as a skill unless there is a root `SKILL.md`.

2. In `discoverGithubSkillDirs`, attempt `fetchRootLevelFallbackSkill` when a single requested skill is not present among discovered directories, not only when there are zero directories:

```ts
const [singleSkillName] = skillFilter;
const hasRequestedDir = singleSkillName !== undefined && remoteSkillDirs.some((d) => d.name === singleSkillName);

if (!isWildcard && skillFilter.length === 1 && !hasRequestedDir) {
  const fallback = await fetchRootLevelFallbackSkill(...);
  if (fallback.handled) return ...;
}
```

3. Add tests for both transports with remote files like:

```text
SKILL.md
README.md
.claude-plugin/plugin.json
```

and source filters `skills: ["humanizer"]`. Expected result: one fetched skill named `humanizer`, with root `SKILL.md` installed under `.rulesync/skills/.curated/humanizer/SKILL.md`.

rudironsoni:
## Summary

Fixes #2117.

- allow explicit single-skill declarative sources to install a root-level SKILL.md even when unrelated top-level metadata directories exist
- keep wildcard behavior and real matching skill directories preferred over root fallback
- add git and GitHub transport regression coverage for root SKILL.md plus .claude-plugin metadata
- ignore TokenSave local state through the generated gitignore source registry

## Root cause

Root fallback was gated on there being no discovered top-level skill directories. Repositories like blader/humanizer include root SKILL.md plus metadata directories such as .claude-plugin/, so discovery saw a directory, skipped fallback, then filtered out the metadata directory and fetched zero skills.

## Validation

- pnpm run check
- pnpm run test
- pnpm run check:supported-tools
- pnpm run check:sync-skill-docs
- pnpm run cspell
- pnpm run secretlint
- pnpm run dev gitignore
- git diff --check


dyoshikawa:
@rudironsoni Thank you!

PR #2124, opened 2026-07-06
test(e2e): enforce exact partition in assertGenerateMatrixCoversTargets and correct overstated coverage comments

### LINKED ISSUE #2120: Follow-ups from PR #2116 review: enforce exact partition in assertGenerateMatrixCoversTargets and correct overstated coverage comments
## Background

PR #2116 ("test(e2e): assert the generate matrix covers every declared tool target") is a high-quality test-infrastructure refactor that was merged. It extracts previously-inline `it.each` target arrays into named `const`s and adds a helper `assertGenerateMatrixCoversTargets` (`src/e2e/e2e-helper.ts`) that derives each processor's `getToolTargets()` (filtering `-legacy` aliases) and asserts the declared set partitions into the hand-written matrix plus an explicit `untested` list — so drift now fails CI.

During review, two **non-blocking (severity: low)** follow-ups surfaced. They do not block the merge but are captured here so they are not lost.

### Finding 1 (low) — partition is not fully enforced

`src/e2e/e2e-helper.ts:203,209` (`assertGenerateMatrixCoversTargets`).

The docstring says declared targets are "partitioned exactly" into `testedTargets` and `untested`, but the implementation only checks `stray` (declared-outside) and `uncovered` (not-covered). It does **not** detect:

- (a) a tool listed in **both** `testedTargets` and `untested` (a contradiction: "tested" and "intentionally untested" at once), or
- (b) duplicates within `testedTargets`.

**Failure scenario:** adding a tool to `untested` but forgetting to remove it from `testedTargets` leaves it in both, breaking the exclusive-partition invariant while CI stays green.

**Suggested fix:** add an intersection check, e.g.

```ts
const overlap = testedTargets.filter((t) => untested.includes(t)).toSorted();
expect(overlap, `Listed in both testedTargets and untested: ${overlap.join(", ")}`).toEqual([]);
```

### Finding 2 (low) — for standard-list specs, "enumeration" and "actual test bodies" are decoupled, so the in-spec comment overstates the guarantee

In table-driven specs (skills / rules / mcp / subagents / commands / ignore, and the hooks main array) the same array feeds both `it.each` and the coverage check, so enumeration == real test bodies.

But for **permissions** (each tool has its own `it`) and the **hooks standalone lists** (`hooksProjectStandaloneTargets` / `hooksGlobalStandaloneTargets` in `src/e2e/e2e-hooks.spec.ts`, and `permissionsGenerateTargets` in `src/e2e/e2e-permissions.spec.ts`), the assertion only verifies "list == processor-declared set" — **not** that a matching `it` body exists for each name.

**Failure scenario:** deleting a tool's dedicated `it` (e.g. permissions) but forgetting to remove its name from `permissionsGenerateTargets` keeps the coverage check green while real coverage is lost (false coverage). The in-spec comment ("a new project-scope permissions tool cannot be added without a matching e2e test") overstates this — only array membership is enforced.

**Suggested fix:** soften those comments to say the enumeration is forced to match the declared set but individual test-body existence is not guaranteed. (The PR description already correctly acknowledges "only the enumeration is derived".)

## Solution / Next Steps

- [ ] Finding 1: add an intersection/duplicate check to `assertGenerateMatrixCoversTargets` so a target appearing in both `testedTargets` and `untested` (or duplicated) fails CI.
- [ ] Finding 2: soften the in-spec comments for permissions and hooks standalone lists to accurately state that only the enumeration is enforced, not per-target test-body existence.

dyoshikawa:
## Background

Follow-ups from the review of PR #2116 ("test(e2e): assert the generate matrix covers every declared tool target"). Two non-blocking (severity: low) findings were captured in #2120 so they would not be lost.

### Finding 1 — enforce the exact partition

`assertGenerateMatrixCoversTargets` documented that declared targets "partition exactly" into `testedTargets` and `untested`, but only checked `stray` (declared-outside) and `uncovered` (not-covered). A target left in **both** `testedTargets` and `untested` — or duplicated within `testedTargets` — broke the exclusive-partition invariant while CI stayed green.

Added an explicit duplicate check and an overlap (intersection) check so those cases now fail CI, matching the documented contract.

### Finding 2 — correct overstated coverage comments

For the permissions generate list and the hooks project/global standalone lists, each tool has its own hand-written `it`, so the completeness check only verifies "enumeration == processor-declared set" — **not** that a matching `it` body exists for each name. Softened those in-spec comments so they no longer overstate the guarantee (deleting a tool's `it` while leaving its name in the list would keep the check green).

## Verification

`pnpm cicheck` passes (all 6752 tests green, format/lint/typecheck clean).

Closes #2120

### DIFF
--- a/src/e2e/e2e-helper.ts
+++ b/src/e2e/e2e-helper.ts
@@ -184,6 +184,11 @@ type ProcessorTargets = {
  * Mirrors the "derive from the implementation, fail on drift" idiom already used
  * by the TOOL_DISPLAY completeness check and the gitignore derivation.
  */
+/** Returns the sorted, de-duplicated set of values that appear more than once. */
+function uniqueDuplicates(targets: readonly string[]): string[] {
+  return Array.from(new Set(targets.filter((t, i) => targets.indexOf(t) !== i))).toSorted();
+}
+
 export function assertGenerateMatrixCoversTargets({
   processor,
   testedTargets,
@@ -200,10 +205,18 @@ export function assertGenerateMatrixCoversTargets({
     .filter((target) => !target.endsWith("-legacy"));
   const declaredSet = new Set<string>(declared);
 
-  const duplicates = testedTargets.filter((t, i) => testedTargets.indexOf(t) !== i).toSorted();
+  // A true "exact partition" requires each side to have no internal duplicates,
+  // so check both `testedTargets` and `untested` symmetrically.
+  const testedDuplicates = uniqueDuplicates(testedTargets);
   expect(
-    Array.from(new Set(duplicates)),
-    `These targets appear more than once in \`testedTargets\`: ${duplicates.join(", ")}`,
+    testedDuplicates,
+    `These targets appear more than once in \`testedTargets\`: ${testedDuplicates.join(", ")}`,
+  ).toEqual([]);
+
+  const untestedDuplicates = uniqueDuplicates(untested);
+  expect(
+    untestedDuplicates,
+    `These targets appear more than once in \`untested\`: ${untestedDuplicates.join(", ")}`,
   ).toEqual([]);
 
   const overlap = testedTargets.filter((t) => untested.includes(t)).toSorted();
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2132, opened 2026-07-06
Kiro permissions: map `grep`/`glob` and cover `aws` tool, shell auto-trust flags, and domain-scoped web permissions


dyoshikawa:
## Summary

rulesync's Kiro permissions translator (`src/features/permissions/kiro-permissions.ts`) writes `.kiro/agents/default.json`. It maps `bash`→`toolsSettings.shell.{allowed,denied}Commands`, `read`→`toolsSettings.read.{allowed,denied}Paths`, `edit`/`write`→`toolsSettings.write.{...}`, and `webfetch`/`websearch` **only** with pattern `*`→`allowedTools` (`web_fetch`/`web_search`). It drops `ask`, non-`*` web patterns, and unknown categories with warnings. This issue proposes (1) mapping the already-canonical `grep`/`glob` categories to Kiro's `toolsSettings`, and (2) a **kiro-scoped override namespace** (per #2127) for Kiro-specific keys. Related upstream context: #2034 (Kiro v3 EA follow-up).

## Motivation / Purpose

Two distinct gaps: first, rulesync already has canonical `grep`/`glob` categories that Kiro can represent, but they are silently dropped. Second, several Kiro-specific permission surfaces (the `aws` tool, shell auto-trust flags, domain-scoped web trust, MCP auto-approve/disabled lists) have no canonical representation and cannot be authored from rulesync.

## Background from Research

- Kiro's [configuration reference](https://kiro.dev/docs/cli/custom-agents/configuration-reference/) documents `toolsSettings` for tools including `glob` and `grep` (with `allowedPaths`/`deniedPaths`) — yet rulesync currently drops the canonical `grep`/`glob` categories instead of mapping them.
- Per the [built-in tools](https://kiro.dev/docs/cli/reference/built-in-tools/) reference, `web_search`/`web_fetch` are governed by domain `trusted`/`blocked` arrays, so domain-scoped web permissions cannot be expressed by rulesync's coarse `*`-only toggle (inferred limitation).
- The [configuration reference](https://kiro.dev/docs/cli/custom-agents/configuration-reference/) documents shell auto-trust flags `toolsSettings.shell.autoAllowReadonly` and `denyByDefault`.
- Kiro has a tool-specific `aws` tool with `allowedServices`/`deniedServices`, documented in the [built-in tools](https://kiro.dev/docs/cli/reference/built-in-tools/) reference — no canonical rulesync category maps to it.
[...]

PR #2170, opened 2026-07-07
feat(permissions): map Kiro grep/glob categories to toolsSettings

dyoshikawa:
## Problem

rulesync already carries canonical `grep`/`glob` permission categories, but the Kiro translator dropped them with an "unsupported category" warning — even though Kiro documents `toolsSettings.grep`/`toolsSettings.glob` with `allowedPaths`/`deniedPaths`.

## Fix

- Map `grep` → `toolsSettings.grep.{allowedPaths,deniedPaths}` and `glob` → `toolsSettings.glob.{allowedPaths,deniedPaths}`, mirroring the existing `read` mapping.
- Emit each table only when a rule is present, so existing configs do not gain empty `grep`/`glob` tables.
- Round-trip both on import.
- Refactored the path-category handling into shared helpers (`rulesFromArrays`, `pathTable`, `applyKiroWebPermission`) to keep the translator under the complexity limit.

## Scope

This is **item 1 of #2132** — the straightforward, override-independent fix. The kiro-scoped override namespace for the Kiro-specific keys (item 2: `aws` tool, shell auto-trust flags, domain-scoped web `trusted`/`blocked`, MCP `autoApprove`/`disabledTools`) is intentionally **left as a follow-up**: it carries several unresolved design questions (what counts as a permission vs. config in the shared `.kiro/agents/default.json`, how MCP `autoApprove`/`disabledTools` reconcile with the canonical `mcp__*` allow/ask/deny model, whether `aws`/`subagent`/`introspect` belong under a permissions override) that warrant a maintainer design call before implementing. The issue is left open for that part.

Source: https://kiro.dev/docs/cli/custom-agents/configuration-reference/

Refs #2132

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -909,6 +909,8 @@ For Kiro, this generates tool permission settings in `.kiro/agents/default.json`
 - `bash` maps to `toolsSettings.shell.allowedCommands` / `toolsSettings.shell.deniedCommands`
 - `read` maps to `toolsSettings.read.allowedPaths` / `toolsSettings.read.deniedPaths`
 - `edit` / `write` map to `toolsSettings.write.allowedPaths` / `toolsSettings.write.deniedPaths`
+- `grep` maps to `toolsSettings.grep.allowedPaths` / `toolsSettings.grep.deniedPaths`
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2127, opened 2026-07-06
Support OpenCode-specific permission categories (external_directory / webfetch / websearch / codesearch) via a tool-scoped `opencode` key in permissions.json


dyoshikawa:
## Summary

Allow OpenCode-specific permission categories (e.g. `external_directory`, `webfetch`, `websearch`, `codesearch`) to be configured in `.rulesync/permissions.json` under a nested tool-scoped key (`opencode`), so they are emitted only into OpenCode's config and never leak into other tools' permission files.

## Motivation / Purpose

The canonical `permission` block in `.rulesync/permissions.json` is **shared across every target tool** — each category (`bash`, `read`, `edit`, `webfetch`, ...) is translated into every tool's permission file. This works for categories that have a cross-tool meaning, but OpenCode exposes permission categories that other tools do not understand (`external_directory`, `codesearch`, and OpenCode's own `webfetch`/`websearch` handling). Putting them in the shared `permission` block would push meaningless/incorrect entries into Claude Code, Codex, etc.

Rulesync already solves exactly this problem for other features via **tool-scoped override keys**:

- Hooks: `opencode: { hooks: {...} }`, `claudecode: { hooks: {...} }`, ... in `HooksConfigSchema` (`src/types/hooks.ts:569-589`). Entries under a tool key are emitted only for that tool.
- Rules frontmatter: per-tool nested objects (`claudecode: {...}`, `cursor: {...}`, ...) in `RulesyncRuleFrontmatterSchema` (`src/features/rules/rulesync-rule.ts`).

Permissions should follow the same established pattern so OpenCode-only permission settings can coexist with the shared block without leaking.

## Details

Proposed shape for `.rulesync/permissions.json` (referencing the rules-frontmatter / hooks tool-override convention):

```jsonc
{
  "permission": {
    "bash": { "git *": "allow", "*": "ask" }
  },
  // OpenCode-only override: emitted only into opencode.json's `permission`
  "opencode": {
    "permission": {
      "external_directory": "deny",
      "webfetch": "allow",
      "websearch": "allow",
      "codesearch": "allow"
    }
  }
}
```

### Changes required

[...]

PR #2172, opened 2026-07-07
feat(permissions): add a vibe override for per-tool sensitive_patterns

### LINKED ISSUE #2146: Vibe permissions: support per-tool `sensitive_patterns` (ASK-override on an allowed tool)
## Summary

rulesync's Vibe permissions translator cannot express Vibe's per-tool `sensitive_patterns` — a list of patterns that escalate to ASK even when the tool's base permission is ALWAYS (allow). The canonical model can only set a pattern to `ask` outright, which weakens the base allow, so an ALWAYS+`sensitive_patterns` combination round-trips lossily. This proposes preserving `sensitive_patterns` via a Vibe-scoped override (or a per-pattern semantic).

This relates to [#2127](https://github.com/dyoshikawa/rulesync/issues/2127), which proposes a tool-scoped override namespace so tool-specific permission keys don't leak into the canonical model.

## Motivation / Purpose

The canonical `.rulesync/permissions.json` model is `{ "permission": { "<category>": { "<glob>": "allow"|"ask"|"deny" } } }` with categories `bash, read, edit, write, webfetch, websearch, grep, glob, notebookedit, agent` (plus `mcp__*`). Each pattern resolves to exactly one of `allow`/`ask`/`deny`.

Vibe supports a strictly richer per-tool semantic: a tool can be ALWAYS-allowed by default while still escalating specific patterns to ASK. The canonical model cannot represent "allowed by default, but ask on these patterns" — it must choose a single verdict per pattern — so this information is lost when translating to/from Vibe.

## Background from Research

- Current mapping: rulesync maps canonical categories to Vibe tools as `bash→bash`, `read→read_file`, `edit`/`write→write_file`, `webfetch→fetch`, `websearch→search_web`. Per-tool `permission` is derived (`allow→always`, `ask→ask`), along with `allowlist`/`denylist`, plus top-level `enabled_tools`/`disabled_tools`. Pattern-level `ask` is currently skipped with a warning. (See the translator at `src/features/permissions/`.)
- Authoritative gap: Vibe's own config class `BaseToolConfig` defines `sensitive_patterns: list[str]`, documented as patterns that trigger ASK even when permission is ALWAYS. (Source: https://github.com/mistralai/mistral-vibe — `vibe/core/tools/base.py`, `BaseToolConfig`.)
- Because the canonical model can only set a pattern to `ask` outright, an ALWAYS base permission combined with `sensitive_patterns` degrades to a plain `ask` on those patterns, weakening the base allow rather than layering an escalation on top of it.

## Proposed Specification

Preserve `sensitive_patterns` when translating Vibe permissions, via one of:

- A Vibe-scoped override namespace in `.rulesync/permissions.json` (aligned with [#2127](https://github.com/dyoshikawa/rulesync/issues/2127)) that carries the per-tool `sensitive_patterns` list alongside the canonical base allow, or
- A per-pattern canonical semantic that can express "allow by default but escalate to ask on match" — a larger change with cross-tool implications.

The Vibe-scoped override is the lower-risk option and keeps the tool-specific concept out of the canonical categories.

## Open Questions

- Scoped override vs. a new canonical per-pattern semantic — do other tools have a comparable "allow-with-escalation" concept that would justify a canonical change rather than a Vibe-only override?
- When importing a Vibe config that has both an ALWAYS base and `sensitive_patterns`, what should the canonical categories reflect (base `allow`) while the override retains the escalation list?
- Should the existing warning on skipped pattern-level `ask` be revised once escalation can be represented?

## References

- https://github.com/mistralai/mistral-vibe — Mistral Vibe source; `vibe/core/tools/base.py` `BaseToolConfig` defines `sensitive_patterns` (patterns that trigger ASK even when permission is ALWAYS). Authoritative source for the gap.
- https://github.com/dyoshikawa/rulesync/issues/2127 — proposes the tool-scoped override namespace this issue would build on.

dyoshikawa:
## Problem

Vibe's per-tool `BaseToolConfig` carries a `sensitive_patterns` list — patterns that escalate to **ASK even when the base permission is ALWAYS** (allow). The canonical model can only set a pattern to a single `allow`/`ask`/`deny`, so this "allow-by-default but ask on these patterns" escalation could not be authored — it degraded to a plain `ask`, weakening the base allow, and round-tripped lossily.

## Fix

Add a tool-scoped `vibe` override namespace (per #2127):

```json
{
  "permission": { "bash": { "*": "allow" } },
  "vibe": { "permission": { "bash": { "sensitive_patterns": ["rm *", "sudo *"] } } }
}
```

- `vibe.permission.<category>.sensitive_patterns` carries the list per canonical category; it is emitted onto the Vibe tool config while the shared `permission` block keeps driving the base permission and allow/deny lists.
- On import, a tool's `sensitive_patterns` round-trips back into the `vibe` override (the base allow stays in the shared block).
- rulesync owns the list for any category named in the override (present list set, empty list cleared); unnamed categories keep whatever the existing `config.toml` had.

Verified against the primary source: [`vibe/core/tools/base.py`](https://github.com/mistralai/mistral-vibe) `BaseToolConfig` defines `sensitive_patterns: list[str]` = "Patterns that trigger ASK even when permission is ALWAYS".

Closes #2146


dyoshikawa:
@dyoshikawa Thank you!

PR #2187, opened 2026-07-08
refactor(generate): unify shared config file writes behind a declared-ownership gateway

saitota:
## Background

#2081/#2109/#2115 settled the write *order* (declared, registry-verified) and gave `.claude/settings.json` an ownership gateway. This PR covers the remaining pieces: `GENERATION_STEP_GRAPH` re-declares ~40 registry-known paths by hand, each shared file has its own merge helper re-spelling preserve/overwrite semantics (where the takt 329c9bf5 and amp 8e69d7ec bugs came from), and no test guards cross-feature key preservation.

## Changes

- Derive `writesSharedFile`/`dependsOn` from the registry plus a single `SHARED_WRITE_FEATURE_ORDER` list; the hand-maintained graph is gone.
- Add a registry-derived contract test: real generation runs one shared-write feature at a time (both scopes) and fails if any step deletes another feature's — or the user's — key paths in any shared file.
- Unify the per-file helpers into `shared-config-gateway.ts`: format codecs, declared conflict policies (`SHARED_CONFIG_OWNERSHIP`), and `applySharedConfigPatch`, which rejects writes outside declared ownership. takt's hand-rolled `provider_options` merge becomes the declared deep-merge.
- Surfaced by the derivation: hermes `subagents` writes `~/.hermes/config.yaml` with no declared shared write — now declared and ordered. Not-yet-migrated shared files sit on an enforced, shrink-only pending list.

No change to generated output (one parse-error message reworded).

## Test Plan

- `pnpm run check` + full vitest green (6,880 tests).
- Mutation-verified: breaking the hermes merge and reverting 329c9bf5 both fail.

Refs #2081, #2109, #2115


dyoshikawa:
@saitota Thank you!

PR #2191, opened 2026-07-09
fix(generate): make shared-write ownership guard reachable and harden derivation (#2189)

### LINKED ISSUE #2189: Follow-up cleanups from PR #2187 (shared-config gateway refactor)
## Background

PR #2187 (`refactor(generate): unify shared config file writes behind a declared-ownership gateway`) replaced four per-file config helper modules (`claudecode-settings-gateway`, `hermes-config`, `shared/takt-config`, `opencode-config`) with a single shared-config gateway (`src/features/shared/shared-config-gateway.ts`) and now derives the generation step graph from the processor registry (`src/lib/shared-file-derive.ts`).

Code and security review found **no high/critical issues** — merge/write semantics are equivalent to or more preserving than before, and the PR was merged. The items below are **non-blocking follow-up cleanups** (one `mid`, the rest `low`/informational) captured so they are not lost.

### 1. [mid] Unreachable ownership guard + vacuous test — `src/lib/shared-file-derive.ts`

`deriveSharedWriteSteps()` (~line 177; JSDoc ~163) is designed to throw when a feature writes a shared file but is absent from `SHARED_WRITE_FEATURE_ORDER`. However, the upstream `deriveSharedFileWriters()` (~line 110) already filters with `if (!SHARED_WRITE_FEATURES.has(entry.feature)) continue;` where `SHARED_WRITE_FEATURES = new Set(SHARED_WRITE_FEATURE_ORDER)`. So `writer.features` is always a subset of the order and **the throw is dead code**.

If a future feature (e.g. `commands`/`skills`) starts returning a shared path from `getSettablePaths`, it would be **silently excluded** from `writesSharedFile`/`dependsOn` and from `assertSharedFilesOrdered` — exactly the "unordered RMW drops another feature's keys" class this PR aims to prevent. Also `shared-file-derive.test.ts:38` `expect(() => deriveSharedWriteSteps()).not.toThrow()` can never fail and gives false confidence.

**Fix:** drop the pre-filter and let order-unregistered features hit the throw, or add a separate test asserting that every feature returning a shared path is present in the order.

### 2. [low] `getExtraSharedWritePaths()` call not try/catch protected — `src/lib/shared-file-derive.ts` (~line 84)

The sibling `getSettablePaths()` call (~line 70) is wrapped in try/catch, but `getExtraSharedWritePaths?.({ global }) ?? []` is not. Since `SHARED_WRITE_STEPS = deriveSharedWriteSteps()` is evaluated at `generate.ts` module load, a future tool whose `getExtraSharedWritePaths` throws would break module loading. The current hermes impl returns a constant so it is safe today; the asymmetry is a robustness trap.

### 3. [low] takt `config.yaml` round-trip now expands YAML anchors/aliases — `src/features/shared/shared-config-gateway.ts`

`stringifySharedConfig` uses `dump(..., { noRefs: true })` plus a `sanitizeSharedConfigValue` deep-copy. The old takt path used `parseTaktConfig` (no sanitize) + plain `dump` (`noRefs: false`), preserving hand-written `&ref`/`*ref` anchors. The new impl inlines/expands them. Semantically equivalent but changes user file formatting (takt only; hermes already used sanitize + noRefs). Worth noting as an exception to "no change to generated output".

### 4. [low] `policyFunction` strings are stringly-typed and unverified — `src/features/shared/shared-config-gateway.ts`

The `custom` policy `policyFunction` strings (`"applyPermissions"`, `"applyIgnoreReadDenies"`) in `SHARED_CONFIG_OWNERSHIP` are not verified to resolve to real exports; `applySharedConfigPatch` just throws for `custom`, and those functions are called directly by the claude features. A rename would silently stale the declarations. Add one test pinning name resolution.

### 5. [low/informational] jsonc nested `__proto__` bypasses sanitize — `src/features/shared/shared-config-gateway.ts:59-60`

`jsonc-parser` sets `__proto__` on the prototype rather than as an own property, so `isPlainObject(value)` returns false and `sanitizeSharedConfigValue` passes such an object through unsanitized. No global pollution and it does not leak to generated files; the jsonc path is read-only (`readOpencodeConfig`) and `opencode.json` is not in the gateway merge set. This is a **pre-existing limitation, not a regression**.

**Defensive improvement:** use a prototype-agnostic `isRecord` check plus `Object.setPrototypeOf(result, Object.prototype)` normalization if jsonc files ever join the deep-merge set.

## Solution / Next Steps

- Address finding #1 (the `mid` one) first: make the ownership guard reachable or add a real test so a future shared-writing feature can't slip through unordered.
- The `low` items (#2–#5) are optional hardening/consistency cleanups — pick up opportunistically.

_Follow-up from PR #2187._

dyoshikawa:
## Summary

Follow-up cleanups from PR #2187 (shared-config gateway refactor), addressing the review findings captured in the issue.

### 1. [mid] Unreachable ownership guard + vacuous test — `src/lib/shared-file-derive.ts`

`deriveSharedFileWriters()` pre-filtered registry entries with `if (!SHARED_WRITE_FEATURES.has(entry.feature)) continue;`, so `writer.features` was always a subset of `SHARED_WRITE_FEATURE_ORDER`. That made the ownership `throw` in `deriveSharedWriteSteps()` **dead code** and the `expect(() => deriveSharedWriteSteps()).not.toThrow()` test **vacuous** — a future feature (e.g. `commands`/`skills`) that started returning a shared path would be silently excluded from the write order instead of forcing an explicit precedence decision.

**Fix:** drop the pre-filter so every feature is considered. Single-writer files are still excluded (`features < 2`), so the derived writer set is unchanged — the inline snapshot test confirms this — while the guard is now genuinely reachable. The test now asserts the property directly (every writer feature has a position in the order) rather than relying only on the throw.

### 2. [low] `getExtraSharedWritePaths()` not try/catch protected — `src/lib/shared-file-derive.ts`

Wrapped the `getExtraSharedWritePaths()` call in try/catch, mirroring the sibling `getSettablePaths()` call, so a future tool whose implementation throws cannot break module load of `generate.ts` (`SHARED_WRITE_STEPS` is evaluated at load time).

### 4. [low] `policyFunction` strings unverified — `src/features/shared/shared-config-gateway.ts`

Added a test pinning every `custom` `policyFunction` name in `SHARED_CONFIG_OWNERSHIP` to a real exported function, so a rename cannot silently stale the declaration.

### Not addressed (informational per the issue)

- **#3** (takt `config.yaml` YAML-anchor expansion) — a documented behavior note, not a defect; no action required.
- **#5** (jsonc nested `__proto__` bypasses sanitize) — a pre-existing limitation, not a regression, on a read-only path (`opencode.json`) that is not in the gateway deep-merge set; deferred as defensive-only.

## Testing

- `pnpm cicheck` — all green (6882 unit tests, fmt, oxlint, typecheck, cspell, secretlint, gitignore/supported-tools/sync-skill-docs drift).
- Affected unit tests pass, including the unchanged writer snapshot and the strengthened guard test.

Closes #2189

### DIFF
--- a/src/features/shared/shared-config-gateway.test.ts
+++ b/src/features/shared/shared-config-gateway.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it } from "vitest";
 import { deriveSharedFileWriters } from "../../lib/shared-file-derive.js";
 import { createMockLogger } from "../../test-utils/mock-logger.js";
 import type { ClaudeSettingsJson } from "../../types/claude-settings.js";
+import * as sharedConfigGateway from "./shared-config-gateway.js";
 import {
   applyIgnoreReadDenies,
   applyPermissions,
@@ -209,6 +210,23 @@ describe("SHARED_CONFIG_OWNERSHIP", () => {
       (key) => SHARED_CONFIG_OWNERSHIP[key] !== undefined || !derivedKeys.has(key),
     );
     expect(stale, "remove these entries from GATEWAY_PENDING_SHARED_FILES").toEqual([]);
+  });
+
+  it("resolves every custom policyFunction name to an exported function", () => {
+    // The `custom` policy references its implementation by string name; a rename
+    // that misses the declaration would otherwise stale silently. Pin the names
+    // so they must resolve to a real exported function of this module.
+    const exports = sharedConfigGateway as Record<string, unknown>;
+    for (const [fileKey, declaration] of Object.entries(SHARED_CONFIG_OWNERSHIP)) {
+      for (const [feature, policy] of Object.entries(declaration.features)) {
+        if (policy?.kind !== "custom") continue;
+        expect(
+          typeof exports[policy.policyFunction],
+          `policyFunction '${policy.policyFunction}' declared for feature '${feature}' on ` +
+            `'${fileKey}' does not resolve to an exported function in shared-config-gateway.ts`,
+        ).toBe("function");
+      }
+    }
   });
 });
 
--- a/src/lib/shared-file-derive.test.ts
+++ b/src/lib/shared-file-derive.test.ts
@@ -32,9 +32,22 @@ const sortedFeatures = (writers: Map<string, Set<Feature>>): Record<string, stri
 
 describe("shared-file write derivation", () => {
   it("every feature that writes a shared file has a position in SHARED_WRITE_FEATURE_ORDER", () => {
+    // deriveSharedFileWriters considers every feature (not just the ordered
+    // ones), so a feature that starts returning a shared path surfaces here and
+    // must get an explicit precedence decision, not a silent arbitrary order.
+    // This asserts the property directly rather than relying only on the throw,
+    // which was unreachable while writers were pre-filtered by the order set.
+    const ordered = [...SHARED_WRITE_FEATURE_ORDER];
+    for (const writer of deriveSharedFileWriters()) {
+      for (const feature of writer.features) {
+        expect(
+          ordered,
+          `feature '${feature}' writes shared file '${writer.key}' but has no position in SHARED_WRITE_FEATURE_ORDER`,
+        ).toContain(feature);
+      }
+    }
     // deriveSharedWriteSteps throws when a shared-file writer feature is not
-    // ordered — a new feature entering the shared-write set must get an
-    // explicit precedence decision, not a silent arbitrary order.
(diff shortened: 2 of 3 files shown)


dyoshikawa:
Merged. Addressed finding #1 (mid) by removing the pre-filter so the shared-write ownership guard is reachable, plus findings #2 and #4 (try/catch hardening and the policyFunction-resolution test). #3 and #5 were informational and left as noted in the PR body.

PR #2194, opened 2026-07-09
feat(permissions): add augmentcode-scoped override for webhook/script policies and tool-response eventType

### LINKED ISSUE #2134: AugmentCode permissions: allow authoring `webhook-policy`/`script-policy` and `tool-response` eventType via a tool-scoped override
## Summary

rulesync's AugmentCode permissions translator (`src/features/permissions/augmentcode-permissions.ts`) emits `toolPermissions[]` entries with `toolName`, an optional bash-only `shellInputRegex`, and `permission.type ∈ allow | deny | ask-user`. AugmentCode additionally supports `webhook-policy` / `script-policy` permission types (with `webhookUrl` / `script`) and a `tool-response` `eventType` (a post-execution check). rulesync preserves these "special" entries verbatim on round-trip but never generates them, and skips them on import with a warning. This issue proposes making them authorable through a **tool-scoped override namespace** per #2127.

## Motivation / Purpose

The canonical rulesync model — `{ "permission": { "<category>": { "<glob>": "allow"|"ask"|"deny" } } }` — has no way to express "delegate this decision to a webhook or a script," nor a pre-execution vs. post-execution dimension. AugmentCode has both. Users who rely on webhook/script policies or post-execution checks currently cannot author them from rulesync; they only survive if already present in an existing file. A tool-scoped `augmentcode` override namespace would let users author these AugmentCode-specific policies without contaminating the canonical categories or other tools' output.

## Background from Research

- rulesync currently generates AugmentCode [`toolPermissions[]`](https://docs.augmentcode.com/cli/permissions) entries with `toolName`, an optional `shellInputRegex` (bash only), and `permission.type` restricted to `allow`, `deny`, or `ask-user`; it maps the 6 canonical categories to Augment tool names and applies first-match-wins ordering. (Derived from the current implementation in `src/features/permissions/augmentcode-permissions.ts`.)
- AugmentCode's `permission.type` also supports [`webhook-policy` and `script-policy`](https://docs.augmentcode.com/cli/permissions) (carrying `webhookUrl` and `script` respectively) — a "delegate the allow/ask/deny decision to a webhook or script" action that the canonical allow/ask/deny vocabulary cannot express.
- AugmentCode supports an [`eventType` of `tool-response`](https://docs.augmentcode.com/cli/permissions), i.e. a check that runs after tool execution rather than before; the canonical model has no pre/post-execution dimension.
- rulesync today only preserves these "special" entries (webhook/script policy, non-default `eventType`, `webhookUrl`, `script`) verbatim on round-trip, never generates them, and skips them with a warning on import — so they are not authorable from rulesync. (Inferred from the current translator behavior.)

## Proposed Specification

Introduce an `augmentcode`-scoped override namespace in `.rulesync/permissions.json` (per #2127) allowing users to author:

- `webhook-policy` entries (with `webhookUrl`).
- `script-policy` entries (with `script`).
- entries with `eventType: tool-response` (post-execution checks), alongside the default pre-execution behavior.

The canonical categories continue to produce the common `allow` / `deny` / `ask-user` `toolPermissions[]`; the `augmentcode` namespace is additive for these AugmentCode-specific policy types. Exact shape should follow the #2127 convention. (Proposed design — details inferred and open to refinement.)

## Open Questions

- How should ordering/first-match-wins interleave canonical-derived entries with `augmentcode`-scoped entries? Should users control relative order?
- Should `webhook-policy` / `script-policy` be modeled as a generic "delegate" action reusable across tools (cf. Amp's `delegate`), or kept tool-scoped?
- Should `eventType: tool-response` be represented as an attribute on existing entries or a separate list?

## References

- https://docs.augmentcode.com/cli/permissions — AugmentCode CLI permissions documentation; documents `toolPermissions[]`, `permission.type` values including `webhook-policy` / `script-policy`, `webhookUrl` / `script`, and the `eventType` (including `tool-response`).
- #2127 — umbrella proposal for a tool-scoped override namespace in `.rulesync/permissions.json`; this issue is the AugmentCode-specific instance.

dyoshikawa:
## Summary

Introduces a tool-scoped `augmentcode` permissions override namespace (per #2127) so AugmentCode's custom-policy entries — which the canonical allow/ask/deny model cannot express — become authorable and round-trippable.

AugmentCode's `toolPermissions[]` supports:
- `permission.type` of `webhook-policy` / `script-policy` (delegating the decision to a `webhookUrl` / `script`), and
- an `eventType` of `tool-response` (a post-execution check rather than the default pre-execution `tool-call`).

Previously these "special" entries were preserved verbatim on round-trip but could never be authored from rulesync, and were skipped with a warning on import.

## Changes

- **`src/types/permissions.ts`** — new `AugmentcodePermissionsOverrideSchema` (a loose `toolPermissions` array of verbatim entries), registered on `PermissionsConfig` as the `augmentcode` key; docstring updated.
- **`src/features/permissions/augmentcode-permissions.ts`**:
  - `fromRulesyncPermissions`: authored override entries are **prepended** ahead of the canonical-generated basic rules (first-match-wins safe, so a webhook/script gate or tool-response check is never shadowed). When the override authors `toolPermissions` it becomes the source of truth, so the existing file's specials are no longer separately preserved; a composite-key dedup prevents any double-emit.
  - `toRulesyncPermissions`: special entries are **extracted verbatim into the `augmentcode` override** instead of being warn-and-dropped, so they round-trip and become user-authorable. Basic entries continue to drive the shared `permission` block.
- **Docs** — `docs/reference/file-formats.md` (+ synced `skills/rulesync/`) gains an "AugmentCode-only override" note.
- **Tests** — cover authoring from the override, extraction on import, omission when there are no specials, and a no-double-emit round-trip.

## Design decision

Per the maintainer's decision recorded on the issue, `delegate` (the "defer to an external program" concept shared with Amp's `webhook`/`script`/`delegate`) is **not** promoted to a canonical action. This stays a tool-scoped override, consistent with Amp's existing deliberate design (`delegate` is kept in the `amp` override, not canonical).

## Ordering / dedup (open-question resolution)

- Authored specials lead, preserving author order; the generated basic rules follow.
- De-duplication uses a composite key (`toolName|shellInputRegex|eventType|type|webhookUrl|script`), keeping the first occurrence, so an authored policy wins over an identical preserved one and the same policy is never emitted twice.

Closes #2134


dyoshikawa:
@dyoshikawa Thank you!

PR #2197, opened 2026-07-09
feat(deps): migrate to js-yaml v5

### LINKED ISSUE #2122: Migrate to js-yaml v5 (held back from the #2114 group bump due to breaking empty-input behavior)
## Background

The dependabot `all-dependencies` group bump (PR #2114) tried to raise `js-yaml` from `4.2.0` to `5.2.1` (a **major** version). js-yaml v5 changed empty-input behavior: `load("")` now **throws** ("expected a document, but the input is empty") instead of returning `undefined`. This broke two parsers and failed CI:

- `parseApmManifest` — `src/lib/apm/apm-manifest.ts:83` (`load(content)` is not wrapped, so empty `apm.yml` throws instead of returning `{ dependencies: [] }`).
- `injectSourceMetadata` — `src/lib/gh/gh-frontmatter.ts:74-77` (empty frontmatter `---\n---` now throws inside the `try`, which is converted to `invalid frontmatter` instead of being treated as an empty block).

To unblock the other 126 updates, js-yaml was pinned back to `4.2.0` and excluded from that group bump (merged in #2114). This issue tracks the **deliberate** migration to js-yaml v5.

## Details

`js-yaml` is used at ~76 `load`/`dump` sites across many features (frontmatter output for cursor rules, subagents, mcp, permissions, commands, skills, plus `gh`/`apm` lib parsing). The existing e2e Tool × Feature matrix passed with only the two empty-input tests failing, which suggests `dump` output for non-empty values is compatible — but the migration should be validated end-to-end rather than assumed.

## Solution / Next Steps

1. Bump `js-yaml` to v5 (and `@types/js-yaml` to a matching version).
2. Guard empty input at every `load` call site that can receive empty content — at minimum:
   - `src/lib/apm/apm-manifest.ts:83` — return `{ dependencies: [] }` when content is blank.
   - `src/lib/gh/gh-frontmatter.ts:74` — treat empty `fmBody` as an empty block (provenance-only) rather than `invalid frontmatter`.
   - Audit the other ~74 `load`/`dump` sites for empty-input and any other v5 breaking changes (option renames, `dump` formatting differences).
3. Add regression tests for the empty-input cases so a future v5 bump does not reintroduce the failure.
4. Verify generated YAML output is byte-compatible across the Tool × Feature matrix (`pnpm cicheck`), since this is a config-generation tool and `dump` format changes would alter users' generated files.
5. Re-enable js-yaml in dependabot updates once the migration lands.

dyoshikawa:
## Summary

Migrates `js-yaml` from 4.2.0 to **5.2.1** (a major version that was held back from the #2114 dependabot group bump due to a breaking empty-input change).

js-yaml v5 changed `load("")` to **throw** (`expected a document, but the input is empty`) instead of returning `undefined`. Several parsers here relied on the old `undefined` sentinel, so a naive bump broke them.

## Approach

**Centralized empty-input guard.** Rather than patch each call site ad hoc, this adds `loadYaml` (`src/utils/yaml.ts`) — a thin wrapper that:
- returns `undefined` for empty / whitespace-only / **comment-only** documents (matching v4), and
- re-throws genuine parse errors unchanged (detected via `YAMLException.reason === "expected a document, but the input is empty"`).

Every js-yaml `load` call now routes through `loadYaml`, so the v4 contract lives in one place and a future bump cannot silently reintroduce the regression.

**Dependency changes:**
- `js-yaml` 4.2.0 → 5.2.1
- Dropped `@types/js-yaml` — v5 is written in TypeScript and ships its own type declarations.

**Scope of breaking changes actually hit:** the codebase already used named imports (`import { load, dump }`), no default import, no custom `Type` API, and none of v5's removed `dump` options (`styles`/`replacer`/`noCompatMode`/`condenseFlow`/`quotingType`/`noArrayIndent`) — only `lineWidth`/`noRefs`/`sortKeys`, all retained. So `dump` needed no changes.

## Bugs fixed

Beyond the two documented breakages:
- `parseApmManifest` (`apm-manifest.ts`) — empty `apm.yml` now returns `{ dependencies: [] }` instead of throwing.
- `injectSourceMetadata` (`gh-frontmatter.ts`) — empty frontmatter block (`---\n---`) is treated as an empty block, not `invalid frontmatter`.

…the migration also fixed several **undocumented** sites with the same "empty file parses to undefined/null" assumption inside a try/catch: `goose-command`, `goose-mcp`, `goose-subagent`, `roo-subagent` (an empty file would have thrown a "Failed to parse …" error under a naive v5 bump).

## Verification

- `loadYaml` unit tests cover empty / whitespace / comment-only / valid / invalid inputs.
- Existing empty-input regression tests (`apm-manifest.test.ts`, `gh-frontmatter.test.ts`) stay green.
- Full unit suite: 6918 pass.
- YAML-related e2e (`mcp`, `permissions`, `commands`, `subagents`, `skills`): 437 pass — these assert on generated YAML byte-for-byte, empirically confirming v5 `dump` output is compatible across the Tool × Feature matrix.
- `pnpm cicheck` green.

> Note: this touches `package.json` / lockfile (a dependency manifest). The maintainer explicitly authorized driving this one through to merge.

Closes #2122


dyoshikawa:
@dyoshikawa Thank you!

PR #2198, opened 2026-07-09
refactor(sources): unify root-fallback decision and short-circuit github root fetch

### LINKED ISSUE #2119: Follow-ups from PR #2118 review: root-fallback DRY, widened github fallback trigger, and .gitignore drive-by
## Background

PR #2118 ("fix(sources): handle root skill with metadata dirs") fixes the case where a remote repository's root `SKILL.md` coexists with metadata directories (e.g. `.claude-plugin/`) so that the root-skill fallback fires correctly. The fix and its tests are correct and have been merged.

During review, three **non-blocking (severity: low)** follow-ups surfaced. They do not block the merge but are captured here so they are not lost.

### Finding 1 (low) — DRY: root-fallback decision logic duplicated

The "should we use the root fallback?" decision is implemented in two places with asymmetric gate conditions:

- `src/lib/sources.ts:463-468` (git path, `groupRemoteFilesBySkillRoot`) uses `hasRootSkillFile && !grouped.has(singleSkillName)` — an explicit `SKILL.md` existence check.
- `src/lib/sources.ts:724-730` (github path, `discoverGithubSkillDirs`) delegates the `SKILL.md` check into the inner `groupRemoteFilesBySkillRoot` call and only gates on `!hasRequestedSkillDir`.

The effect is equivalent today, but the two branches can drift if only one side is changed later.

**Suggested fix:** extract a shared helper such as `shouldUseRootFallback({ skillFilter, isWildcard, hasRootSkillFile, hasRequestedSkillDir })` and call it from both paths.

### Finding 2 (low) — github path: fallback trigger condition widened

The fallback trigger at `src/lib/sources.ts:724` was widened. Previously `fetchRootLevelFallbackSkill` was called only when `remoteSkillDirs.length === 0`; now it is called whenever the requested skill dir is absent.

`fetchRootLevelFallbackSkill` fetches all root file contents via `getFileContent` (`sources.ts:552-563`) **before** checking whether a root `SKILL.md` exists. So in the case "`skills/` has multiple real subdirs, the requested name matches none, and root has NO `SKILL.md`", it now performs a wasted full API fetch of root files that are then discarded (minor perf waste).

There is also a behavior change: when real skill dirs and an incidental root `SKILL.md` coexist and a nonexistent skill name is requested, nothing was installed before, but now the root is installed under the requested name. This is a rare edge case — **confirm it is intended.**

**Suggested fix:** short-circuit before the full root fetch when a root `SKILL.md` is not present, and confirm/lock the new behavior with a test.

### Finding 3 (low) — `.gitignore` drive-by change mismatched with PR description

`.gitignore:219` adds `.tokensave/`, which is unrelated to "handle root skill with metadata dirs". The PR body says it is ignored "through the generated gitignore source registry", but it was actually added directly to the hand-written section above the `# Generated by Rulesync` block — not via `HAND_MAINTAINED_GITIGNORE_ENTRIES` in `src/cli/commands/gitignore-entries.ts:36`.

The placement itself is fine, but the "via registry" wording is inaccurate.

**Suggested fix:** if `.tokensave/` should be owned by the registry, move it into `HAND_MAINTAINED_GITIGNORE_ENTRIES` and regenerate; otherwise keep the manual entry but correct the description. Prefer splitting unrelated drive-by changes out of feature/fix PRs in the future.

## Solution / Next Steps

- [ ] Finding 1: extract a shared `shouldUseRootFallback` helper used by both the git and github paths.
- [ ] Finding 2: skip the full root file fetch when no root `SKILL.md` exists, and confirm the intended behavior with a test.
- [ ] Finding 3: reconcile `.tokensave/` ignore ownership (registry vs. manual) and correct any inaccurate wording.

dyoshikawa:
## Summary

Addresses the three non-blocking (low) follow-ups from the PR #2118 review.

### Finding 1 — DRY: shared root-fallback decision

The "install the repo root as the single requested skill" decision was implemented twice with asymmetric gates:
- git path (`groupRemoteFilesBySkillRoot`) — explicit `hasRootSkillFile && !grouped.has(name)` check.
- github path (`discoverGithubSkillDirs`) — only gated on `!hasRequestedSkillDir`, delegating the SKILL.md check downstream.

Extracted a shared `shouldUseRootFallback({ skillFilter, isWildcard, hasRootSkillFile, hasRequestedSkillDir })` used by both paths so they cannot drift.

### Finding 2 — perf + behavior lock

#2118 widened the github fallback trigger to fire whenever the requested dir is absent, so `fetchRootLevelFallbackSkill` performed a **full root-file fetch** even when the root had no `SKILL.md` and nothing would be installed. Now the root `SKILL.md` is detected from the directory listing already in hand and gated via the shared helper, short-circuiting the fetch when there is nothing to install.

Added two tests to lock the intended behavior:
- Root `SKILL.md` **is** installed under the requested name when it coexists with real skill dirs and the requested skill is absent (locks the widened-trigger edge case flagged in the review).
- **No** root files are fetched (`getFileContent` not called) when the requested skill is absent and there is no root `SKILL.md`.

### Finding 3 — `.gitignore` `.tokensave/`

`.tokensave/` is a third-party byproduct that no tool emits, so per the gitignore guidelines (`HAND_MAINTAINED_GITIGNORE_ENTRIES` is for tool-external paths) the hand-written section is the correct owner. Left as-is intentionally; no code change.

## Verification

- New + existing `sources.test.ts`: 41 pass.
- `pnpm cicheck` green (6920 unit tests).

Closes #2119

### DIFF
--- a/src/lib/sources.test.ts
+++ b/src/lib/sources.test.ts
@@ -1247,6 +1247,72 @@ describe("resolveAndFetchSources", () => {
     );
   });
 
+  it("installs the root SKILL.md under the requested name when real skill dirs coexist and the requested skill is absent", async () => {
+    // Locks the (intended) widened-fallback behavior: when the repository has
+    // real skill dirs plus an incidental root SKILL.md and the caller requests a
+    // skill name that matches no dir, the root is installed under the requested
+    // name.
+    mockClientInstance.listDirectory.mockImplementation(
+      async (_owner: string, _repo: string, path: string) => {
+        if (path === ".") {
+          return [
+            { name: "other-skill", path: "other-skill", type: "dir", size: 0 },
+            { name: "SKILL.md", path: "SKILL.md", type: "file", size: 50 },
+            { name: "README.md", path: "README.md", type: "file", size: 20 },
+          ];
+        }
+        return [];
+      },
+    );
+    mockClientInstance.getFileContent.mockImplementation(
+      async (_o: string, _r: string, path: string) => {
+        if (path === "SKILL.md") return "# Root Skill";
+        if (path === "README.md") return "docs";
+        return "";
+      },
+    );
+
+    const result = await resolveAndFetchSources({
+      logger,
+      sources: [{ source: "org/repo:.", skills: ["nonexistent"] }],
+      projectRoot: testDir,
+    });
+
+    expect(result.fetchedSkillCount).toBe(1);
+    expect(writeFileContent).toHaveBeenCalledWith(
+      join(testDir, RULESYNC_CURATED_SKILLS_RELATIVE_DIR_PATH, "nonexistent", "SKILL.md"),
+      "# Root Skill",
+    );
+  });
+
+  it("does not fetch root files when the requested skill is absent and there is no root SKILL.md", async () => {
+    // The fallback (and its full root-file fetch) must be short-circuited when the
+    // directory listing shows no root SKILL.md — nothing would be installed, so no
+    // root content should be fetched.
+    mockClientInstance.listDirectory.mockImplementation(
+      async (_owner: string, _repo: string, path: string) => {
+        if (path === ".") {
+          return [
+            { name: "other-skill", path: "other-skill", type: "dir", size: 0 },
+            { name: "README.md", path: "README.md", type: "file", size: 20 },
+          ];
+        }
+        return [];
+      },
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2203, opened 2026-07-10
refactor(rules): consolidate rovodev root-mirror hooks into one contract

### LINKED ISSUE #2111: Follow-ups from PR #2108: decouple/guard rovodev root-mirror spoke hooks
## Background

PR #2108 was a behavior-preserving refactor that moved rovodev-specific quirks (root-mirror generation, root-mirror deletion glob, and local-root deletion glob) out of the shared hub (`rules-processor.ts`) and into per-tool spoke hooks on `RovodevRule`. Behavior was verified as preserved (`ROVODEV_RULE_FILE_NAME === "AGENTS.md"` matches the previous hardcode; deletion globs and guard conditions are equivalent). The PR was reviewed and merged. No high/critical or security findings. These are the non-blocking (mid/low) follow-ups captured for later.

### mid

1. **Implicit coupling between the `meta.mirrorsRootToAgentsMd` flag and two separate hooks** — `src/features/rules/rules-processor.ts:239` plus `getRootMirrorFiles` (generation) and `getRootMirrorDeletionGlobs` (deletion) on `RovodevRule`. Before the refactor, generation and deletion were bound to a single `factory.class === RovodevRule` condition, so they were symmetric. Now they are independent optional methods. If a future tool sets `mirrorsRootToAgentsMd: true` and implements only one hook (e.g., defines the generation hook but forgets the deletion hook), it would get an asymmetric "mirror is generated but never cleaned up on deletion" behavior with no compile-time or runtime guard — only the JSDoc "Set alongside" comment protects against it. Consider consolidating the three hooks into a single optional interface/object, or detecting a missing hook when the meta flag is set and warning.

### low

2. **No direct unit tests for the three new public static methods** — `src/features/rules/rovodev-rule.ts` (`getRootMirrorFiles`, `getRootMirrorDeletionGlobs`, `getLocalRootDeletionGlob`). They are currently covered only indirectly via `rules-processor.test.ts` integration tests. Behavior preservation is assured, but as newly extracted pure-function public APIs, adding 1–2 direct unit tests would improve regression resistance.

## Solution / Next Steps

- Address the mid item to remove the footgun for future tools that opt into root mirroring — either unify the three hooks under one contract or assert hook presence when `mirrorsRootToAgentsMd` is set.
- Optionally add direct unit tests for the three extracted static methods on `RovodevRule`.

dyoshikawa:
## Summary

Follow-up from PR #2108 (the `mid` item in #2111): the `meta.mirrorsRootToAgentsMd` flag was coupled to **two independent optional hooks** on the tool class — `getRootMirrorFiles` (generation) and `getRootMirrorDeletionGlobs` (deletion). Before PR #2108 both were bound to a single `factory.class === RovodevRule` condition, so they were symmetric. After the refactor they became independent optional methods, so a future tool that set `mirrorsRootToAgentsMd: true` but implemented only one hook would silently get asymmetric "mirror is generated but never cleaned up on deletion" behavior — guarded only by a JSDoc "Set alongside" comment.

## Changes

- Removed the `meta.mirrorsRootToAgentsMd` flag and the two separate `getRootMirrorFiles` / `getRootMirrorDeletionGlobs` hooks.
- Introduced a single optional `getRootMirror()` contract that bundles both halves:
  - `getMirrorFiles({ outputRoot, rootRule, content })`
  - `getMirrorDeletionGlobs({ outputRoot })`
- **The method's mere presence is now the single source of truth** for "this tool mirrors its root to `./AGENTS.md`". Generation (`rules-processor.ts`), mirror deletion (`rules-processor.ts`), and ownership registration (`generate.ts`) all key off `factory.class.getRootMirror`, so the generation/deletion pair cannot drift out of symmetry — a tool cannot define one half without the other.
- Added direct unit tests (the `low` item in #2111) for `getRootMirror` (both halves) and `getLocalRootDeletionGlob`, which were previously covered only indirectly via `rules-processor.test.ts` integration tests.

`getLocalRootDeletionGlob` is intentionally left as a standalone hook: it is gated by `localRootMode: "separate-local-file"` (a different mechanism) and has no generation/deletion partner, so it carries no symmetry footgun.

Behavior is preserved — rovodev is the only tool that mirrors, and it always returns the contract, so all three gate sites remain equivalent to the previous flag-based checks. Verified by the unchanged Tool × Feature e2e rules matrix (`e2e-rules.spec.ts`, 105 tests) plus the new unit tests; full `pnpm cicheck` is green.

Closes #2111

### DIFF
--- a/src/features/rules/rovodev-rule.test.ts
+++ b/src/features/rules/rovodev-rule.test.ts
@@ -472,4 +472,68 @@ describe("RovodevRule", () => {
       expect(RovodevRule.isTargetedByRulesyncRule(rulesyncRule)).toBe(true);
     });
   });
+
+  describe("getRootMirror", () => {
+    const primaryRoot = () =>
+      new RovodevRule({
+        outputRoot: testDir,
+        relativeDirPath: ".rovodev",
+        relativeFilePath: "AGENTS.md",
+        fileContent: "root body",
+        validate: true,
+        root: true,
+      });
+
+    it("getMirrorFiles mirrors the primary root to a project-root ./AGENTS.md", () => {
+      const mirrors = RovodevRule.getRootMirror().getMirrorFiles({
+        outputRoot: testDir,
+        rootRule: primaryRoot(),
+        content: "mirrored content",
+      });
+
+      expect(mirrors).toHaveLength(1);
+      const [mirror] = mirrors;
+      expect(mirror).toBeInstanceOf(RovodevRule);
+      expect(mirror?.isRoot()).toBe(true);
+      expect(mirror?.getRelativeDirPath()).toBe(".");
+      expect(mirror?.getRelativeFilePath()).toBe("AGENTS.md");
+      expect(mirror?.getFileContent()).toBe("mirrored content");
+    });
+
+    it("getMirrorFiles returns empty when the rule is not the primary root", () => {
+      const nonPrimary = new RovodevRule({
+        outputRoot: testDir,
+        relativeDirPath: ".",
+        relativeFilePath: "AGENTS.md",
+        fileContent: "already at root",
+        validate: true,
+        root: true,
+      });
+
+      expect(
+        RovodevRule.getRootMirror().getMirrorFiles({
+          outputRoot: testDir,
+          rootRule: nonPrimary,
+          content: "content",
+        }),
+      ).toEqual([]);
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2204, opened 2026-07-10
refactor(lib): harden shared-file-writer derivation against drift

### LINKED ISSUE #2110: Follow-ups from PR #2109: harden shared-file-writer derivation against drift
## Background

PR #2109 introduced `src/lib/shared-file-derive.ts`, which derives the multi-writer set from `PROCESSOR_REGISTRY` x `getSettablePaths` (across project and global scope, plus a `getExtraSharedWritePaths` contract) and asserts it matches `GENERATION_STEP_GRAPH`. This makes a missing shared-file-writer declaration fail CI instead of being caught only in review. Refs #2081.

The PR was reviewed and merged. No high/critical correctness or security findings were raised. The items below are the non-blocking (mid/low) follow-ups captured for later.

### mid

1. **`SHARED_WRITE_FEATURES` manual allowlist limits source-of-truth completeness** — `src/lib/shared-file-derive.ts:22-28`. Derivation targets are hand-fixed to `ignore/mcp/hooks/permissions/rules` (5 features). If `commands`/`subagents`/`skills` later start writing to shared config files and someone forgets to add them here, the drift this PR aims to prevent would silently re-appear. Consider an exhaustiveness check against the `Feature` enum, or an explicit exclusion list with a comment/test documenting why each feature is in or out.

2. **`settablePathsForScope` catch early-return swallows `getExtraSharedWritePaths`** — `src/lib/shared-file-derive.ts` (the try/catch that does `return paths;`). When `getSettablePaths({ global })` throws for a scope, the subsequent `getExtraSharedWritePaths` is skipped. Currently harmless with existing tools, but a tool whose global `getSettablePaths` throws while it also has a global-only extra shared path would silently drop that path. Collect extra paths outside the try/catch, independent of `getSettablePaths` success.

### low

3. **`getExtraSharedWritePaths` is a duck-typed contract with no base-class/interface declaration** — `codexcli-hooks.ts`, `kilo-rule.ts`, etc. Coupled only via the optional call `cls.getExtraSharedWritePaths?.(...)`. New tool authors cannot discover the hook, and signature mismatches are not caught at compile time. Consider a base-class default `static getExtraSharedWritePaths(): SharedWritePath[] { return []; }`.

4. **Type looseness (`as unknown as`, `unknown` return) defeats derivation type safety** — `src/lib/shared-file-derive.ts:41-46, 100`. If the `getSettablePaths` return shape changes, derivation may silently return empty (missed shared files). Extract a type-guard for the expected `settable` shape (`relativeFilePath`/`root`/`alternativeRoots`).

5. **Test cases 2 and 3 are logically subsumed by case 1** — `src/lib/shared-file-derive.test.ts:31-47`. If the full-equality case passes, "no undeclared" and "no missing declaration" necessarily hold. Acceptable if intentional for clearer failure messages; add a comment noting that.

6. **(out of PR scope) Kilo writes mcp/rules to `kilo.json` but permissions to `kilo.jsonc`** — `kilo-permissions.ts:114-119` vs `kilo-mcp.ts:271-282`. The PR's fix (removing permissions from the `kilo.json` shared set) correctly matches reality. Worth confirming whether splitting one tool's config across two files (`kilo.json`/`kilo.jsonc`) is intentional.

## Solution / Next Steps

- Address the two mid items to harden the derivation against future drift and silent path loss (exhaustiveness/exclusion for the feature allowlist; move extra-path collection out of the try/catch).
- Optionally tackle the low items to improve discoverability and type safety of the `getExtraSharedWritePaths` contract.
- Separately confirm the intended Kilo `kilo.json`/`kilo.jsonc` split.

dyoshikawa:
## Summary

Addresses the two `mid` follow-ups from the PR #2109 review captured in #2110, hardening `src/lib/shared-file-derive.ts` against future drift. Refs #2081.

### mid #1 — exhaustiveness for the shared-write feature classification
`SHARED_WRITE_FEATURE_ORDER` is hand-maintained. If `commands`/`subagents`/`skills` later start writing shared config files and someone forgets to account for them, the drift this derivation is meant to prevent could silently re-appear.

- Added an explicit `NON_SHARED_WRITE_FEATURES` exclusion list (`commands`, `skills`), each documented with *why* it is excluded (both write their own dedicated per-item artifacts — command files / skill directories — never a file another feature also writes).
- Added a test asserting `SHARED_WRITE_FEATURE_ORDER ∪ NON_SHARED_WRITE_FEATURES` partitions `ALL_FEATURES` exactly. Adding a new `Feature` now fails the test until it is consciously classified as a shared writer or an excluded one. (A duplicate makes the classified list longer than `ALL_FEATURES`, so the single partition assertion also rejects overlap — no redundant case.)

### mid #2 — `getExtraSharedWritePaths` no longer swallowed by a throwing `getSettablePaths`
`settablePathsForScope` early-returned from the `getSettablePaths` `catch`, which skipped the subsequent `getExtraSharedWritePaths` collection. Currently harmless with existing tools, but a tool whose global `getSettablePaths` throws while it also has a global-only extra shared path would silently drop that path.

- The two hooks are now collected independently: a throwing `getSettablePaths` leaves `settable` undefined and falls through to the extra-path collection instead of returning early. Both hooks remain fail-open individually (the derivation runs at module load, so a throwing tool must not take down every `generate`).
- Exported `settablePathsForScope` and added a test that a throwing `getSettablePaths` still yields the tool's `getExtraSharedWritePaths` entries.

## Notes
- The `low` items (#3 base-class default for `getExtraSharedWritePaths`, #4 type-guard for the `settable` shape, #5 test-case-subsumption comment) and item #6 (confirm the intended Kilo `kilo.json`/`kilo.jsonc` split) are out of scope for this PR and left for the maintainer.

Full `pnpm cicheck` is green (typecheck, lint, 6926 tests, content checks).

Closes #2110

### DIFF
--- a/src/lib/shared-file-derive.test.ts
+++ b/src/lib/shared-file-derive.test.ts
@@ -1,11 +1,13 @@
 import { describe, expect, it } from "vitest";
 
-import type { Feature } from "../types/features.js";
+import { ALL_FEATURES, type Feature } from "../types/features.js";
 import { GENERATION_STEP_GRAPH } from "./generate.js";
 import {
   deriveSharedFileWriters,
   deriveSharedWriteSteps,
+  NON_SHARED_WRITE_FEATURES,
   SHARED_WRITE_FEATURE_ORDER,
+  settablePathsForScope,
 } from "./shared-file-derive.js";
 
 const graphWritersByFile = (): Map<string, Set<Feature>> => {
@@ -51,6 +53,34 @@ describe("shared-file write derivation", () => {
     // evaluates at module load), so an unordered writer fails loudly at runtime,
     // not just here.
     expect(() => deriveSharedWriteSteps()).not.toThrow();
+  });
+
+  it("classifies every Feature as either a shared writer or an explicit non-shared-writer", () => {
+    // Guards #2110 mid#1: SHARED_WRITE_FEATURE_ORDER is hand-maintained, so a
+    // feature that later starts writing a shared file could be silently
+    // forgotten. Requiring the two lists to partition ALL_FEATURES exactly means
+    // adding a Feature forces classifying it in one of them (a duplicate would
+    // make `classified` longer than ALL_FEATURES, so this also rejects overlap).
+    const classified = [...SHARED_WRITE_FEATURE_ORDER, ...NON_SHARED_WRITE_FEATURES].toSorted();
+    expect(classified).toEqual([...ALL_FEATURES].toSorted());
+  });
+
+  it("collects getExtraSharedWritePaths even when getSettablePaths throws", () => {
+    // Guards #2110 mid#2: the two hooks are independent, so a throwing
+    // getSettablePaths must not suppress a tool's (possibly scope-only) extra
+    // shared path.
+    const cls = {
+      getSettablePaths: () => {
+        throw new Error("boom");
+      },
+      getExtraSharedWritePaths: () => [
+        { relativeDirPath: ".config/example", relativeFilePath: "example.json" },
+      ],
+    };
+
+    expect(settablePathsForScope(cls, true)).toEqual([
+      { relativeDirPath: ".config/example", relativeFilePath: "example.json" },
+    ]);
   });
 
   it("orders every writer pair of every shared file by dependsOn edges", () => {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2217, opened 2026-07-10
fix: normalize Codex MCP server names

### LINKED ISSUE #2215: Normalize MCP name for Codex
Hey.

Currently Codex fails with MCP server names containing spaces, or other symbols for that matter:

```toml
[mcp_servers."Postgres MCP - Production - Read Only"]
command = "node"
args = [ "./scripts/mcp/postgres-ro.mjs" ]
```

```
⚠ MCP client for `Postgres MCP - Production - Read Only` failed to start: MCP startup failed: Invalid MCP server name 'Postgres MCP - Production - Read Only': must match pattern ^[a-zA-Z0-9_-]+$
```

While this is an issue on Codex side, I have tried reporting it and couldn't get an improvement on their side: https://github.com/openai/codex/issues/19179

These names are much more readable. Our project and team mostly uses Cursor, which does support these names fully, and only a few people use Codex or other tools, so it would be unwise for us to rename those MCPs for everyone.

Instead, would it be possible for Rulesync to automatically normalize MCP name for Codex? Replace spaces and dashes with `_` and perform some other conversions too, so that Codex would see `postgres_mcp_production_read_only` instead of `Postgres MCP - Production - Read Only`?

autaut03:
## Summary

- normalize invalid MCP server names for Codex CLI
- warn and let the last server win when normalized names collide
- warn and skip names that cannot be normalized safely

## Testing

- pnpm cicheck

Fixes #2215

### DIFF
--- a/src/features/mcp/codexcli-mcp.test.ts
+++ b/src/features/mcp/codexcli-mcp.test.ts
@@ -325,6 +325,40 @@ args = ["server.js"]
       expect(codexcliMcp.getRelativeFilePath()).toBe("config.toml");
     });
 
+    it("should normalize invalid Codex MCP server names and warn when the last collision wins", async () => {
+      const warnSpy = vi.spyOn(console, "warn").mockImplementation(() => {});
+      const rulesyncMcp = new RulesyncMcp({
+        relativeDirPath: RULESYNC_RELATIVE_DIR_PATH,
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@autaut03 Thank you!

PR #2222, opened 2026-07-13
fix: make security-scan resilient to empty/stalled OpenRouter responses

dyoshikawa:
## Background

The scheduled **Security Scan** workflow has been failing intermittently (e.g. run [`29218107477`](https://github.com/dyoshikawa/rulesync/actions/runs/29218107477), and earlier 6/15 & 6/22 scheduled runs). The `Run security scan` step exits with code 1.

Root cause, from the CI logs:

1. OpenRouter returned an **empty/truncated response body** (HTTP 200 with no content) while scanning `repomix-output-features-hooks.xml` — after the request had **stalled for ~18 minutes**.
2. The OpenRouter SDK (`@openrouter/sdk@0.13.32`) `JSON.parse`s that empty body inside its internal `matchFunc`, throwing `SyntaxError: Unexpected end of JSON input`.
3. That surfaced as an **unhandled promise rejection** originating entirely inside the SDK (`matchFunc` → `$do` → `processTicksAndRejections`), which bypassed the per-file `try/catch` in the scan loop and crashed Node with exit code 1 — aborting the entire batch even though the per-file error had already been recorded.

## Changes

- **Per-request timeout**: bound each `client.chat.send` call to 3 minutes so a stalled response can no longer hang the job (previously ~18 min).
- **Retry with backoff**: retry transient failures — including the empty-body `JSON.parse` failure that the SDK's status-code-based retry does not cover — up to 3 times. The retry delay is injectable so tests stay fast.
- **`unhandledRejection` safety net**: log and continue instead of crashing, so a stray SDK rejection can't abort the batch after the failure was already handled per file. Genuine total failures still surface via the existing `"All scans failed"` throw.

## Tests

- Added `runSecurityScan` tests: retry-then-succeed, and give-up-after-N-attempts.
- Existing failure-path tests updated to inject a zero retry delay.
- `pnpm cicheck` passes (code + content).

## Notes

No GitHub issue is associated — this was found by investigating the failing scheduled workflow directly. This PR only touches the security-scan scripts under `scripts/`; it does not change any workflow YAML or dependency manifests.

### DIFF
--- a/scripts/security-scan-lib.test.ts
+++ b/scripts/security-scan-lib.test.ts
@@ -488,6 +488,7 @@ describe("runSecurityScan", () => {
         content: "some code",
         model: "test-model",
         prompt: "analyze this",
+        retryDelayMs: 0,
       }),
     ).rejects.toThrow("No content returned from OpenRouter");
   });
@@ -507,6 +508,7 @@ describe("runSecurityScan", () => {
         content: "some code",
         model: "test-model",
         prompt: "analyze this",
+        retryDelayMs: 0,
       }),
     ).rejects.toThrow();
   });
@@ -526,6 +528,7 @@ describe("runSecurityScan", () => {
         content: "some code",
         model: "test-model",
         prompt: "analyze this",
+        retryDelayMs: 0,
       }),
     ).rejects.toThrow();
   });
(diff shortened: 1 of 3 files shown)


dyoshikawa:
Merged. This makes the scheduled Security Scan resilient to empty/stalled OpenRouter responses (per-request timeout + retry) and stops one flaky response from aborting the whole run, while still failing the job when a file cannot be scanned after retries.

PR #2223, opened 2026-07-13
refactor(features): migrate the JSON/JSONC shared-config writers onto the ownership gateway

saitota:
## Background

#2187 introduced the shared-config gateway but migrated only the three files with cross-feature entry-level conflicts; the other 17 registry-derived shared files stayed on `GATEWAY_PENDING_SHARED_FILES`, each tool class still re-spelling its own parse/merge/preserve semantics — the class of duplication behind 329c9bf5/8e69d7ec.

## Changes

- Migrate the JSON writers (zed, qwencode, augmentcode, devin, kiro — 14 classes over 7 files) onto `SHARED_CONFIG_OWNERSHIP` + `applySharedConfigPatch`. Blocks whose value depends on existing entries (zed `private_files`, the per-tool permission preserves) are recomputed in the tool class and applied as a whole owned key — the hermes `plugins` idiom.
- Migrate the JSONC writers (amp, opencode, kilo — 5 files). The gateway grows the two semantics they need: an owned key patched as `undefined` is retracted from the document (empty `amp.permissions` regeneration, filter-less `tools`), and a per-file `jsoncParseErrors` policy so amp keeps refusing to write over a file it could not fully parse.
- `sharedConfigFileKey()` resolves scope-dependent paths (`.zed/` vs `.config/zed/`) and json/jsonc twins to their declaration. The global-only `.config/kilo/kilo.json` single writer is declared as well, validated via `deriveSharedFileWriters({ minWriters: 1 })`.
- `GATEWAY_PENDING_SHARED_FILES` shrinks 17 → 5 (the TOML files, pending a TOML codec).

No change to generated output (two parse-error messages reworded).

## Test Plan

- `pnpm cicheck` green (6,937 tests).
- CLI-verified on a seeded project (amp/opencode/kilo/zed/qwencode): user-authored keys survive regeneration, reruns are idempotent, empty regeneration retracts `amp.permissions`, and a malformed `settings.jsonc` aborts with the file named and the user's file untouched.

Refs #2187


dyoshikawa:
@saitota Thank you!

PR #2251, opened 2026-07-14
feat(permissions): type Codex CLI approval/sandbox override keys as enums

dyoshikawa:
## Background

The `codexcli` permission override (`.rulesync/permissions.json` → `.codex/config.toml`) modeled its autonomy keys as bare strings:

```json
"codexcli": {
  "approval_policy": "on-request",
  "approvals_reviewer": "auto_review",
  "sandbox_mode": "workspace-write"
}
```

Because these are free-form `z.string()` values, a typo (e.g. `sandbox_mode: "read-write"`) passes validation silently and only surfaces as a Codex CLI error later. Codex actually accepts a small, fixed set of values for each, so they can be typed.

## Changes

Model the three keys with `zod` enums, sourced from the [Codex CLI config reference](https://learn.chatgpt.com/docs/config-file/config-reference):

- `approval_policy` — `untrusted` | `on-request` | `never` (the granular `{ granular = { … } }` table form is preserved via the union with `looseObject`)
- `sandbox_mode` — `read-only` | `workspace-write` | `danger-full-access`
- `approvals_reviewer` — `user` | `auto_review` | `guardian_subagent` (`guardian_subagent` is a legacy value Codex still accepts, kept so existing configs round-trip; the table form is preserved via the union)

Invalid values are now rejected at validation time and the accepted values appear as `enum` constraints in the generated `permissions-schema.json` (regenerated via `pnpm run generate:schema`), giving editor autocomplete and schema validation.

## Tests

Added two cases to `codexcli-permissions.test.ts` covering that every documented enum value (plus the granular table form) validates, and that out-of-range values are rejected. `pnpm cicheck` (code + content) passes.


dyoshikawa:
@dyoshikawa Thank you!

PR #2256, opened 2026-07-15
feat(init,gitignore): scaffold permissions.json and delimit the gitignore block

dyoshikawa:
## Summary

Three related improvements to the `init` and `gitignore` commands.

### 1. `init` scaffolds `.rulesync/permissions.json`

`rulesync init` now creates a sample `.rulesync/permissions.json` (matching the
example in the docs) alongside the already-generated `hooks.json`. `permissions`
is added to the default `features` list in the generated `rulesync.jsonc`, and
the post-init "Next steps" message now references the file. The generated sample
validates and round-trips through `rulesync generate` (verified against
`claudecode` → `.claude/settings.json`).

### 2. `gitignore` includes `**/.codex/rules/`

Codex CLI's `.codex/rules/` directory holds the `rulesync.rules` bash-permission
file produced by `createCodexcliBashRulesFile` in `codexcli-permissions.ts`. That
file is written outside `getSettablePaths`, so the directory is not derived
automatically — it is now added as a hand-maintained entry under the `codexcli` /
`permissions` tag.

### 3. `gitignore` block is delimited by start/end markers

The auto-generated block is now wrapped with explicit markers:

```
# Generated by Rulesync
...entries...
# End of Rulesync
```

Previously the block had only a start header, and removal relied on a heuristic
(known-entry matching + two-consecutive-blank-lines) to guess where the block
ended, which occasionally failed. With an explicit end marker the managed range
is removed deterministically. Legacy marker-less blocks (written before this
change) are still cleaned up via the retained heuristic fallback, so existing
`.gitignore` files migrate seamlessly on the next run.

## Testing

- `pnpm cicheck` (code + content) — all green (6970 tests)
- Added tests covering the start/end markers and deterministic block removal
- Manually ran `rulesync init` and `rulesync generate --features permissions` in
  a scratch directory to confirm end-to-end behavior


dyoshikawa:
@dyoshikawa Thank you!

PR #2267, opened 2026-07-15
feat: add codexcli base_permission_profile and deprecate sandbox_mode

dyoshikawa:
## Summary

- Adds a new `codexcli.base_permission_profile` override key (`:read-only` | `:workspace`) to `.rulesync/permissions.json`. It is emitted as the managed `[permissions.rulesync]` profile's `extends` baseline in `.codex/config.toml`:

  ```toml
  [permissions.rulesync]
  extends = ":workspace"
  ```

- Defaults to `:workspace` when unspecified. `:danger-full-access` is not accepted because Codex rejects extending it at config load time (per the [Codex permissions docs](https://learn.chatgpt.com/docs/permissions)).
- Deprecates `codexcli.sandbox_mode` / `sandbox_workspace_write`: Codex prioritizes the legacy sandbox settings over permission profiles when both are present, so authoring them disables the generated `rulesync` profile. They still emit (with a deprecation warning) so existing configs round-trip.
- Emits `approval_policy = "on-request"` and `approvals_reviewer = "auto_review"` as defaults when neither the `codexcli` override nor the existing `config.toml` sets them (existing user-set values are never clobbered).
- On import, the profile's `extends` now round-trips into `codexcli.base_permission_profile` instead of being synthesized into a workspace-wide `edit: { ".": "allow" }` rule; the previous "emit `extends` only for workspace-wide write rules" heuristic is removed since the baseline is now always explicit.

## Changes

- `src/types/permissions.ts` — new `CodexBasePermissionProfileSchema` enum; `base_permission_profile` added to the `codexcli` override; `sandbox_mode` / `sandbox_workspace_write` marked `@deprecated`
- `src/features/permissions/codexcli-permissions.ts` — extends baseline emission with `:workspace` default, deprecation warnings, `approval_policy` / `approvals_reviewer` defaults, import round-trip of `extends`
- Unit tests + e2e happy-path assertions updated for the new defaults and round-trip
- `docs/reference/file-formats.md` (and synced `skills/rulesync/file-formats.md`) updated


dyoshikawa:
@dyoshikawa Thank you!

PR #2273, opened 2026-07-16
feat: move Homebrew formula update from CI job into goal-release flow

### LINKED ISSUE #2271: Move Homebrew formula update from a separate CI PR into the draft-release skill
# Background

Today, the Homebrew tap formula (`Formula/rulesync.rb`) is updated by a dedicated `homebrew` job in `.github/workflows/publish.yml`. After the release PR merges and the publish job promotes the draft release, the job downloads the release's `SHA256SUMS`, regenerates the formula with `scripts/generate-homebrew-formula.ts`, and opens a **separate auto-merge PR** (`homebrew-formula/<tag>`) against `main`.

This works, but it has downsides:

- Every release produces an extra PR that must ride through CI and auto-merge, adding noise and another failure point (e.g. auto-merge blocked when `HOMEBREW_FORMULA_TOKEN` is unavailable, leaving a manual-merge warning).
- The workflow job carries non-trivial complexity (credential-scoping for the formula token, idempotent PR reuse, gating conditions mirroring the publish job).
- The release flow is split across two places: the `draft-release` skill drives the release PR, while the formula update happens later and out-of-band in CI.

# Details

Proposal: stop creating the Homebrew formula update as a separate PR from CI. Instead, fold the formula creation/modification into the `draft-release` skill (or command), so the AI performs the formula update as part of the release work itself — ideally landing in the release PR.

Current pieces involved:

- `.claude/skills/draft-release/SKILL.md` — drives the release: version bump, release PR, draft GitHub release.
- `.github/workflows/publish.yml` — `homebrew` job: downloads `SHA256SUMS` from the published release, runs the generator, pushes `homebrew-formula/<tag>`, opens and auto-merges the PR.
- `scripts/generate-homebrew-formula.ts` — pure generator taking `<version> <SHA256SUMS path> <output path>`; requires sha256 digests for the four platform binaries (`darwin-arm64`, `darwin-x64`, `linux-arm64`, `linux-x64`).

Key constraint to resolve: the formula embeds sha256 checksums of the release binaries, which are only produced and uploaded by `publish-assets.yml` **after** the release PR merges. At `draft-release` time those checksums do not exist yet, so the naive "just generate the formula in the release PR" does not work as-is.

# Solution / Next Steps

Possible directions (to be decided during implementation):

1. **Build binaries locally during draft-release**: have the AI (via the skill) build the four platform binaries locally (Bun supports cross-compilation targets), compute their sha256 sums, and generate the formula into the release PR. Requires `publish-assets.yml` to produce byte-identical binaries — likely fragile; would need verification or a switch to uploading the locally built binaries as the release assets.
2. **Two-phase within the skill flow**: keep asset building in CI, but move the formula update from an automated CI PR into the release skill flow — e.g. `goal-release` waits for `publish-assets` to finish, then the AI regenerates the formula and commits it directly to `main` (or amends the flow so no separate human-visible PR is needed).
3. **Hybrid**: `draft-release` writes the formula with placeholder checksums in the release PR, and a minimal CI step (or a follow-up AI step) patches in the real sha256 values after assets are published.

Next steps:

- Decide which direction fits the release pipeline (option 2 seems the most robust given the checksum ordering constraint).
- Update `.claude/skills/draft-release/SKILL.md` (and possibly `goal-release`) to include the formula work.
- Remove or slim down the `homebrew` job in `.github/workflows/publish.yml` accordingly.
- Keep `scripts/generate-homebrew-formula.ts` as the single source of formula generation logic.

dyoshikawa:
## Summary

Closes #2271

Moves the Homebrew tap formula update out of the dedicated `homebrew` CI job in `.github/workflows/publish.yml` and into the `/goal-release` command flow (direction 2 from the issue: keep asset building in CI, move the formula update into the release flow driven by the AI).

## Changes

- **`.rulesync/commands/goal-release.md`**: adds a new Step 5 "Update the Homebrew Formula". After the release PR merges, the command waits for the `Publish Assets` workflow run to succeed, downloads the release's `SHA256SUMS`, regenerates `Formula/rulesync.rb` with `scripts/generate-homebrew-formula.ts`, and — because `main` is branch-protected (the `pull_request` rule blocks direct pushes even for admins; the admin bypass is `bypass_mode: pull_request`) — lands the change via a PR that is merged immediately with `gh pr merge --admin --merge --delete-branch`. Skips cleanly when the formula is already up to date, and refuses to commit anything other than `Formula/rulesync.rb`.
- **`.rulesync/skills/draft-release/SKILL.md`**: adds a note that the formula is intentionally not updated at draft time (the checksums only exist after `Publish Assets` runs) and points manual-release users at the post-merge step.
- **`.github/workflows/publish.yml`**: removes the `homebrew` job (checksum download, formula regeneration, `homebrew-formula/<tag>` branch push, auto-merge PR, and the `HOMEBREW_FORMULA_TOKEN` credential handling), leaving a pointer comment. The `publish` job is unchanged.
- **`scripts/generate-homebrew-formula.ts`** / **`Formula/rulesync.rb`**: update the generated header comment to reference the new flow instead of the removed CI job. The generator itself stays the single source of formula generation logic.

## Why this shape

The issue's key constraint is that the formula embeds sha256 checksums of release binaries that only exist after the release PR merges and `publish-assets.yml` uploads them, so the formula cannot land in the release PR itself (direction 1 would require byte-identical local rebuilds — fragile). Direction 2 was chosen: the formula update becomes a synchronous, AI-driven step of `/goal-release`. Compared to the old CI job this removes the `HOMEBREW_FORMULA_TOKEN` dependency, the auto-merge failure mode, and the out-of-band job complexity; the remaining PR is created and merged immediately within the release flow (a direct push to `main` is not possible under the branch ruleset).

## Notes

- Generated command/skill outputs (`.claude/`, `.opencode/`, etc.) are gitignored; only the `.rulesync/` sources are tracked, and `pnpm generate` was run.
- `pnpm cicheck` passes (7059 tests, content checks green).


dyoshikawa:
Thanks! Merged as part of the automated /goal-pr flow. 🎉

Issue #2270, opened 2026-07-16
codexcli: emit a practical baseline permissions profile by default (git/.git writes, SSH agent socket, temp dirs, ~/.codex) with an opt-out flag


dyoshikawa:
## Background

The Codex CLI permissions profile that rulesync currently generates is too restrictive for day-to-day development. Without a set of baseline carve-outs, Codex CLI hits permission denials constantly (git operations, temp-dir usage, reading its own config, SSH agent access), making it effectively unusable for real work. rulesync should emit a practical baseline by default so that a generated `config.toml` works out of the box.

## Details

The following entries should be emitted by default in the generated Codex CLI `config.toml`:

```toml
[features]

[permissions.default.network.unix_sockets]
"${SSH_AUTH_SOCK}" = "allow"

[permissions.rulesync.filesystem]
":minimal" = "read"
":tmpdir" = "write"
":slash_tmp" = "write"
"${HOME}/.codex/**" = "read"
"${HOME}/.codex/auth.json" = "deny"
glob_scan_max_depth = 8

[permissions.rulesync.filesystem.":workspace_roots"]
".git/index" = "write"
".git/index.lock" = "write"
".git/objects/**" = "write"
".git/COMMIT_EDITMSG" = "write"
".git/refs/**" = "write"
".git/logs/**" = "write"
```

Rationale for each group:

- `unix_sockets` allow for `$SSH_AUTH_SOCK`: git push/fetch over SSH needs the agent socket.
- `.git/**` write carve-outs under `:workspace_roots`: commit/stage operations write to `.git/index`, `.git/objects`, refs, and logs; without these, basic git workflows are denied.
- `:minimal = "read"`: already emitted today as the fixed baseline (platform default read roots).
- `:tmpdir` / `:slash_tmp` write: many tools require a writable temp dir.
- `${HOME}/.codex/**` read with `auth.json` deny: Codex reads its own config tree, but credentials stay protected.
- `glob_scan_max_depth = 8`: matches the Codex CLI default and bounds `**` deny-glob scanning.

### Environment variable resolution

`${SSH_AUTH_SOCK}` and `${HOME}` in the entries above are placeholders: at generation time, rulesync must resolve them and write the actual values into the output (Codex CLI does not expand env vars in these keys). If a variable is unset (e.g. `SSH_AUTH_SOCK` on a machine without an SSH agent), skip the corresponding entry instead of writing a literal `${...}` key.

## Solution / Next Steps

[...]

PR #2274, opened 2026-07-16
feat(codexcli): emit .git write carve-outs by default; document env-dependent permission entries in FAQ

### LINKED ISSUE #2272: codexcli: emit .git write carve-outs by default; document env-dependent permission entries (SSH agent socket, ~/.codex, temp dirs) in FAQ
## Background

Supersedes #2270 with a revised direction.

The Codex CLI permissions profile that rulesync currently generates is too restrictive for day-to-day development: without extra carve-outs, Codex CLI hits permission denials constantly (git commit/stage operations, SSH agent access, temp-dir usage, reading its own config), making it hard to use for real work.

#2270 originally proposed emitting a full practical baseline by default behind an opt-out flag. On reflection, that scope is too broad for a built-in default: several of the entries depend on the user's environment or are workflow-specific preferences. The revised plan splits the solution in two:

1. **Implement in rulesync (default output): only the `.git` write carve-outs.** These are universal — every developer using Codex CLI in a git repository needs them, they contain nothing environment-specific, and rulesync's whole purpose is generating working configs.
2. **Document everything else in the FAQ** (`docs/faq.md`) so users can add the entries themselves. In particular, **network settings are out of rulesync's management scope by design** — users keep full editing freedom there.

### Part 1 — Implementation: emit `.git` write carve-outs by default

Add the following to the generated `[permissions.rulesync.filesystem.":workspace_roots"]` table in the Codex CLI `config.toml` output:

```toml
[permissions.rulesync.filesystem.":workspace_roots"]
".git/**" = "write"
".git/config" = "read"
```

Rationale:

- Codex CLI's own workspace-write default policy makes `.git` read-only inside workspace roots (`append_default_read_only_project_root_subpath_if_no_explicit_rule(..., ".git")` in `codex-rs`), so without an explicit write rule, basic git workflows (commit/stage writes to `.git/index`, `.git/objects`, refs, logs) are denied.
- `".git/**" = "write"` opens the whole `.git` subtree for writes (a trailing `/**` on a read/write entry means subtree access in Codex's parser).
- `".git/config" = "read"` keeps the repository config read-only as a security guard: writable `.git/config` would let a sandboxed process set keys like `core.fsmonitor` or `core.hooksPath` that lead to arbitrary code execution outside the sandbox. Codex resolves the more specific path with priority (officially documented: "A more specific path can also reopen a narrower subtree inside a broader deny", and the same specificity rule applies here), so this read rule wins over the broader write rule for that one file.

Implementation notes:

- Emit these in `src/features/permissions/codexcli-permissions.ts`, merging with (not clobbering) user-defined rules; a user-specified value for the same key takes precedence.
- Decide whether an opt-out is needed. If so, a boolean in the `codexcli` tool options of the permissions config (e.g. `codexcli.git_write_rules`, default `true`; only an explicit `false` suppresses the output). A narrowly named flag is preferable to the broad `workaround_fix` name from #2270.
- Make sure import (`rulesync import`) round-trips cleanly: default-emitted entries should not duplicate on re-generate.
- Update docs and e2e happy-path coverage per the Tool × Feature matrix guidelines.

### Part 2 — FAQ: document the user-managed entries

Add a section to `docs/faq.md` (e.g. "Codex CLI denies git pushes, SSH agent access, or temp-dir writes with a generated permissions profile") documenting the following additions users can make themselves.

**Network settings stay out of rulesync's management scope.** rulesync will not emit or manage these keys; the FAQ documents them and generation must preserve user-authored network keys untouched (verify the existing pass-through in `codexcli-permissions.ts` covers `dangerously_allow_all_unix_sockets`, which is not currently part of rulesync's `CodexNetwork` type — user-added keys must survive re-generation).

```toml
[permissions.rulesync.network]
enabled = true
# Simplest option: allow all unix sockets. Codex names this "dangerously_*"
# because it is broad, but it avoids hardcoding an env-dependent socket path.
# (Verified: `dangerously_allow_all_unix_sockets: Option<bool>` in
# codex-rs/config/src/permissions_toml.rs.)
dangerously_allow_all_unix_sockets = true

# Stricter alternative: allow only the SSH agent socket.
# Replace with the actual value of $SSH_AUTH_SOCK on your machine;
# Codex does not expand env vars in these keys.
# [permissions.rulesync.network.unix_sockets]
# "/path/to/ssh-agent.sock" = "allow"

[permissions.rulesync.filesystem]
":tmpdir" = "write"
":slash_tmp" = "write"
"~/.codex/**" = "read"
"~/.codex/auth.json" = "deny"
glob_scan_max_depth = 8
```

Explain in the FAQ entry:

- Unix socket access: git push/fetch over SSH needs the agent socket. `dangerously_allow_all_unix_sockets = true` is the simple env-independent option; a per-socket `unix_sockets` allow entry with the resolved `$SSH_AUTH_SOCK` path is the stricter one. Either way this is user territory — rulesync never writes it.
- `:tmpdir` / `:slash_tmp` write: many build tools require a writable temp dir.
- `~/.codex/**` read with `auth.json` deny: Codex reads its own config tree while credentials stay protected. Tilde paths are officially supported: the Codex permissions reference documents `~/path` as "A path under the current user's home directory" (scoped subpaths supported), and the parser expands `~` via the user's home directory (`codex-rs/utils/absolute-path/src/lib.rs`), so no manual `$HOME` resolution is needed.
- `glob_scan_max_depth = 8` matches the Codex CLI default and bounds `**` deny-glob scanning.

Note: `docs/faq.md` is currently not mirrored by `scripts/sync-skill-docs.ts` output checks in the same way as reference docs, but verify with `pnpm cicheck:content` after editing.

## Solution / Next Steps

1. Implement Part 1 (`.git/**` write + `.git/config` read emitted by default) in `codexcli-permissions.ts` with tests and docs.
2. Write the Part 2 FAQ entry in `docs/faq.md`, and verify user-authored network keys survive re-generation.
3. #2270 is closed as superseded by this issue.

dyoshikawa:
## Summary

Closes #2272

Implements both parts of the issue: the universal `.git` write carve-outs are now emitted by default in the generated Codex CLI permissions profile, and the environment-dependent entries (SSH agent socket, temp dirs, `~/.codex`) are documented in the FAQ for users to add themselves — with network settings explicitly kept out of rulesync's management scope.

## Part 1 — Default `.git` write carve-outs

The generated `[permissions.rulesync.filesystem.":workspace_roots"]` table now always receives:

```toml
".git/**" = "write"
".git/config" = "read"
```

- Codex's `:workspace` baseline keeps `.git` read-only inside workspace roots, denying basic git workflows (commit/stage writes to `.git/index`, `.git/objects`, refs, logs). The `".git/**"` write rule reopens the subtree.
- `".git/config"` stays read-only as a security guard: a writable `.git/config` would let a sandboxed process set `core.fsmonitor` / `core.hooksPath` and execute code outside the sandbox. Codex resolves the more specific path with priority, so the read rule wins for that one file.
- **User rules win per key** — the defaults are filled with `??=` after the user's canonical rules are mapped.
- **Opt-out**: new `codexcli.git_write_rules` boolean override (default `true`; only an explicit `false` suppresses). Like `base_permission_profile` it is consumed by the profile builder and never written as a top-level `config.toml` key. (Named per the issue's preference for a narrow name over #2270's `workaround_fix`.)
- **Clean import round-trip**: like `:minimal`, the exact default pattern/value pairs are not imported into the rulesync model (they would leak Codex-specific rules into the shared canonical model and other tools' outputs, and are re-added on every export). Customized `.git` values (e.g. `".git/config" = "write"`) import normally.

## Part 2 — FAQ + user network key preservation

- New FAQ entry in `docs/faq.md` ("Codex CLI denies SSH agent access, temp-dir writes, or reading its own config…") documenting the user-managed additions: `network.enabled = true` with `dangerously_allow_all_unix_sockets = true` (simple option) or a per-socket `unix_sockets` allow with the resolved `$SSH_AUTH_SOCK` path (stricter; Codex does not expand env vars), `":tmpdir"`/`":slash_tmp"` write, `~/.codex/**` read + `~/.codex/auth.json` deny (tilde is expanded by Codex itself), and `glob_scan_max_depth = 8`.
- **Preservation verified and hardened**: unknown network keys such as `dangerously_allow_all_unix_sockets` were already preserved verbatim by `preserveUnmanagedProfileKeys` (now covered by a test). `network.enabled`, however, was a fully managed key and a user-authored `enabled = true` was deleted on regeneration whenever rulesync computed no network table — breaking the FAQ's recommended setup. It is now preserved (with a warning) when the freshly computed profile does not set it; rulesync still sets `enabled = true` itself when the canonical model has an allow domain.
- `docs/reference/file-formats.md` updated accordingly (carve-out semantics, `git_write_rules` override key, network-preservation nuance) and mirrored to `skills/rulesync/` via `sync-skill-docs`.

## Tests

- 8 new unit tests: default emission, per-key user precedence, opt-out, no top-level `git_write_rules` key, import skip of default-valued entries, import of customized values, import→regenerate round-trip without duplication, and preservation of `dangerously_allow_all_unix_sockets` + user-authored `enabled` (including alongside deny-only managed domains).
- e2e Tool×Feature happy-path extended: the codexcli permissions generate case now asserts the default carve-outs (ran `vitest --config vitest.e2e.config.ts src/e2e/e2e-permissions.spec.ts` locally — 60 passed).
- `pnpm cicheck` passes.


dyoshikawa:
Thanks! Merged as part of the automated /goal-pr flow. 🎉

PR #2286, opened 2026-07-16
feat: fall back to a stable hash name for non-representable Codex MCP server names

### LINKED ISSUE #2219: Follow-ups from PR #2217: Codex MCP server-name normalization edge cases
## Background

PR #2217 ("fix: normalize Codex MCP server names", resolving #2215) added `normalizeCodexMcpServerName` in `src/features/mcp/codexcli-mcp.ts`. It normalizes Codex MCP server names to match `^[a-zA-Z0-9_-]+$` by lowercasing, replacing `[^a-z0-9]+` with `_`, trimming leading/trailing `_`, and skipping names that normalize to empty or that collide with prototype-pollution keys.

The PR was reviewed and merged. The security review found no issues, and no findings were merge-blockers (all mid/low). This issue tracks the non-blocking follow-ups for later.

## Details

**1. [mid] Non-ASCII-only server names are silently dropped**

`.replace(/[^a-z0-9]+/g, "_")` removes all non-ASCII characters. A name made entirely of non-ASCII characters (e.g. Japanese `日本語サーバー`, Cyrillic, emoji) normalizes to an empty string, returns `null`, and the whole server is skipped with a warning. Since this project's primary audience is Japanese users, a Japanese-named MCP server silently disappearing from Codex output is impactful.

Lossless conversion is impossible because Codex requires ASCII-only names, but a full discard is the harshest possible outcome. Consider a fallback instead — e.g. append a stable hash suffix (`server_<hash>`) so the entry survives — plus a dedicated test and a warning message guiding the user to rename manually.

**2. [low] generate→import round-trip is non-reversible**

`convertToCodexFormat` normalizes names, but `convertFromCodexFormat` does not restore the original name. Importing back from the generated Codex config overwrites the readable original name (e.g. `Postgres MCP - Production - Read Only`) with the normalized `postgres_mcp_production_read_only`. This is inherent to lossy normalization; at minimum document in the docs that importing from Codex flows the normalized name back into the source.

**3. [low] Collision last-wins warning interacts confusingly with `removeEmptyEntries`**

In `fromRulesyncMcp`, `convertToCodexFormat` resolves collisions last-wins, then `removeEmptyEntries` runs afterward. If the winning (last-processed) server's config is empty, the winner is dropped, so both colliding servers disappear — yet the "Only the last processed server will be used" warning still fires, which is inconsistent with the actual behavior.

**4. [low] Test coverage gap for the empty-string normalization path**

The empty-string guard (pure-symbol or non-ASCII-only names like `"!!!"`) is not directly tested. The added skip test uses `"constructor!"`, which exercises the post-normalization prototype-pollution check rather than the empty-string guard. Add a case like `"!!!"`.

**5. [low] Docs not updated**

`docs/reference/file-formats.md` (Codex MCP section) does not document the new auto-normalization, the collision last-wins behavior, or the skip-on-unnormalizable behavior. Per `feature-change-guidelines`, these user-visible behavior changes should be documented.

## Solution / Next Steps

- Decide on a fallback strategy for names that normalize to empty (finding 1) — a hash-suffixed name is the most user-friendly option — and add coverage.
- Add the missing test case for the empty-string path (finding 4).
- Reconcile the collision warning with `removeEmptyEntries` so the message matches the actual outcome (finding 3).
- Update `docs/reference/file-formats.md` to document the normalization, collision, and skip behaviors, and note the non-reversible round-trip (findings 2 and 5).

dyoshikawa:
## Summary

Follow-ups from PR #2217's Codex MCP server-name normalization, resolving all findings recorded in the issue.

## Changes

- **F1 (fallback strategy — the design decision the issue was held open for):** a server name with no ASCII-representable characters at all (e.g. a fully Japanese name like `日本語サーバー`) or that normalizes to a prototype-pollution key now falls back to a **stable sha256-derived name** (`mcp_<hash8>`) instead of the whole server being silently dropped. The issue itself identified the hash-suffix fallback as the most user-friendly option, and it matters for this project's primary Japanese audience. The warning tells the user which fallback name was used and to rename the source server for a readable Codex name.
- **F3:** the collision warning is reworded from "Only the last processed server will be used" to an overwrite statement (`"Y" (processed last) overwrites "X"`), which stays accurate even when `removeEmptyEntries` later drops the surviving entry — that path fires its own dropped-entry warning.
- **F4:** tests now cover the empty-normalization paths directly (fully non-ASCII name, pure-symbol `!!!`) plus the pollution-key path (`constructor!`), asserting the stable fallback names and the warning.
- **F2 + F5:** `docs/reference/file-formats.md` gains a Codex CLI server-name note documenting the auto-normalization rules, collision last-wins behavior, hash fallback, and the one-way (non-reversible) round-trip (synced to `skills/rulesync/`).

Closes #2219

### DIFF
--- a/docs/reference/file-formats.md
+++ b/docs/reference/file-formats.md
@@ -696,6 +696,8 @@ You can control which individual tools from an MCP server are enabled or disable
 
 > **Qwen Code note:** MCP servers are written to the `mcpServers` key of `.qwen/settings.json` (project) / `~/.qwen/settings.json` (global, via `--global`). Qwen supports stdio (`command`/`args`), SSE (`url`), and HTTP (`httpUrl`) transports. Rulesync maps the canonical per-server `enabledTools` ⇄ Qwen's `includeTools` (allowlist) and `disabledTools` ⇄ Qwen's `excludeTools` (denylist). Other top-level keys in `settings.json` are preserved on round-trip.
 
+> **Codex CLI server-name note:** Codex requires MCP server names matching `[a-zA-Z0-9_-]+`, so Rulesync auto-normalizes non-conforming names on generate (lowercase, runs of other characters become `_`, leading/trailing `_` trimmed) — e.g. `Postgres MCP - Production - Read Only` becomes `postgres_mcp_production_read_only`. If two names normalize to the same Codex name, the last processed server overwrites the earlier one (with a warning). A name with no representable characters at all (e.g. a fully Japanese name) falls back to a stable hash-derived name like `mcp_1a2b3c4d` instead of being dropped; rename the server in `.rulesync/mcp.json` to pick a readable Codex name. This normalization is one-way: importing back from the generated `config.toml` yields the normalized name, not the original.
+
 ### Codex-specific: pass shell env vars to MCP servers (`envVars`)
 
 Codex CLI supports a per-server array of shell env var names to inherit when launching the MCP server process. The source schema uses `envVars` (camelCase, matching the project convention used by sibling fields like `enabledTools`/`disabledTools`); the codex generator renames it to `env_vars` (snake_case) for codex's native `config.toml` format.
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2279, opened 2026-07-16
Codex CLI: drop the `".git/config" = "read"` carve-out and keep only `".git/**" = "write"`


dyoshikawa:
## Background

rulesync currently emits two default `.git` carve-outs for Codex CLI under `:workspace_roots` (opt-out via `codexcli.git_write_rules: false`):

- `".git/**" = "write"` — reopens the `.git` subtree, which Codex's `:workspace` baseline keeps read-only (commit/stage operations write to `.git/index`, `.git/objects`, refs, and logs).
- `".git/config" = "read"` — keeps the repository config read-only as a security guard, since a writable `.git/config` would let a sandboxed process set keys like `core.fsmonitor` or `core.hooksPath` that execute code outside the sandbox.

The security guard was documented as deliberately partial from the start (`.git/hooks/` remains writable anyway), see `docs/reference/file-formats.md` and the FAQ (`docs/faq.md`).

## Details

In practice, it turned out that git is not really usable without write access to `.git/config`. Many everyday git commands write to it, for example:

- `git remote add` / `git remote set-url`
- `git push -u` / `git branch --set-upstream-to` (records `branch.<name>.remote`/`merge`)
- `git config` itself (local scope)
- `git clone` / `git submodule` flows

So the `".git/config" = "read"` override blocks basic workflows, while the protection it adds is already acknowledged as partial (hooks stay writable). The cost/benefit does not hold up.

Making it worse, the default is effectively always emitted and cannot be suppressed by a broader user rule: `applyDefaultGitWriteRules` fills the defaults with a per-exact-key `??=`, so authoring `".git/**": "allow"` under `write` does NOT prevent the `".git/config" = "read"` default from being injected — and since Codex gives the more specific path priority, `.git/config` stays read-only even though the user explicitly granted write on the whole `.git` subtree. The only ways to get rid of it are an exact-key `".git/config"` rule, `codexcli.git_write_rules: false` (which also drops the `.git/**` write rule), the `:read-only` baseline, or a direct `":workspace_roots"` string rule. Deleting the default resolves this gotcha as well.

Additionally, the FAQ's recommended `codexcli.permission` example (the "Filesystem entries: author them in `.rulesync/permissions.json`" section in `docs/faq.md`) should also carry explicit write rules for the working tree and tool directories, so the generated profile includes:

```toml
"./**" = "write"
".git/**" = "write"
".agents/**" = "write"
".codex/**" = "write"
```

[...]

PR #2290, opened 2026-07-16
docs: add :root write to the Codex CLI FAQ permissions example

### LINKED ISSUE #2285: FAQ: add `":root" = "write"` to the Codex CLI permissions example — `npx {package}` is unusable without it
## Background

The FAQ section "Codex CLI denies SSH agent access, temp-dir writes, or reading its own config with a generated permissions profile" (`docs/faq.md`) recommends authoring extra filesystem entries in `.rulesync/permissions.json` under the tool-scoped `codexcli.permission` block. The current example covers `:tmpdir` / `:slash_tmp` writes and `~/.codex/**` read with an `auth.json` deny.

## Details

In practice, `npx {package} ...` (and similar package-runner commands) does not work properly under the generated profile unless write access beyond the workspace and temp dirs is granted. `npx` downloads and unpacks packages into the npm cache under the home directory (`~/.npm/_npx`), and other everyday tools likewise write to home-directory caches (`~/.cache`, `~/.local`, corepack/pnpm stores, etc.), which the `:workspace` baseline keeps read-only.

The fix that proved necessary is Codex's special path `:root` (the filesystem root) with `write` access:

```jsonc
"codexcli": {
  "permission": {
    "write": { ":root": "allow", ":tmpdir": "allow", ":slash_tmp": "allow" }
  }
}
```

which generates `":root" = "write"` in the profile's filesystem table (`:root` starts with `:`, so it lands directly under `[permissions.rulesync.filesystem]`, not under `:workspace_roots`).

## Solution / Next Steps

1. Add `":root": "allow"` to the `write` block of the FAQ's `codexcli.permission` example, and extend the following "This generates into the profile as ..." sentence with `":root" = "write"`.
2. Add a bullet to the "What each entry does" list: **`:root` write** — package runners such as `npx` unpack into the npm cache under the home directory (`~/.npm/_npx`), and many dev tools write to home-directory caches; the `:workspace` baseline denies these writes, which breaks the commands outright.
3. Note the trade-off in the same bullet: `":root" = "write"` effectively opens filesystem writes system-wide, so readers who want tighter scoping can grant narrower home-directory patterns instead (e.g. `"~/.npm/**"`, `"~/.cache/**"`) at the cost of chasing each tool's cache path.
4. `skills/rulesync/faq.md` is synchronized by `scripts/sync-skill-docs.ts`, so regenerate rather than hand-editing.

Related: #2279 also extends the same FAQ example (workspace-tree write rules such as `"./**"`, `".git/**"`, `".agents/**"`, `".codex/**"`); if both land, update the example once with the combined rule set.

dyoshikawa:
## Summary

The FAQ section on Codex CLI permission denials recommends authoring extra filesystem entries in the `codexcli.permission` block, but the example only covered temp-dir writes and `~/.codex` reads. In practice `npx {package}` (and other package runners) is unusable under the generated `:workspace`-based profile because it unpacks into home-directory caches (`~/.npm/_npx`, `~/.cache`, corepack/pnpm stores) that the baseline keeps read-only.

## Changes

- Added `":root": "allow"` to the `write` block of the FAQ example and to the "generates into the profile as …" sentence (`":root" = "write"`).
- Added a **`:root` write** bullet to the "What each entry does" list explaining why package runners break without it, plus the trade-off (system-wide writes) and the narrower per-cache-path alternative (`"~/.npm/**"`, `"~/.cache/**"`).
- `skills/rulesync/faq.md` regenerated via `scripts/sync-skill-docs.ts`.

Note: #2279 proposes further edits to the same example (workspace-tree write rules); it is not included here, so the example can be updated again when that lands.

Closes #2285

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -69,18 +69,19 @@ dangerously_allow_all_unix_sockets = true
   },
   "codexcli": {
     "permission": {
-      "write": { ":tmpdir": "allow", ":slash_tmp": "allow" },
+      "write": { ":root": "allow", ":tmpdir": "allow", ":slash_tmp": "allow" },
       "read": { "~/.codex/**": "allow", "~/.codex/auth.json": "deny" },
     },
   },
 }
 ```
 
-This generates into the profile as `":tmpdir" = "write"`, `":slash_tmp" = "write"`, `"~/.codex/**" = "read"`, and `"~/.codex/auth.json" = "deny"`, and round-trips through `rulesync import`. Note that a tool-scoped category replaces the shared one wholesale for Codex CLI: if your shared `permission` block already has `read`/`write` rules that should also apply to Codex CLI, repeat them inside `codexcli.permission`.
+This generates into the profile as `":root" = "write"`, `":tmpdir" = "write"`, `":slash_tmp" = "write"`, `"~/.codex/**" = "read"`, and `"~/.codex/auth.json" = "deny"`, and round-trips through `rulesync import`. Note that a tool-scoped category replaces the shared one wholesale for Codex CLI: if your shared `permission` block already has `read`/`write` rules that should also apply to Codex CLI, repeat them inside `codexcli.permission`.
 
 What each entry does:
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2291, opened 2026-07-16
fix: merge per-path Codex filesystem rules across read/edit/write categories

### LINKED ISSUE #2283: Codex CLI: merge per-path read/write rules — `read: allow` + `write: deny` should emit `"{path}" = "read"`
## Background

Codex CLI's filesystem permission model is a single access level per path (`read` / `write` / `deny`), while rulesync's canonical model has separate `read`, `edit`, and `write` categories. When the same path pattern appears in more than one canonical category, the two models need a semantic merge — but the current generator does not perform one.

## Details

In `convertRulesyncToCodexProfile` (`src/features/permissions/codexcli-permissions.ts`, ~line 332), the `read`/`edit`/`write` categories are iterated in object order and each rule is written into the filesystem table with a plain assignment. For the same path key, the last category processed silently overwrites the earlier one.

Concretely, with:

```jsonc
{
  "permission": {
    "read": { "{path}": "allow" },
    "write": { "{path}": "deny" }
  }
}
```

`read: allow` maps to `"read"` and `write: deny` maps to `"deny"`; whichever category happens to come last in the JSON wins. In the typical order above the final emitted rule is `"{path}" = "deny"` — blocking reads the user explicitly allowed. The intended combination "readable but not writable" is exactly what Codex's `"read"` access level expresses.

## Solution / Next Steps

1. Merge rules per path instead of last-writer-wins: when the same pattern has `read: allow` and `write` (and/or `edit`) `deny`, emit `"{path}" = "read"`.
2. Define the full merge matrix explicitly, e.g.:
   - read allow + write allow → `"write"`
   - read allow + write deny → `"read"` (this issue's case)
   - read deny + anything → `"deny"` (deny read implies nothing is accessible; a contradictory `write: allow` should warn)
   - single-category rules keep the current mapping (`mapReadAction` / `mapWriteAction`, ~line 951).
3. Decide how `edit` interacts with `write` when both are present for the same path (they currently share `mapWriteAction`).
4. Check the import side (`convertCodexProfileToRulesync` / `addRulesyncFilesystemRule`) so the merged output round-trips: `"{path}" = "read"` already imports as `read: allow`, which loses the explicit `write: deny` — consider whether importing `"read"` should also emit `edit`/`write: deny` for symmetry, or document the asymmetry.
5. Add unit tests for the merge matrix in `codexcli-permissions.test.ts`, plus an e2e round-trip case.

dyoshikawa:
## Summary

Codex CLI models a single filesystem access level per path (`deny` < `read` < `write`), while rulesync's canonical model has separate `read`/`edit`/`write` categories. `convertRulesyncToCodexProfile` iterated the categories independently and wrote each rule with a plain assignment, so when the same pattern appeared in multiple categories the last-processed category silently overwrote the earlier one — `read: allow` + `write: deny` emitted `"{path}" = "deny"`, blocking reads the user explicitly allowed, when the intended "readable but not writable" is exactly what Codex's `"read"` level expresses.

## Changes

- New `mergeFilesystemCategoryRules` collects the three categories and emits one access level per pattern, order-independently:
  - `edit`/`write` collapse onto Codex's write side; when both carry the same pattern, the more restrictive action wins (`deny` > `ask` > `allow`).
  - `read: allow` + write-side `allow` → `"write"`
  - `read: allow` + write-side non-allow → `"read"` (the issue's headline case)
  - `read` non-allow → `"deny"`; a contradictory write-side `allow` ("writable but not readable" is not expressible in Codex) emits a warning.
  - Single-category patterns keep the existing `mapReadAction`/`mapWriteAction` mappings — no behavior change for configs without cross-category overlaps.
- **Round-trip asymmetry documented** (issue step 4): `"{path}" = "read"` imports back as `read: allow` only; the write-side deny is implied by Codex's access level and deliberately not re-materialized (synthesizing `write: deny` on import would pollute round-trips of plain single-category read rules). Recorded in the merge function's doc comment.
- Unit tests cover the merge matrix (read+write allow/deny combinations, contradiction warning, edit-vs-write restrictive collapse).

Closes #2283

### DIFF
--- a/src/features/permissions/codexcli-permissions.test.ts
+++ b/src/features/permissions/codexcli-permissions.test.ts
@@ -76,6 +76,7 @@ describe("CodexcliPermissions", () => {
             "/data/readable/**": "allow",
             "/data/full/**": "allow",
             "/data/blocked/**": "deny",
+            "/data/asked/**": "allow",
           },
           write: {
             // read allow + write deny → "read" (this was the last-wins bug).
@@ -84,6 +85,8 @@ describe("CodexcliPermissions", () => {
             "/data/full/**": "allow",
             // read deny + write allow → contradiction, warn + "deny".
             "/data/blocked/**": "allow",
+            // read allow + write ask → "read" (ask maps to the deny side).
+            "/data/asked/**": "ask",
           },
         },
       }),
@@ -99,6 +102,7 @@ describe("CodexcliPermissions", () => {
     expect(fileContent).toContain('"/data/readable/**" = "read"');
     expect(fileContent).toContain('"/data/full/**" = "write"');
     expect(fileContent).toContain('"/data/blocked/**" = "deny"');
+    expect(fileContent).toContain('"/data/asked/**" = "read"');
     expect(logger.warn).toHaveBeenCalledWith(
       expect.stringContaining('Codex CLI cannot express "writable but not readable"'),
     );
--- a/src/features/permissions/codexcli-permissions.test.ts
+++ b/src/features/permissions/codexcli-permissions.test.ts
@@ -64,6 +64,71 @@ describe("CodexcliPermissions", () => {
     expect(fileContent).toContain('"example.com" = "deny"');
   });
 
+  it("should merge per-path rules across read/edit/write categories instead of last-category-wins", async () => {
+    const logger = createMockLogger();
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2292, opened 2026-07-16
feat: support base_permission_profile :danger-full-access via default_permissions

### LINKED ISSUE #2282: Codex CLI: support `base_permission_profile: ":danger-full-access"` via `default_permissions` instead of `extends`
## Background

The `codexcli.base_permission_profile` override controls which built-in Codex profile the managed `[permissions.rulesync]` profile extends. Codex ships three built-in profiles (`:read-only`, `:workspace`, `:danger-full-access`), but its config loader rejects `extends = ":danger-full-access"`, so rulesync currently accepts only the two extendable baselines (`CODEX_BASE_PERMISSION_PROFILES` in `src/types/permissions.ts`, ~line 549). As a result there is no way to tell rulesync "run Codex with full access" through the canonical permission model.

## Details

Codex cannot *extend* `:danger-full-access`, but it can *select* it directly: `default_permissions = ":danger-full-access"` at the top level of `.codex/config.toml` is valid. So the restriction is only about the `extends` mechanism, not about using the profile at all.

Proposal: accept `":danger-full-access"` as a `base_permission_profile` value, and when it is set:

- emit `default_permissions = ":danger-full-access"` (instead of `default_permissions = "rulesync"`), and
- do **not** emit the `[permissions]` / `[permissions.rulesync]` tables at all — with full access there is no sandbox for filesystem/network rules to refine, and an extends-based managed profile cannot be built anyway.

## Solution / Next Steps

1. `src/types/permissions.ts` — add `":danger-full-access"` to the accepted `base_permission_profile` values. Keep the extendable-baseline set (`:read-only`, `:workspace`) separate, since `CODEX_EXTENDABLE_BASELINES` in `src/features/permissions/codexcli-permissions.ts` still governs what `extends` may reference.
2. `src/features/permissions/codexcli-permissions.ts` — when the override is `":danger-full-access"`, skip `convertRulesyncToCodexProfile` output entirely: write `default_permissions = ":danger-full-access"` and omit the `[permissions.rulesync]` profile (currently `default_permissions` is hardcoded to the `rulesync` profile name around line 204). The default `.git` carve-outs are moot in this mode. Warn (or document) that canonical `read`/`edit`/`write`/`webfetch` rules are not representable and will be ignored for Codex CLI while this baseline is selected.
3. Import side — a `config.toml` with `default_permissions = ":danger-full-access"` and no rulesync profile should round-trip into `codexcli.base_permission_profile: ":danger-full-access"` (the current import derives the override from the profile's `extends`, ~line 242, which will not exist in this mode).
4. Tests: `src/features/permissions/codexcli-permissions.test.ts` and the e2e permissions spec (generate + import round-trip for the new value).
5. Docs: update the `base_permission_profile` sections in `docs/reference/file-formats.md` (and the FAQ if it mentions the two-value restriction); `skills/rulesync/` is regenerated by `scripts/sync-skill-docs.ts`.

dyoshikawa:
## Summary

Codex CLI ships three built-in permission profiles, but its config loader rejects `extends = ":danger-full-access"`, so rulesync accepted only the two extendable baselines — leaving no way to express "run Codex with full access" through the canonical model. Codex CAN select the built-in directly, though: `default_permissions = ":danger-full-access"` is valid at the top level (verified against the [Codex permissions docs](https://learn.chatgpt.com/docs/permissions): "It must match a profile under `[permissions]` or a built-in", while `extends` "cannot extend `:danger-full-access`").

## Changes

- `codexcli.base_permission_profile` now accepts `":danger-full-access"`. The extendable-baseline set stays `[":read-only", ":workspace"]` via the new `CODEX_EXTENDABLE_BASELINE_PROFILES` constant (still the single source for the import-side `extends` check).
- When `:danger-full-access` is selected, generation emits `default_permissions = ":danger-full-access"` and **skips the managed `[permissions.rulesync]` profile entirely** — with the sandbox removed there is nothing for filesystem/network rules to refine, so canonical `read`/`edit`/`write`/`webfetch` rules are ignored for Codex CLI with a warning. A stale managed profile from a previous generate is pruned; hand-written sibling profiles are preserved. The other `codexcli` override keys (`approval_policy` etc.) keep flowing through `computeCodexcliOverridePatch` unchanged.
- **Import**: a `config.toml` with `default_permissions = ":danger-full-access"` (no managed profile) round-trips into `codexcli.base_permission_profile: ":danger-full-access"` from the top-level key, complementing the existing `extends`-based derivation.
- Tests cover generation (managed-profile skip + prune, sibling preservation, ignored-rules warning) and the import round-trip; the enum-rejection test is updated since `:danger-full-access` is now a valid override value.
- Docs: the `base_permission_profile` sections in `docs/reference/file-formats.md` describe the direct-selection mode (synced to `skills/rulesync/`).

Closes #2282


dyoshikawa:
@dyoshikawa Thank you!

PR #2294, opened 2026-07-17
fix: drop the Codex CLI .git/config read carve-out and keep only .git/** write

### LINKED ISSUE #2279: Codex CLI: drop the `".git/config" = "read"` carve-out and keep only `".git/**" = "write"`

dyoshikawa:
## Summary

Drops the default `".git/config" = "read"` carve-out that rulesync emitted into the Codex CLI managed profile's `:workspace_roots` table, keeping only `".git/**" = "write"`.

In practice git is not usable without write access to `.git/config`: everyday commands such as `git remote add`/`set-url`, `git push -u` (records `branch.<name>.remote`/`merge`), local-scope `git config`, and clone/submodule flows all write to it. Meanwhile the protection the read guard added was already partial by design (`.git/hooks/` stays writable, a maintainer decision on #2272), so the cost/benefit did not hold. Making it worse, the default was injected per exact key (`??=`), so even a user-authored `".git/**": "allow"` write rule could not suppress it — Codex resolves the more specific path with priority, keeping `.git/config` read-only against the user's explicit intent.

Users who want stricter isolation can still author their own override (e.g. `read: { ".git/config": "allow" }`) in the canonical permissions; the `codexcli.git_write_rules: false` opt-out stays as is.

## Changes

- `src/features/permissions/codexcli-permissions.ts`: remove `".git/config": "read"` from `CODEX_GIT_WRITE_RULES` and update the rationale comments (generate and import both key off the same constant, so the import-skip logic needed no code change).
- `src/types/permissions.ts`: update the `git_write_rules` doc comment.
- Tests: `codexcli-permissions.test.ts` (default emission, import of legacy `".git/config" = "read"` as a user rule, round-trip) and `e2e-permissions.spec.ts`.
- `docs/faq.md`: rewrite the carve-out paragraph, and extend the `codexcli.permission` example with the practical workspace write set (`"."`, `".git/**"`, `".agents/**"`, `".codex/**"`) plus an explanatory bullet — a tool-scoped category replaces the shared one wholesale, and Codex's `:workspace` baseline keeps `.git`, `.agents`, and `.codex` read-only inside workspace roots ([codex-rs `FileSystemSandboxPolicy::workspace_write`](https://github.com/openai/codex)).
- `docs/reference/file-formats.md`: rewrite the carve-out paragraphs (no longer claims `.git/config` stays read-only).
- `skills/rulesync/*`: regenerated via `scripts/sync-skill-docs.ts`.

Note: the issue proposed `"./**" = "write"` for the workspace-subtree entry, but Codex rejects `:workspace_roots` subpath keys containing `.` or `..` components ([Codex permissions reference](https://developers.openai.com/codex/permissions) — "Other subpaths must be relative descendants and cannot contain `.` or `..` components"; the documented base form is `"."`), so the example uses `".": "allow"` instead.

Closes #2279


dyoshikawa:
@dyoshikawa Thank you!

PR #2311, opened 2026-07-17
docs: recommend :minimal = write alongside :root = write in Codex CLI FAQ

dyoshikawa:
## Summary

The Codex CLI FAQ entry recommended adding `":root": "allow"` to the canonical write rules so package runners and home-directory caches work, but that alone is not enough: rulesync always emits `":minimal" = "read"` as a fixed baseline, and Codex treats that entry as a write-narrowing rule — the policy no longer qualifies for full disk write access (`FileSystemSandboxPolicy::has_full_disk_write_access`), so writes that `:root` appears to allow can still be denied.

This PR updates the FAQ to:

- Add `":minimal": "allow"` to the example write block alongside `":root": "allow"`.
- Explain in the `:root` bullet why `:minimal` must be raised to write together with `:root`.
- Document the import caveat: `:minimal` is never imported regardless of its value (it is rulesync's fixed baseline), so after `rulesync import` the write rule must be re-authored or the next generate drops back to `":minimal" = "read"`.

`skills/rulesync/faq.md` is regenerated by `scripts/sync-skill-docs.ts`.


dyoshikawa:
@dyoshikawa Thank you!

PR #1108, opened 2026-02-18
feat(commands): support subdirectory import and generate

saitota:
This PR addresses #1079.

## Problem
Some tools (for example, Claude Code) support command files in subdirectories such as `.claude/commands/pj/test.md`, but `rulesync import` previously only loaded top-level command files.

At the same time, not all target tools support subdirectory commands. So we need per-tool behavior control for both import and generate.

## Solution
Add a `supportsSubdirectory` flag to `ToolCommandFactory.meta` and use it as the single policy switch in `CommandsProcessor`.

|                    | `supportsSubdirectory: true`                         | `supportsSubdirectory: false`                    |
| ------------------ | ---------------------------------------------------- | ----------------------------------------------- |
| Import             | Recursively load command files (`**/*.md` / `**/*`) | Load only top-level command files (`*.md`)      |
| Generate           | Preserve nested path (e.g. `pj/test.md`)            | Flatten to filename (e.g. `pj/test.md` -> `test.md`) |
| Collision handling | N/A (path is preserved)                              | Warn when multiple nested paths map to one filename (later one wins) |

### Set to `true`
| Tool | Rationale |
| ---- | --------- |
| `claudecode` | Official docs explicitly describe nested slash commands ([Anthropic Docs](https://docs.anthropic.com/en/docs/claude-code/slash-commands)) |
| `geminicli` | Official docs describe subdirectory namespace behavior ([Gemini CLI Docs](https://googlegemini.wiki/gemini-cli/custom-commands/)) |
| `roo` | Official docs describe recursive loading for slash commands ([Roo Code Docs](https://docs.roocode.com/features/slash-commands)) |
| `opencode` | Source behavior indicates recursive command globbing ([OpenCode repository](https://github.com/opencode-ai/opencode)) |

### Set to `false`
| Tool | Rationale |
| ---- | --------- |
| `codexcli` | Official docs describe top-level command usage ([Codex CLI Docs](https://openai.github.io/codex-cli/)) |
| `factorydroid` | Official CLI docs describe flat command structure ([Factory Docs](https://docs.factory.ai/reference/cli-reference)) |
| `kilo` | Docs describe filename-based command usage, no subdirectory behavior documented ([Kilo Docs](https://kilo.ai/docs/features/slash-commands/workflows)) |
| `cursor` | Current docs/examples are flat; no subdirectory command behavior documented ([Cursor Docs](https://docs.cursor.com/)) |
[...]

PR #2314, opened 2026-07-17
feat(commands): add flattenedCommandNaming option for collision-free flattened command names

Millon15:
## Problem

For tools with `supportsSubdirectory: false` (Cursor, Antigravity IDE, Codex CLI, …), nested commands are flattened to their basename (introduced in #1108). When two nested commands share a basename, they map to the same output file and the last one processed silently wins (with a warning since the collision-warn was added).

This is easy to hit in real repos that namespace commands by category. Our project hit it with 12 of 45 nested commands, e.g.:

- `plan/run.md`, `qa/run.md`, `ralphex/run.md` → all `run.md`
- `update/docs.md`, `update/confluence/docs.md`, `update/google/docs.md`, `update/slack/docs.md` → all `docs.md`

The warning helps, but there is currently no way to get *all* nested commands generated for flat-namespace tools — some are always lost.

## Solution

Add an opt-in, config-file-only option `flattenedCommandNaming`:

| Value | Behavior |
| --- | --- |
| `"basename"` (default) | Current behavior, unchanged: `pj/test.md` → `test.md` |
| `"path"` | Join directory segments into the filename: `pj/test.md` → `pj-test.md`, `update/confluence/docs.md` → `update-confluence-docs.md` |

- Tools with `supportsSubdirectory: true` (Claude Code, OpenCode, Roo, …) are unaffected — nested paths are preserved as before.
- The existing collision warning still fires if joined names ever collide (e.g. a literal `pj-test.md` next to `pj/test.md`).
- Config-file-only (like `sources`), no CLI flag — it is a project-level generation policy, mirroring how `ConfigResolverResolveParams` already excludes `sources`.

## Changes

- `src/types/features.ts` — `FlattenedCommandNamingSchema` (`"basename" | "path"`)
- `src/config/config.ts`, `src/config/config-resolver.ts` — schema field, defaults, local-config merge, getter
- `src/features/commands/commands-processor.ts` — naming strategy in `flattenRelativeFilePath`
- `src/lib/generate.ts` — pass the option through
- `docs/guide/configuration.md` + `skills/rulesync/configuration.md` — option documented (kept in sync)
- Tests: 3 new cases in `commands-processor.test.ts` (path naming, no-collision assertion, subdirectory tools unaffected); Config mocks extended in the two generate test suites

## Test Plan

- `pnpm vitest run src/features/commands/commands-processor.test.ts` — 47/47
- `pnpm test` — 323 files, 7270 tests passed
- `pnpm run check` — oxfmt, oxlint, tsgo all clean; `pnpm cspell` clean
- Real-world check: applied the same change as a patch to rulesync 12.0.0 in our monorepo — 45 nested commands now generate 59 unique files per flat-namespace target with zero collisions, `generate --check` idempotent

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -50,6 +50,14 @@ Example:
   "simulateSubagents": false, // Generate simulated subagents
   "simulateSkills": false, // Generate simulated skills
 
+  // Naming for command files flattened for tools without subdirectory
+  // command support (e.g. Cursor): "basename" (default) keeps only the
+  // filename, so `pj/test.md` and `ops/test.md` collide and the last one
+  // wins; "path" joins the directory segments into the filename
+  // (`pj/test.md` -> `pj-test.md`), keeping flattened names unique.
+  // Tools that support subdirectories (e.g. Claude Code) are unaffected.
+  "flattenedCommandNaming": "basename",
+
   // When true (default), `rulesync gitignore` only emits entries for the
   // tools listed in `targets`. Set to false to emit entries for all supported
   // tools regardless of `targets`.
--- a/skills/rulesync/configuration.md
+++ b/skills/rulesync/configuration.md
@@ -50,6 +50,14 @@ Example:
   "simulateSubagents": false, // Generate simulated subagents
   "simulateSkills": false, // Generate simulated skills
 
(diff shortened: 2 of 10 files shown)


dyoshikawa:
@Millon15 Thank you!

PR #2316, opened 2026-07-19
fix: honor config-file silent and verbose in logger configuration

### LINKED ISSUE #2312: Config-file `silent: true` is ignored: logger only honors CLI flags, plus warn paths that bypass silent entirely
# `silent: true` in the config file is completely ignored — logger is configured from CLI flags only

## Background

A user sets `silent: true` in their rulesync config file (documented in `docs/guide/configuration.md` as "Silent mode - suppress all output (except errors)"), yet `rulesync generate` still prints all of its normal output:

```
Permissions feature manages 'Read' tool and will overwrite 1 existing Read deny entries. Permissions take precedence.
Codex CLI permissions support only read/edit/write/webfetch categories. Skipping: bash
Preserving existing "network.enabled" from config. Review this value manually as it may enable network access beyond the Rulesync-managed domain rules.
Preserving unmanaged keys in the "rulesync" permissions profile: network.dangerously_allow_all_unix_sockets. Review them manually; rulesync carries them forward verbatim.
Written 6 rules
    AGENTS.md
    backend/AGENTS.md
    ...
🎉 All done! Written 6 file(s) total (6 rules)
```

These messages are a mix of `logger.warn`, `logger.info`, and `logger.success` calls. `ConsoleLogger` itself honors silent correctly (`src/utils/logger.ts` — `isSuppressed()` gates info/success/warn/debug), so the problem is not the logger implementation but how (and from what source) it gets configured.

### Root cause: config-file `silent`/`verbose` never reach the logger

- `wrapCommand` (`src/cli/wrap-command.ts:53-56`) configures the logger **only from CLI flags**:

  ```ts
  logger.configure({
    verbose: Boolean(globalOpts.verbose) || Boolean(options.verbose),
    silent: Boolean(globalOpts.silent) || Boolean(options.silent),
  });
  ```

  This happens before config resolution, and the logger is never re-configured afterward.
- The config file value is parsed and resolved correctly (`src/config/config.ts:102`, `src/config/config-resolver.ts:331` — `pick({ cli: silent, file: configByFile.silent, ... })`) and exposed as `Config.getSilent()` (`src/config/config.ts:616`), but **`getSilent()` has zero call sites in the entire codebase**. So `silent: true` in `rulesync.jsonc` is a documented no-op; only `--silent` works.
- `verbose` has the same wiring gap, partially masked: `Config.getVerbose()` IS consulted in `src/lib/convert.ts` and `src/lib/import.ts` to gate a few extra messages, but the logger's own `verbose` state (which gates `logger.debug`) still comes from CLI flags only. So `verbose: true` in the config file enables some extra messages but not `logger.debug` output — inconsistent either way.

### Other spots that bypass silent even when `--silent` IS passed (requested sweep)

While investigating, I swept for similar output paths that escape the configured logger entirely:

1. **Unconfigured module-level loggers** — these are `new ConsoleLogger()` with defaults (`silent: false`) and are used as fallbacks when no logger is threaded through, so their warnings ignore both config-file and CLI silent:
   - `src/features/permissions/qwencode-permissions.ts:43` (used via `params.logger ?? moduleLogger` at line 439)
   - `src/features/permissions/augmentcode-permissions.ts:28` (passed as `logger: moduleLogger` at line 503)
2. **`warnWithFallback(undefined, ...)`** falls through to raw `console.warn`, bypassing silent:
   - `src/features/mcp/codexcli-mcp.ts:125` (explicit `undefined`)
   - Any other `warnWithFallback` call site where the optional `logger` param is not threaded through by the caller (`src/config/config-resolver.ts:183`, `src/features/mcp/goose-mcp.ts:256`, `src/features/mcp/grokcli-mcp.ts:158,199`, etc.) — worth auditing which call chains actually drop the logger.
3. **`BaseLogger.onConflictingFlags`** (`src/utils/logger.ts:61`) uses raw `console.warn` — arguably fine since it fires exactly when `--silent` is involved in a flag conflict, but it also prints in silent mode.

## Solution / Next Steps

1. **Wire resolved config into the logger.** After `ConfigResolver` produces the final config in each command handler (generate, import, etc.), re-configure the logger:

   ```ts
   logger.configure({ verbose: config.getVerbose(), silent: config.getSilent() });
   ```

   CLI flags must still win over the config file — the resolver already implements that precedence (`pick({ cli, file, fallback })`), so re-configuring from the resolved config is sufficient. Alternatively, centralize this in one place (e.g. right after config resolution inside each command) rather than per-call-site.
2. **Fix the escape hatches** so `silent` is respected end to end:
   - Thread the real logger into the `qwencode` / `augmentcode` permission paths instead of falling back to unconfigured module loggers (or configure the fallbacks from the resolved config).
   - Audit `warnWithFallback` call chains and thread the logger through where it is currently dropped (`codexcli-mcp.ts:125` and friends).
3. **Add regression tests**: a test asserting that with `silent: true` in the config file (no CLI flag), `rulesync generate` emits nothing on stdout/stderr except errors; and unit tests for the previously-bypassing warn paths.
4. Decide whether `verbose: true` in the config file should also enable `logger.debug` output (it should, for consistency) and cover it in the same wiring.

cm-dyoshikawa:
## Summary

Fixes the documented no-op reported in #2312: `silent: true` in `rulesync.jsonc` was parsed and resolved correctly, but the logger was configured from CLI flags only (`wrapCommand`), and `Config.getSilent()` had zero call sites — so only `--silent` worked. `verbose` had the same wiring gap for `logger.debug`.

## Changes

- **Centralized re-configuration**: `ConfigResolver.resolve` now re-configures the supplied logger from the resolved `verbose`/`silent` as soon as the base and local config files are merged — before the resolver's own warnings fire — so config-file settings apply end to end. CLI flags still win through the existing `pick` (CLI > local > base > default) precedence. Re-configuration only happens when the caller threads a logger through, since those call sites also pass their CLI flags (keeping precedence intact).
- **Logger threading**: `convert`, `install`, and `gitignore` now pass their logger (and, for gitignore, the CLI `verbose`/`silent` flags) into `ConfigResolver.resolve`; `generate` and `import` already did.
- **Escape hatches closed**: added a shared `fallbackLogger` in `src/utils/logger.ts` that `wrapCommand` configures from CLI flags and `ConfigResolver.resolve` keeps in sync with the resolved config. `warnWithFallback(undefined, ...)` now routes through it instead of raw `console.warn`, and the unconfigured module-level `ConsoleLogger` fallbacks in the qwencode/augmentcode permission translators now alias it — so those warn paths honor `silent` too (issue items 1 and 2).
- **`verbose: true` in the config file now enables `logger.debug`** (issue item 4).

## Tests

- `ConfigResolver` re-configuration: config-file values applied, CLI flags winning, config-file `verbose` enabling debug, and no fallback mutation when no logger is supplied.
- `warnWithFallback` routing and fallback silent suppression.
- Updated existing mocks/spies that assumed the old one-argument `resolve` calls or raw `console.warn` fallback.

Closes #2312

### DIFF
--- a/src/cli/commands/convert.test.ts
+++ b/src/cli/commands/convert.test.ts
@@ -86,6 +86,7 @@ describe("convertCommand", () => {
 
       expect(ConfigResolver.resolve).toHaveBeenCalledWith(
         expect.objectContaining({ targets: ["cursor", "claudecode"] }),
+        { logger: mockLogger },
       );
     });
   });
@@ -101,6 +102,7 @@ describe("convertCommand", () => {
           targets: ["cursor", "claudecode"],
           features: ["*"],
         }),
+        { logger: mockLogger },
       );
 
       expect(RulesProcessor).toHaveBeenCalledWith(
@@ -125,6 +127,7 @@ describe("convertCommand", () => {
         expect.objectContaining({
           features: ["rules", "mcp"],
         }),
+        { logger: mockLogger },
       );
     });
 
@@ -172,6 +175,7 @@ describe("convertCommand", () => {
 
       expect(ConfigResolver.resolve).toHaveBeenCalledWith(
         expect.objectContaining({ global: true }),
+        { logger: mockLogger },
       );
     });
   });
--- a/src/cli/commands/convert.ts
+++ b/src/cli/commands/convert.ts
@@ -44,11 +44,14 @@ export async function convertCommand(logger: Logger, options: ConvertOptions): P
   // Pass both source and destinations as `targets` so per-target feature maps
   // in `rulesync.jsonc` are honored for every tool involved. Default features
   // to `*` so every feature that both tools support is attempted.
-  const config = await ConfigResolver.resolve({
-    ...options,
-    targets: [fromTool, ...toTools],
-    features: options.features ?? ["*"],
-  });
+  const config = await ConfigResolver.resolve(
+    {
+      ...options,
+      targets: [fromTool, ...toTools],
+      features: options.features ?? ["*"],
+    },
+    { logger },
+  );
 
   const isPreview = config.isPreviewMode();
   const modePrefix = isPreview ? "[DRY RUN] " : "";
--- a/src/cli/commands/gitignore.ts
+++ b/src/cli/commands/gitignore.ts
@@ -166,6 +166,8 @@ const extractRulesyncManagedEntries = (content: string): string[] => {
 export type GitignoreCommandOptions = {
   readonly targets?: string[];
   readonly features?: RulesyncFeatures;
+  readonly verbose?: boolean;
+  readonly silent?: boolean;
 };
 
 const groupEntriesByDestination = ({
@@ -211,7 +213,10 @@ export const gitignoreCommand = async (
 ): Promise<void> => {
   const gitignorePath = join(process.cwd(), ".gitignore");
   const gitattributesPath = join(process.cwd(), ".gitattributes");
-  const config = await ConfigResolver.resolve({});
+  const config = await ConfigResolver.resolve(
+    { verbose: options?.verbose, silent: options?.silent },
+    { logger },
+  );
 
   const resolvedEntries = resolveGitignoreEntries({
     targets: options?.targets,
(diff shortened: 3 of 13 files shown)


cm-dyoshikawa:
Status: iteration-2 review round is clean (zero mid-or-above findings; the three iteration-1 mid findings around the conflicting-flags warning were fixed in 9fa06cb9) and all 9 CI checks are green. Merge is blocked for this session's credentials (cm-dyoshikawa has write access only; the main ruleset requires approval from someone other than the last pusher, and self-approval is not possible). Ready for a one-click merge by the maintainer.

dyoshikawa:
@cm-dyoshikawa Thank you!

PR #2341, opened 2026-07-22
Preserve Factory Droid absolute and bare hook commands

### LINKED ISSUE #2327: Factory Droid hooks: absolute and bare-executable commands are corrupted by the $FACTORY_PROJECT_DIR prefix
## Background

Found while reviewing PR #2320 (Factory Droid hooks file shape). That PR's premise turned out to be wrong and it was not merged, but verifying its claims surfaced a genuine pre-existing defect on `main` in the shared hooks converter. Issue #2317 happened to use an absolute hook command (`/tmp/notify.sh`) in its repro, which is exactly the input that triggers this bug.

`applyCommandPrefix` in `src/features/hooks/tool-hooks-converter.ts` (lines 116-137) prefixes a hook command with the tool's project-directory variable. It decides whether to prefix using only two conditions: the command does not start with `$`, and — when `prefixDotRelativeCommandsOnly` is set — the command starts with `.`. There is no check for an absolute path, and no check for a bare executable name.

`FACTORYDROID_CONVERTER_CONFIG` (`src/features/hooks/factorydroid-hooks.ts:33-39`) sets `projectDirVar: "$FACTORY_PROJECT_DIR"` but does **not** set `prefixDotRelativeCommandsOnly`, so every command that does not start with `$` gets prefixed. Claude Code is unaffected because `CLAUDECODE_CONVERTER_CONFIG` sets `prefixDotRelativeCommandsOnly: true` (`src/features/hooks/claudecode-hooks.ts:50-51`); every other tool has an empty `projectDirVar` and is unaffected.

## Details

For the `factorydroid` target, two classes of perfectly valid hook commands are corrupted:

1. **Absolute paths.** `/tmp/notify.sh` becomes `"$FACTORY_PROJECT_DIR"//tmp/notify.sh` — the prefix is glued on and the leading slash produces a double slash. The hook points at a path inside the project that does not exist.
2. **Bare executables.** `npx prettier --write .` becomes `"$FACTORY_PROJECT_DIR"/npx prettier --write .` — rulesync tries to execute `npx` from the project root instead of resolving it on `PATH`. The inline comment on `prefixDotRelativeCommandsOnly` (`tool-hooks-converter.ts:54`) explicitly says bare executables like `npx prettier ...` should be left intact, so this is a stated intent that the Factory Droid config does not honor.

Only dot-relative commands (`./scripts/foo.sh`, `.rulesync/hooks/notify.sh`) are handled correctly today, which is why the existing tests all pass — every Factory Droid hooks test uses a relative command.

There is also a round-trip concern: `toRulesyncHooks` strips the prefix at `tool-hooks-converter.ts:352-362`, so a corrupted `"$FACTORY_PROJECT_DIR"//tmp/notify.sh` imports back as `/tmp/notify.sh`. The generate → import cycle looks clean while the generated file on disk is broken, which makes the bug easy to miss.

## Solution / Next Steps

- Add an absolute-path guard to `applyCommandPrefix` so a command starting with `/` (and a Windows drive-letter or UNC path) is never prefixed, regardless of `prefixDotRelativeCommandsOnly`. This is unconditionally correct for every tool — an absolute path is already fully resolved.
- Decide whether `FACTORYDROID_CONVERTER_CONFIG` should set `prefixDotRelativeCommandsOnly: true` to match Claude Code and the documented intent about bare executables. Confirm against the Factory hooks reference whether Droid resolves bare commands on `PATH` before changing this, since it alters generated output for existing users.
- Add regression tests covering an absolute command and a bare-executable command for the `factorydroid` target, plus a generate → import round-trip assertion that the on-disk command is unchanged (not just that the canonical form survives).

dyoshikawa:
## Summary

- never prefix absolute hook commands across POSIX and Windows path forms
- limit Factory Droid project-directory prefixing to dot-relative commands
- add on-disk and import round-trip coverage for absolute and bare commands

## Verification

- pnpm cicheck

Closes #2327

### DIFF
--- a/src/features/hooks/factorydroid-hooks.test.ts
+++ b/src/features/hooks/factorydroid-hooks.test.ts
@@ -7,6 +7,7 @@ import { setupTestDirectory } from "../../test-utils/test-directories.js";
 import { ensureDir, writeFileContent } from "../../utils/file.js";
 import { FactorydroidHooks } from "./factorydroid-hooks.js";
 import { RulesyncHooks } from "./rulesync-hooks.js";
+import { canonicalToToolHooks, type ToolHooksConverterConfig } from "./tool-hooks-converter.js";
 
 describe("FactorydroidHooks", () => {
   let testDir: string;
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2349, opened 2026-07-23
feat: support rules in declarative sources

### LINKED ISSUE #1598: feat: support rules filtering in declarative sources
# feat: support `rules` filtering in declarative sources

## Summary

Declarative sources (`sources` in `rulesync.jsonc`) currently support fetching and filtering skills from remote repositories via the `skills` field. It would be useful to have analogous support for rules — allowing users to selectively install rules from external sources alongside skills.

### Proposed config surface

```jsonc
{
  "sources": [
    // Fetch specific rules and skills from a source
    {
      "source": "owner/repo",
      "skills": ["skill-creator"],
      "rules": ["typescript-conventions", "testing-guidelines"]
    },
    // Fetch all rules from a source
    { "source": "owner/repo", "rules": ["*"] }
  ]
}
```

## Motivation

Teams often maintain shared rule repositories, but not every rule applies to every team. Today `rulesync fetch` can download rules from a remote repo, but it's all-or-nothing — there's no way to select a subset.

In practice this creates a problem for organizations where multiple teams aligned to a business domain serve different technical functions (e.g., iOS, Android, backend), but share a common rules repo. Many rules are generally applicable, but some are platform- or stack-specific. Without selective filtering, teams face two options:

1. **Fetch everything** — teams end up with irrelevant rules in their AI context (iOS rules in a backend repo, etc.)
2. **Reorganize the source repo into subdirectories** — this is fragile over time as teams' software stacks converge or diverge, making any fixed division stale

Selective rule filtering via declarative sources solves this cleanly: the shared repo keeps rules in a flat, topic-oriented structure, and each consuming team declares exactly which rules they need in their `rulesync.jsonc`. Teams can adjust their selections as their stacks evolve without reorganizing the source.

## AI Assessed Implementation challenges

Based on a review of the current sources pipeline, there are several non-trivial areas to address:

### 1. Skills are directories, rules are files

This is the fundamental structural mismatch. Skills are fetched as directories (`.rulesync/skills/.curated/{name}/` containing `SKILL.md` + supporting files), while rules are single `.md` files in `.rulesync/rules/`. The existing fetch pipeline (`src/lib/sources.ts`) assumes directory-based content — listing directories, recursively fetching all files per directory, and computing integrity hashes over directory contents.

Relevant code:
- `sources.ts` groups fetched files by first path component (~line 623-629), assuming each skill is a subdirectory
- `computeSkillIntegrity` in `sources-lock.ts` hashes directory contents
- Transport layer (`fetchSkillFiles`, `listDirectoryRecursive`) is oriented around recursive directory listing

### 2. Lockfile schema is skills-only

`LockedSource` in `sources-lock.ts` hardcodes a `skills` record:

```ts
skills: z.record(z.object({ integrity: z.string() }))
```

Adding rules would require either:
- A new `rules` record alongside `skills` in the locked source (preferred — additive, backwards compatible)
- A `lockfileVersion` bump if the shape changes in a breaking way

### 3. No "curated" concept for rules

Skills have a clear local-vs-curated precedence model (`.rulesync/skills/` wins over `.rulesync/skills/.curated/`). Rules have no equivalent — all rules currently live flat in `.rulesync/rules/`. This needs a design decision:

- **Option A**: Add `.rulesync/rules/.curated/` mirroring the skills model. Local rules with the same name take precedence.
- **Option B**: Write fetched rules directly into `.rulesync/rules/` with a naming convention or metadata to distinguish them.

Option A seems cleaner and consistent with the skills precedent.

### 4. Source path conventions

Skills assume a `skills/` directory in the source repo by default. What should the default path be for rules? Likely `rules/`, but the `path` field in `SourceEntry` currently points to a single directory. If a source provides both skills and rules, the fetch needs to read from two different directories in the same repo (e.g., `skills/` and `rules/`).

Possible approaches:
- Separate `skillsPath` / `rulesPath` fields (explicit but verbose)
- Convention-based: always look in `skills/` and `rules/` relative to the source `path`
- Current `path` applies to skills; add a `rulesPath` override

### 5. RulesProcessor integration

`src/features/rules/rules-processor.ts` currently assumes all rules are locally authored. It would need to discover and load curated rules from the new `.curated/` directory, applying the same precedence logic (local wins over curated, first-declared source wins for duplicates).

### 6. Gitignore and cleanup

`.rulesync/rules/.curated/` would need to be added to the gitignore entries generated by `rulesync gitignore` (`src/cli/commands/gitignore.ts`), and `cleanPreviousCuratedSkills`-equivalent logic would be needed for rules.

## Suggested approach

1. Add `rules` field to `SourceEntrySchema` (mirrors `skills`)
2. Create `.rulesync/rules/.curated/{rule-name}.md` output path
3. Extend lockfile with a `rules` record per source (additive, no version bump needed)
4. Generalize the fetch pipeline to handle file-based content (rules) alongside directory-based content (skills)
5. Update `RulesProcessor` to discover curated rules with local-wins precedence
6. Update gitignore generation and cleanup logic

## Related

- Declarative sources were introduced for skills — this extends the same model to rules
- The `fetch` command already supports `--features rules,skills` for one-off fetches; `install` would gain parity

dyoshikawa:
## Summary

- add selective `rules` and `rulesPath` support to declarative GitHub, git, and npm sources
- stage remote rules under `.rulesync/rules/.curated/` with local-first and first-source precedence
- extend source lockfiles, frozen installs, source-wide transactional updates, stale artifact cleanup, and generated gitignore entries
- verify cached rule integrity, check out exact locked git SHAs, and reject unsafe curated symlink trees
- document rule source configuration and reject rule-only fields in `--mode gh`

Closes #1598

## Test plan

- `pnpm cicheck` (329 test files, 7,420 tests)


dyoshikawa:
@dyoshikawa Thank you!

PR #2351, opened 2026-07-23
feat: add Rulesync feature scaffolding

### LINKED ISSUE #2348: Add feature file scaffolding to the rulesync add command
## Background

Rulesync currently creates a complete set of sample files through `rulesync init`, but there is no focused command for adding a single Rulesync feature file later. Users must remember the expected path, filename, frontmatter, and initial structure for each feature and create the file manually.

The existing `rulesync add <source>` command installs a declarative rule or skill source, so feature scaffolding must be introduced without breaking that workflow.

## Details

Extend `rulesync add` so it can create an initial template for a requested Rulesync feature. For example:

```bash
rulesync add rule --name overview
rulesync add rule --name overview.md
```

Both forms should create the normalized rule path `.rulesync/rules/overview.md`. Multi-file features should accept a name and create the file in the canonical Rulesync location. This should cover applicable feature types such as rules, commands, subagents, skills, and checks, with the skill layout handled as `.rulesync/skills/<name>/SKILL.md`. Generated content should contain valid minimal frontmatter or configuration and provide a useful editable starting point.

MCP, hooks, ignore, and permissions are single-file Rulesync features. They should not accept arbitrary names and should target their canonical file paths. If the target file already exists, the command must ask the user whether it may overwrite the file with the initial template. Declining the prompt must leave the existing file unchanged. Non-interactive execution must fail safely instead of overwriting silently unless an explicit overwrite option is designed and supplied. The same collision policy should be defined for named multi-file features.

Feature keywords and source identifiers must be parsed unambiguously so existing commands such as `rulesync add anthropics/skills --skills skill-creator` remain backward compatible.

## Solution / Next Steps

- Define the supported singular feature keywords and CLI syntax for feature scaffolding while preserving `rulesync add <source>`.
- Extract or centralize the sample templates currently embedded in initialization so `rulesync init` and `rulesync add` cannot drift.
- Resolve output paths through the feature classes and Rulesync path constants rather than duplicating path literals.
- Normalize names and extensions, validate names to prevent path traversal, and create parent directories as needed.
- Add an interactive confirmation before replacing an existing singleton file with its initial template; preserve the file when confirmation is declined.
- Define safe non-interactive and explicit-overwrite behavior.
- Add colocated command tests for each feature type, optional `.md` normalization, skill directory layout, existing-file confirmation and refusal, invalid names, non-interactive behavior, and compatibility with declarative source installation.
- Document the new syntax and examples in the CLI reference.

dyoshikawa:
## Summary

- extend `rulesync add` with named scaffolds for rules, commands, subagents, skills, and checks
- add singleton scaffolds for MCP, hooks, ignore, and permissions with safe overwrite prompting and `--force`
- share scaffold templates with `rulesync init` and document source-keyword disambiguation
- cover every feature path, collision behavior, unsafe names, source compatibility, and CLI execution

Closes #2348

## Test plan

- `pnpm cicheck` (329 test files, 7,440 tests)
- `pnpm exec vitest run --config vitest.e2e.config.ts src/e2e/e2e-init.spec.ts --silent=false`

### DIFF
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -156,7 +156,36 @@ rulesync gitignore --targets copilot --features rules,commands
 
 ## Add Command
 
-The `add` command appends one source to `rulesync.jsonc` and immediately runs the declarative source resolver. It preserves JSONC comments, installs selected rules into `.rulesync/rules/.curated/`, installs selected skills into `.rulesync/skills/.curated/`, and updates `rulesync.lock` or `rulesync-npm.lock.json`.
+The `add` command can scaffold one Rulesync feature file or append one declarative source to `rulesync.jsonc`.
+
+### Feature scaffolding
+
+Use a feature keyword to create a valid, editable starter file:
+
+```bash
+# Named Markdown features
+rulesync add rule --name overview
+rulesync add command --name review-pr.md
+rulesync add subagent --name planner
+rulesync add skill --name project-context
+rulesync add check --name security
+
+# Singleton features
+rulesync add mcp
+rulesync add hooks
+rulesync add ignore
+rulesync add permissions
+```
+
+Named features accept a name with or without the `.md` suffix. Skills use the directory layout `.rulesync/skills/<name>/SKILL.md`; the other named features create `<name>.md` in their canonical Rulesync directory. Names cannot contain path separators.
+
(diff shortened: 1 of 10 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2357, opened 2026-07-23
fix(rules): preserve global non-root rules for folded targets

### LINKED ISSUE #2358: Global non-root rules are ignored for folded targets
## Summary

Global non-root rules are discarded for targets that fold them into a root file.

## Motivation / Purpose

Codex and Pi declare `foldsNonRootIntoRoot`, but global loading removes their
non-root rules before conversion can fold or route them.

## Details

- A global root plus non-root Codex or Pi rule emits only the root body.
- Pi rules using `systemPrompt: append` do not reach `APPEND_SYSTEM.md`.
- Global loading should retain non-root rules when the target either exposes a
  global `nonRoot` path or declares `foldsNonRootIntoRoot`.

## Additional Context

Draft fix: #2357

wongcken:
## Summary

Preserve global non-root rules for targets that fold them into their root rule file.

## Problem

In global mode, non-root rules are retained only when a target exposes a physical
`nonRoot` output path. Folded targets such as Codex and Pi intentionally expose no
such path, so their non-root rules are discarded before `foldsNonRootIntoRoot` is
evaluated.

This also prevents Pi rules using `systemPrompt: append` from reaching
`APPEND_SYSTEM.md`.

## Reproduction

Generate a root and non-root rule for Codex or Pi in global mode. Only the root body
is emitted; the non-root rule is reported as ignored. The same inputs are folded
correctly in project mode.

## Fix

Retain global non-root rules when the target either exposes a global `nonRoot` path
or supports global scope and declares `foldsNonRootIntoRoot`. Junie global conversion
now maps both root and non-root inputs to `.junie/AGENTS.md` so the processor can fold
them consistently.

## Tests

- Added a derived processor regression matrix covering all seven global-capable
  folded targets.
- Added direct Junie global non-root conversion coverage.
- Added E2E coverage for global Codex, Junie, and Pi folding.
- Added E2E coverage for global Pi `APPEND_SYSTEM.md` routing.
- `pnpm run cicheck`

Closes #2358

### DIFF
--- a/src/features/rules/rules-processor.ts
+++ b/src/features/rules/rules-processor.ts
@@ -1399,10 +1399,12 @@ As this project's AI coding tool, you must follow the additional conventions bel
       );
     }
 
-    // In global mode, return root rule + non-root rules if the target supports global nonRoot
+    // In global mode, retain non-root rules when the target can emit or fold them globally
     if (this.global) {
       const globalPaths = factory.class.getSettablePaths({ global: true });
-      const supportsGlobalNonRoot = "nonRoot" in globalPaths && globalPaths.nonRoot !== null;
+      const supportsGlobalNonRoot =
+        ("nonRoot" in globalPaths && globalPaths.nonRoot !== null) ||
+        factory.meta.foldsNonRootIntoRoot === true;
 
       const nonRootRules = rulesyncRules.filter(
         (rule) =>
--- a/src/features/rules/rules-processor.ts
+++ b/src/features/rules/rules-processor.ts
@@ -1404,7 +1404,7 @@ As this project's AI coding tool, you must follow the additional conventions bel
       const globalPaths = factory.class.getSettablePaths({ global: true });
       const supportsGlobalNonRoot =
         ("nonRoot" in globalPaths && globalPaths.nonRoot !== null) ||
-        factory.meta.foldsNonRootIntoRoot === true;
+        (factory.meta.supportsGlobal && factory.meta.foldsNonRootIntoRoot === true);
 
       const nonRootRules = rulesyncRules.filter(
         (rule) =>
--- a/src/e2e/e2e-rules.spec.ts
+++ b/src/e2e/e2e-rules.spec.ts
@@ -902,6 +902,94 @@ globs: ["src/**/*"]
     expect(generatedContent).not.toContain("Non-Root Rule Content");
   });
 
+  it.each([
+    { target: "codexcli", outputPath: join(".codex", "AGENTS.md") },
+    { target: "pi", outputPath: join(".pi", "agent", "AGENTS.md") },
+  ] as const)(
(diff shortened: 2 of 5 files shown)


src/features/rules/rules-processor.ts
```diff
@@ -1386,10 +1386,12 @@ As this project's AI coding tool, you must follow the additional conventions bel
...
     }
 
-    // In global mode, return root rule + non-root rules if the target supports global nonRoot
+    // In global mode, retain non-root rules when the target can emit or fold them globally
     if (this.global) {
       const globalPaths = factory.class.getSettablePaths({ global: true });
-      const supportsGlobalNonRoot = "nonRoot" in globalPaths && globalPaths.nonRoot !== null;
+      const supportsGlobalNonRoot =
+        ("nonRoot" in globalPaths && globalPaths.nonRoot !== null) ||
+        factory.meta.foldsNonRootIntoRoot === true;
```
dyoshikawa: This condition also becomes true for Junie (`supportsGlobal: true`, `foldsNonRootIntoRoot: true`), but `JunieRule.fromRulesyncRule({ global: true })` explicitly throws for every non-root rule. As a result, global Junie generation with a root plus detail rule changes from generating the root and warning about the ignored detail to failing the entire conversion before the fold runs. Please update the Junie global mapping so both rules target `.junie/AGENTS.md` and can be folded, with a regression test for Junie (preferably a table covering every global-capable folded target), or use explicit metadata that only opts in targets whose global conversion accepts non-root rules.

dyoshikawa (commented):
This introduces a merge-blocking regression for Junie global generation. The security review found no actionable issues and all five CI jobs are green, but the current test matrix does not exercise this affected folded target. Please make Junie global conversion fold non-root rules into `.junie/AGENTS.md` (or narrow this condition with explicit global-fold support metadata) and add Junie, ideally every global-capable folded target, to the regression matrix before merging.

src/features/rules/rules-processor.ts
wongcken: Good catch — confirmed and fixed in 5c99641e.

- Junie global conversion now maps both root and non-root inputs to `.junie/AGENTS.md`, preserving `root: false` for fold candidates.
- The processor fold clause now also requires `supportsGlobal`, so project-only folded targets such as Hermes Agent and Warp keep their previous behavior.
- The processor regression matrix is derived from the global target list and explicitly asserts all seven global-capable folded targets: Codex CLI, Deep Agents, Goose, Grok CLI, Junie, Pi, and Reasonix.
- Added a Junie converter regression, Junie E2E coverage, and a root-only global negative case.

`pnpm run cicheck` passes (7,465 tests), as do the focused Codex/Junie/Pi E2E tests. Refreshed CI is running now.

dyoshikawa:
@wongcken Thank you!

Issue #2342, opened 2026-07-22
Complete Hermes Agent feature support


rudironsoni:
## Summary

Complete the `hermesagent` target's feature support and model `.rulesync/commands` separately from installed skills.

Hermes exposes installed skills as dynamic slash commands, but that does not make Rulesync commands and skills equivalent. `.rulesync/skills/*/SKILL.md` should continue to install as Hermes skills, while `.rulesync/commands/*.md` should become explicit user-facing command entry points backed by a Rulesync-managed Hermes plugin.

## Motivation/Purpose

Rulesync currently reports several Hermes features as unsupported:

```text
Target 'hermesagent' does not support the feature 'ignore'. Skipping.
Target 'hermesagent' does not support the feature 'commands'. Skipping.
Target 'hermesagent' does not support the feature 'checks'. Skipping.
```

The existing target also needs consistent scope diagnostics for global-only and project-only features. Hermes's plugin API provides native extension points for commands, tool-call interception, tool-result transformation, and verification hooks, so these features can be represented without collapsing commands into skills.

## Details

- Generate global `.rulesync/commands/*.md` as native Hermes slash commands through a Rulesync-managed plugin.
- Preserve `.rulesync/skills/` as Hermes skills, including their existing dynamic slash-command behavior.
- Reject command/skill slash-name collisions rather than silently shadowing one surface with the other.
- Generate project-scoped ignore support through a plugin that blocks ignored file reads and writes and filters search results.
- Generate project-scoped checks through Hermes's verification hook, preserving check description, severity, tool guidance, and body.
- Keep Rulesync-owned plugin cleanup gated by explicit ownership markers.
- Complete the existing Hermes hooks and permissions adapters so they load source files through the normal processor contract.
- Emit scope-aware warnings and retry hints for project-only versus global-only features.
- Update supported-tool documentation and preserve Tool x Feature matrix coverage in E2E tests.

Relevant Hermes documentation:

- https://hermes-agent.nousresearch.com/docs/user-guide/plugins/
- https://hermes-agent.nousresearch.com/docs/user-guide/skills/
- https://hermes-agent.nousresearch.com/docs/user-guide/hooks/

## Additional Context

This follows up #2018 and the later Hermes feature work. It is not a good first issue because it spans feature processors, generated Python plugins, shared config ownership, cleanup semantics, documentation, and end-to-end matrix coverage.

dyoshikawa:
## Upstream update (re-check on 2026-07-23)

[...]

PR #2385, opened 2026-07-24
feat(hermesagent): complete v0.19 configuration support

rudironsoni:
## Summary

- activate generated Hermes project plugins by merging `plugins.enabled` into
  `$HERMES_HOME/config.yaml` and setting
  `HERMES_ENABLE_PROJECT_PLUGINS=true` in `$HERMES_HOME/.env`
- preserve Hermes v0.19 native hook events, MCP OAuth/lifecycle fields, skill
  metadata, and native permission structures during generation and import
- make global command plugins resolve payloads relative to their installation,
  so custom `HERMES_HOME` profiles work
- fix project and global Hermes subagent import source paths and RuleSync
  destinations
- include Hermes activation writes in generation counts, dry-run output,
  `--check`, JSON output, and MCP results
- document the Hermes Agent v0.19.0 (`v2026.7.20`) compatibility contract

Project activation preserves unrelated user configuration and fails on an
explicit `plugins.disabled` conflict. Native-only Hermes hooks remain isolated
under `hermesagent.hooks` on import.

Refs #2342 and the v0.19 follow-up analysis in
https://github.com/dyoshikawa/rulesync/issues/2342#issuecomment-5060207185.

## Runtime validation

Generated project and global fixtures were loaded through a clean checkout of
Hermes Agent v0.19.0 at commit
`3ef6bbd201263d354fd83ec55b3c306ded2eb72a`. The upstream loaders accepted:

- `.hermes.md` project context
- project and global `rulesync-*` plugins, including registered commands
- all configured native hook events against Hermes `VALID_HOOKS`
- stdio and HTTP MCP servers with OAuth and lifecycle fields
- native permission, approval, and website-block policies
- Hermes-native skill metadata

Both project and global RuleSync import round-trips passed from clean
directories. Project/global generation and `--check` also passed with isolated
`HOME_DIR`/`HERMES_HOME` values.

## Repository validation

- `pnpm run check`
- `pnpm run cicheck:content`
- `pnpm run docs:build`
- `pnpm run cspell`
- `git diff --check`
- Hermes-focused unit tests: 50 passed
- Hermes command/subagent E2E suites: passed
- Hermes rules, ignore, checks, MCP, skills, hooks, and permissions E2E suites:
  passed
- full unit run: 7,529 passed; the unrelated Amp Bun Shell test could not run
  because `bun` is absent from `PATH` (`spawn bun ENOENT`)
- full E2E run: the existing convert dry-run assertion still fails because it
  expects `[DRY RUN]` on stdout but receives an empty string; all Hermes suites
  pass


src/features/shared/hermes-project-plugin-activation.ts
```diff
@@ -0,0 +1,186 @@
...
+  const configPath = join(homeDirectory, HERMESAGENT_CONFIG_FILE_PATH);
+  const dotenvPath = join(homeDirectory, HERMESAGENT_DOTENV_FILE_PATH);
+  const existingConfig = (await readFileContentOrNull(configPath)) ?? "";
+  const existingDotenv = (await readFileContentOrNull(dotenvPath)) ?? "";
+  const expectedConfig = mergeEnabledPlugins({
+    existingContent: existingConfig,
+    filePath: configPath,
+    pluginNames,
+  });
+  const expectedDotenv = enableHermesProjectPluginsInDotenv(existingDotenv);
```
dyoshikawa: High: this turns a repository-specific trust decision into a global, persistent setting. Hermes deliberately disables project-local plugins by default and documents enabling them only for trusted repositories. Persisting `HERMES_ENABLE_PROJECT_PLUGINS=true` in the user's global `.env`, while globally enabling these plugin names, means a later untrusted repository can provide a same-named `.hermes/plugins/<name>/__init__.py` and have arbitrary Python loaded under the user's account. Please keep normal project generation from changing this global gate; use an explicit warning/opt-in or a session-scoped setting for the trusted repository instead.

src/features/shared/hermes-project-plugin-activation.ts
```diff
@@ -0,0 +1,186 @@
...
+}: {
+  pluginNames: readonly HermesProjectPluginName[];
+  dryRun: boolean;
+  logger: Logger;
+}): Promise<FeatureGenerateResult> {
+  if (pluginNames.length === 0) {
+    return { count: 0, paths: [], hasDiff: false };
+  }
+
+  const homeDirectory = getHomeDirectory();
```
dyoshikawa: Mid: this always resolves activation under the default home directory and never reads `HERMES_HOME`. With a custom Hermes profile, generation therefore updates `~/.hermes/config.yaml` and `.env` instead of the active profile, contrary to the compatibility documentation added in this PR. Please resolve Hermes's configuration root from `HERMES_HOME` when set and cover that path in tests.

dyoshikawa (commented):
I found one merge-blocking security issue and two correctness gaps. The current CI checks are green, but normal project generation would persistently remove Hermes's project-plugin trust boundary for every future repository. Please address that before merging and fix the profile path described inline. The MCP integration also remains incomplete: the Hermes factory still declares `supportsEnabledTools` and `supportsDisabledTools` as `false`, so `McpProcessor` strips those fields before the adapter can produce `tools.include` and `tools.exclude`; please update the metadata and add processor/E2E coverage.

dyoshikawa:
@rudironsoni Thank you!

PR #2394, opened 2026-07-26
fix(hermesagent): honor HERMES_HOME globally

### LINKED ISSUE #2389: bug(hermesagent): honor HERMES_HOME for all global configuration
## Background

PR #2385 documents support for custom Hermes profiles through `HERMES_HOME` and makes project-plugin activation resolve that profile. However, the other global Hermes configuration adapters still target the default `~/.hermes` tree.

## Details

With `HERMES_HOME` set, global generation and import for MCP, hooks, permissions, skills, commands, and subagents can write to or read from the default profile instead of the active Hermes profile. In particular, permission allow/deny rules may not reach the profile used by Hermes, leaving intended protections inactive while modifying a different profile. The activation helper already resolves `HERMES_HOME`, which makes the inconsistency visible.

## Solution / Next Steps

Introduce a shared Hermes configuration-root resolver and apply it consistently to every global Hermes adapter and shared-config writer/importer. Add global generate/import end-to-end coverage using a custom `HERMES_HOME`, including permissions deny rules and representative MCP/hooks configuration. Keep project-scoped paths unchanged and synchronize the compatibility documentation with the final behavior.

Found during review of #2385.

rudironsoni:
## Summary

- treat `HERMES_HOME` as the global Hermes profile root across MCP, hooks, permissions, skills, commands, subagents, and conversion
- keep project-scoped `.hermes` paths unchanged and canonical RuleSync sources under `$HOME/.rulesync`
- share profile-root resolution with plugin activation and cover generation, import, conversion, check, and cleanup behavior
- keep the E2E runner under `NODE_ENV=test`, using `NODE_ENV=e2e` only for subprocesses that assert user-visible output
- reject Hermes path de-prefixing outside `.hermes` and resolve the home-directory fallback lazily
- synchronize Hermes compatibility documentation

## Test plan

- `mise exec -- pnpm cicheck`
- Hermes profile unit tests: 8 passed
- Hermes profile E2E tests: 3 passed
- convert E2E tests: 14 passed
- full E2E suite: 789 passed
- `git diff --check`

Closes #2389

### DIFF
--- a/src/features/shared/hermes-project-plugin-activation.ts
+++ b/src/features/shared/hermes-project-plugin-activation.ts
@@ -1,7 +1,6 @@
-import { join, resolve } from "node:path";
+import { join } from "node:path";
 
 import {
-  HERMESAGENT_CONFIG_FILE_NAME,
   HERMESAGENT_CONFIG_FILE_PATH,
   HERMESAGENT_PROJECT_PLUGINS_ENV_VAR,
 } from "../../constants/hermesagent-paths.js";
@@ -12,6 +11,10 @@ import {
   readFileContentOrNull,
   writeFileContent,
 } from "../../utils/file.js";
+import {
+  getHermesagentRelativeFilePath,
+  resolveHermesagentOutputRoot,
+} from "../../utils/hermesagent.js";
 import type { Logger } from "../../utils/logger.js";
 import type { FeatureGenerateResult } from "../../utils/result.js";
 import { isPlainObject } from "../../utils/type-guards.js";
@@ -114,11 +117,14 @@ export async function activateHermesProjectPlugins({
     return { count: 0, paths: [], hasDiff: false };
   }
 
-  const configuredHermesHome = process.env.HERMES_HOME?.trim();
-  const configRoot = configuredHermesHome ? resolve(configuredHermesHome) : getHomeDirectory();
-  const relativeConfigPath = configuredHermesHome
-    ? HERMESAGENT_CONFIG_FILE_NAME
-    : HERMESAGENT_CONFIG_FILE_PATH;
+  const configRoot = resolveHermesagentOutputRoot({
+    outputRoot: getHomeDirectory(),
+    global: true,
+  });
+  const relativeConfigPath = getHermesagentRelativeFilePath({
+    global: true,
+    relativeFilePath: HERMESAGENT_CONFIG_FILE_PATH,
+  });
   const configPath = join(configRoot, relativeConfigPath);
   const existingConfig = (await readFileContentOrNull(configPath)) ?? "";
   const expectedConfig = mergeEnabledPlugins({
--- a/src/features/shared/hermes-project-plugin-activation.ts
+++ b/src/features/shared/hermes-project-plugin-activation.ts
@@ -11,10 +11,7 @@ import {
   readFileContentOrNull,
   writeFileContent,
 } from "../../utils/file.js";
-import {
-  getHermesagentRelativeFilePath,
-  resolveHermesagentOutputRoot,
-} from "../../utils/hermesagent.js";
+import { getHermesagentHome, getHermesagentRelativeFilePath } from "../../utils/hermesagent.js";
 import type { Logger } from "../../utils/logger.js";
 import type { FeatureGenerateResult } from "../../utils/result.js";
 import { isPlainObject } from "../../utils/type-guards.js";
@@ -117,10 +114,7 @@ export async function activateHermesProjectPlugins({
     return { count: 0, paths: [], hasDiff: false };
   }
 
-  const configRoot = resolveHermesagentOutputRoot({
-    outputRoot: getHomeDirectory(),
-    global: true,
-  });
+  const configRoot = getHermesagentHome() ?? getHomeDirectory();
   const relativeConfigPath = getHermesagentRelativeFilePath({
     global: true,
     relativeFilePath: HERMESAGENT_CONFIG_FILE_PATH,
--- a/src/lib/convert.ts
+++ b/src/lib/convert.ts
@@ -277,7 +277,7 @@ function buildCommandsStrategy(ctx: ConvertContext) {
     allTargets,
     createProcessor: ({ toolTarget, dryRun }) =>
       new CommandsProcessor({
-        outputRoot,
+        outputRoot: resolveToolOutputRoot({ outputRoot, toolTarget, global }),
         toolTarget,
         global,
         dryRun,
@@ -426,7 +426,13 @@ function buildChecksStrategy(ctx: ConvertContext) {
     itemLabel: "check file(s)",
     allTargets,
     createProcessor: ({ toolTarget, dryRun }) =>
-      new ChecksProcessor({ outputRoot, toolTarget, global, dryRun, logger }),
+      new ChecksProcessor({
+        outputRoot: resolveToolOutputRoot({ outputRoot, toolTarget, global }),
+        toolTarget,
+        global,
+        dryRun,
+        logger,
+      }),
     loadSource: (p) => p.loadToolFiles(),
     toRulesync: (p, files) => p.convertToolFilesToRulesyncFiles(files),
     fromRulesync: (p, files) => p.convertRulesyncFilesToToolFiles(files),
--- a/src/lib/convert.ts
+++ b/src/lib/convert.ts
@@ -10,9 +10,9 @@ import { SkillsProcessor } from "../features/skills/skills-processor.js";
 import { SubagentsProcessor } from "../features/subagents/subagents-processor.js";
 import type { Feature } from "../types/features.js";
 import type { ToolTarget } from "../types/tool-targets.js";
-import { resolveToolOutputRoot } from "../utils/kimi-code.js";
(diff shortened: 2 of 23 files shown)


src/lib/convert.ts
```diff
@@ -10,9 +10,9 @@ import { SkillsProcessor } from "../features/skills/skills-processor.js";
 import { SubagentsProcessor } from "../features/subagents/subagents-processor.js";
 import type { Feature } from "../types/features.js";
 import type { ToolTarget } from "../types/tool-targets.js";
-import { resolveToolOutputRoot } from "../utils/kimi-code.js";
 import type { Logger } from "../utils/logger.js";
 import { isPackagingToolTarget } from "../utils/plugin-root.js";
+import { resolveToolOutputRoot } from "../utils/tool-output-root.js";
```
dyoshikawa, line 15: This is the blocker. `buildCommandsStrategy` (around line 280) still does `new CommandsProcessor({ outputRoot, ... })` with the raw value, while rules/ignore/mcp/subagents/skills/hooks/permissions all go through `resolveToolOutputRoot({ outputRoot, toolTarget, global })`. You fixed exactly this in `generate.ts`, but `convert.ts` was missed.

It matters now because the path side strips the `.hermes` prefix when `HERMES_HOME` is set while the root side stays at `$HOME`. With `HERMES_HOME=/opt/hermes-profile`, `rulesync convert --from claudecode --to hermesagent --features commands --global` writes `$HOME/rulesync/commands/*.json`, `$HOME/plugins/rulesync-commands/{plugin.yaml,__init__.py,.rulesync-owned}` and `$HOME/config.yaml`, and leaves `$HERMES_HOME` empty. `--global` is the only path here since hermesagent commands are `supportsGlobal: true, supportsProject: false`.

The `$HOME/config.yaml` part is the worst of it: `HermesagentCommandAuxiliaryFile.shouldMergeExistingFileContent()` returns true for that path once the prefix is stripped, so an unrelated `~/config.yaml` gets parsed as YAML, merged, and written back. Keys survive the deep merge, but comments and formatting are gone. On main the same command lands in `$HOME/.hermes/` and leaves `~/config.yaml` alone, so this branch widens the blast radius rather than narrowing it.

Fix is the same shape as `generate.ts`:

```ts
outputRoot: resolveToolOutputRoot({ outputRoot, toolTarget, global }),
```

`buildChecksStrategy` has the same omission — harmless today since `amp` is the only global checks tool and it has no home override, but worth doing at the same time for symmetry. Please also add a convert case to `e2e-hermesagent-home.spec.ts`; the spec covers generate/import only, which is why this went unnoticed.

vitest.e2e.config.ts
```diff
@@ -4,6 +4,9 @@ export default defineConfig({
   test: {
     globals: true,
     environment: "node",
+    env: {
+      NODE_ENV: "e2e",
```
dyoshikawa: I would rather not have this. Vitest applies `test.env` to the runner process itself, and the subprocesses inherit it through `{ ...process.env }` in `e2e-helper.ts`, so `isEnvTest()` becomes false for the entire e2e suite. That kills the guard in `getHomeDirectory()` (`src/utils/file.ts`) that throws when `HOME_DIR` is not set — from here on, an e2e test that runs `--global` and forgets `env: { HOME_DIR }` silently writes into the developer's or runner's real `$HOME` instead of failing loudly. Given what the previous comment shows about `~/config.yaml` being read-modify-written, that is not a theoretical concern.

Nothing breaks today (every existing `global: true` call does pass `HOME_DIR`), but the new spec asserts on file contents only and never reads stdout, so it does not need this. The existing convention is to pass `NODE_ENV: "e2e"` per spawn where output actually matters, as in `e2e-rules.spec.ts`. Could you revert this commit and follow that pattern if you need it anywhere?

src/utils/hermesagent.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export function getHermesagentRelativeDirPath({
+  global,
+  relativeDirPath,
+}: {
+  global: boolean;
+  relativeDirPath: string;
+}): string {
+  return global && getHermesagentHome()
+    ? relative(HERMESAGENT_GLOBAL_DIR, relativeDirPath)
```
dyoshikawa: Minor, but stripping the prefix with `relative()` is escapable in a way the kimi-code equivalent is not: anything not under `.hermes` comes back as `../…` and then gets joined onto the output root. `getHermesagentRelativeFilePath` makes it easier to hit, since a bare filename gives `dirname` of `"."` and `relative(".hermes", ".")` is `".."`.

Every current caller passes a `.hermes/…` constant so this is latent, and the escape guard in `ai-file.ts` would catch a write, but `kimi-code.ts` avoids the problem structurally by prefixing instead of stripping. A prefix assertion here (or de-prefixing the constants to match kimi-code) would make it impossible to get wrong later.

src/features/shared/hermes-project-plugin-activation.ts
```diff
@@ -114,11 +117,14 @@ export async function activateHermesProjectPlugins({
     return { count: 0, paths: [], hasDiff: false };
   }
 
-  const configuredHermesHome = process.env.HERMES_HOME?.trim();
-  const configRoot = configuredHermesHome ? resolve(configuredHermesHome) : getHomeDirectory();
-  const relativeConfigPath = configuredHermesHome
-    ? HERMESAGENT_CONFIG_FILE_NAME
-    : HERMESAGENT_CONFIG_FILE_PATH;
+  const configRoot = resolveHermesagentOutputRoot({
+    outputRoot: getHomeDirectory(),
```
dyoshikawa: `getHomeDirectory()` is now evaluated eagerly, whereas before it was only called when `HERMES_HOME` was unset. It throws under `NODE_ENV=test` without `HOME_DIR`, so a unit test that sets only `HERMES_HOME` will now blow up even though the home value is discarded. No production impact — just worth passing a thunk or resolving it inside the branch.

dyoshikawa (commented):
Thanks for the thorough work here. I verified the premise of #2389 independently: on main, every global Hermes adapter hardcodes the `.hermes` prefix while `hermes-project-plugin-activation.ts` already resolves `HERMES_HOME`, so the inconsistency is real. Dropping your new e2e spec onto main makes it fail, and it passes on this branch, so it is a genuine regression test. No existing tests were modified or weakened (the test diff is two new files, +304 lines only), docs and `skills/rulesync/` stay byte-identical, and the supported-tools tables are untouched as they should be. The path-resolution design is sound and the negative assertions in the e2e spec (`~/.hermes/config.yaml` must not exist, `wrong-profile` must not be imported) are exactly the right way to pin this down.

One blocker before this can go in. `buildCommandsStrategy` in `src/lib/convert.ts` still passes the raw `outputRoot` while all seven other strategies in that file wrap it in `resolveToolOutputRoot`. You fixed the same omission in `generate.ts`, but `convert.ts` was missed, and now that the `.hermes` prefix gets stripped when `HERMES_HOME` is set, that gap starts writing into `$HOME` directly instead of the profile root. Details inline. It needs the one-line fix plus a convert case in the e2e spec, since the current coverage is generate/import only, which is why this slipped through.

I would also drop the `NODE_ENV: "e2e"` change in `vitest.e2e.config.ts` — it disables a safety net for the whole suite and the new spec does not need it. Rest of the comments are minor and can be follow-ups.

CI is green across all five jobs including Windows, and I found nothing concerning from a security standpoint: no network calls, no new process spawning, no CI/workflow/dependency changes, and the only env vars read are `HERMES_HOME` and the test-only `HOME_DIR`.

vitest.e2e.config.ts
rudironsoni: Fixed in `de0f70fb`. Removed the suite-wide `NODE_ENV=e2e`; the runner keeps `NODE_ENV=test`, and only the existing convert dry-run spawn that asserts stdout opts into `NODE_ENV=e2e`. Full E2E passes: 789 tests.

src/utils/hermesagent.ts
rudironsoni: Fixed in `de0f70fb`. `getHermesagentRelativeDirPath` now rejects `..`, `..${sep}` descendants, and absolute escapes before joining. Unit coverage includes an outside directory and a bare filename.

src/lib/convert.ts
rudironsoni: Fixed in `de0f70fb`. Both conversion strategies now use `resolveToolOutputRoot`, and the new global Claude Code to Hermes commands E2E case verifies command and plugin output stays under `$HERMES_HOME` while `$HOME/config.yaml` remains byte-identical.

src/features/shared/hermes-project-plugin-activation.ts
rudironsoni: Fixed in `de0f70fb`. Project-plugin activation now resolves `getHermesagentHome() ?? getHomeDirectory()`, so the fallback is lazy. The regression test deletes `HOME_DIR` and succeeds with only `HERMES_HOME` set.

rudironsoni:
@dyoshikawa review again, please?

dyoshikawa:
@rudironsoni Thank you!

Issue #1247, opened 2026-03-02
Global Sync - Clarify Intended Behavior


sheldonhull:
## Context

Upgraded from 3.x to latest. I use rulesync primarily for global sync: distributing rules, skills, and an agents file from a single source repo to multiple tool locations.

## Setup

Source repo contains:

- Global rules (some with root: true and target-specific tweaks, e.g. targets: [claudecode] + root: true alongside targets: [opencode] + root: true)
- Global skills (replacing commands, since Claude merged that feature)
- Global agents file

I tested two paths:

Set the JSON config to default to global.
Run generate per target individually.

Observed behavior (different from 3.x)

Global rules no longer copy to ~/.claude/ or $HOME/.github-instructions.
root: true combined with per-target targets no longer produces target-specific rule variants.

## Expected behavior
Run a single sync command against my source repo and distribute rules + skills to every configured destination. Equivalent to a task sync wrapper that calls rulesync and outputs everything where it belongs.

## Question

The config table documents many global options, but the README still states "Only Claude Code is supported for global mode commands." Which is accurate?

Specifically:

- Does global sync intend to support writing rules to all target tool locations (Claude, OpenCode, etc.)?
- Is root: true + per-target targets a supported pattern for target-specific global rules, as this worked in earlier versions and not certain if was unintentional or intended behavior.

tmak:
This is related to https://github.com/dyoshikawa/rulesync/issues/1240

cm-dyoshikawa:
@sheldonhull Thank you for your feedback.

> Only Claude Code is supported for global mode commands.

is incorrect. Multiple agent global configs should be supported.

PR #2395, opened 2026-07-26
feat(rules): compose multiple root rules safely

### LINKED ISSUE #2393: Allow multiple root rules to compose safely in project and global modes
## Summary

Allow multiple `.rulesync/rules/*.md` files with `root: true` to target the same tool in both project and global modes.

Rulesync should render each source rule through the target adapter, compose compatible root or single-file outputs, and preserve native modular layouts without emitting malformed or platform-dependent files.

## Problem

Project mode currently rejects two matching root rules before conversion:

```text
Multiple root rulesync rules found for target 'hermesagent': .rulesync/rules/overview.md, .rulesync/rules/personal-assistant.md
```

Global mode can retain multiple root rules, but project and global modes should use the same target-native composition behavior.

## Proposed behavior

1. Filter source rules for the selected target.
2. Allow every matching `root: true` rule in project and global modes.
3. Render each rule through the existing target adapter.
4. Compose compatible root or explicitly folded single-file outputs in deterministic source-discovery order with one blank line between fragments.
5. Keep rules resolving to distinct native modular paths as separate files.
6. Reject modular rules that normalize to the same exact output path, because concatenating independently rendered metadata blocks is not safe.
7. Reject generated paths that differ only by case so output is deterministic on case-sensitive and case-insensitive filesystems.

Fixed-root targets such as Claude Code, Codex CLI, and Hermes Agent therefore produce one composed `CLAUDE.md`, `AGENTS.md`, or `.hermes.md`. Filename-based targets such as Cursor, Roo, AugmentCode, and Takt retain separate native rule files unless their generated paths collide, in which case generation fails clearly.

`localRoot`, target filtering, and target-specific rendering remain unchanged.

## Acceptance criteria

- Multiple matching root rules are accepted in project and global modes.
- Compatible root and single-file outputs contain every matching body without duplicate writes or content loss.
- Distinct native output paths remain separate.
- Wildcard and target-specific root rules compose when both match.
- Nonmatching target roots remain excluded.
- Ordering is deterministic and fragments are separated by one blank line.
- Metadata-bearing modular collisions fail before malformed files are written.
- Case-only output collisions fail consistently across platforms.
- Unit coverage exercises every supported project and global rules target plus normalized, overridden, and case-only collision cases.
- End-to-end coverage includes fixed-root and filename-based targets in both scopes.
- Documentation describes the shared project/global behavior and collision policy.

## Related context

- #1247 made root uniqueness target-aware but retained the one-root-per-target restriction.
- #695 established that native modular rule directories should remain modular where the tool supports them.
- #1765, #1827, and #1980 established folding into a single root file for targets that do not natively load loose rule files.

rudironsoni:
## Summary

- allow multiple matching `root: true` rules in project and global modes
- compose compatible root and single-file outputs while preserving deterministic source order
- preserve distinct target-native paths for modular targets such as Cursor, AugmentCode, and Takt
- reject unsafe same-path collisions involving a source `root: true` rule before independently rendered metadata can be corrupted
- warn and preserve existing last-writer behavior for pre-existing non-root exact-path and case-insensitive collisions
- use one explicit collision policy while retaining each generated rule's source provenance
- check collisions after conventions, MCP files, and root mirrors are added
- document lexicographic source-path ordering and synchronize the documentation skill mirror
- cover every supported rules target plus focused project and global E2E composition cases

## Behavior change

Multiple root fragments targeting the same compatible output are composed with a blank line between fragments. Explicitly supported plain-Markdown modular outputs are also composed. Unsafe same-path collisions involving a source `root: true` rule fail before writing and report the generated path and originating `.rulesync/rules/*.md` files.

Pre-existing collisions between non-root modular rules remain backward compatible: Rulesync retains both generated outputs and warns that the last write wins wherever exact or case-insensitive paths collide.

## Test plan

- `mise exec -- pnpm exec vitest run src/features/rules/rules-processor.test.ts --silent=false`
- `NODE_ENV=e2e mise exec -- pnpm exec vitest run --config vitest.e2e.config.ts src/e2e/e2e-rules.spec.ts --silent=false`
- `mise exec -- pnpm cicheck`
- `git diff --check`

Closes #2393


src/features/rules/rules-processor.ts
```diff
@@ -1128,11 +1131,37 @@ export class RulesProcessor extends FeatureProcessor {
       }
     }
 
+    const caseFoldedPaths = new Map<string, string>();
```
dyoshikawa: This is the blocker. The check walks every entry in `groups`, not just the groups that are actually being composed, so two completely unrelated rules whose generated names differ only by case now abort generation.

Concretely, `.rulesync/rules/overview.md` (root) plus `API.md` and `api.md` (both non-root) generates three correct files on main for both cursor and claudecode. On this branch it throws `Generated rule output paths differ only by case for target 'cursor': '.cursor/rules/API.mdc', '.cursor/rules/api.mdc'`. Nothing about that config involves multiple root rules, and on a case-sensitive filesystem those two files coexist perfectly well today.

It also fails loudly across the whole run rather than for the one target: the rules step in `src/lib/generate.ts` has no try/catch around it, unlike the commands and subagents steps, so a single case-only pair kills every target and every feature in that `rulesync generate` invocation. There is no flag to opt out, so the only fix available to the user is renaming source files.

Could you either scope this to paths that participate in an actual merge group, downgrade it to a warning, or put it behind an opt-in strictness flag? This criterion came from your own issue rather than from a reported problem, so I would rather not ship it as a hard failure by default.

src/features/rules/rules-processor.ts
```diff
@@ -1128,11 +1131,37 @@ export class RulesProcessor extends FeatureProcessor {
...
+      caseFoldedPaths.set(caseFoldedPath, path);
+    }
+
     const survivors = new Set<ToolRule>();
-    for (const group of groups.values()) {
-      // The root-path group prefers the root rule as its merge target; other
-      // groups fold into their first rule in source order.
-      const target = group.find((rule) => rule.isRoot()) ?? group[0];
+    for (const [path, group] of groups) {
+      if (group.length === 1) {
```
dyoshikawa, line 1185: This fast path skips the `.trim()` that main applied to every group, so trimming is now conditional on whether a collision happened, which is arbitrary.

It shows up on pi, which routes `pi.systemPrompt: append` rules to a separate path: a root rule plus an append rule gives two groups of one rule each. On main you get `AGENTS.md` = `"# RootA"` and `.pi/APPEND_SYSTEM.md` = `"# Appended"`; on this branch both come out with a trailing `\n\n\n`. Existing pi/warp/deepagents users will see spurious diffs on regeneration and `generate --check` will report drift. Trimming in the single-rule branch too should be enough, plus a test.

src/features/rules/rules-processor.ts
```diff
@@ -1128,11 +1131,37 @@ export class RulesProcessor extends FeatureProcessor {
...
+        }
+        continue;
+      }
+
+      // Root-path groups prefer the root rule as their merge target. Explicitly
+      // folding tools use their first rule when no root exists.
+      const rootRule = group.find((rule) => rule.isRoot());
+      if (!rootRule && !mergeNonRootRules) {
+        throw new Error(
+          `Multiple generated rules resolve to output path '${path}' for target '${this.toolTarget}', but this target does not support composing modular rule files`,
```
dyoshikawa: The message only names the generated path. For targets that kebab-normalize filenames (devin, antigravity-ide) the user has no way to work back to which `.rulesync/rules/*.md` files caused it — `CodingGuidelines.md` and `coding_guidelines.md` both become `coding-guidelines.md`. The error this PR removes did include the source paths via `formatRulePaths(...)`; please carry that over.

Separately, this is a behavior break worth calling out in the PR description: on main that devin config silently emitted two ToolFiles on the same path and the last write won. Failing is better, but it turns a working-if-lossy setup into a hard error, and per the other comment that error currently aborts the entire run.

src/e2e/e2e-rules.spec.ts
```diff
@@ -855,6 +922,22 @@ This is a global test rule for E2E testing.
 
       const generatedContent = await readFileContent(join(homeDir, outputPath));
       expect(generatedContent).toContain("Global Test Rule");
+
+      if (target === "augmentcode" || target === "takt") {
```
dyoshikawa: I would rather keep this one as a pure matrix test. The per-target global happy path is the guarantee that every native global rules target still generates at all, and now it also writes a second `root: true` file and asserts composition, with a hardcoded `augmentcode || takt` branch for the modular ones.

The practical problem is the allowlist: when someone adds a new modular global rules target they will fall through to `expect(generatedContent).toContain("Additional Root Fragment")` and get a confusing failure instead of a clear "add your target to the modular list" signal. Your new project-mode test at line 144 is a standalone `it.each` and is exactly the right shape — could you do the same here and leave the matrix test alone?

docs/reference/file-formats.md
```diff
@@ -56,6 +56,8 @@ This is Rulesync, a Node.js CLI tool that automatically generates configuration
 ...
 ```
 
+Multiple files can set `root: true` for the same target in project and global modes. Rulesync renders each file through the target adapter, then combines compatible root or single-file outputs in deterministic source-discovery order with one blank line between fragments. Targets that map source rules to distinct native paths keep those files separate. If modular rules normalize to the same output path, or if generated paths differ only by case, generation fails instead of emitting malformed or platform-dependent files.
```
dyoshikawa: "deterministic source-discovery order" is accurate (it comes from `findFilesByGlobs`'s `toSorted()`, so lexicographic by path) but it does not tell a reader how to actually control the order. Your own tests use `10-`/`20-` prefixes for exactly this reason. One extra sentence saying fragments are ordered by file path, so prefixes control composition order, would make this actionable.

src/features/rules/rules-processor.ts
```diff
@@ -1094,20 +1094,23 @@ export class RulesProcessor extends FeatureProcessor {
...
-   * The root rule (if any) becomes the merge target and leads the merged content;
-   * otherwise the first rule is used so a rule set without a root overview still
-   * produces a single, complete file. Mutates `toolRules` in place.
+   * The root rule becomes the merge target for root groups. Explicitly folding
+   * tools use the first rule when no root exists. Source discovery order is
+   * preserved and fragments are separated by one blank line. Mutates `toolRules`
+   * in place.
    */
-  private foldNonRootRulesIntoRootRule(toolRules: ToolRule[]): void {
+  private mergeRulesByOutputPath(
```
dyoshikawa: Minor: `.claude/rules/coding-guidelines.md` asks for a single object argument when a method takes multiple, and the neighbouring private methods (`applyLocalRootRules`, `applyRootRuleSections`, `buildMcpInstructionFiles`) all follow that. `mergeRulesByOutputPath({ toolRules, mergeNonRootRules })` would match.

dyoshikawa (commented):
Before the code: #2393 was filed by the same account as this PR and nobody else has commented on it, so the acceptance criteria in it are your design rather than something we agreed on. Removing the one-root-per-target restriction is a deliberate guardrail that came out of #1247, and this PR also turns two previously-silent situations into hard failures. That is a product decision I need to make explicitly, so please treat the direction as open even though the implementation is solid. Happy to discuss it on the issue.

On the implementation itself, the core is genuinely good. I verified the premise (main really does throw on two matching root rules), and I looked hard at the test diff since that is where things usually go wrong: the six removed lines are the two `rejects.toThrow` assertions, and they were replaced with stronger positive assertions on count, bodies and order, not softened. The e2e matrix tests are untouched, the docs mirrors are in sync, and I confirmed that root composition cannot produce a file with frontmatter in the middle — every output that can land in a merge group is frontmatter-free, and every frontmatter-bearing output is `isRoot() === false` and therefore hits the reject path. Nice invariant.

The blocker is the case-only collision check. It is not scoped to groups that are actually merging, so it fires on ordinary unrelated rules and takes down the whole generate run. Details inline. Beyond that there is a trim regression for folding targets, an error message that cannot be traced back to source files, and a couple of smaller things.

One more that is not in the diff but becomes visible because of it: `applyRootRuleSections` injects the reference section and Additional Conventions into `toolRules.find(rule => rule.isRoot())`, i.e. only the first root. Fixed-root targets are fine since they end up with one file, but on cursor/augmentcode/takt with multiple surviving roots those sections now land in whichever rule sorts first by filename rather than in the overview. On cursor that can mean the memory-file references end up in a rule that is not always applied. Worth handling or at least documenting.

CI is green on all five jobs and I found nothing concerning security-wise: no network calls, no env/credential access, no CI or dependency changes.

dyoshikawa:
Correction to my review: scratch the paragraph about `applyRootRuleSections` landing the reference section in the wrong file. I checked and it does not happen — cursor, augmentcode, aiassistant, takt and devin all set `root: false` in `fromRulesyncRule`, so `applyRootRuleSections` hits its early return on those targets and injects nothing, identically on main and on this branch. Fixed-root targets have already been merged to a single file by then. Nothing for you to do there.

It is latent rather than active, though: if a target ever emitted two `isRoot() === true` rules on distinct paths, that `find()` would silently pick one, and this PR is what makes multiple roots reachable. A short comment noting the assumption would be worth having.

The rest of the review stands.

rudironsoni:
Thanks for the detailed review. I agree that the case-only check was too broad and that its policy had not been established. I removed it, so unrelated outputs such as `API.md` and `api.md` retain their existing behavior. The narrower exact-path guard remains because independently rendered modular files resolving to the same path were already subject to silent last-writer-wins content loss. That error now identifies the original `.rulesync/rules/*.md` sources.

I also restored trimming for singleton folding groups, separated the global composition coverage from the target matrix, documented the lexicographic source-path ordering, changed the helper to an object argument, and documented the single-surviving-root invariant in `applyRootRuleSections`.

The focused processor suite passes 106 tests, the rules E2E suite passes 120 tests, and the full `pnpm cicheck` passes.

On the product decision, I agree that #2393 should not be treated as previously accepted direction. I do think multiple matching roots are the better supported behavior under this narrower policy.

The restriction retained by #1247 validates source cardinality before target conversion. That was a useful guardrail when Rulesync could not distinguish whether two roots would conflict, but the target adapters already contain the information needed to make that decision at the actual output boundary. This change generalizes #1247's target-aware behavior into output-aware behavior:

- Fixed-root targets compose rendered fragments into their one native root file.
- Modular targets preserve distinct native files.
- Multiple modular rules resolving to the same exact output path fail explicitly instead of losing content.
- Single-root configurations and unrelated case-only paths keep their existing behavior.

This lets users split always-on instructions into independently maintained source files without forcing them to concatenate those sources manually for single-file tools. It also preserves each target's native model instead of imposing one source-file cardinality rule across every adapter. Ordering is deterministic and documented, and the collision rule now covers only the concrete content-loss case.

With that narrowed scope, would you be comfortable treating multiple matching roots as supported behavior?

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `8d4596ac`. I removed the case-only hard failure, so `overview.md` plus unrelated `API.md` and `api.md` rules remain valid for both Cursor and Claude Code. I added parameterized unit coverage for those two targets. Exact same-path modular collisions remain explicit errors.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `8d4596ac`. Singleton groups are now trimmed when `mergeNonRootRules` is enabled, preserving the previous Pi, Warp, and Deepagents behavior. The new Pi regression covers one root group and one `systemPrompt: append` group and asserts the exact trimmed contents of both outputs.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `8d4596ac`. Conversion now retains `ToolRule` to `RulesyncRule` provenance, and exact-path collision errors report every originating `.rulesync/rules/*.md` path through `formatRulePaths`. I also updated the PR description to call out the previous last-writer-wins behavior and the new hard error for exact-path collisions.

src/e2e/e2e-rules.spec.ts
rudironsoni: Fixed in `8d4596ac`. I restored the global target matrix to its single-root happy path and moved multi-root coverage into a standalone `it.each` for Claude Code, AugmentCode, and Takt. The rules E2E suite passes all 120 tests.

docs/reference/file-formats.md
rudironsoni: Fixed in `8d4596ac`. The docs now state that fragments are ordered lexicographically by source file path and explicitly recommend filename prefixes such as `10-` and `20-` to control composition order. I regenerated the `skills/rulesync/` mirror from the canonical docs.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `8d4596ac`. `mergeRulesByOutputPath` now takes one object argument containing `toolRules`, `mergeNonRootRules`, and the source-rule provenance map, matching the neighboring private methods and the repository convention.

src/features/rules/rules-processor.ts
```diff
@@ -1144,10 +1160,30 @@ export class RulesProcessor extends FeatureProcessor {
...
+          }
+          survivors.add(rule);
+        }
+        continue;
+      }
+
+      // Root-path groups prefer the root rule as their merge target. Explicitly
+      // folding tools use their first rule when no root exists.
+      const rootRule = group.find((rule) => rule.isRoot());
+      if (!rootRule && !mergeNonRootRules) {
```
dyoshikawa: False negative. Because the guard is `!rootRule && ...`, any group containing a root rule skips the safety check entirely and falls through to the unconditional concatenation below — including on targets that cannot compose modular files.

Kiro global mode reproduces it: `getSettablePaths({ global: true })` puts the root at `.kiro/steering/product.md` and non-root steering files in the same `.kiro/steering/` directory, so a non-root rule named `product.md` is merged into the root and the result is

```
# Root Body

---
inclusion: fileMatch
fileMatchPattern: src/**/*.ts
---
# Non Root Body
```

Main emitted one intact file here (last writer wins), so this is a new corruption path. The group needs to be checked for whether its non-root members are safe to fold, not just for the presence of a root. A test asserting that no merged output contains a `---` block after the first line would lock this down.

src/features/rules/rules-processor.ts
```diff
@@ -1144,10 +1160,30 @@ export class RulesProcessor extends FeatureProcessor {
...
+        const sourceRules = group
+          .map((rule) => sourceRuleByToolRule.get(rule))
+          .filter((rule): rule is RulesyncRule => rule !== undefined);
+        throw new Error(
+          `Multiple generated rules resolve to output path '${path}' for target '${this.toolTarget}', but this target does not support composing modular rule files. Source rules: ${formatRulePaths(sourceRules)}`,
```
dyoshikawa: False positive, and the message is inaccurate for the targets it fires on. Two non-root `agentsmd` rules sharing an `agentsmd.subprojectPath` both render as plain Markdown with no frontmatter (`buildToolRuleParamsAgentsmd` writes the body as-is), so they compose fine, yet this throws:

```
Multiple generated rules resolve to output path 'packages/app/AGENTS.md' for target 'agentsmd', but this target does not support composing modular rule files. ...
```

amp, factorydroid, kilo and opencode behave the same way. Worse, `generateRulesCore` in `src/lib/generate.ts` does not wrap per-target generation in try/catch, so a single collision — e.g. devin normalizing `CodingGuidelines.md` and `coding_guidelines.md` to the same kebab-case path — aborts the entire generate run for every other target and feature. Gating on whether the rendered output carries frontmatter, or degrading to a warn-and-skip for the affected target, would be more consistent.

docs/reference/file-formats.md
```diff
@@ -58,6 +58,8 @@ This is Rulesync, a Node.js CLI tool that automatically generates configuration
 ...
 ```
 
+Multiple files can set `root: true` for the same target in project and global modes. Rulesync renders each file through the target adapter, then combines compatible root or single-file outputs in deterministic source-discovery order with one blank line between fragments. Fragments are ordered lexicographically by source file path, so filename prefixes such as `10-` and `20-` control composition order. Targets that map source rules to distinct native paths keep those files separate. If modular rules normalize to the same output path, generation fails instead of emitting malformed files or silently overwriting content.
```
dyoshikawa: "Fragments are ordered lexicographically by source file path" is not what the implementation does. `loadRulesyncFiles` returns `[...discoveredFiles, ...discoveredCuratedFiles]`, so `.curated/` rules always sort after local ones regardless of filename — a local `20-local.md` and a curated `.curated/05-curated.md` compose as local-then-curated. The order is deterministic, just not lexicographic overall. Either describe it as local rules first then curated, each group lexicographic, or sort the merged list in the implementation.

Also, "generation fails instead of emitting malformed files" is not accurate while the root-group case flagged in `rules-processor.ts` still merges.

docs/guide/global-mode.md
```diff
@@ -48,5 +48,5 @@ Currently, supports rules generation for Claude Code, GitHub Copilot, and OpenCo
 > Currently, when in the directory enabled global mode:
 >
 > - `rulesync.jsonc` only supports `global`, `features`, `delete` and `verbose`. `Features` can be set `"rules"` and `"commands"`. Other parameters are ignored.
-> - Tools support only a single `root: true` file in global mode as a target, e.g. you can't have 2 root files targeting Claude.
+> - Multiple `root: true` files can target the same tool in project and global modes. Compatible root or single-file outputs are combined in lexicographic source file path order with a blank line between files, so filename prefixes can control composition order. Distinct native paths remain separate, and exact modular path collisions fail explicitly.
```
dyoshikawa: Same ordering issue as `docs/reference/file-formats.md`: "lexicographic source file path order" does not hold across local and `.curated/` rules, since curated files are always appended last. Please keep this line and its `skills/rulesync/` mirror in sync with whatever wording you settle on.

dyoshikawa (commented):
Thanks for the rework — the case-only check is gone, the folding-target trim regression is fixed, and the error message now names the originating `.rulesync/rules/*.md` files. Those all look right, and I re-checked my own earlier claim about `applyRootRuleSections`: the "at most one root survives" comment does hold, since every rule class derives its root path from a fixed per-scope constant.

Two things still block this, both in `mergeRulesByOutputPath`.

The reject condition is `!rootRule && !mergeNonRootRules`, so a group that happens to contain a root rule is merged unconditionally, even on targets that cannot compose modular files. Kiro in global mode is the concrete case: the global root is `.kiro/steering/product.md` and non-root steering files land in the same directory, so a non-root rule named `product.md` gets concatenated onto the root and the generated file ends up with an `inclusion:` frontmatter block sitting in the middle of it. That is exactly the malformed output this PR sets out to prevent, and it is new — on main the same config produced one intact file via last-writer-wins. kiro-cli and kiro-ide share the class.

In the other direction the reject is too broad. Two non-root `agentsmd` rules sharing an `agentsmd.subprojectPath` produce plain frontmatter-less Markdown that composes perfectly well, but they now throw, and the message says the target "does not support composing modular rule files", which is not true for that target. The same applies to amp, factorydroid, kilo and opencode. And since `generateRulesCore` has no per-target try/catch, one devin kebab-case collision (`CodingGuidelines.md` + `coding_guidelines.md` — plausible in a real repo, and the shape your own test uses) fails the whole generate run including every other target and feature. Deciding by "does this output carry frontmatter" rather than "is there a root in the group" would fix both directions at once.

The new ordering contract in the docs also does not match `loadRulesyncFiles`, which concatenates `[...discoveredFiles, ...discoveredCuratedFiles]` — see the inline notes.

Smaller things, no need to block on them but worth a pass: `it.each(["devin", "antigravity-ide"] as const)` uses `$toolTarget` in the title, which only expands for object arrays, so both cases print the same name; the four new tests around lines 1287-1443 sit under the global-mode describe but do not pass `global`; and the `group.length === 1` branch produces the same result as the general merge path below it, so it can collapse to just the `!mergeNonRootRules` early-continue.

One process note rather than a code one: #2393 is your own issue, so the "allow multiple roots" direction has not actually been agreed on our side yet. I'm fine with it in principle given the fail-closed framing, but it is a behavior change for existing configs that previously generated silently, so it needs a release-note callout.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `d7fd3b3e`. Same-path collision safety now checks whether a group contains non-root rules and whether the target adapter explicitly supports composing them. Kiro does not opt in, so a global root `product.md` colliding with a metadata-bearing non-root `product.md` now fails with both source paths instead of appending frontmatter into the root output. I added a focused Kiro regression for this exact case.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `d7fd3b3e` using the capability-gating option. `agentsmd`, Amp, Factory Droid, Kilo, and OpenCode explicitly opt into same-path non-root composition because those outputs are plain Markdown; metadata-bearing formats remain hard errors with source-path diagnostics. A new nested `AGENTS.md` regression verifies two rules with the same `agentsmd.subprojectPath` compose in source order. This also preserves the existing Devin and Takt collision protections.

docs/reference/file-formats.md
rudironsoni: Fixed in `d7fd3b3e`. The reference now states the actual discovery order: local rules first in lexicographic source-path order, followed by non-overridden `.curated/` rules in lexicographic order, with filename prefixes controlling order within each set. I also corrected the collision wording to distinguish composable plain-Markdown outputs from metadata-bearing modular formats, and regenerated the `skills/rulesync/` mirror.

docs/guide/global-mode.md
rudironsoni: Fixed in `d7fd3b3e`. The global-mode guide now documents local rules first and non-overridden `.curated/` rules second, with lexicographic ordering within each set. The synchronized `skills/rulesync/global-mode.md` mirror contains the same wording.

rudironsoni:
@dyoshikawa do not merge, let's test this thoroughly and assert the intended product decision first!

dyoshikawa (commented):
I ran another pass over this, and I want to answer your "assert the intended product decision first" directly rather than merge it.

The composition half is sound. I re-derived the invariant your new comment at `applyRootRuleSections` claims and it holds across every adapter: `buildToolRuleParamsDefault` always routes `root: true` to the target's fixed root path, so a target can only ever have one root group, and every frontmatter-emitting adapter writes plain Markdown on its root branch — so concatenating root fragments can't inject a `---` block mid-body. Ordering is genuinely deterministic (`findFilesByGlobs` sorts, then local before `.curated/`), which matches what you wrote in the docs. And I checked the test diff against main: the only two assertions you removed are the ones for the error this PR intentionally retires, and their replacements assert more than the originals did. No silent weakening, nothing suspicious anywhere in the diff.

What I'm not ready to sign off on is the second half. The hard error for modular same-path collisions is a separate decision from "multiple roots may compose", and it's the part #2393 never asked for. Removing the old `Multiple root rulesync rules found` guard newly exposes exactly two cases (takt's `takt.name` override, global kiro's fixed `product.md`); the non-root collisions the new error covers were already reachable on main, where they warned-or-overwrote rather than aborted. So this PR is converting a pre-existing silent-overwrite path into a generate-stopping exception, for targets that had nothing to do with the root-composition feature. That's defensible — an error beats losing a rule file — but it's my call to make, not something to fold in, and right now it's coupled to the feature so I can't take one without the other. I'd rather see the collision policy split out into its own issue and PR.

Two more things worth settling before either half lands. The guard is case-sensitive, so the most likely real-world collision slips through it, and the new test pins that silence as spec — details on the line. And it runs mid-way through `convertRulesyncFilesToToolFiles`, before `appendSeparateConventionsRule` and the root-mirror pushes, so it doesn't actually cover everything that reaches the writer; a check over the final file list would be both simpler and complete, and would let the merge function stop throwing.

Smaller stuff: the new `mergeRulesByOutputPath` JSDoc drops the tool-specific rationale the old comment carried; `.rulesync/rules/overview.md:2` still has the old "less than or equal to one file" wording that you updated in both `file-formats.md` copies, and that file is what people copy as a template; and the docs now promise local-before-`.curated/` ordering with no test pinning it. Nothing blocking, but the overview.md one is a real inconsistency now.

So: holding as you asked. Once we've settled whether the collision error ships here or separately, the rest is close.

src/features/rules/rules-processor.ts
```diff
@@ -1144,10 +1170,33 @@ export class RulesProcessor extends FeatureProcessor {
...
+      // composable, while adapters must explicitly declare colliding non-root
+      // outputs safe because their metadata formats vary by target.
+      const rootRule = group.find((rule) => rule.isRoot());
+      const hasNonRootRule = group.some((rule) => !rule.isRoot());
+      if (!mergeNonRootRules && !composeSamePathNonRootRules && hasNonRootRule) {
+        const sourceRules = group
+          .map((rule) => sourceRuleByToolRule.get(rule))
+          .filter((rule): rule is RulesyncRule => rule !== undefined);
+        throw new Error(
+          `Multiple generated rules resolve to output path '${path}' for target '${this.toolTarget}', but this target cannot safely compose colliding modular rule files. Source rules: ${formatRulePaths(sourceRules)}`,
```
dyoshikawa: This is the piece I want to decide separately from the feature.

`convertToolFilesToRulesyncFiles` in this same file already has a policy for "several sources collide on one path": warn and continue, with "the last one wins wherever they collide". This introduces the opposite policy for the generate direction, and `generate` writes each target's files before moving to the next — so a collision on a later target leaves the earlier targets updated and the failing one stale, with no way forward until the user renames a file.

The configurations that newly hard-fail include ones that succeeded on main: `CodingGuidelines.md` plus a `.curated/coding-guidelines.md` from `rulesync fetch` both normalize to `coding-guidelines.md` on devin and antigravity-ide, copilot folds `api.md` and `api.instructions.md` together, and global kiro pins root to `product.md`. None of those have anything to do with multiple roots.

My preference: keep the throw for groups that involve a `root: true` rule — that's what the guard you removed was protecting, and it's the only thing this feature newly exposes — and warn for non-root collisions using the wording at the import site, so pre-existing configurations keep generating. Or split the collision policy into its own issue and PR so it can be judged on its own.

src/features/rules/rules-processor.test.ts
```diff
@@ -1318,6 +1349,324 @@ Content that would fail parsing`;
...
+        {
+          toolTarget: "claudecode",
+          expectedPaths: [
+            "CLAUDE.md",
+            join(".claude", "rules", "API.md"),
+            join(".claude", "rules", "api.md"),
+          ],
+        },
+      ] as const)(
+        "should preserve unrelated $toolTarget output paths that differ only by case",
```
dyoshikawa: This test is the one I'd push back on hardest. Pinning `.claude/rules/API.md` and `.claude/rules/api.md` as two separate expected outputs makes the current behaviour spec, and on macOS and Windows those are one file — the writer writes both in turn and one rule silently disappears, which is the exact failure the rest of this PR is about.

It's not a regression, and I know I asked you to drop the broad case check earlier. But `convertToolFilesToRulesyncFiles` compares case-insensitively for precisely this reason ("on a case-insensitive filesystem `Docs.md` and `docs.md` are one file"), so the two directions now disagree inside one file. Merging case-only variants would change output content, so I don't want that — a detection-only second pass keyed on `toLowerCase()` that warns with the import site's wording would cover it. At minimum, assert that the warning fires instead of asserting silence.

src/features/rules/rules-processor.ts
```diff
@@ -912,24 +919,30 @@ export class RulesProcessor extends FeatureProcessor {
...
+        return toolRule;
       })
       .filter((rule): rule is ToolRule => rule !== null);
 
-    if (meta.foldsNonRootIntoRoot) {
-      this.foldNonRootRulesIntoRootRule(toolRules);
-    }
+    this.mergeRulesByOutputPath({
+      toolRules,
+      mergeNonRootRules: meta.foldsNonRootIntoRoot === true,
```
dyoshikawa: Two booleans give four states but only three mean anything, and `fold: true` with `compose: true` is silently tolerated. Deriving a single `collisionPolicy: "fold" | "compose" | "reject"` from `meta` once at the call site would make the invalid combination unrepresentable and read better at the branch below.

Related: `sourceRuleByToolRule` exists only to build the error string. Keeping `{ toolRule, rulesyncRule }` pairs through the `map` above would drop the Map entirely.

src/features/rules/rules-processor.ts
```diff
@@ -1109,20 +1125,30 @@ export class RulesProcessor extends FeatureProcessor {
...
+   * Project and global modes can compose multiple root rules into one target
+   * file. This is also used for tools whose rules engine reads only one root file and therefore
+   * folds non-root rule bodies into it. Adapters whose colliding modular outputs
+   * are plain Markdown opt into composition; other modular collisions are rejected
+   * because their rendered metadata formats may conflict.
    *
-   * The root rule (if any) becomes the merge target and leads the merged content;
-   * otherwise the first rule is used so a rule set without a root overview still
-   * produces a single, complete file. Mutates `toolRules` in place.
+   * The root rule becomes the merge target for root groups. Explicitly folding
```
dyoshikawa: "The root rule becomes the merge target for root groups" isn't quite what the code does — the `find(rule => rule.isRoot())` below runs for every group, not only root ones.

Also, the old comment carried the reasoning for why this exists at all: dcode reads only `.deepagents/AGENTS.md`, Warp reads root and subdirectory `AGENTS.md` but never `.warp/memories/`, and last-writer-wins would silently drop content. That's the part someone touching this in six months needs; please keep it. The sentence starting "This is also used for tools whose rules engine..." also runs on a bit.

src/features/rules/rules-processor.ts
```diff
@@ -335,6 +338,7 @@ export const toolRuleFactories = new Map<RulesProcessorToolTarget, ToolRuleFacto
         extension: "md",
         supportsGlobal: true,
         ruleDiscoveryMode: "toon",
+        composesSamePathNonRootRules: true,
```
dyoshikawa: amp, factorydroid, kilo and opencode reach the same composition path as agentsmd through the `agentsmd.subprojectPath` re-route, but only agentsmd is covered by a test. Generalizing the "should compose plain Markdown modular rules with the same output path" case into an `it.each` over all five is cheap and fits how we cover the tool matrix elsewhere.

src/features/rules/rules-processor.test.ts
```diff
@@ -1318,6 +1349,324 @@ Content that would fail parsing`;
...
+              result.every(
+                (rule) => rule.getFileContent().split("# Second Root Fragment").length <= 2,
+              ),
+              toolTarget,
+            ).toBe(true);
+          }
+        },
+      );
+
+      it.each(["devin", "antigravity-ide"] as const)(
```
dyoshikawa: `$toolTarget` doesn't interpolate for primitive `it.each` cases — that form needs `%s`, and object cases are what support `$name`. Right now the reporter prints the literal `$toolTarget` for both runs.

While you're here: this block and the neighbouring project-mode cases sit inside `describe("RulesProcessor with global flag") > describe("convertRulesyncFilesToToolFiles in global mode")`, which isn't where a project-mode test belongs. The top-level `convertRulesyncFilesToToolFiles` describe is the right home.

rudironsoni:
Implemented the requested product-policy split in `ee1a42a9`.

Unsafe same-path groups involving a source `root: true` rule still fail. Explicitly safe plain-Markdown outputs compose. Pre-existing non-root exact-path and case-insensitive collisions now warn and continue with last-writer behavior. Collision detection runs over the final generated list, the factory uses one collision policy, and source provenance stays paired through conversion.

I also aligned the template, canonical docs, generated skill mirrors, PR body, test placement, five-adapter composition matrix, case-warning coverage, and local-before-curated ordering coverage.

Validation after rebasing onto current `upstream/main`:

- processor suite: 118 passed
- rules E2E suite: 121 passed
- `pnpm cicheck`: 342 files and 7,778 tests passed, with all content checks green
- `git diff --check`: passed

The broader decision to hard-fail ordinary modular collisions is intentionally deferred.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `ee1a42a9`. Hard errors are now limited to unsafe collisions involving source `root: true` rules. Ordinary non-root collisions warn and continue, while explicitly safe plain-Markdown adapters still compose.

src/features/rules/rules-processor.test.ts
rudironsoni: Fixed in `ee1a42a9`. Final generated paths are checked case-insensitively after conventions, MCP files, and mirrors are added. Case-only variants remain separate, but Rulesync now warns that the last write wins where the filesystem treats them as one path.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `ee1a42a9`. The two booleans were replaced with one `collisionPolicy`, and conversion retains `{ toolRule, rulesyncRule }` pairs directly, removing the provenance Map and the invalid combined state.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `ee1a42a9`. The comment now accurately describes root selection and restores the dcode, Warp, and last-writer-loss rationale.

src/features/rules/rules-processor.ts
rudironsoni: Fixed in `ee1a42a9`. The plain-Markdown same-path composition regression now covers `agentsmd`, Amp, Factory Droid, Kilo, and OpenCode.

src/features/rules/rules-processor.test.ts
rudironsoni: Fixed in `ee1a42a9`. Project-mode cases now live under a top-level conversion describe, and primitive parameterized titles use `%s`.

src/features/rules/rules-processor.ts
```diff
@@ -335,6 +338,7 @@ export const toolRuleFactories = new Map<RulesProcessorToolTarget, ToolRuleFacto
         extension: "md",
         supportsGlobal: true,
         ruleDiscoveryMode: "toon",
+        collisionPolicy: "compose",
```
dyoshikawa, line 341: This is the blocker, and it only appears after a rebase.

Since this branch's merge-base, main taught `AmpRule.fromRulesyncRule` to prepend a `globs:` YAML block to non-root files — Amp's native conditional-load gate, from #2410. So amp is no longer a plain-Markdown adapter, and `compose` ends up concatenating two frontmatter-bearing bodies.

I merged current main into a scratch worktree and ran it. Two non-root rules that both carry `agentsmd: { subprojectPath: "packages/app" }` and non-empty `globs` produce a single `packages/app/AGENTS.md`:

```
---
globs:
  - packages/app/**/*.ts
---
# first

---
globs:
  - packages/app/**/*.tsx
---
# second
```

That is the mid-body frontmatter this PR exists to prevent. Amp reads only the leading block, so the second rule's gate is silently dropped, its body gets loaded under the first rule's globs, and a stray `---` block is left sitting in the text. On main today the same config produces two separate well-formed files and last-writer-wins on disk: one rule lost, but nothing corrupt. The rebase turns a lossy-but-valid outcome into a malformed one.

I swept every target in `toolRuleFactories` for this shape and amp is the only one affected — the other four `compose` targets and every `fold`/`preserve` target are fine. So either drop amp from `compose`, or make the compose path strip each fragment's frontmatter and re-emit one merged `globs:` gate. Either way it needs the test below.

src/features/rules/rules-processor.test.ts
```diff
@@ -1318,6 +1349,343 @@ Content that would fail parsing`;
...
+          body: "# Non Root Body",
+        }),
+      ];
+
+      await expect(processor.convertRulesyncFilesToToolFiles(rulesyncRules)).rejects.toThrow(
+        `Multiple generated rules resolve to output path '${join(".kiro", "steering", "product.md")}' for target 'kiro', but this target cannot safely compose a collision involving a root rule. Source rules: ${join(RULESYNC_RULES_RELATIVE_DIR_PATH, "overview.md")}, ${join(RULESYNC_RULES_RELATIVE_DIR_PATH, "product.md")}`,
+      );
+    });
+
+    it.each(["agentsmd", "amp", "factorydroid", "kilo", "opencode"] as const)(
```
dyoshikawa, line 1553: This is why CI stays green on the amp problem: none of the rules here set `globs`, so nothing emits frontmatter and the composed output is plain Markdown for all five targets.

Please add a case where both rules carry non-empty `globs`. On this branch as-is it passes for all five; after a rebase onto main it fails for amp, which is exactly the signal you want.

src/features/rules/rules-processor.test.ts
```diff
@@ -2006,6 +2374,30 @@ targets: ["opencode", "agentsmd"]
   });
 
   describe("loadRulesyncFiles with curated rules", () => {
+    it("should compose local root fragments before curated root fragments", async () => {
```
dyoshikawa, line 2478: Not blocking — happy to take this in a follow-up.

`.curated/` is where `rulesync add`/`fetch`/`install` drop rules pulled from other repositories, and the guard this PR removes was the thing that made a third-party root fragment landing in `CLAUDE.md`/`AGENTS.md` impossible to miss: it aborted the run. Now it composes with no output at all. Root files are loaded unconditionally by every agent, so silently promoting remote content into one is worth a `logger.warn` when a `.curated/` fragment is merged into a root output. `formatRulePaths` already gives you the wording.

docs/reference/file-formats.md
```diff
@@ -58,6 +58,8 @@ This is Rulesync, a Node.js CLI tool that automatically generates configuration
 ...
 ```
 
+Multiple files can set `root: true` for the same target in project and global modes. Rulesync renders each file through the target adapter, then combines compatible root or plain-Markdown single-file outputs in deterministic source-discovery order with one blank line between fragments. Local rules are ordered lexicographically by source file path and composed before non-overridden `.curated/` rules, which are also ordered lexicographically; filename prefixes such as `10-` and `20-` control composition order within each set. Targets that map source rules to distinct native paths keep those files separate. Explicitly supported plain-Markdown modular rules that normalize to the same output path are combined. Unsafe collisions involving a source `root: true` rule fail. Other exact or case-insensitive modular collisions remain separate and produce a warning that the last write wins wherever the filesystem treats their paths as the same.
```
dyoshikawa: "Explicitly supported plain-Markdown modular rules" stops being accurate for amp once you rebase — see the `collisionPolicy` comment. Whichever way you resolve that, this sentence and its `skills/rulesync/file-formats.md` mirror need to follow.

dyoshikawa (commented):
Thanks for the rework — I went back through all six points from the last round and they are genuinely addressed. The hard error is scoped to groups involving a source `root: true` rule now, so the devin/copilot/`.curated/` collisions that used to abort warn and keep generating; the case-insensitive pass runs over the final file list after conventions, MCP files and the root mirrors, and the test asserts the warning instead of asserting silence; the two booleans collapsed into one `collisionPolicy`; the provenance Map is gone; the JSDoc carries the dcode/Warp rationale again; the composition case covers all five re-routed targets; and the project-mode cases moved out of the global describe.

On the product decision I was holding on: scoping the throw to root-involved collisions and warning for the rest is exactly where I said I wanted it, so I am no longer asking to split the collision policy into its own PR. Consider that settled.

I also re-derived the composition invariant rather than take it on trust — I swept every target in `toolRuleFactories` with two `root: true` rules, in both project and global scope, and checked each generated file for a frontmatter block appearing mid-body. Nothing malformed, nothing throws. That half is solid.

What blocks it is the branch being stale, not the design. This is 44 commits behind main (merge-base 04d8e5db) and GitHub reports it as conflicting. I merged current main into a scratch worktree — only the two `file-formats.md` copies conflicted, the source auto-merged cleanly — and one of the new `compose` targets breaks in exactly the way this PR sets out to prevent. Details on the line.

So: please rebase on main and re-check the amp case, with a regression test as part of the fix — `pnpm cicheck` does not catch it as things stand (tsc clean, all 7816 unit tests green on the merged tree). The `.curated/` note is smaller and I am happy to take it separately if you prefer.

src/features/rules/rules-processor.ts
```diff
@@ -1132,40 +1154,102 @@ export class RulesProcessor extends FeatureProcessor {
...
+
+      const rootConversion = group.find(({ toolRule }) => toolRule.isRoot());
+      const allGeneratedRulesAreRoots = group.every(({ toolRule }) => toolRule.isRoot());
+      const hasSourceRoot = group.some(
+        ({ rulesyncRule }) => rulesyncRule.getFrontmatter().root === true,
+      );
+      const shouldCompose =
+        collisionPolicy === "fold" || collisionPolicy === "compose" || allGeneratedRulesAreRoots;
+
+      if (!shouldCompose && hasSourceRoot) {
```
dyoshikawa, line 1215: This throw and the amp problem share a root cause worth naming: the decision is structural (what policy the target carries) rather than content-based (whether the rendered bodies actually contain a `---` block). That cuts both ways.

Too strict here: kiro in global mode with a non-root `product.md` that has no `globs` renders as plain Markdown — no `inclusion:` block, nothing to corrupt — and it still aborts with "cannot safely compose", which isn't true for that input. Too loose over at `collisionPolicy: "compose"`, which is how amp slipped through the moment #2410 gave it frontmatter. A content check (skip composing any fragment whose rendered body starts with `---`) would fix both at once, and would not silently break again the next time one of agentsmd/factorydroid/kilo/opencode gains a frontmatter branch.

One knock-on effect to weigh: `generateRulesCore` in `src/lib/generate.ts` has no try/catch around its per-target loop — unlike `generateIgnoreCore`, which does. So this exception aborts rules generation for every remaining target, after the earlier ones have already been written. The old `Multiple root rulesync rules found` guard had the same property, but it never fired for the kiro-global shape, so for those users this turns a working (if lossy) generate into a hard stop with no way forward but renaming a file. At minimum the message should say that.

src/features/rules/rules-processor.ts
```diff
@@ -1132,40 +1154,102 @@ export class RulesProcessor extends FeatureProcessor {
...
+    };
+
+    for (const file of outputFiles) {
+      const path = join(file.getRelativeDirPath(), file.getRelativeFilePath());
+      const key = path.toLowerCase();
+      const previous = seen.get(key);
+      if (previous) {
+        const previousPath = join(previous.getRelativeDirPath(), previous.getRelativeFilePath());
+        this.logger.warn(
+          `Both ${describeSource(previous)} and ${describeSource(file)} generate to '${previousPath}' and '${path}' (compared case-insensitively, as on macOS and Windows); the last one wins wherever they collide.`,
```
dyoshikawa: For the common case — an exact path collision, not a case-only one — `previousPath` and `path` are the same string, so this reads "generate to '.devin/rules/coding-guidelines.md' and '.devin/rules/coding-guidelines.md'". Worth branching on `previousPath === path` and printing one path there; the import site's `import to <path>` phrasing is a good model.

Separately, `describeSource` falls back to the output path for files that aren't in `convertedRules` — the conventions rule, MCP instruction files, root mirrors — so a single message can end up mixing source paths and output paths. Minor, but it makes the warning harder to act on.

dyoshikawa (commented):
Two more from a second pass, both smaller than the amp one but worth folding into the same round.

dyoshikawa:
@rudironsoni I've pushed two commits here (using the maintainer-edit permission) so this can land without another round trip — the design was in good shape and I didn't want to hold it on a staleness issue that grew while the review was running.

The merge commit brings the branch up to date with main; only the two `file-formats.md` mirrors conflicted (your `root:` comment rewording vs main's Roo Code addition — kept both).

The fix commit addresses the amp interaction I flagged: since #2410, amp prepends a `globs:` frontmatter gate to non-root files, so `collisionPolicy: "compose"` would have buried the second fragment's gate mid-body. Rather than special-casing amp, composition safety is now decided by content — a colliding group only composes when every rendered fragment is plain Markdown, and otherwise falls back to preserve-and-warn (or the root-collision error, unchanged). That also means the guard won't silently break again if another compose target gains a frontmatter branch later. Regression test added, plus the exact-path collision warning no longer prints the same path twice, and the docs paragraph now mentions the frontmatter carve-out.

Everything else from the review stands as written by you. Merging once CI is green.

dyoshikawa:
@rudironsoni Thank you!

PR #2431, opened 2026-07-27
fix(generate): stop creating empty shared config files

### LINKED ISSUE #2430: Untracked .antigravity/settings.json, .devin/config.json and .factory/settings.json after generate: gitignore them or stop emitting empty merge targets
## Background

Running `rulesync generate` in this repository leaves untracked files behind in `git status`:

```
$ git status --porcelain
?? .antigravity/settings.json
?? .devin/config.json
?? .factory/settings.json
```

None of these three paths is covered by the project `.gitignore`, even though every neighbouring output of the same tools is (`**/.devin/rules/`, `**/.devin/hooks.v1.json`, `**/.factory/mcp.json`, `**/.factory/hooks.json`, …).

This is not an oversight in the derivation logic — it is deliberate. `DERIVED_PATHS_NOT_GITIGNORED` in `src/cli/commands/gitignore-derive.ts` explicitly excludes them, with the documented rationale:

> Project-scope outputs that rulesync merges into rather than fully owns (user-managed settings files), so they are deliberately not gitignored even though a feature emits them.

The full exclusion set today:

```
**/.amp/settings.json          **/.amp/settings.jsonc
**/.antigravity/settings.json  **/.claude/settings.json
**/.claude/settings.local.json **/.codex/config.toml
**/.devin/config.json          **/.factory/settings.json
**/.grok/config.toml           **/.vibe/config.toml
**/reasonix.toml               **/.vscode/settings.json
**/.zed/settings.json          **/kilo.json
**/kilo.jsonc                  **/opencode.json
```

So the practical effect is that any repository using these targets accumulates untracked shared-config files that a user must either commit or ignore by hand.

## Details

Two separate observations came out of looking at the actual files.

### 1. The three paths are excluded from `rulesync gitignore` output

`deriveAllGitignoreEntries()` filters out every entry in `DERIVED_PATHS_NOT_GITIGNORED`, so `pnpm dev gitignore` will never add them. `src/cli/commands/gitignore-entries.test.ts` asserts both directions of that exclusion (no emitted entry is in the set; no set member is stale), so this is a deliberate, test-locked decision rather than drift.

### 2. The generated content is effectively empty

```
$ cat .antigravity/settings.json
{}
$ cat .devin/config.json
{ "mcpServers": {} }
$ cat .factory/settings.json
{}
```

rulesync created these files even though it had nothing to contribute to them. That is arguably the more interesting half of the problem: a merge-target file that rulesync does not own should probably not be created at all when the merge payload is empty. Writing `{}` gives the user a file to manage without giving them anything in return, and it is exactly what produces the `git status` noise above.

## Solution / Next Steps

Two candidate directions. They are not mutually exclusive, but they imply different answers to "does rulesync own this file?", so the maintainer should pick a stance first.

### Option A — gitignore these paths (the direct request)

Remove `**/.antigravity/settings.json`, `**/.devin/config.json` and `**/.factory/settings.json` from `DERIVED_PATHS_NOT_GITIGNORED` so they flow through the normal derivation, then run `pnpm dev gitignore` to refresh the project `.gitignore`.

- Pro: `git status` stays clean; consistent with every other output of the same tools.
- Con: it contradicts the documented rationale. If a user hand-authors settings in `.factory/settings.json` (e.g. `commandBlocklist`, `sandbox`, `enabledPlugins`), gitignoring it means their configuration silently stops being version-controlled and shared with the team.
- If this option is taken, the same question applies to the other 13 entries in the set — a partial change would leave the rule incoherent. `.vscode/settings.json` in particular must **not** be gitignored (it is a general-purpose editor file rulesync only patches a few keys into), so at minimum the set needs to be split into "shared with non-rulesync writers" vs "rulesync is the only realistic writer".

### Option B — do not emit an empty merge-target file

Skip writing a shared-config file when the merged payload contributes nothing (`{}`, or only empty containers such as `{"mcpServers": {}}`). This removes the `git status` noise at the source without changing the ownership model, and leaves user-authored settings version-controlled.

- Pro: preserves the documented rationale; no `.gitignore` change needed; fixes the problem for every target in the exclusion set at once.
- Con: needs a per-feature definition of "contributes nothing" — the check has to run on the merge payload, not on the final file, so that an existing user-authored file is never deleted or truncated.

### Recommendation

Start with **Option B**, since the concrete complaint is untracked noise from files that carry no rulesync content, and it does not require re-litigating the ownership model. Treat Option A as a follow-up only if empty-payload suppression turns out not to cover the real-world cases.

Either way, add a regression test that a `generate` run with no permissions/MCP content for these targets leaves the working tree clean.

dyoshikawa:
## Background

`rulesync generate` left untracked, empty files behind:

```
?? .antigravity/settings.json   # {}
?? .devin/config.json           # { "mcpServers": {} }
?? .factory/settings.json       # {}
```

These are shared config files rulesync *merges into* rather than owns, so they are deliberately excluded from `rulesync gitignore` via `DERIVED_PATHS_NOT_GITIGNORED`. The result was that every generate run handed the user a file to manage without putting anything in it.

This PR takes **Option B** from the issue (do not emit an empty merge-target file), which preserves the documented ownership model and needs no `.gitignore` change.

## Changes

Two causes, both fixed:

### 1. `readOrInitializeFileContent` created the file as a side effect

`readOrInitializeFileContent(filePath, initial)` wrote `initial` to disk when the file was missing, so a merge-target file was materialized *before* `writeAiFiles` ever decided whether it had anything to write.

Every generator now reads with `(await readFileContentOrNull(filePath)) ?? <default>` instead — the pattern the copilot / zed / qwencode permissions generators had already adopted for exactly this reason. Three call sites that discarded the result entirely (antigravity/devin/grokcli hooks) are simply dropped. The util itself is removed so the eager-write path cannot come back.

Side benefit: `--dry-run` / `--check` no longer touch the filesystem in ~30 generators.

### 2. `writeAiFiles` created files for empty payloads

`FeatureProcessor#writeAiFiles` now skips a file that does not exist yet when:

- the generated payload is empty, **and**
- the path is one of the shared, user-managed config files — `AiFile#shouldSkipCreationWhenPayloadEmpty()`.

An already-existing file is still rewritten, so user-authored content is never dropped, and every other file is still created regardless.

The path list moves to `src/constants/shared-config-paths.ts` as `SHARED_USER_MANAGED_CONFIG_PATHS`, and `DERIVED_PATHS_NOT_GITIGNORED` is now derived from it — one source of truth for "rulesync merges into this file but does not own it", so the gitignore exclusions and the creation rule cannot drift apart. (The predicate deliberately does *not* key off `isDeletable()`: that flag also protects global-scope files and files the orphan sweep cannot rediscover, which are unrelated concerns.)

`fileContentIsEmptyPayload` in `src/utils/content-equivalence.ts` decides emptiness on the *parsed* document (JSON/JSONC/YAML/TOML), so `{}`, `{"permissions":{}}` and `{"mcpServers":{}}` all count as empty while any scalar counts as content. It is deliberately conservative — a non-empty array, an object with a non-plain prototype (`Date`/`TomlDate`, or a `__proto__` entry that jsonc-parser resolves by replacing the prototype), a value already on the current path, and anything unparseable all count as content, because "not empty" only ever means "write the file".

## Verification

Generating every target/feature from this repo's `.rulesync/` into a clean directory, before vs. after — exactly the three paths from the issue, and nothing else:

```
- ./.antigravity/settings.json   # {}
- ./.factory/settings.json       # {}
- ./.vscode/settings.json        # {}
```

`.devin/config.json` is still written in that particular run because it genuinely carries `mcpServers` content; with an empty MCP payload it is skipped too.

Tests added:

- `src/e2e/e2e-permissions.spec.ts` — the regression test the issue asked for: an empty permissions payload leaves `.antigravity/settings.json`, `.factory/settings.json` and `.vscode/settings.json` absent.
- `src/constants/shared-config-paths.test.ts` — path matching, including any-depth matches and partial-segment non-matches.
- `src/utils/content-equivalence.test.ts` — JSON/JSONC/YAML/TOML, comment-only YAML, `null`, arrays, `__proto__`, a self-referential YAML anchor, and a date-only TOML document.
- `src/types/feature-processor.test.ts` — the four `writeAiFiles` branches, delegating to the real `AiFile.prototype` predicate rather than a stub.

`pnpm cicheck` passes.

Closes #2430

### DIFF
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -125,7 +125,7 @@ rulesync generate --check --targets "*" --features "*"
 
 ### Shared config files are never created empty
 
-Some outputs are files Rulesync merges into rather than owns, because the tool (or you) keeps unrelated settings there — `.claude/settings.json`, `.vscode/settings.json`, `.antigravity/settings.json`, `.factory/settings.json`, `.devin/config.json`, `.codex/config.toml`, `.zed/settings.json`, `opencode.json`, `kilo.json(c)`, `reasonix.toml`, `.grok/config.toml`, `.vibe/config.toml`, `.amp/settings.json(c)`. These are deliberately **not** added to `.gitignore` by `rulesync gitignore`, so that settings you hand-author in them stay version-controlled.
+Some outputs are files Rulesync merges into rather than owns, because the tool (or you) keeps unrelated settings there: `.amp/settings.json(c)`, `.antigravity/settings.json`, `.claude/settings.json`, `.claude/settings.local.json`, `.codex/config.toml`, `.devin/config.json`, `.factory/settings.json`, `.grok/config.toml`, `.vibe/config.toml`, `.vscode/settings.json`, `.zed/settings.json`, `kilo.json(c)`, `opencode.json(c)`, and `reasonix.toml`. These are deliberately **not** added to `.gitignore` by `rulesync gitignore`, so that settings you hand-author in them stay version-controlled.
 
 Because they stay committable, `generate` will not **create** one of them just to hold an empty payload: if Rulesync has nothing to contribute (e.g. no permissions map to that tool), the file is left absent instead of being written as `{}`. A file that already exists is always rewritten as usual, so nothing you authored is dropped. Every other generated file is written even when empty, since for a file Rulesync owns its existence is part of the output.
 
--- a/docs/reference/cli-commands.md
+++ b/docs/reference/cli-commands.md
@@ -123,6 +123,12 @@ rulesync generate --dry-run --targets claudecode --features rules
 rulesync generate --check --targets "*" --features "*"
 ```
 
+### Shared config files are never created empty
+
+Some outputs are files Rulesync merges into rather than owns, because the tool (or you) keeps unrelated settings there — `.claude/settings.json`, `.vscode/settings.json`, `.antigravity/settings.json`, `.factory/settings.json`, `.devin/config.json`, `.codex/config.toml`, `.zed/settings.json`, `opencode.json`, `kilo.json(c)`, `reasonix.toml`, `.grok/config.toml`, `.vibe/config.toml`, `.amp/settings.json(c)`. These are deliberately **not** added to `.gitignore` by `rulesync gitignore`, so that settings you hand-author in them stay version-controlled.
+
+Because they stay committable, `generate` will not **create** one of them just to hold an empty payload: if Rulesync has nothing to contribute (e.g. no permissions map to that tool), the file is left absent instead of being written as `{}`. A file that already exists is always rewritten as usual, so nothing you authored is dropped. Every other generated file is written even when empty, since for a file Rulesync owns its existence is part of the output.
+
 ## Gitignore Command
 
 The `gitignore` command adds generated AI tool configuration files to `.gitignore`. By default, it emits entries only for the tools listed in the `targets` of your `rulesync.jsonc` (controlled by the `gitignoreTargetsOnly` option, which defaults to `true`). Set `gitignoreTargetsOnly` to `false` to emit entries for all supported tools instead. You can also filter the output per-invocation with `--targets` / `--features`, which take precedence over the config.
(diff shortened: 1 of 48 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2414, opened 2026-07-27
Follow up Hermes Agent upstream updates: MCP ssl_verify/skip_preflight/sampling dropped and clobbered, stale delegate_task toolsets argument


dyoshikawa:
## Summary

No Hermes Agent release newer than the one rulesync is validated against has shipped (`v2026.7.20` / v0.19.0 is still `Latest`), but three drifts exist: three MCP per-server keys are dropped **and destroyed on regenerate**, and the generated subagent/command plugins still pass a `toolsets` argument that `delegate_task` no longer accepts.

This does not overlap with open issue #2389 (`honor HERMES_HOME for all global configuration`), whose scope is global config **root resolution** rather than the per-feature schemas below.

## Recent Releases

`gh release list -R NousResearch/hermes-agent` confirms `v2026.7.20` (Hermes Agent v0.19.0, 2026-07-20) is still `Latest` as of 2026-07-26 — i.e. rulesync's documented compatibility baseline is current.

| Version | Tag | Date | Notes |
| --- | --- | --- | --- |
| v0.19.0 "Quicksilver" | `v2026.7.20` | 2026-07-20 | [release](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.7.20) — smart approvals default + `approvals.deny` user deny rules, `mcp__server__tool` naming, hosted MCP OAuth + `redirect_uri`/`redirect_host`, live subagent transcripts + durable background delegation, stacked slash-skill invocations, `pre_tool_call` approve escalation re-landed. Rolls up the v0.18.1/v0.18.2 windows. |
| v0.18.2 | `v2026.7.7.2` | 2026-07-07 | [release](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.7.7.2) — WhatsApp Baileys dependency patch only. No config surface. |
| v0.18.1 | `v2026.7.7` | 2026-07-07 | [release](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.7.7) — infrastructure patch tag, uncurated; contents documented in v0.19.0. |
| v0.18.0 "Judgment" | `v2026.7.1` | 2026-07-01 | [release](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.7.1) |
| v0.17.0 | `v2026.6.19` | 2026-06-19 | [release](https://github.com/NousResearch/hermes-agent/releases/tag/v2026.6.19) |
[...]

PR #2434, opened 2026-07-27
fix(hermesagent): remove stale delegation toolsets

adity982:
## Summary

- stop emitting the removed `toolsets` field in generated Hermes subagent specs
- omit `toolsets` from both subagent and command `delegate_task` dispatch payloads
- add regression assertions for generated JSON and Python plugin content

## Validation

- focused Vitest cases for both Hermes generators (2 passed)
- targeted `oxfmt --check`
- `pnpm typecheck`
- repository pre-commit secretlint hook

Addresses the stale `delegate_task` toolsets portion of #2414.

### DIFF
--- a/src/features/commands/hermesagent-command.test.ts
+++ b/src/features/commands/hermesagent-command.test.ts
@@ -73,6 +73,7 @@ describe("HermesagentCommand", () => {
 
     expect(init?.getFileContent()).toContain("ctx.register_command(slug, handler, description)");
     expect(init?.getFileContent()).toContain('ctx.dispatch_tool(\n            "delegate_task"');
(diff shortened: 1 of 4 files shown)


dyoshikawa:
Thanks for taking this on! The analysis here is correct — `delegate_task` no longer accepts a model-facing `toolsets` argument, and the generated plugins shouldn't emit it.

That said, this has already been resolved on `main` by #2451 (merged 2026-07-28), which addresses the same portion of #2414 along with the MCP `ssl_verify` / `skip_preflight` / `sampling` gaps. The `toolsets` removal in this PR is fully covered by that change, so I'm closing this as superseded. Appreciate the effort!

PR #2436, opened 2026-07-27
fix(agentsmd): import nested AGENTS.md files and stop competing for .agents/skills/

### LINKED ISSUE #2428: Follow up AGENTS.md standard: nested AGENTS.md dropped on import, and .agents/skills/ is written by both agentsmd and agentsskills
## Summary

The AGENTS.md standard has had **no normative change in 4+ months**, so nothing new was introduced upstream. What this research found instead are two pre-existing divergences: nested `AGENTS.md` files — the standard's one and only scoping mechanism — are silently dropped on import, and `.agents/skills/` is written by both the `agentsmd` and `agentsskills` targets, so the same command with the targets reordered produces different on-disk content.

## Releases / spec changes checked

- The spec repo is `agentsmd/agents.md` (not `openai/agents.md`; that name redirects). `gh release list` → **no releases, no tags, ever**. `gh api repos/agentsmd/agents.md` → `pushed_at: 2026-03-12T14:26:14Z`.
- Every commit since 2026-03-01:
  - `d1ac7f06` 2026-03-12 — Resolve merge conflict in `CompatibilitySection.tsx` (#163)
  - `342016eb` 2026-03-10 — docs: update Gemini CLI instructions (#137)
  - `906511e0` 2026-03-10 — Add Augment Code CLI as a compatible agent (#156)
  - `37f86ac3` 2026-03-10 — Add Junie agent to the agents list (#145)

  All four are agent-list / site-rendering changes. **No normative spec change in the last 4+ months.**
- https://agents.md/ and the repo README were re-fetched 2026-07-26. The spec still defines: root `AGENTS.md`, plain Markdown with no frontmatter, and nested `AGENTS.md` in subdirectories/packages with nearest-file-wins precedence. It defines **no** `.agents/` subdirectory conventions, **no** global/user scope, and **no** version identifier.

### 1. rules — nested `AGENTS.md` files are silently not imported, and `subprojectPath` is not round-tripped

- **Upstream**: nested `AGENTS.md` in subdirectories is the standard's only scoping mechanism — "Agents automatically read the nearest file in the directory tree, so the closest one takes precedence." — https://agents.md/ (repo snapshot [`d1ac7f06`](https://github.com/agentsmd/agents.md/commit/d1ac7f063d20e70015ed6732664049ae4ba9d74e), 2026-03-12).
- **rulesync**: `project`; export works, import is lossy.
  - Export is fine: `buildToolRuleParamsAgentsmd` (`src/features/rules/tool-rule.ts`) redirects a non-root rule declaring `agentsmd.subprojectPath` to `<subprojectPath>/AGENTS.md`.
  - Import does not: `AgentsMdRule.fromFile` (`src/features/rules/agentsmd-rule.ts`) only accepts the literal root `AGENTS.md` or a path under `.agents/memories/`, and the processor enumerates exactly `root file` + `<nonRoot.relativeDirPath>/**/*.md` (`src/features/rules/rules-processor.ts`), i.e. `.agents/memories/**/*.md`.
  - Even if discovery were added, `AgentsMdRule.toRulesyncRule()` delegates to `toRulesyncRuleDefault()`, which never reconstructs `agentsmd.subprojectPath`, so the directory scope would still be lost.
  - **Verified empirically** in a scratch project containing `AGENTS.md` and `packages/api/AGENTS.md`: `rulesync import --targets agentsmd --features rules` → `Imported 1 file(s) total (1 rules)`, producing only `.rulesync/rules/overview.md`. `packages/api/AGENTS.md` was silently dropped.
- **History**: #1827 covered this and was closed as already-addressed on the *export* side. A re-check comment posted 2026-07-23 on that closed issue records the remaining import-side gap; nothing open tracks it.

### 2. skills — `.agents/skills/` is written by both `agentsmd` and `agentsskills`, and the winner depends on CLI target order

- **Upstream**: `.agents/skills/` (project) and `~/.agents/skills/` (user) are the documented cross-client interoperability convention — [adding skills support](https://agentskills.io/client-implementation/adding-skills-support) (guide added `07278fe1`, 2026-03-05). The `SKILL.md` optional fields `license` / `compatibility` / `metadata` / `allowed-tools` are normative — [specification](https://agentskills.io/specification).
- **rulesync**: `agentsmd` skills are **simulated, project-only** (`src/features/skills/agentsmd-skill.ts`, `relativeDirPath = .agents/skills` via `src/constants/agentsmd-paths.ts`); `agentsskills` skills are **native, project + global**. **Both resolve to the identical path.** `SimulatedSkill` carries only `name` + `description` (`src/features/skills/simulated-skill.ts`) and its `toRulesyncSkill()` throws.
- **Verified empirically** with one input skill declaring `agentsskills.license: Apache-2.0` and `targets: ["*"]`:
  - `generate --targets agentsmd,agentsskills --features skills --simulate-skills` → `Written 2 skills` listing `.agents/skills/demo/SKILL.md` **twice**; the final file keeps `license: Apache-2.0`.
  - `generate --targets agentsskills,agentsmd --features skills --simulate-skills` → the same double write; the final file **has lost `license`**.

  So the same command with the targets reordered yields different on-disk content, and the `agentsmd` simulated writer strips standard optional frontmatter that the `agentsskills` native writer emits.

## Proposed Follow-up

1. Discover nested `AGENTS.md` files below the project root on import (excluding `node_modules/`, `.git/` and other dependency/build trees, with a depth bound), and preserve each file's parent directory as `agentsmd.subprojectPath` in `toRulesyncRule()` so the nearest-file scope round-trips.
2. Decide ownership of `.agents/skills/`. Either drop the simulated `agentsmd` skills convention now that `agentsskills` covers the same path natively (removing `skills: { skillClass: AgentsmdSkill }` from the `agentsmd` `additionalConventions` entry in `src/features/rules/rules-processor.ts`, per `.claude/rules/feature-change-guidelines.md`), or make `agentsmd` skills delegate to the native `AgentsSkillsSkill` writer. At minimum, detect and fail when two targets resolve to the same skill output path instead of writing it twice.

## Open questions / unconfirmed

- **Global `~/.agents/AGENTS.md`.** `agentsmd` rules are `supportsGlobal: false`, so `generate --targets agentsmd --global` emits nothing. `~/.agents/AGENTS.md` *is* a real cross-tool location — rulesync already models it for `cline` and `goose` (closed issues #1744 / #2207) — but **the AGENTS.md standard itself defines no global scope** (confirmed against https://agents.md/ and the repo README on 2026-07-26). Whether the generic `agentsmd` target should emit it is a rulesync design decision, not upstream conformance.
- The spec is unversioned: no tags, no releases, no version identifier in the document. There is nothing for rulesync to pin against.

## Relationship to existing issues

- #2393 (OPEN, "Allow multiple root rules to compose safely in project and global modes") is scoped to multi-`root: true` composition into one fixed-root file — **no overlap** with nested non-root discovery or skills-path ownership.
- The `agentsskills`-side conformance gaps found in the same pass are filed separately.

## References

- https://agents.md/ — the spec itself: root `AGENTS.md`, no frontmatter, nested files with nearest-wins precedence, and the absence of any `.agents/` or global convention (gap 1, and the global open question).
- https://github.com/agentsmd/agents.md/commit/d1ac7f063d20e70015ed6732664049ae4ba9d74e — the most recent commit to the spec repo (2026-03-12), establishing that nothing normative has changed since.
- https://agentskills.io/client-implementation/adding-skills-support — the `.agents/skills/` + `~/.agents/skills/` discovery convention that both targets now claim (gap 2).
- https://agentskills.io/specification — the normative optional `SKILL.md` fields the simulated `agentsmd` writer strips (gap 2).

dyoshikawa:
## Background

The AGENTS.md standard has had no normative change in 4+ months (spec repo `agentsmd/agents.md`, no releases or tags, last commit 2026-03-12 and site-only). Re-checked against <https://agents.md/> on 2026-07-27: the standard defines a root `AGENTS.md`, plain Markdown with no frontmatter, and nested `AGENTS.md` in subdirectories with nearest-file-wins precedence — and defines **no** `.agents/` subdirectory conventions and **no** global scope.

Two pre-existing divergences, both reproduced before fixing.

## 1. Nested `AGENTS.md` files were silently dropped on import

Nested files are the standard's only scoping mechanism: *"Agents automatically read the nearest file in the directory tree, so the closest one takes precedence and every subproject can ship tailored instructions."*

Export already honored them via `agentsmd.subprojectPath`. Import did not — it enumerated the root file plus `.agents/memories/**` and nothing else:

```
$ ls AGENTS.md packages/api/AGENTS.md
$ rulesync import --targets agentsmd --features rules
Imported 1 file(s) total (1 rules)     # packages/api/AGENTS.md is gone, no warning
```

`AgentsMdRule` now exposes `getNestedFileGlobs`, a new optional hook the `RulesProcessor` uses to enumerate rule files **by pattern** rather than at a fixed path (the existing `getExtraFixedFiles` hook only covers known paths). Each match imports to `.rulesync/rules/<directory-with-hyphens>.md` carrying `agentsmd.subprojectPath`, so generate puts it back where it came from.

Two deliberate constraints:

- **Import only.** The matches are hand-authored files anywhere in the tree, not files under a rulesync-owned directory, so enumerating them for `--delete` would sweep away work rulesync never wrote.
- **Hidden directories and `node_modules/` are skipped.** An `AGENTS.md` inside a dot-directory is another tool's generated output (rulesync writes several itself); one under `node_modules/` describes somebody else's project.

`fromFile` also stops treating a modular file literally named `AGENTS.md` under `.agents/memories/` as the project root file — it previously read `<root>/AGENTS.md` instead of the memory file.

After:

```
$ rulesync import --targets agentsmd --features rules
Imported 2 file(s) total (2 rules)
$ cat .rulesync/rules/packages-api.md
---
root: false
targets: ['*']
globs: [packages/api/**/*]
agentsmd:
  subprojectPath: packages/api
---
# API
$ rulesync generate --targets agentsmd --features rules
    AGENTS.md
    packages/api/AGENTS.md
```

## 2. The simulated `agentsmd` skills writer degraded `.agents/skills/`

`.agents/skills/` is not an AGENTS.md convention — it is the [Agent Skills](https://agentskills.io/specification) project location. Nine targets write there, because they all implement that convention; each native one writes its own documented frontmatter, so enabling several and reordering `--targets` can change which optional keys survive. That is inherent to several tools sharing one path.

`agentsmd` is the exception, and the one this PR fixes. Its skills support is **simulated** — the AGENTS.md standard defines no skills at all — so it has no frontmatter model of its own, yet it overwrote the native output with a bare `name`/`description` pair:

```
$ rulesync generate --targets agentsmd,agentsskills --features skills --simulate-skills
Written 2 skills
    .agents/skills/demo/SKILL.md
    .agents/skills/demo/SKILL.md     # written twice
```

...and with the targets reversed, the simulated writer ran last and stripped `license`, `compatibility`, `metadata` and `allowed-tools`.

It now emits exactly what `agentsskills` emits, through the same `toSpecConformantAgentSkillFields` helper, so both orders produce one identical file:

```
$ rulesync generate --targets agentsskills,agentsmd --features skills --simulate-skills
Written 1 skill
    .agents/skills/demo/SKILL.md     # license and allowed-tools intact
```

The rule enforced here is narrow and defensible: **a simulated writer must not degrade a path a native target owns.** Making the eight native targets share one frontmatter block instead is a design decision about which rulesync block feeds a shared file, not a bug fix — a partial answer would make their `generate` → `import` round trip lossy, so it is deliberately out of scope.

Dropping the `agentsmd` skills convention outright was the issue's other option. `agentsmd` is the **only** simulated-skills target, so removing it would leave `--simulate-skills` with nothing to do — a much larger product change than this issue calls for.

## Tests

- `src/features/rules/agentsmd-rule.test.ts` — the glob set, nested import, the subproject round trip, and that neither the root file nor a `.agents/memories/AGENTS.md` is mistaken for a subproject.
- `src/features/skills/agentsmd-skill.test.ts` — the simulated writer's frontmatter equals the native writer's for the same input.
- `src/e2e/e2e-rules.spec.ts` — nested import → generate round trip, with `node_modules/` and `.agents/` proven excluded.

### Security review outcomes

The scan walks the whole project, so it was reviewed for that. It does **not** follow symbolic links — a link committed to a repository could otherwise pull a file from outside the project (a key, a dotfile) into version-controlled `.rulesync/` — and it honors `.gitignore`, so a vendored dependency's rule file is not copied into a tracked directory. A subproject that would claim the reserved `overview.md` name gets an `-agents` suffix (compared case-insensitively), because overwriting the root rule would make the next `--delete` remove the project's `AGENTS.md`.

`pnpm cicheck` passes.

Closes #2428


dyoshikawa:
@dyoshikawa Thank you!

Issue #2426, opened 2026-07-27
Follow up Replit upstream updates: skill allowed-tools/compatibility string forms rejected, custom_instruction/instructions.md not emitted


dyoshikawa:
## Summary

Replit's skills surface explicitly conforms to the Agent Skills specification, which defines `allowed-tools` and `compatibility` as **strings** — but `replit-skill.ts` types them as an array and an object, so a spec-conformant skill is rejected. (Same defect class as `pi`, tracked separately.) Replit also reads a second project-root rules file, `custom_instruction/instructions.md`, that rulesync does not emit.

## Recent Releases

Replit publishes weekly changelog entries rather than versioned releases. Every entry from **2026-03-06 → 2026-07-24** (`https://docs.replit.com/updates/YYYY/MM/DD/changelog`) was downloaded and grepped for skills / mcp / instructions / hooks / commands / permissions. The only relevant entry is **2026-06-12 (Agent Customization)**; everything else was MCP catalog additions, Slack/mobile/billing.

Note: the docs tree was reorganized — `core-concepts/agent/*` → `features/agent/*`. The rulesync research map (`.claude/skills/rulesync-feature-research/references/replit.md`) still points at `https://docs.replit.com/core-concepts/agent/skills` and should be refreshed.

Dry-run baseline: project → writes `replit.md` only; ignore/commands/checks/subagents/mcp/hooks/permissions all report "does not support". Global → writes 41 skill dirs under `.agents/skills`.

### 1. skills — `allowed-tools` string form rejected

- **Upstream**: [the Agent Skills specification](https://agentskills.io/specification), which [Replit's skills page](https://docs.replit.com/features/agent/skills) explicitly conforms to, defines `allowed-tools` as a space-separated **string** (experimental).
- **rulesync**: `src/features/skills/replit-skill.ts` uses `z.optional(z.array(z.string()))`, as does the `replit` block in `src/features/skills/rulesync-skill.ts`. Verified: constructing a `ReplitSkill` with `"allowed-tools": "Bash(git:*) Read"` throws `Invalid input: expected array, received string`. Sibling adapters (`agentsskills-skill.ts`, `deepagents-skill.ts`, `claudecode-skill.ts`, `rovodev-skill.ts`) already accept the union.

### 2. skills — `compatibility` string form rejected

- **Upstream**: the spec defines `compatibility` as a 1–500 character **string**.
[...]

PR #2437, opened 2026-07-27
fix(replit): accept the Agent Skills spec's string forms for allowed-tools and compatibility

dyoshikawa:
## Background

Replit's [skills page](https://docs.replit.com/features/agent/skills) states conformance to the [Agent Skills specification](https://agentskills.io/specification), which types `allowed-tools` as a space-separated **string** and `compatibility` as a 1–500 character **string**. `ReplitSkill` typed them as an array and an object, so a spec-conformant `SKILL.md` was rejected outright:

```
"allowed-tools": "Bash(git:*) Read"   → Invalid input: expected array, received string
compatibility: "Requires git and docker" → expected object, received string
```

## Changes

Both fields now accept the union — in `src/features/skills/replit-skill.ts` and in the `replit` block of `src/features/skills/rulesync-skill.ts` — so the legacy rulesync spellings keep working unchanged.

Output follows the `DeepagentsSkill` precedent in both directions:

- **generate**: a canonical rulesync list is joined into the space-separated string the spec requires, rather than emitted as a YAML sequence.
- **import**: the string is split back into the list, so a generate → import round trip leaves `.rulesync/skills/**` in the shape it started in.

Sibling adapters (`agentsskills`, `deepagents`, `opencode`, `rovodev`) already accepted the union; this brings `replit` in line.

Also refreshes the stale `core-concepts/agent/*` URLs to the current `features/agent/*` paths — the docs tree was reorganized upstream, which the issue flagged. The fix goes into the tracked source `.rulesync/skills/rulesync-feature-research/references/replit.md` (the `.claude/skills/` copy is gitignored generated output, so editing that would have been undone by the next `rulesync generate`), plus the same stale link in two source comments.

## Deliberately not included

The issue's third gap, `custom_instruction/instructions.md`, is left for a maintainer decision. It is documented only on the Enterprise [custom templates](https://docs.replit.com/teams/custom-templates) and [custom design system](https://docs.replit.com/teams/custom-design-system) pages; upstream says nothing about whether non-Enterprise projects honor it, and the issue itself asks whether it should become the default destination for non-root `replit` rules or an opt-in `replit:` frontmatter key. Adding a second rules output on that basis is a design call, not a bug fix, so the issue stays open for it.

## Tests

New cases cover constructing a `ReplitSkill` from the spec's string forms without throwing, the string forms surviving `fromRulesyncSkill` untouched, the list being joined on the way out, and the string being split back on the way in. The existing expectation that a list stays a list on output is updated to the joined form, which is the behaviour change.

`pnpm cicheck` passes.

Refs #2426


dyoshikawa:
@dyoshikawa Thank you!

PR #2438, opened 2026-07-27
feat(kimi-code): emit the config.toml [mcp] timeout defaults and [tools] switch

### LINKED ISSUE #2425: Follow up Kimi Code upstream updates: config.toml [mcp] default timeouts (0.29.1) and [tools] global enable/disable (0.29.0)
## Summary

rulesync added the `kimi-code` target on 2026-07-23, and the two Kimi Code releases immediately around that date added `config.toml` sections rulesync does not emit: `[mcp]` global default timeouts (0.29.1) and `[tools]` global enable/disable (0.29.0). Everything else was re-verified as current.

Upstream identity: repo <https://github.com/MoonshotAI/kimi-code> (npm `@moonshot-ai/kimi-code`), docs <https://moonshotai.github.io/kimi-code/> (sources in `docs/en/**`). Latest release at time of research: **0.29.1 (2026-07-24)**.

## Recent Releases

rulesync's implementation landed 2026-07-23 (`d228120c feat: add Kimi Code support`, plus `358b9ff9`, `08280bb5`, `ad311829` the same day), so the only upstream release strictly newer is 0.29.1; 0.29.0 landed one day before.

| Version | Date | Config-surface relevant change |
| --- | --- | --- |
| **0.29.1** | 2026-07-24 | **Global default MCP timeouts in `config.toml`** + env vars; experimental secondary-model bindings for subagents (`model_preference`, `[secondary_model]`) |
| **0.29.0** | 2026-07-22 | `web:` agents defined in Markdown files; `web:` `SYSTEM.md` main-agent prompt override; **`web:` global tool enable/disable via `config.toml` (`[tools]`)** |
| 0.28.1 | 2026-07-20 | ACP credential start; no config-surface change |
| 0.28.0 | 2026-07-20 | `kimi server` → `kimi web` (breaking, CLI only) |
| 0.27.0 | 2026-07-17 | `/copy`; no config-surface change |
| 0.26.0 | 2026-07-16 | polish only |

Dry-run baseline: project → `.kimi-code/AGENTS.md`, `.kimi-code/mcp.json`, `.kimi-code/agents/*.md`, `.kimi-code/skills/**`; `ignore`, `commands`, `checks` reported unsupported; `hooks` and `permissions` reported global-only. Global → the `~/.kimi-code/` twins plus `~/.kimi-code/config.toml` (hooks + permissions).

### 1. mcp — global default MCP timeouts in `config.toml` (`[mcp]`)

- **Upstream**: `[mcp] startup_timeout_ms` and `[mcp] tool_timeout_ms` set the default connect and per-tool-call timeouts for *all* MCP servers — [config files reference](https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#mcp), added in **0.29.1 (2026-07-24)**. Live in the main engine, not v2-gated: `McpConfigSchema` at `packages/agent-core/src/config/schema.ts` and `mcp:` on the top-level config schema.
- **rulesync**: `project` + `global` for `mcp.json` only. `src/features/mcp/kimi-code-mcp.ts` writes `.kimi-code/mcp.json` / `~/.kimi-code/mcp.json`; nothing in `src/` emits `startup_timeout_ms` / `tool_timeout_ms` (grep returns no hits). Only `KimiCodeHooks` and `KimiCodePermissions` patch `config.toml`, via `KIMI_CODE_CONFIG_SHARED_FILE_KEY` in `src/features/shared/shared-config-gateway.ts`. A user can repeat `startupTimeoutMs`/`toolTimeoutMs` per server, but cannot set one default for all servers.

### 2. permissions — global tool allow/deny switch in `config.toml` (`[tools]`)

- **Upstream**: `[tools] enabled` (allowlist) and `[tools] disabled` (denylist) remove tools from every agent in every session — a different enforcement layer from `[[permission.rules]]`, since it hides the tool from the model rather than prompting. [Config files reference](https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#tools), added in **0.29.0 (2026-07-22)**. Entry matching uses the same syntax as an agent file's `tools`/`disallowedTools` (exact built-in names, `mcp__server__*` globs).
- **rulesync**: `global` only for permissions, and only `default_permission_mode` + `[[permission.rules]]` are written (`src/features/permissions/kimi-code-permissions.ts`). `[tools]` is never emitted.
- **Caveat (confirmed, not speculation)**: the section is registered only in the v2 engine — `packages/agent-core-v2/src/agent/toolPolicy/configSection.ts` — and the v1 top-level config schema (`packages/agent-core/src/config/schema.ts`) has no `tools` key. The changelog entry is prefixed `web:`. So today it takes effect under `kimi web` and experimental `kimi -p`, not the interactive TUI, even though the docs page describes it in general terms. Worth implementing, but rank it below gap 1.

## Proposed Follow-up

1. Let the kimi-code MCP adapter emit an `[mcp]` section into the shared global `config.toml` (the same `applySharedConfigPatch` path hooks and permissions already use, with a new `feature` key), fed from a `kimi-code` override block in `.rulesync/.mcp.json` — e.g. `{"kimi-code": {"startupTimeoutMs": 45000, "toolTimeoutMs": 90000}}`. Global scope only, since `config.toml` has no project counterpart.
2. Extend the `kimi-code` permissions override to accept `tools: { enabled: string[], disabled: string[] }` and patch it into the same global `config.toml`, alongside `default_permission_mode`.

## Feature-by-feature verification (non-gaps)

- **rules** — upstream discovery (`packages/agent-core/src/profile/context.ts`): `$KIMI_CODE_HOME/AGENTS.md` (default `~/.kimi-code/AGENTS.md`) → `~/.agents/AGENTS.md` → per-directory `.kimi-code/AGENTS.md` and `AGENTS.md`/`agents.md` from project root to cwd. rulesync writes `.kimi-code/AGENTS.md` (project) and `~/.kimi-code/AGENTS.md` (global, `KIMI_CODE_HOME`-aware via `src/utils/kimi-code.ts`). Both confirmed by dry-run.
- **ignore** — correctly unsupported. Grepping the repo for `kimiignore|\.aiignore|ignoreFile|ignore_file` across `docs`, `packages`, `apps` returns nothing; `Glob`/`Grep` respect `.gitignore` implicitly via ripgrep only.
- **mcp (`mcp.json`)** — `~/.kimi-code/mcp.json` and `.kimi-code/mcp.json`; transports stdio/http/sse; fields `env`, `cwd`, `headers`, `bearerTokenEnvVar`, `enabled`, `startupTimeoutMs`, `toolTimeoutMs`, `enabledTools`, `disabledTools` are all copied by `src/features/mcp/kimi-code-mcp.ts`. `ws` is warned+skipped, matching upstream.
- **commands** — correctly unsupported. There is no standalone user command directory; `docs/en/reference/slash-commands.md` shows that all user-authored slash commands are Skills (`/skill:<name>`, or bare `/<name>`). The only Markdown-command surface is `commands` inside a **plugin manifest** (`kimi.plugin.json`), namespaced `/<plugin>:<command>` and installed per-user from a marketplace/GitHub URL — not a project-checkout surface.
- **subagents** — upstream `.kimi-code/agents/`, `.agents/agents/` (project) and `$KIMI_CODE_HOME/agents/`, `~/.agents/agents/` (user). Frontmatter `name`, `description`, `whenToUse`, `override`, `model_preference`, `tools`, `disallowedTools`, `subagents`. `src/features/subagents/kimi-code-subagent.ts` types all but `model_preference`, and the schema is `z.looseObject`, so `model_preference` round-trips through the `kimi-code:` section in both directions. No capability loss.
- **skills** — upstream frontmatter `name`, `description`, `type` (`prompt`/`inline`/`flow`), `whenToUse`, `disableModelInvocation`, `arguments`; directory form and flat form; project `.kimi-code/skills/` + `.agents/skills/`, user `$KIMI_CODE_HOME/skills/` + `~/.agents/skills/`. `src/features/skills/kimi-code-skill.ts` covers every field and both forms, and treats `.agents/skills` as import-only (correct — it is the cross-tool shared tree).
- **hooks** — upstream declares `[[hooks]]` only in `~/.kimi-code/config.toml`; the project-local file `.kimi-code/local.toml` accepts **only** `[workspace] additional_dir` (`packages/agent-core/src/config/workspace-local.ts`), so global-only is right. All 16 upstream events are handled: 14 canonical mappings in `src/types/hooks.ts` plus `PermissionResult` and `Interrupt` as native passthrough. Fields `event`/`matcher`/`command`/`timeout` (1–600s) all emitted, respecting upstream's 4-field-only restriction.
- **permissions (`[[permission.rules]]`)** — `decision`/`scope`/`pattern`/`reason` plus `default_permission_mode` all emitted. Tool-name mapping (`CATEGORY_TO_KIMI_CODE_TOOL`) matches `docs/en/reference/tools.md`. The `scope` enum (`turn-override`/`session-runtime`/`project`/`user`) is a **precedence label inside the user config file**, not a project-scope file — so global-only is correct.
- **checks** — Kimi Code has no equivalent to the Amp-style review-check surface.

## Deliberately not reported

- **`SYSTEM.md`** (`$KIMI_CODE_HOME/SYSTEM.md`, 0.29.0) permanently replaces the *entire* built-in main-agent system prompt (`docs/en/customization/agents.md`). It is a prompt-replacement surface, not an instruction-injection surface like `AGENTS.md`, and it is v2/web-gated today. Emitting rulesync rules there would destroy default behavior, so it is not treated as a rules gap.
- **`model_preference` / `[secondary_model]`** (0.29.1) is experimental and gated behind `KIMI_CODE_EXPERIMENTAL_SECONDARY_MODEL=1` / `KIMI_CODE_EXPERIMENTAL_FLAG=1`; the TUI ignores the field. It already round-trips through rulesync's `looseObject` subagent schema, so there is no capability loss — only an untyped key.
- **`auth: "oauth"`** on http/sse `mcp.json` entries exists in the upstream schema and is dropped by rulesync's fixed field whitelist (import preserves it, generate drops it — asymmetric). Not reported: the schema comment calls it a "Backward-compatible UI marker", nothing in the upstream codebase writes it, it is absent from the public docs field table, and it only changes behavior when a static `headers` block is also present. Same for stdio `executor`, documented in-source as "Reserved for future kaos-backed stdio launchers".
- **Sub-skills** (nested `SKILL.md` bundles surfaced as `/parent.child`, gated by a `has-sub-skill` / `hasSubSkill` frontmatter flag — `packages/agent-core/src/skill/scanner.ts`) are undocumented in `docs/en/customization/skills.md`. rulesync copies nested skill files verbatim as `otherFiles` and the frontmatter schema is loose, so nothing is lost.
- **Plugin manifests** (`kimi.plugin.json` with `skills`, `commands`, `mcpServers`, `hooks`) would be the only route to `commands` support for Kimi. That would be a new `kimi-code-plugin` target rather than a gap in the existing one, mirroring `claudecode-plugin` / `antigravity-plugin` — out of scope for this issue, but worth a separate decision.

## References

- https://moonshotai.github.io/kimi-code/en/configuration/config-files.html — the `[mcp]` and `[tools]` sections (gaps 1–2).
- https://github.com/MoonshotAI/kimi-code — the upstream repo; `packages/agent-core/src/config/schema.ts` (v1 config schema, confirming `[mcp]` is live and `[tools]` is absent), `packages/agent-core-v2/src/agent/toolPolicy/configSection.ts` (where `[tools]` is registered — the basis for the gap-2 caveat), `packages/agent-core/src/config/workspace-local.ts` (confirming hooks are global-only), `packages/agent-core/src/profile/context.ts` (rules discovery order).
- `docs/en/customization/{mcp,hooks,skills,agents,plugins}.md` and `docs/en/reference/{tools,slash-commands}.md` in the same repo — used to confirm the non-gap rows.

dyoshikawa:
## Background

Kimi Code shipped two `config.toml` sections in the releases either side of the day rulesync's `kimi-code` target landed, and rulesync writes neither.

## 1. `[mcp]` global timeout defaults (0.29.1)

`startup_timeout_ms` and `tool_timeout_ms` set the default connect and per-tool-call timeouts for **every** MCP server — including ones rulesync did not write. A per-server `startupTimeoutMs` / `toolTimeoutMs` in `mcp.json` still wins; this only moves the floor. ([config-files reference](https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#mcp))

Authored in the existing `kimi-code` block of `.rulesync/mcp.json`:

```jsonc
{
  "mcpServers": { "remote": { "type": "http", "url": "https://example.com/mcp" } },
  "kimi-code": { "startupTimeoutMs": 45000, "toolTimeoutMs": 90000 }
}
```

```toml
# ~/.kimi-code/config.toml
[mcp]
startup_timeout_ms = 45000
tool_timeout_ms = 90000
```

The servers stay in `mcp.json`, so this is the MCP feature writing a file the `hooks` and `permissions` features already share. Two small pieces of plumbing were needed, both mirroring what already exists:

- **`getAuxiliaryFiles` on the MCP processor** — the same hook the hooks processor already has (Vibe's `config.toml` flag is emitted through it). It lets one target contribute a file beyond its own settable path, and keeps the write inside the normal write phase so `--dry-run` stays honest.
- **`getExtraSharedWritePaths` on `KimiCodeMcp`** — so the shared-write derivation sees the MCP feature as one of `config.toml`'s writers. The one-line change to the derivation snapshot in `shared-file-derive.test.ts` is the reviewable record of that.

Import reads the section back into the same override block, so generate → import round-trips rather than silently dropping what generate wrote.

## 2. `[tools]` global enable/disable (0.29.0)

A **separate enforcement layer** from `[[permission.rules]]`: a rule prompts, whereas `[tools]` removes the tool from every agent in every session, so the model never sees it. ([config-files reference](https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#tools))

```jsonc
// .rulesync/permissions.jsonc
{
  "permission": { "bash": { "git status *": "allow" } },
  "kimi-code": { "tools": { "enabled": ["Bash", "Read"], "disabled": ["mcp__github__*"] } }
}
```

Entries pass through verbatim: the section uses agent-file tool syntax (exact built-in names, `mcp__server__*` globs), not the canonical `category`/`pattern` shape, so mapping them would be lossy in both directions.

**Caveat, carried into the docs:** Kimi registers this section in its v2 engine (`packages/agent-core-v2/src/agent/toolPolicy/configSection.ts`), and the v1 top-level schema has no `tools` key. So today it takes effect under `kimi web` and experimental `kimi -p`, not the interactive TUI, even though the docs page describes it in general terms. Worth emitting — the docs say when it applies.

## Shared file safety

Both sections are global-only (`config.toml` has no project counterpart) and both go through `applySharedConfigPatch` with disjoint owned keys, so the `hooks` array, the `permission` rules and the user's own keys all survive. Verified by hand against a `config.toml` already carrying hooks, permission rules and an unmanaged key.

## Tests

- `src/e2e/e2e-mcp.spec.ts` — generate writes `[mcp]` into `config.toml` while the servers stay in `mcp.json`, and import brings the defaults back into the `kimi-code` block.
- `src/e2e/e2e-permissions.spec.ts` — generate writes `[tools]`, and import brings it back.
- `src/lib/shared-file-derive.test.ts` — the derivation snapshot gains `mcp` as a writer of `.kimi-code/config.toml`.

`pnpm cicheck` passes.

Closes #2425


dyoshikawa:
@dyoshikawa Thank you!

Issue #2423, opened 2026-07-27
bug(vibe): every generated hooks.toml entry is rejected after Vibe v2.21.0 renamed hook types (+ removed flag, wrong permission tool names, MCP toggles clobbered)


dyoshikawa:
## Summary

Vibe v2.21.0 (2026-07-17) graduated hooks from experimental and **renamed every hook `type`**. `HookType` is a strict enum, so **every `.vibe/hooks.toml` entry rulesync currently emits is rejected** and dropped with a `HookConfigIssue` in the TUI. Four further drifts follow: a removed config flag rulesync still writes, three wrong permission tool names, dropped MCP per-server keys that also get clobbered, and a global skills root that is never discovered. Commands are unsupported although Vibe has user-defined slash commands.

Product identified: **Mistral Vibe** — https://github.com/mistralai/mistral-vibe (rulesync targets `.vibe/`, `AGENTS.md`, `.vibeignore`, `.vibe/config.toml`, `.vibe/hooks.toml`, `.vibe/agents/*.toml`, `.vibe/skills/`).

## Recent Releases

Checked 2026-04-01 → 2026-07-26: v2.22.0 (2026-07-21), **v2.21.0 (2026-07-17)**, v2.20.0 (2026-07-15), v2.19.1 (2026-07-09), v2.19.0 (2026-07-03), v2.18.4/.3/.2/.1/.0 (2026-06-25 … 07-01), v2.17.1/.0 (2026-06-19), v2.16.1/.0 (2026-06-15/16), v2.15.0 (2026-06-12), v2.14.1/.0 (2026-06-04/08), v2.13.0 (2026-05-29), v2.12.1, v2.11.1, v2.11.0 (2026-05-25), v2.10.1/.0, v2.9.x (2026-04-28 … 05-11), v2.8.x, v2.7.x. Also went back to v2.14.1 (2026-06-08) for the `search_replace` → `edit` rename.

### 1. hooks — event type names renamed (BREAKING, v2.21.0); every emitted hook is rejected

- **Upstream** CHANGELOG v2.21.0: "Hooks graduated from experimental; `hooks.toml` is now stable. Every hook `type` is renamed: `post_agent_turn` → `post_agent`, `before_tool` → `pre_tool`, `after_tool` → `post_tool`." In `vibe/core/hooks/models.py`:

  ```python
  class HookType(StrEnum):
      POST_AGENT = auto()
      PRE_TOOL = auto()
      POST_TOOL = auto()

[...]

PR #2441, opened 2026-07-27
fix(vibe): follow the v2.21.0 hook rename and three further upstream drifts

dyoshikawa:
## 1. Every emitted hook was rejected (breaking, v2.21.0)

Vibe v2.21.0 graduated hooks from experimental and **renamed every hook `type`**:

> Every hook `type` is renamed: `post_agent_turn` → `post_agent`, `before_tool` → `pre_tool`, `after_tool` → `post_tool`

`HookType` is a strict enum and `HookConfig.model_validate` turns a failure into a `HookConfigIssue`, so **every** `.vibe/hooks.toml` entry rulesync wrote was dropped at load with a TUI warning. The mapping and the tool-event gate are updated to the current names.

## 2. `enable_experimental_hooks` was removed

Same release, "Removed": the flag and its `VIBE_ENABLE_EXPERIMENTAL_HOOKS` env var — "No flag is required — declaring a hook is enough." The key is gone from `VibeConfigSchema`.

Rulesync wrote it into `.vibe/config.toml` as an auxiliary file on every hooks generation. That whole path is removed: the `VibeConfigToml` class, `VibeHooks.getAuxiliaryFiles`, `getExtraSharedWritePaths`, and the `hooks` entry in the file's `SHARED_CONFIG_OWNERSHIP`. `.vibe/config.toml` now has two writers (`mcp`, `permissions`) instead of three, which the shared-write derivation snapshot records.

## 3. Three permission tool names were wrong

Vibe's builtin names come from `BaseTool.get_name()` — the snake_case of each tool class:

| canonical | was | now |
| --- | --- | --- |
| `webfetch` | `fetch` | `web_fetch` |
| `websearch` | `search_web` | `web_search` |
| `edit` | `write_file` | `edit` |
| `write` | `write_file` | `write_file` |

`fetch` and `search_web` do not exist upstream, and `BaseToolConfig` allows extra keys, so those blocks were silently inert rather than erroring. Worse, `edit` and `write` both collapsed onto `write_file`: the real `edit` tool never received a permission or allow/deny list, while `write_file` — create-only since v2.14.0 — inherited edit rules.

## 4. Four per-server MCP keys were dropped and clobbered

`prompt`, `sampling_enabled`, `disabled` and `disabled_tools` are on `_MCPBase`, and `mcp_settings.py` persists `disabled` / `disabled_tools` into `config.toml` when a user toggles a server or a single tool from the `/mcp` panel. Rulesync's field allowlist omitted all four, and `mcp_servers` is replaced as a whole array on each generate — so **a user's `/mcp` toggle on a rulesync-managed server was wiped on the next `rulesync generate`**. They are passthrough fields now.

## 5. The shared global skills root was not discovered

Vibe's `user_skills_dirs` returns both `~/.vibe/skills` and `~/.agents/skills`, and the README lists both under "Global skills directories". Rulesync registered `.agents/skills` as an alternative root for project scope only, so `rulesync import --global` never saw it. It now applies at either scope.

## Not included

The issue's sixth item — commands, which Vibe expresses as skills carrying `user-invocable: true` — is a feature addition rather than a drift from a shipped release, and routing commands onto the skills surface is a design decision (the shape `devin` uses). Left on the issue.

## Tests

The existing Vibe hook/permission suites are updated to the corrected names, which is where the behaviour change shows: `disabled_tools` for a denied `edit` category is now `["edit"]` rather than `["write_file"]`, and the test that previously asserted "edit and write merge onto write_file" now asserts they stay on their own tools — that assertion was pinning the bug. The e2e hooks cases assert the new `type` strings and that no `.vibe/config.toml` is written for hooks at either scope.

`pnpm cicheck` passes.

Refs #2423


dyoshikawa:
@dyoshikawa Thank you!

Issue #2422, opened 2026-07-27
bug(rovodev): per-tool permissions emitted at toolPermissions.<tool> instead of toolPermissions.tools.<tool>, so they never apply


dyoshikawa:
## Summary

rulesync writes Rovo Dev per-tool permission levels **one level too shallow** — `toolPermissions.<tool>` instead of the documented `toolPermissions.tools.<tool>` — so generated permissions never take effect and a real user config round-trips to nothing. Three smaller MCP/checks drifts follow from the same pass.

## Release / version situation

Rovo Dev updates independently of the Atlassian CLI package, and Atlassian does not publish per-build Rovo Dev release notes on a public page. The [ACLI changelog](https://developer.atlassian.com/cloud/acli/changelog/) was last updated 2026-03-25 and states "Rovo Dev can now be updated independently of Atlassian CLI"; in-product release notes are only reachable via the `/changelog` slash command. `developer.atlassian.com/cloud/acli/release-notes/` and `/cloud/acli/reference/rovodev/` both 404.

**Therefore the baseline below is the current state of the living support docs, not a versioned changelog.** All upstream claims were verified against Atlassian's live support docs (fetched raw with `curl` and converted to text; the pages are server-rendered, so there was no SPA problem).

### Verified upstream surface (2026-07-26)

```yaml
toolPermissions:
  # Default permission for tools not explicitly listed (default: "ask")
  default: "ask"
  tools:
    create_file: "ask"
    delete_file: "ask"
    move_file: "ask"
    find_and_replace_code: "ask"
    open_files: "allow"
    expand_code_chunks: "allow"
    expand_folder: "allow"
    grep: "allow"
    createTechnicalPlan: "allow"
    getJiraIssue: "allow"
    createJiraIssue: "ask"
    updateJiraIssue: "ask"
    getConfluencePage: "allow"
    createConfluencePage: "ask"
    updateConfluencePage: "ask"
  bash:
    default: "ask"
    commands:
      - command: "ls.*"
        permission: "allow"
    runInSandbox: false        # macOS only, internal users
  allowedExternalPaths: []
```

Both [use tools in Rovo Dev CLI](https://support.atlassian.com/rovo/docs/use-tools-in-rovo-dev-cli/) and [manage Rovo Dev CLI settings](https://support.atlassian.com/rovo/docs/manage-rovo-dev-cli-settings/) agree that per-tool levels live under `toolPermissions.tools.*`.

[...]

PR #2442, opened 2026-07-27
fix(rovodev): nest per-tool permissions under toolPermissions.tools and translate MCP transport

dyoshikawa:
Two adapter bugs where rulesync wrote a key Rovo Dev does not read, plus the ownership and fail-closed fixes that fell out of reviewing them.

## 1. Per-tool permissions were written one level too shallow

Rovo Dev documents per-tool levels under `toolPermissions.tools.<tool>`. Rulesync wrote them directly under `toolPermissions.<tool>`, where Rovo Dev ignores them — so **generated permissions never applied**, and importing a real user config round-tripped to nothing. Confirmed against the live [use tools](https://support.atlassian.com/rovo/docs/use-tools-in-rovo-dev-cli/) page, which shows:

```yaml
toolPermissions:
  tools:
    create_file: ask
    open_files: allow
```

Generate now writes the nested block, and import reads it — falling back to the legacy flat shape only for a key the nested block says nothing about, so a `config.yml` an earlier rulesync produced still imports.

Not implemented from the issue's gap 1: `toolPermissions.default`. The issue's YAML block shows it, but the live docs page does not, so the claim could not be confirmed from a primary source.

## 2. MCP servers carried the canonical `type` instead of `transport`

Rovo Dev documents the per-server transport key as `transport` (`stdio` | `http` | `sse`), with no `type` alias. `RovodevMcp.fromRulesyncMcp` wrote `getMcpServers()` verbatim, so a canonical `"type": "http"` — the spelling this repo's own `.rulesync/mcp.json` uses — reached the file under a key Rovo Dev does not read.

Translated in both directions (`local` → `stdio`, `streamable-http` → `http`, `sse` → `sse`), matching the kilo / roo / goose / amp / codexcli adapters. Two entries are skipped with a warning rather than written as a guess: a `ws` server (no Rovo Dev equivalent — same as the Kimi Code adapter) and a `disabled` one, since Rovo Dev turns servers off through `mcp.disabledMcpServers` in `config.yml` and `mcp.json` has no flag for it. On import, a `transport` value outside Rovo Dev's vocabulary is dropped rather than carried into the canonical config, whose transport field is a strict enum — passing one through would make `.rulesync/mcp.json` fail to parse for *every* target.

## 3. Ownership: a revoked rule now actually goes away

Fixing (1) surfaced a worse problem than the wrong depth. `bash`, `allowedExternalPaths`, and the eight per-tool keys were merged into whatever was already in `config.yml`, so deleting a rule from `.rulesync/permissions.*` left the old value live — and the stale flat copies were then resurrected by the import fallback and propagated to every other tool. Those keys are rulesync-owned now: dropped before the current set is written, with the keys rulesync has no category for (`createTechnicalPlan`, the Jira/Confluence tools) left untouched.

Two guards around that:

- A generate that produces nothing this adapter can express leaves the block alone rather than emptying it — there would be no rule of ours to replace the user's levels with.
- Removing an owned key logs a warning, because `allowedExternalPaths` is also written from inside a session by `/directories`.

## 4. Collisions resolve fail-closed

Both directions had a path where the looser of two levels won:

- **Import.** A category maps onto four tool keys, and Rovo Dev rewrites a single key when the user answers "always allow" to one prompt. Last-write-wins meant `grep: allow` next to three denies imported as a blanket `read: allow`. They fold onto the strictest level now, matching the augmentcode and opencode adapters.
- **Generate.** `edit` and `write` drive the same mutation tools, so a conflicting catch-all cannot be represented. Resolving to `edit` wrote an allow when the author had also said `write: deny`; the stricter level is kept instead, and the warning names it.

## Left open on the issue

- **`.rovodev/.review-agent.md` (checks)** — needs a new `RovodevCheck` adapter; a feature addition rather than a drift fix.
- **`mcp.disabledMcpServers`** — honouring canonical `disabled` means making the MCP feature a second writer of the `config.yml` the permissions feature already owns, with sub-key seeding so `mcpConfigPath` / `allowedMcpServers` survive. Worth doing, but its own change.
- **hooks** — still blocked upstream: `/hooks` and `eventHooks` are referenced but the schema has never been published.

## Follow-up worth filing

`strictestAction` here is the fourth implementation of `deny > ask > allow` in `src/features/permissions/` (grokcli, opencode, amp). A shared helper would be the right home for it.

## Tests

52 unit assertions across the two adapters, including the legacy-flat import, the nested-wins direction where the nested value is the *looser* one (so a regression is visible), the strictest fold, ownership deletion vs. preservation of unmapped keys, both conflict directions, the five transport translations, and the skip cases. The e2e permissions case moved to the nested depth.

`pnpm cicheck` and the permissions/MCP e2e suites pass locally.

Refs #2422


dyoshikawa:
@dyoshikawa Thank you!

Issue #2421, opened 2026-07-27
Follow up Takt upstream updates: workflow_overrides quality gates (checks), assistant.init_files, remaining workflow security policies


dyoshikawa:
## Summary

Takt has a `workflow_overrides.quality_gates` surface (string gates plus `type: command` gates, shipped 0.43.0, 2026-05-29) that maps directly onto rulesync's `checks` feature but is unsupported. Separately, `assistant.init_files` (0.42.0) is the *only* surface that feeds project rules into interactive mode and rulesync registers nothing there, and the workflow security-policy toggles other than `workflow_mcp_servers` remain unauthorable.

Upstream repo: https://github.com/nrslib/takt (no GitHub Releases published; versions are git tags + `CHANGELOG.md`).

## Recent Releases

Tag dates from `gh api repos/nrslib/takt/git/refs/tags`:

| Version | Date | Notes relevant to rulesync |
| --- | --- | --- |
| 0.52.0 | 2026-07-19 | BREAKING: `{current_report}`/`{previous_report}`/`{peer_reports}`/`{report_history}` placeholders removed (rulesync never emits them); `auto_routing` rework; `auto_requeue_max_attempts`, `ignore_exceed` config keys |
| 0.51.0 | 2026-07-11 | `auto_routing`, `session: compact`, `finding_contract.manager.provider/model`; partial `provider_profiles` now overlay per-provider defaults |
| 0.50.0 | 2026-07-06 | `takt-mcp` (Takt **as** an MCP server) and `takt-acp` entrypoints; `when()` rule conditions; Codex `approval_policy` pinned to `never` |
| 0.49.0 | 2026-06-28 | `takt exec` + presets under `.takt/exec/presets/` and `~/.takt/exec/presets/`, `~/.takt/exec.yaml`; `session_key` |
| 0.48.0 | 2026-06-21 | `provider_options.claude\|codex.base_url`; `team_leader.inspect_tools` |
| 0.47.0 | 2026-06-18 | Finding Contract; `provider_routing` (personas/tags/steps); `provider_options.extends` replaces `$ref` |
| 0.46.0 | 2026-06-13 | OTLP export; `provider_options.opencode.allowed_tools`; `provider_options.kiro.agent` |
| 0.44.0 | 2026-06-03 | `kiro` provider |
[...]

Issue #2145, opened 2026-07-06
Takt permissions: support `step_permission_overrides`, `required_permission_mode`, and `provider_options.*` sandbox/network keys


dyoshikawa:
## Summary

rulesync's Takt permissions translator currently maps only the coarse `provider_profiles.<provider>.default_permission_mode` (one of `readonly` < `edit` < `full`). This proposal is to additionally support Takt's finer-grained permission concepts: `step_permission_overrides.<step>`, a step-level `required_permission_mode` floor, and the sandbox/network toggles under `provider_options.*` — most likely via a Takt-scoped override namespace in `.rulesync/permissions.json`.

This relates to [#2127](https://github.com/dyoshikawa/rulesync/issues/2127), which proposes a tool-scoped override namespace so tool-specific permission keys don't leak into the canonical model.

## Motivation / Purpose

The canonical `.rulesync/permissions.json` model is `{ "permission": { "<category>": { "<glob>": "allow"|"ask"|"deny" } } }` with per-pattern granularity across categories (`bash, read, edit, write, webfetch, websearch, grep, glob, notebookedit, agent`, plus `mcp__*`).

Takt's model, by contrast, is a single coarse mode per provider profile, so the current translation is lossy in both directions:

- Any `deny` present collapses the whole profile to `readonly`.
- Otherwise, an `edit`/`write` `allow` collapses to `edit`.
- Otherwise, a `bash` `allow` collapses to `full`.
- Otherwise `readonly`.

All per-pattern granularity is discarded. Supporting Takt's own richer keys would let users express step-level and provider-level intent that Takt actually understands, instead of losing it in the round-trip.

## Background from Research

- Takt maps only `provider_profiles.<provider>.default_permission_mode` with the ordering `readonly` < `edit` < `full`; the collapse rules above are how rulesync currently derives that single mode. (See the current translator at `src/features/permissions/` and Takt's configuration docs: https://github.com/nrslib/takt/blob/main/docs/configuration.md)
- Takt exposes per-workflow-step overrides via `step_permission_overrides.<step>`, which set a permission mode for an individual workflow step and exist at both project and global tiers. The canonical model has no notion of a "workflow step", so this is fundamentally tool-specific. (Inferred from Takt configuration docs: https://github.com/nrslib/takt/blob/main/docs/configuration.md — please verify exact key shape.)
[...]

PR #2443, opened 2026-07-28
feat(takt): support checks as quality gates and author the workflow security policies

dyoshikawa:
Covers gaps 1 and 3 of #2421. Gap 2 is left open — see the bottom.

## 1. checks → Takt quality gates

`.rulesync/checks/*.md` had nowhere to go for Takt: the target was unsupported, and a dry-run printed `Target 'takt' does not support the feature 'checks'`. Takt's equivalent is the [quality gates](https://github.com/nrslib/takt/blob/main/docs/workflows.md) in the `workflow_overrides` block of `.takt/config.yaml` (project) / `~/.takt/config.yaml` (global). Both scopes carry the block — confirmed against `GlobalConfigSchema` upstream, which omits only `assistant`.

A check becomes one gate:

- by default a **string gate**, the body text, which Takt injects into the agent step prompt as a completion directive (`description` is used when the body is empty, the file stem when neither is set);
- with `command` in the check's `takt` frontmatter block, a **command gate** — `{type: command, name, command, cwd, timeout_ms}`, matching upstream's `.strict()` `CommandQualityGateInputSchema`, so the block's schema is closed here too. `name` defaults to the file stem, since Takt's logs identify a gate by it.

`steps` / `personas` in that block scope a gate to named workflow steps or personas; an unscoped gate applies everywhere, and one naming both is written to both. `quality_gates_edit_only` is a property of the block rather than of one gate, so a check setting it turns it on for all of them.

`workflow_overrides` is owned by the checks feature in the shared config: rewritten from `.rulesync/checks/` on every generate — so deleting a check deletes its gate — while every other key of `config.yaml` is preserved and the file is never deleted.

Two edges of that ownership, both documented and both warned about:

- When checks remain but none target Takt, the block is retracted rather than left stale — a deleted check's command gate would otherwise keep running after every step. The adapter cannot tell a block a previous generate wrote from one a user wrote by hand, so both go; that is what owning a key means here, and the warning names rulesync as the owner.
- It cannot fire when `.rulesync/checks/` is emptied altogether: the feature has no source, so nothing runs and the gates stay. Deleting them is a manual step in that case. Closing this properly means running a feature's generate for a target that has no source files at all, which is a pipeline-wide change rather than a Takt one.

### Two processor extensions this needed

The checks processor was one-tool-file-per-check in both directions, which cannot express "all checks collapse into one shared file":

- `fromRulesyncChecks` — optional, and deliberately **not** declared on `ToolCheck`, so the processor can detect which tools implement it. Same shape as the hooks processor's `fromRulesyncHooks`.
- `toRulesyncChecks` — defaults to `[this.toRulesyncCheck()]`, so one shared file can import as many checks as it holds. Gate text (or a command gate's `name`) becomes the file name.

Neither changes behaviour for the existing Amp and Hermes adapters.

Because checks now reach a file that the MCP and permissions features also write, `checks` moves out of `NON_SHARED_WRITE_FEATURES` into `SHARED_WRITE_FEATURE_ORDER`, and its generation step gains the derived shared-write metadata. It owns `workflow_overrides` outright, so its position is not a conflict decision.

## 2. permissions → the workflow security policies

Only `workflow_mcp_servers` was reachable, and only as a transport allowlist the MCP adapter derives. Takt's other default-deny policies — `workflow_arpeggio.{custom_data_source_modules, custom_merge_inline_js, custom_merge_files}`, `workflow_runtime_prepare.custom_scripts`, `workflow_command_gates.custom_scripts`, `sync_conflict_resolver.auto_approve_tools`, `allow_git_hooks`, `allow_git_filters` — could not be authored at all. #2145 listed them as an open question and the implementation did not cover them.

They now round-trip through the `takt` block of `.rulesync/permissions.*`. Modelled key by key rather than left to the loose passthrough: each admits one class of user-supplied code, so a typo would leave a capability denied while looking authored, and only the boolean shapes Takt accepts are written — a wrong type makes Takt's loader reject the whole config rather than ignore the key.

`workflow_command_gates.custom_scripts` reads as a companion to the gates above, so ownership had to be decided: it stays with permissions, because Takt validates it against gates declared in *workflow YAML* (`validateWorkflowCommandGates`), not against gates coming from `workflow_overrides`.

## Left open

**Gap 2, `assistant.init_files`.** Not attempted here, and not for lack of value — it is the only surface that feeds project rules into Takt's interactive mode. It needs the *rules* feature to become a writer of `config.yaml`, which today it is not: rules emit standalone facet files and the feature has no shared-config write path. That is a change to the rules pipeline rather than to the Takt adapter, and it comes with its own decisions (which facets to register within the 16-file limit, and how to preserve user-authored entries). Worth its own PR. Note it is also project-scope only — `GlobalConfigSchema` omits `assistant`.

## Tests

20 new adapter assertions (string and command gates, the description/stem fallbacks, step and persona scoping, edit-only, an invalid `takt` block, key preservation vs. owned-block replacement, the import split, an unimportable gate shape, and a round-trip), 3 for the security policies including a dropped wrong-typed value,, the retraction and the warnings, and Takt joins the checks e2e matrix in both scopes.

`pnpm cicheck` passes; the full e2e suite (800 tests) passes locally.

Refs #2421


dyoshikawa:
@dyoshikawa Thank you!

Issue #2420, opened 2026-07-27
Follow up AugmentCode upstream updates: PromptSubmit hook event, hook args/metadata clobbered, code_review_guidelines.yaml checks, .agents/commands import root


dyoshikawa:
## Summary

auggie 0.27.0 (2026-05-14) added a `PromptSubmit` hook event rulesync silently drops; hook handler `args` and matcher-group `metadata` are unauthorable **and overwritten** on every generate because `.augment/settings.json` owns the `hooks` key; Augment Code Review's `.augment/code_review_guidelines.yaml` is unsupported by the `checks` feature; and the `.agents/commands/` discovery root is not read on import.

## Recent Releases

Source: npm `@augmentcode/auggie` stable tags plus the matching changelog pages on augmentcode.com.

| Version | npm publish | config-relevant content |
| --- | --- | --- |
| 0.24.0 | 2026-04-14 | image view tool, MCP OAuth cleanup — no config surface |
| 0.25.x | 2026-04-30 … 05-04 | `auggie login --no-tui` only |
| 0.26.0 | 2026-05-06 | no config surface |
| **0.27.0** | **2026-05-14** | **"Added PromptSubmit hooks and support for updating input from PreToolUse hooks"**; plugin marketplaces gain branch/tag refs |
| 0.28.0 | 2026-05-22 | general-purpose builtin subagent; lazy skill includes can nest skills |
| 0.29.0 | 2026-06-05 | `auggie-guide` builtin subagent; `auggie mcp add -H` header flag |
| 0.30.0 | 2026-06-19 | PreToolUse/PostToolUse now also run inside sub-agent sessions (no new event) |
| 0.31.0 | 2026-06-25 | `${...}` variable expansion in MCP + hook config files |
| 0.32.0 | 2026-07-03 | MCP restart/health only — explicitly no config change |
| 0.33.0 | 2026-07-17 | MCP `tools/list_changed` refresh, provenance; "skills" renamed "prompt modules" **in the knowledgebase only** |

Docs pages read: `/cli/config`, `/cli/rules`, `/cli/custom-commands`, `/cli/skills`, `/cli/subagents`, `/cli/hooks`, `/cli/permissions`, `/cli/plugins`, `/cli/setup-auggie/workspace-indexing`, `/codereview/review-guidelines`.

[...]

Issue #2404, opened 2026-07-27
Follow up Goose upstream updates: goose review checks surface, retired .gooseignore, custom agents (.agents/agents/*.md)


dyoshikawa:
## Summary

Goose has three surfaces rulesync has not followed: the `goose review` checks surface (`.agents/checks/*.md`, shipped v1.35.0, 2026-05-22), the first-class Markdown custom-agent format (`.agents/agents/*.md`, shipped v1.34.0, 2026-05-13) that rulesync's recipe-YAML subagent mapping does not reach, and `.gooseignore`, which upstream has **retired** but rulesync still generates.

Note on sources: `block/goose` now redirects to **`aaif-goose/goose`** (`gh api repos/block/goose` → `full_name: aaif-goose/goose`), and the canonical docs host is now `https://goose-docs.ai` (`GOOSE_DOCS_ROOT` default in `documentation/docs/guides/config-files.md`); `https://block.github.io/goose/docs/` still serves the same content.

## Recent Releases

- **v1.44.0** (2026-07-23) — [release](https://github.com/aaif-goose/goose/releases/tag/v1.44.0) — Stop hook gains `working_dir`; hooks docs formalize the deny contract / matcher regex / `tool_input` keys; **stale `.gooseignore` references removed from docs** (#10359); `goose review` security fix (GHSA-r5pp-p5r8-466r).
- **v1.43.0** (2026-07-14) — [release](https://github.com/aaif-goose/goose/releases/tag/v1.43.0) — custom agents guide docs (#9293 / #9947); streamable HTTP deeplink generator.
- **v1.41.0** (2026-07-03) — global hints `~/.agents/AGENTS.md` (already followed, #2207).
- **v1.40.0** (2026-07-02) / **v1.39.0** (2026-06-25) — open-plugin MCP `.agents/plugins/<name>/.mcp.json` (already followed, #2078).
- **v1.38.0** (2026-06-17) — plugins install to `~/.agents/plugins` (#9088).
- **v1.35.0** (2026-05-22) — **`goose review` local code-review command first released** (PR #9114, merged 2026-05-18).
- **v1.34.0** (2026-05-13) — **Agents CRUD / filesystem agent discovery** (PR #9084, merged 2026-05-11).

### 1. checks — `goose review` with `.agents/checks/*.md` is unsupported

[...]

Issue #2399, opened 2026-07-27
Follow up Cursor upstream updates: .cursor/permissions.json (IDE auto-run surface) and .cursor/BUGBOT.md checks


dyoshikawa:
## Summary

Cursor's IDE/Agent auto-run permission surface (`.cursor/permissions.json`) was materially expanded during 2026-05/06 by the Auto-review run mode and the SDK's `local.autoReview` routing, and Bugbot's `.cursor/BUGBOT.md` code-review instruction files remain an actively developed surface. rulesync supports neither: `permissions` covers only the CLI's `.cursor/cli.json`, and `cursor` is absent from the `checks` feature.

## Recent Releases

Source: [cursor.com/changelog](https://cursor.com/changelog) (paginated at `/changelog/page/N`) plus the official docs.

- **Cursor Router** — 2026-07-22 — [changelog](https://cursor.com/changelog) — model routing; dashboard-controlled, no repo config file.
- **v3.11 Side Chats & Conversation Search** — 2026-07-10 — [changelog](https://cursor.com/changelog) — announces conversation-level hooks (`beforeSubmitPrompt`, `afterAgentResponse`, `afterAgentThought`, `stop`, `subagentStart`) for cloud agents.
- **v3.10 Team Marketplaces & MCPs** — 2026-06-30 — [changelog](https://cursor.com/changelog) — team MCP servers distributed from the dashboard, not a repo file.
- **Bugbot update** — 2026-06-10 — [changelog page 2](https://cursor.com/changelog/page/2) — 3× faster reviews, `/review`, GitHub/GitLab sync. Bugbot remains actively developed.
- **SDK update** — 2026-06-04 — [changelog page 3](https://cursor.com/changelog/page/3) — `local.autoReview` routing configured through **`permissions.json`** with `autoRun.allow_instructions` / `autoRun.block_instructions`.
- **v3.6 Auto-review run mode** — 2026-05-29 — [changelog page 3](https://cursor.com/changelog/page/3) — new approval mode driven by a classifier, with custom classifier instructions.

### 1. permissions — `.cursor/permissions.json` (IDE / Agent auto-run surface) unsupported

[...]

PR #2444, opened 2026-07-28
fix(augmentcode): support PromptSubmit, stop clobbering hook args and metadata, import .agents/commands

dyoshikawa:
Covers gaps 1, 2 and 4 of #2420. Gap 3 is left open — see the bottom.

## 1. `PromptSubmit` was filtered out as unsupported

Auggie 0.27.0 added the event; rulesync's `AUGMENTCODE_HOOK_EVENTS` still listed six, and `CANONICAL_TO_AUGMENTCODE_EVENT_NAMES` had no `beforeSubmitPrompt` entry, so `buildEffectiveHooks` dropped such a hook — silently, and including one authored through the tool-specific `augmentcode` override. It is now mapped, and added to the matcher-less set alongside the session events, matching auggie's own two enums.

## 2. `args` and `metadata` were clobbered on every generate

This is the part worth reading. Auggie's validator accepts `args` on a command hook and `metadata` (`includeConversationData` / `includeMCPMetadata` / `includeUserContext`) on the matcher group — the latter changes the JSON payload the hook script receives. Neither existed canonically, and neither was emitted.

Because `.augment/settings.json` declares `hooks: { kind: "replace-owned-keys", ownedKeys: ["hooks"] }`, that is not merely a missing feature: a hand-written `args` or `metadata` in an existing settings file was **erased on every generate**, and import dropped it too. So the round trip actively destroyed working configuration.

Both round-trip now. The converter gains two mechanisms alongside the existing boolean/string passthroughs:

- `arrayPassthroughFields` — per-hook string arrays, emitted and imported only when every entry is a string, so a malformed value cannot reach a config auggie would reject.
- `groupPassthroughFields` — fields that belong to the *matcher group* rather than to a hook. The canonical model is a flat list of definitions, so export reads the first definition of the group carrying one and import puts it back on every definition of that group. That is the same shape `sequential` already uses for a Qwen group property.

Only the augmentcode adapter registers them, so no other tool's output changes.

## 3. `.agents/commands/` was not read on import

Auggie resolves commands over `.augment`, `.claude` and `.agents` in both the workspace and the home directory. `AugmentcodeCommand.getSettablePaths` returns only `.augment/commands`, so `rulesync import --targets augmentcode` missed commands auggie actually loads.

The issue proposed extending `ToolCommandSettablePaths` with an import-only root list. This uses the processor's existing `loadAdditionalImportFiles` hook instead — it was written for inline/aggregate sources rather than for extra directory roots, but it is the seam that already exists, and it keeps the change inside the adapter. The declarative alternative, matching the `importDirPaths` (subagents) and `alternativeSkillRoots` (skills) fields the same tool already uses for its `.agents` roots, is the tidier home for this and is worth a follow-up.

Two consequences of reading that root, both handled: the duplicate check compares basenames as well as paths, because rulesync writes `.agents/commands/` for the `agentsmd` target with namespaces flattened — otherwise `git/commit.md` and `commit.md` would import as two commands and silently double the user's set. And the import warns, naming the shared root, because a command taken from it becomes a rulesync command written for every target. Generation stays on `.augment/commands/`, and `.agents/commands/` is never written to or swept for orphans, since files there may belong to another tool. A command found under either root gets the same rulesync-side identity: its path under the commands root. Auggie's third root, `.claude/commands/`, is deliberately left alone — it is Claude Code's own output, which rulesync already imports as that target.

One more thing this exposed: auggie namespaces a nested command by its directory (`/frontend:component`), but the target declared `supportsSubdirectory: false`, so a nested command was flattened to its basename on generate. That is fixed, which means an earlier Rulesync's flattened files are still on disk under their old names — `--delete` removes them — and `--delete` now sweeps `.augment/commands/` recursively.

## Left open

**Gap 3, `.augment/code_review_guidelines.yaml`.** Not attempted here. It is a new `AugmentcodeCheck` adapter, and — as the issue itself notes — one of a family of three tracked `checks` gaps (#2399 Cursor BUGBOT, #2404 Goose). It also needs a decision the other two do not: canonical `critical` has no counterpart in the `high | medium | low` scale, so it has to collapse, and that is worth deciding once for the family rather than three times.

## Tests

Nine hook assertions (the new event, the dropped matcher on it, `args` + `metadata` through generate, both back on import, a non-string `args` ignored, a control character in an imported `args` dropped, the two metadata-widening warnings, and a hook whose type the tool cannot express not supplying the group field), three for the commands root (nested files found, a foreign-frontmatter file skipped, missing root is not an error), and one at the processor level for each of the basename dedupe and the logger reaching the adapters.

`pnpm cicheck` passes; the hooks and commands e2e suites pass locally.

Refs #2420


dyoshikawa:
@dyoshikawa Thank you!

Issue #2419, opened 2026-07-27
Follow up Reasonix upstream updates: sse MCP transport rewritten to http, retired trusted_read_only_tools/plan_mode_allowed_tools, ignore unsupported


dyoshikawa:
## Summary

Reasonix v1.17.18 (2026-07-22) re-implemented the `sse` MCP transport and retired three plan-mode config surfaces. rulesync moves in the opposite direction on all of them: it rewrites `type = "sse"` to `http` (producing a server that cannot connect), and it still round-trips two retired keys. Separately, `ignore` is unsupported even though Reasonix has documented read-block surfaces and the permissions adapter already contains the hook for it.

Upstream repo: https://github.com/esengine/DeepSeek-Reasonix (default dev branch `main-v2`). Last rulesync follow-up: #2235 (closed 2026-07-16).

## Recent Releases

Release notes are commit lists plus a bilingual highlights body, and the repo `CHANGELOG.md` only carries `Unreleased` + `1.0.0`, so **doc diffs between tags** were used as the primary evidence.

- v1.17.11 (2026-07-12) … v1.17.17 (2026-07-21) — desktop/CLI fixes, MCP trust-flow simplification, subagent fleets (`task`/`fleet`, `write_paths`), Remote SSH.
- **v1.17.18** (2026-07-22) — [release](https://github.com/esengine/DeepSeek-Reasonix/releases/tag/v1.17.18) — the release where three config-surface changes land, verified by bisecting `docs/GUIDE.md` and `docs/SPEC.md` across tags v1.17.12 → v1.17.21:
  1. `sse` MCP transport goes from "recognised but deferred … Configuring it returns a clear error" to a fully implemented transport.
  2. `[[plugins]].trusted_read_only_tools` (plus `default_tools_approval_mode`, `tools.<raw>.approval_mode`, `approvals_reviewer`) declared **retired**.
  3. `[agent].plan_mode_allowed_tools` removed from the documented config surface.
- v1.17.19 (2026-07-22), v1.17.20 (2026-07-24), v1.17.21 (2026-07-25/26) — MCP installation/activation lifecycle, machine-readable CLI (`docs/CLI.md`), default-trust for project MCP + hooks (#6889). No new user-authored config keys in rulesync's feature dimensions.

Bisect evidence (grep counts on `docs/GUIDE.md` at each tag):

| marker | v1.17.12 | v1.17.15 | v1.17.17 | v1.17.18 | v1.17.19 | v1.17.21 |
[...]

PR #2445, opened 2026-07-28
fix(reasonix): write the sse transport verbatim, and stop authoring retired keys

dyoshikawa:
Covers gaps 1, 2 and 3 of #2419. Gap 4 is left open — see the bottom.

## 1. `sse` was rewritten to `http`, producing a server that cannot connect

`resolveReasonixType` collapsed `sse` onto `http` on the belief that the transport was deprecated. Reasonix v1.17.18 re-implemented it — SPEC §3.3 describes it as "the legacy 2024-11-05 HTTP+SSE transport", a persistent GET stream receiving an announced relative POST endpoint — so the rewrite pointed the client at Streamable HTTP and the server would not connect. That is a wrong generated value, not a lossy round trip.

It is written verbatim now. `streamable-http` → `http` and `local` → `stdio` stay.

## 2. `trusted_read_only_tools` was re-emitted on every generate

Retired in the same release, along with `default_tools_approval_mode`, `tools.<raw>.approval_mode` and `approvals_reviewer` — installing a server is the authorization decision now. Reasonix ignores the key when loading and strips it the next time it saves that entry, so rulesync writing it back meant the two writers churned against each other.

It is now neither written nor imported. Importing it would put a Reasonix-only dead key into the canonical `mcpServers`, which every MCP target writes out, so it would surface in `.mcp.json`, `.cursor/mcp.json` and the rest. Because rulesync owns the `plugins` key, the next generate also drops it from an older `reasonix.toml` — which loses nothing Reasonix still reads, and is announced when a canonical config an older rulesync imported into still carries it.

The permissions side of the same problem is deliberately the opposite: `[agent].plan_mode_allowed_tools` *is* lifted on import, because the `reasonix` override block is tool-scoped and cannot leak into another tool's config the way a canonical MCP field would.

## 3. The `reasonix` permissions override still offered `plan_mode_allowed_tools`

The key left the documented config surface in v1.17.18 — plan-mode tool access is the permissions layer's job now — so presenting it as an authorable knob was misleading. Removed from `REASONIX_OVERRIDE_AGENT_KEYS`.

`plan_mode_read_only_commands` stays, since it still exists, but the code and the docs now say what upstream says about it: "legacy compatibility only; Plan bash now uses Permissions", retained for round-tripping rather than as the way to grant anything.

The strip happens on the *merged* `[agent]` table rather than only on what the override authored. Stripping just the authored value would leave a wider list already in `reasonix.toml` untouched, making "narrow this list" the one edit that could never land.

## 4. Two things reviewing this turned up

`ws` is in the canonical transport enum and was written verbatim — the same unloadable `type` the sse rewrite produced. Servers whose transport Reasonix does not implement are skipped with a warning now, including a `ws://`/`wss://` URL that states no transport at all.

## Left open

**Gap 4, the `ignore` adapter.** Not attempted here. Unlike the three above it is not upstream drift — Reasonix has never had an ignore adapter — and adding one makes the `ignore` feature a **second writer** of the `reasonix.toml` that the MCP and permissions adapters already patch, with the deny-entry ownership that implies. The issue is right that the permissions adapter already carries the "existing Read deny entries (possibly from the ignore feature)" guard, so the seam is there; it is the ownership decision that wants its own change. Worth noting the target choice the issue makes is the right one: `[permissions] deny` takes glob specifiers, while `[sandbox] forbid_read` is absolute paths with no documented glob support.

## Tests

The assertions that pinned the old behaviour now pin the new one, each carrying the reason: `sse` survives and round-trips, the retired MCP field is neither written nor imported and its removal is announced, the retired `[agent]` key is stripped from the merged table with a warning while import still lifts it, and a server whose transport Reasonix cannot load is skipped — by explicit `type` or by URL scheme.

`pnpm cicheck` passes; the MCP and permissions e2e suites pass locally.

Refs #2419


dyoshikawa:
@dyoshikawa Thank you!

Issue #2418, opened 2026-07-27
Follow up Grok CLI upstream updates: .grok/commands/ and .grok/rules/ surfaces unmodeled, skill invocation flags dropped, http hook type


dyoshikawa:
## Summary

Grok CLI has two file surfaces rulesync does not model at all — `.grok/commands/*.md` (a native slash-command directory) and `.grok/rules/*.md` (non-root rules) — plus dropped skill invocation-control frontmatter and an unsupported `http` hook handler type. All findings are confirmed against the public source at version **0.2.112**, since the changelog does not name these directories.

Also worth correcting: **`.grokignore` does not exist**. Grep for `grokignore` across all 2309 `.rs` files in the 0.2.112 source returns nothing; `crates/codegen/xai-grok-agent/src/prompt/ignore.rs` builds its matcher from the repo `.gitignore` plus `core.excludesFile` only. The `.grokignore` claim in closed issue #1983 came from secondary blog sources. rulesync's `ignore —` is correct.

## Recent Releases

Primary sources:

- Public source repo: https://github.com/xai-org/grok-build — shallow clone at commit `b41c75a578f98bddbd326ab02cd53618451d97ee` (2026-07-26), `SOURCE_REV 91d8cf3`, `crates/codegen/xai-grok-version/Cargo.toml` version = **0.2.112**.
- Changelog: https://x.ai/build/changelog
- Docs: https://docs.x.ai/build/overview, `/settings`, `/settings/reference`, `/features/hooks`, `/features/skills-plugins-marketplaces`

The last rulesync follow-up sweep stopped around v0.2.94 (2026-07-09), so the delta window checked is **v0.2.95 → v0.2.112**. Changelog items in that window (workflows, `/doctor`, `[model_providers.<id>]`, `combine_queued_prompts`, `managed_config.toml` team defaults, OIDC auth pinning, `/usage`, dashboard/TUI work) are runtime/UI/enterprise settings outside rulesync's feature dimensions.

Dry-run baseline: project → 1 MCP (`.grok/config.toml`) + 5 subagents (`.grok/agents/*.md`) + 41 skills (`.grok/skills/**`) + 1 hooks (`.grok/hooks/rulesync.json`) + 1 permissions (`.grok/config.toml`); **no rules file, no commands**. Global → same plus 1 rule (`.grok/AGENTS.md`). Processor gates containing `grokcli`: rules, mcp, skills, subagents, hooks, permissions — **not** commands, ignore, or checks.

[...]

PR #2447, opened 2026-07-28
feat(grokcli): write non-root rules to .grok/rules, and stop dropping skill flags and http hooks

dyoshikawa:
Covers gaps 2, 3 and 4 of #2418. Gap 1 is left open — see the bottom.

## 1. Non-root rules had nowhere to go

Grok scans a rules directory alongside the AGENTS.md family: `.grok/rules/*.md` in each project directory it walks up to the git root, and `~/.grok/rules/*.md` in the home scope — flat, sorted by name. Rulesync folded every topic rule into the single root `AGENTS.md`, which the adapter's own comment says was verified against grok 0.2.54. That is no longer what the tool does.

Non-root rules now write to `.grok/rules/` (project) / `~/.grok/rules/` (global), so the target drops out of the fold-into-root list in the rules processor. The gitignore derivation picks the directory up automatically.

For anyone upgrading: an earlier Rulesync left those topics inlined in `AGENTS.md`, so a regenerate splits them back into their own files and the root file shrinks.

## 2. Skill invocation flags were silently dropped

`user-invocable` and `disable-model-invocation` are canonical fields — other adapters emit them, and the docs list the tools that honour them. Grok honours both: `user-invocable: false` hides a skill from the skill tool, `disable-model-invocation: true` stops the model reaching for it on its own. But this adapter built its output frontmatter as exactly `{ name, description }`, so authoring either did nothing here and said nothing about it. Both directions carry them now.

## 3. `http` hooks were dropped on a stale premise

The adapter declared `supportedHookTypes: new Set(["command"])` with the comment "The shared Claude-style converter only round-trips `command` hooks". That stopped being true when Claude Code declared `["command", "prompt", "http", "mcp_tool", "agent"]`. Grok's own `HookHandlerType` is `command | http` — an http handler POSTs to its `url` — so the two are declared and the other canonical types still stay out.

## Left open

**Gap 1, `.grok/commands/`.** Not attempted. Grok discovers a flat `*.md` slash-command directory in the same config dirs it scans for skills, with Claude-command-compatible frontmatter — but `commands` is not a feature this target supports at all today, so it is a new adapter plus a processor registration rather than a drift fix. It is also the one gap the issue could only confirm from source rather than docs, which is worth a second look before building on it.

Also worth keeping from the issue: `.grokignore` does not exist, so `ignore —` for this target is correct, not a gap.

## Tests

The three assertions that pinned the old rules behaviour now pin the new one — the `.grok/rules` non-root path in both scopes, reading a non-root file from it, and writing a topic rule into it — plus the processor's fold-target list, which no longer names `grokcli`. The full unit suite passes, as do the rules, skills and hooks e2e suites.

`pnpm cicheck` passes.

Refs #2418


dyoshikawa:
@dyoshikawa Thank you!

Issue #2417, opened 2026-07-27
bug(kilo): a { enabled: false } MCP toggle entry aborts the entire --targets kilo run (+ sandbox block, steps type, global non-root rules)


dyoshikawa:
## Summary

A Kilo config containing a `{ "enabled": false }` MCP toggle entry — a documented upstream form — makes the **entire `--targets kilo` run abort with a ZodError**, not just the MCP feature. Alongside that: the `sandbox` config block is unauthorable, `steps` is typed as an array where upstream expects a number (so real `.kilo/agents/*.md` files fail to import), and global rules are root-only.

## Recent Releases

Bulk-pulled every `v*` release body since 2026-03-15 via `gh api repos/Kilo-Org/kilocode/releases?per_page=100&page=N` (124 releases). There is no `CHANGELOG.md` at the repo root; release bodies are the changeset log.

- **v7.4.16** (2026-07-24) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.16) — PR #12494 surfaces which rule auto-approved a tool call (agent / project config / global config / YOLO).
- **v7.4.13** (2026-07-20) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.13) — indexing file-extension allowlist (#12306); no rulesync feature dimension.
- **v7.4.6 / v7.4.7** (2026-07-13) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.6) — PR #12075 "Support configuring network destinations that sandboxed tools can reach while network access is otherwise restricted" → the `sandbox.allowed_hosts` key.
- **OpenCode v1.16.2 sync**, PR #12088, [commit ef6b152](https://github.com/Kilo-Org/kilocode/commit/ef6b152ff8534aa98bad1eecee6999be798b19e8) (2026-07-13) — current shape of `packages/core/src/v1/config/config.ts`.
- **v7.4.3** (2026-07-09) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.3) — PR #12004 `/reload` reboots the instance to pick up config/skills/agents/commands/MCP from disk (no new file surface).
- **v7.4.1** (2026-07-03) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.1) — PR #12042 nested `.gitignore`/`.kilocodeignore` respected during indexing (confirms `.kilocodeignore` is still the name).
[...]

PR #2448, opened 2026-07-28
fix(kilo): accept a bare MCP toggle entry, and type steps as the number it is

dyoshikawa:
Fixes the two outright bugs in #2417 — gaps 1 and 3. Gaps 2, 4 and 5 are left open; see the bottom.

## 1. A `{"enabled": false}` MCP entry took down the whole `--targets kilo` run

Kilo types `mcp` as `Record<string, Info | { enabled: boolean }>`: a bare toggle entry switches off a server another config layer defines — the global config, a marketplace, a VS Code import. Rulesync's union required `type`, so the entry failed to parse.

The damage was wider than the MCP feature. `kilo.jsonc` is the file the rules feature also patches, and the `KiloMcp` constructor parses the merged content, so a config with one such entry aborted the entire run with a ZodError and **nothing was written** — including rules, which have nothing to do with it.

The union gains the toggle arm. Such an entry imports as a canonical server carrying only `disabled: true`, with no transport of its own, and writing it back keeps that layer switched off.

## 2. `steps` was typed as a list of step objects

Upstream documents a positive integer: the cap on agentic iterations before a text-only response is forced (`maxSteps` is the deprecated alias). So importing a real `.kilo/agents/*.md` with `steps: 30` threw, and `kilo: { steps: 30 }` in a rulesync subagent was rejected — with the tests locking the wrong shape in.

## Refactor carried along

`convertFromKiloFormat` was a single closure doing tool-map splitting, transport branching and field mapping; adding a third arm pushed it past the repo's complexity ceiling. It is three named helpers now — `splitKiloServerTools`, `kiloServerToRulesync`, and the mapping itself — with the three union arms read through one shape, since they differ only by `type`.

## Left open

- **Gap 2, the `sandbox` block.** Needs a new authoring surface on the `kilo` permissions override (`enabled`, `network`, `writable_paths`, `allowed_hosts`), and upstream's scope rule is part of the design: a project config may only tighten, and `allowed_hosts` / `writable_paths` are honoured from the global config only. Worth doing deliberately rather than as a rider.
- **Gap 4, global non-root rules (`~/.kilo/rules/*.md`).** Needs the rules adapter to gain a global `nonRoot` path *and* the matching `getExtraSharedWritePaths` registration, so the files land in the global `kilo.jsonc` `instructions` array. Project scope already works.
- **Gap 5, `.kilo/rules-{mode}/`.** A mode dimension the rule model does not have; the issue rightly files it as lower priority.

## Tests

Two MCP assertions for the toggle entry (disabled and enabled), and the `steps` assertions rewritten onto the number the schema now takes — schema parse and the `fromRulesyncSubagent` round trip.

`pnpm cicheck` passes; the MCP and subagents e2e suites pass locally.

Refs #2417


dyoshikawa:
@dyoshikawa Thank you!

Issue #2416, opened 2026-07-27
Follow up deepagents-cli upstream updates: Hooks v2 schema, project scope, and the 2026-09-01 legacy-format removal


dyoshikawa:
## Summary

`deepagents-code` 0.1.45/0.1.46 (2026-07-22 / 07-24) landed **Hooks v2**: a new object-shaped `hooks.json` schema with PascalCase event names, a new project scope, and an explicit removal date of **September 1, 2026** for the flat-list format rulesync emits. The migration path is lossy — four of the eleven events rulesync currently emits are not migrated. This is a scheduled break rather than a live one, because Hooks v2 is not yet wired to the lifecycle.

**Package identity note**: the interactive CLI this target follows is the PyPI package **`deepagents-code`** (binary `dcode`), not the older `deepagents-cli` package. `deepagents-cli` is stale at 0.2.2 (2026-06-07); `deepagents-code` is at 0.1.47 (2026-07-24). This was already established in #2067.

## Recent Releases

Full [`deepagents-code` changelog](https://github.com/langchain-ai/deepagents/blob/main/libs/code/CHANGELOG.md) reviewed; entries touching a rulesync feature surface:

- **0.1.32** (2026-07-06) — legacy hook events `tool.use` / `tool.result`. Already followed (#2156).
- **0.1.35** (2026-07-09) — `[startup].mode` default approval-mode config key.
- **0.1.37** (2026-07-13) — env-var expansion inside MCP config (`command`/`url`/`args`/`env`/`headers`); `/tools` slash command.
- **0.1.38** (2026-07-14) — `memory.auto_save` config flag.
- **0.1.40** (2026-07-16) — plugin marketplace; project-MCP "always allow".
- **0.1.42** (2026-07-17) — **plugins GA**.
- **0.1.45** (2026-07-22) — **Hooks v2 execution engine + typed hooks data models** (#4880, #4870); `--allow-fs-tools`.
- **0.1.46** (2026-07-24) — **Hooks v2 capability snapshots, session transcripts, hardened command execution** (#4916, #4918, #4917); Auto approval mode GA.
- **0.1.47** (2026-07-24) — `yolo` in the Shift+Tab approval cycle.

### Upstream

Verified in the shipped **0.1.47 sdist** (not just `main`):

1. **New schema** — `deepagents_code/hooks/models/config.py`:

   ```python
   class CommandHandlerSpec(_ConfigModel):
       type: Literal["command"]
       command: str
[...]

PR #2449, opened 2026-07-28
docs(deepagents): correct the rules anchor in the research reference

dyoshikawa:
## Summary

Corrects the `rules` row of the client-anchors table in the deepagents research reference (`.rulesync/skills/rulesync-feature-research/references/deepagents.md`). The row claimed a `.deepagents/memories` non-root surface that `deepagents-rule.ts` has never emitted; it now states what the adapter actually does — project `.deepagents/AGENTS.md` and global `~/.deepagents/<agent>/AGENTS.md`, root-only, with non-root content folded into the root file.

Refs #2416 — this is the only actionable slice of that issue today. The main gap it tracks (Hooks v2 emission) stays gated on upstream: re-verified at deepagents-code 0.1.49 that hook dispatch still goes through the legacy flat-list loader (`hooks/__init__.py` re-exports `dispatch_hook` from `legacy.py`), which treats an object-shaped `hooks.json` as an empty config — so emitting v2 today would silently disable every hook. The issue stays open as the tracking issue; details in an issue comment.

## Changes

- `.rulesync/skills/rulesync-feature-research/references/deepagents.md`: fix the stale `rules` anchor row (table realigned by the formatter).

### DIFF
--- a/.rulesync/skills/rulesync-feature-research/references/deepagents.md
+++ b/.rulesync/skills/rulesync-feature-research/references/deepagents.md
@@ -18,10 +18,10 @@
 
 Common adapter paths: `rulesync-source-map.md`.
 
-| Surface     | Anchor                                                                                                     |
-| ----------- | ---------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2415, opened 2026-07-27
Follow up Zed upstream updates: untranslated MCP entries, unrecognized write/default permission keys, sandbox_permissions, global ignore, commands, Windows global path


dyoshikawa:
## Summary

Seven Zed drifts, several of which make rulesync emit config Zed silently ignores: MCP servers are forwarded untranslated (a `disabled` server is emitted **enabled**), canonical `write` maps to a tool name Zed does not recognize, there is no way to set `agent.tool_permissions.default`, `agent.sandbox_permissions` is unauthorable, ignore has no global scope, commands are unsupported although Zed skills *are* the slash-command surface, and every global path is wrong on Windows.

## Recent Releases

`gh release list -R zed-industries/zed` (stable channel), 2026-05 → 2026-07:

| Version | Date | Relevant change |
| --- | --- | --- |
| v1.4.2 | 2026-05-27 | Skills shipped; **Rules Library removed**; global `AGENTS.md` added |
| v1.5.3 | 2026-06-03 | Skill Creator; deprecated-Rules fixes |
| v1.7.2 / v1.8.2 | 2026-06-17 / 2026-06-24 | Agent **sandboxing** prompts + network permission handling |
| v1.9.0 | 2026-07-01 | `skill` tool wired into the Tool Permissions settings UI |
| v1.10.0 | 2026-07-08 | Windows agent sandboxing protects Git metadata |
| v1.12.0 | 2026-07-23 | Skill deletion UX (latest stable) |

Docs commits: `docs/src/ai/instructions.md` added 2026-06-04 (#57614, AI docs restructure); `docs/src/ai/sandboxing.md` added 2026-06-23 (#59774), security review 2026-07-10 (#60291).

Dry-run baseline:

```
--targets zed --features "*" --dry-run
  → .rules, .zed/settings.json (ignore + mcp + permissions)
  → commands / subagents / hooks / checks all "does not support the feature … Skipping."
--targets zed --features "*" --global --dry-run
  → .config/zed/AGENTS.md, .config/zed/settings.json (mcp + permissions), 41 skills
  → NO ignore file (global ignore unsupported)
```

### 1. mcp — `context_servers` entries are forwarded untranslated

[...]

Issue #1850, opened 2026-06-12
Follow up Zed upstream updates: permissions web-search tool name should be search_web (not web_search)


dyoshikawa:
## Summary

The Zed permissions adapter (added after #1705/#1750) maps the canonical `websearch` category to a Zed tool named `web_search`, but Zed's built-in web-search tool is actually named `search_web`. The generated `agent.tool_permissions.tools.web_search` entry is silently ignored by Zed, and a real `search_web` entry does not round-trip back to `websearch`.

## Recent Releases

- Zed `agent.tool_permissions` surface and the built-in tool list (`search_web`) are current as of 2026-06. Tool names verified against the authoritative tools doc. Sources: https://github.com/zed-industries/zed/blob/main/docs/src/ai/tools.md , https://zed.dev/docs/ai/tool-permissions

## Gaps

- **permissions — wrong tool name for web search (`bug`).** `src/features/permissions/zed-permissions.ts:56` maps `websearch: "web_search"`. Zed has no tool named `web_search`; the correct name is `search_web` (confirmed in `docs/src/ai/tools.md` and the tool-permissions supported-tools table). Effect: a generated `agent.tool_permissions.tools.web_search` entry is inert (Zed ignores it), and on reverse parse a real `search_web` entry falls through unmapped. The other mappings (`bash→terminal`, `read→read_file`, `edit→edit_file`, `webfetch→fetch`) are verified correct.

## Proposed Follow-up

- Change line 56 of `src/features/permissions/zed-permissions.ts` to `websearch: "search_web"`. The reverse map (`ZED_TO_CANONICAL_TOOL_NAMES`) is derived automatically, so no other code change is needed.
- Add a permissions test asserting `websearch ⇄ search_web` round-trips in `zed-permissions.test.ts`.

## References

- https://github.com/zed-industries/zed/blob/main/docs/src/ai/tools.md — built-in tool names, including `search_web`.
- https://zed.dev/docs/ai/tool-permissions — `agent.tool_permissions` supported-tools table.
- rulesync source: `src/features/permissions/zed-permissions.ts:56`.

PR #2450, opened 2026-07-28
fix(zed): translate MCP servers, map the global default and write_file, use the Windows config dir

dyoshikawa:
## Summary

Fixes gaps 1, 3, 4 and 7 of #2415 (Refs #2415 — gaps 2, 5 and 6 are feature additions and stay open there). Zed silently ignores config keys it does not know, so several canonical fields used to produce settings Zed reads as something else — most seriously, a server the user disabled in rulesync was emitted **enabled**.

## Changes

**Gap 1 — MCP translation (`zed-mcp.ts`).** `context_servers` entries are now translated into the shapes Zed's untagged enum actually parses, instead of forwarding canonical fields verbatim:

- `disabled: true` → `enabled: false` (and imports back as `disabled: true`); a server that says nothing about `disabled` writes no `enabled` key.
- `httpUrl` is normalized to the `url` Zed reads; an array `command` is flattened to Zed's single command string with the rest prepended to `args`.
- Canonical-only keys (`type`/`transport`, `alwaysAllow`, `trust`, `cwd`, `networkTimeout`, the Kiro lists…) are dropped — the drop list is derived from `McpServerSchema` itself, so unknown fields (a remote server's `oauth`, an extension server's `settings`) pass through untouched.
- A server Zed cannot start is skipped with a warning via the shared `mcp-transport.ts` helpers: `sse`/`ws` transports (Zed has neither), a remote server with no url, a local one with no command.
- A transport-less server is written as Zed's extension-provided variant, and on import lands in the tool-scoped `zed.mcpServers` block rather than the shared map (mirroring the Kilo/OpenCode handling from #2448).

**Gap 3 — global tool-permission default (`zed-permissions.ts`).** The canonical `*` category's catch-all rule now sets `agent.tool_permissions.default` (rung 6 of Zed's precedence ladder, and the documented mechanism for MCP tools) instead of an inert `tools["*"]` entry. Round-tripped on import; a stale `tools["*"]` entry written by an earlier version is cleaned up when the canonical config carries a `*` category; pattern-scoped `*`-category rules are dropped with a warning; a user-set `default` survives when canonical says nothing.

**Gap 4 — `write` mapping.** Canonical `write` maps to Zed's `write_file` tool (same class of bug as the fixed `web_search` → `search_web`, #1850). Also corrected the stale `websearch → web_search` claim in the docs.

**Gap 7 — Windows global path.** `getZedGlobalDir()` resolves the user config dir to `%APPDATA%\Zed` on Windows (per Zed's docs) for rules, MCP and permissions. Both spellings are declared to the shared-write derivation via `getExtraSharedWritePaths` and to the gateway ownership table, so the cross-check tests stay platform-complete; the zed-related e2e specs now use the platform-aware path.

## Verification

- `pnpm cicheck` green (unit tests include new cases for every behavior above).
- `e2e-mcp` / `e2e-permissions` / `e2e-rules` specs pass locally.
- Upstream claims re-verified against zed.dev docs (MCP shapes, tool-permission list and `default`, `%APPDATA%\Zed` paths).


dyoshikawa:
@dyoshikawa Thank you!

Issue #2236, opened 2026-07-14
Follow up Hermes Agent upstream updates: MCP advanced per-server fields (auth/oauth, mTLS, connect_timeout) dropped on export


dyoshikawa:
## Summary

Hermes Agent's `mcp_servers` entries recognize advanced per-server fields — `auth` (e.g. `oauth`), `client_cert`/`client_key` (mTLS), `connect_timeout`, `supports_parallel_tool_calls`, and `tools.{prompts,resources}` filters — but rulesync's `hermesagent` MCP adapter whitelists only a subset, so a user cannot express an OAuth- or mTLS-authenticated remote MCP server (or those tuning fields) for Hermes.

## Recent Releases

- MCP config reference (advanced `mcp_servers` fields): https://hermes-agent.nousresearch.com/docs (user-guide → features → MCP), NousResearch/hermes-agent, active through v0.18.2 (2026-07-07)
- Prior MCP follow-up (#2155, closed) added `tools.include/exclude` filtering; the auth/mTLS/timeout fields below remain dropped.

## Gaps

- **mcp (lossy export)** — `convertServerToHermes` (`src/features/mcp/hermesagent-mcp.ts`) whitelists only `command/args/env/url/headers/enabled/timeout/tools.{include,exclude}`. Hermes-recognized fields `auth` (`oauth`), `client_cert`, `client_key`, `connect_timeout`, `supports_parallel_tool_calls`, and `tools.{prompts,resources}` are dropped, so an OAuth/mTLS-authenticated remote MCP server cannot be authored for Hermes. Underlying cause is partly that canonical `McpServerSchema` (`src/types/mcp.ts`, looseObject) has no first-class fields for `auth`/`client_cert`/`connect_timeout`.

## Proposed Follow-up

- Pass the Hermes-recognized fields through in `convertServerToHermes` / `convertFromHermesFormat`.
- Consider adding `auth` / `connect_timeout` (and mTLS `client_cert`/`client_key`) to the canonical `McpServerSchema` so they can be authored portably rather than only via passthrough.

Severity: minor/niche, but a genuine can't-express gap for authenticated remote MCP servers.

## References

- https://hermes-agent.nousresearch.com/docs — MCP `mcp_servers` advanced fields (auth/oauth, mTLS, connect_timeout, supports_parallel_tool_calls, tools.prompts/resources).
- Local: `src/features/mcp/hermesagent-mcp.ts`, `src/types/mcp.ts`.

PR #2451, opened 2026-07-28
fix(hermesagent): carry ssl_verify/skip_preflight/sampling through MCP, drop the dead delegate_task toolsets argument

dyoshikawa:
## Summary

Fixes gaps 1 and 2 of #2414 (Refs #2414 — gap 3, project-scope skills via `skills.external_dirs`, is a design decision and stays open there).

## Changes

**Gap 1 — MCP keys dropped and destroyed on regenerate.** `copyHermesAdvancedFields` (`src/features/mcp/hermesagent-mcp.ts`) now carries three per-server keys Hermes documents but rulesync could neither author nor preserve:

- `ssl_verify` (`true`/`false` or a PEM CA-bundle path) — landed in the same upstream mTLS PR as `client_cert`/`client_key` but was missed by the #2236 fix, so a hand-written value on a rulesync-managed server was destroyed on regenerate (the merge is server-name-wholesale).
- `skip_preflight` (bool, HTTP preflight probe bypass, v0.19.0).
- `sampling` (server-initiated LLM request policy) — cloned as an opaque mapping with prototype-pollution keys stripped, so new upstream sub-keys keep working.

All three now flow in both directions (canonical → `config.yaml`, and import → the `hermesagent.mcpServers.<name>` override block), covered by the extended #2236 round-trip test plus a new dedicated test.

**Gap 2 — dead `delegate_task` `toolsets` argument.** `delegate_task` has no model-facing `toolsets` parameter (a subagent inherits the parent's enabled toolsets; upstream `tools/delegate_tool.py` states this outright, documented in v0.19.0). The generated subagent and command plugins no longer pass `toolsets` in the dispatch payload, and the emitted `.hermes/rulesync/subagents/*.json` spec no longer advertises one — it was dead schema a reader or importer would wrongly trust. Negative assertions added to both plugin tests.

Docs: the Hermes Agent MCP note in `docs/reference/file-formats.md` (and the synced skill copy) now lists the three fields.

## Verification

- Upstream claims re-verified against `mcp-config-reference.md` (server-key table) and `tools/delegate_tool.py` (no model-facing `toolsets`) on NousResearch/hermes-agent main.
- `pnpm cicheck` green.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2410, opened 2026-07-27
Follow up Amp upstream updates: globs: frontmatter never emitted, skill-bundled mcp.json, plugin-API custom agents, subtree checks


dyoshikawa:
## Summary

Four Amp surfaces are unfollowed: `globs:` frontmatter on non-root guidance files (rulesync renders the value only as advisory TOON prose, so Amp's native conditional loading never engages), skill-bundled `mcp.json` — the layout Amp explicitly recommends over user settings — plugin-API custom agents, and subtree-scoped check directories.

## Recent Releases

Primary sources: [ampcode.com/manual](https://ampcode.com/manual), [plugin API reference](https://ampcode.com/manual/plugin-api), and the [news feed](https://ampcode.com/news).

- **2026-07-23** — [Event-Driven Orbs](https://ampcode.com/news/event-driven-orbs) — orbs react to external events; Amp generates a project plugin from a natural-language request.
- **2026-07-21** — [Schedule](https://ampcode.com/news/schedule) — agents set their own schedules (runtime, in-thread).
- **2026-07-09** — [The Dial](https://ampcode.com/news/the-dial) — modes `low` / `medium` / `high` / `ultra`; deprecates smart/deep/rush/large.
- **2026-06-19** — [Custom Agents](https://ampcode.com/news/custom-agents) — plugins can create agents (`amp.experimental.createAgent`), exposed via `registerTool` / `registerAgentMode`.
- **2026-01-29** — [Slashing Custom Commands](https://ampcode.com/news/slashing-custom-commands) — custom commands (`.agents/commands/`, `~/.config/amp/commands/`) removed in favor of skills.
- **2026-01-08** — [Lazy-Load MCP with Skills](https://ampcode.com/news/lazy-load-mcp-with-skills) — skill-bundled `mcp.json`.
- **2025-09-23** — [Globs in AGENTS.md](https://ampcode.com/news/globs-in-AGENTS.md) — `globs:` frontmatter for @-mentioned guidance files (older; never followed).

### 1. rules — `globs:` frontmatter on non-root rule files

- **Upstream**: a file @-mentioned from `AGENTS.md` may carry YAML frontmatter:

  ```yaml
  ---
  globs:
    - '**/*.ts'
    - '**/*.tsx'
  ---
  ```

[...]

PR #2455, opened 2026-07-28
fix(amp): emit the globs frontmatter Amp's conditional loading actually reads

dyoshikawa:
## Summary

Fixes gap 1 of #2410 (Refs #2410 — gaps 2, 3 and 4 are design decisions and stay open there).

Amp gates an @-mentioned guidance file on `globs:` YAML frontmatter — *"Mentioned files with `globs` will only be included if Amp has read a file matching any of the globs"*; without it the file is **always** included. Rulesync's docstrings claimed globs support, but the generated `.agents/memories/*.md` files carried no frontmatter at all: the canonical `globs` value surfaced only as the advisory `applyTo` prose in the root file's TOON table, which Amp never enforces, so every non-root rule was loaded unconditionally.

## Changes

- `AmpRule.fromRulesyncRule` prepends `globs:` YAML frontmatter (via `stringifyFrontmatter`) to a non-root rule that carries globs — including a subtree `AGENTS.md` produced via `agentsmd.subprojectPath`. Canonical globs pass through verbatim (Amp implicitly prefixes `**/` unless the glob starts with `./` or `../`). Rules without globs stay frontmatter-free (no gate to write). The TOON `applyTo` prose is kept as redundant advisory context, per the issue's proposal.
- `AmpRule.fromFile` parses the frontmatter back into the canonical `globs`, and `toRulesyncRule` strips it from the rule body, so generate → import round-trips keep the condition instead of flattening it into an always-loaded rule.
- Docs: new Amp note in the rules section of `docs/reference/file-formats.md` (skill copy synced).

## Verification

- Upstream behavior quoted from [ampcode.com/manual](https://ampcode.com/manual) and [Globs in AGENTS.md](https://ampcode.com/news/globs-in-AGENTS.md) (re-verified in the issue two days ago).
- `pnpm cicheck` green; `e2e-rules` spec passes locally. New unit tests: frontmatter emitted with globs, omitted without, restored on import with the body stripped.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2409, opened 2026-07-27
Follow up Roo Code upstream updates (final state, EOL): command allow/deny permissions, global commands, skill modeSlugs, AGENTS.local.md


dyoshikawa:
## Summary

Roo Code's VS Code extension was **shut down on 2026-05-15** — `v3.54.0` is the final release and `RooCodeInc/Roo-Code` is archived. These are therefore final-state gaps against a frozen upstream, not a moving target: a file-configurable command allow/deny permission surface rulesync does not emit, global-scope commands, dropped skill `modeSlugs`, and the `AGENTS.local.md` personal override.

`docs.roocode.com` now 301-redirects to `https://roocodeinc.github.io/Roo-Code/`, which carries the shutdown notice and points users at the ZooCode community fork and Cline.

## Recent Releases

- **v3.54.0** (2026-05-15) — [release](https://github.com/RooCodeInc/Roo-Code/releases/tag/v3.54.0) — final release; shutdown commits (remove cloud/telemetry/MCP marketplace/web app, move docs to GitHub Pages). No CHANGELOG entry; the body is just "Release v3.54.0".
- **v3.53.0** (2026-04-23) — [release](https://github.com/RooCodeInc/Roo-Code/releases/tag/v3.53.0) — sunsetting announcement plus model additions only.
- **v3.52.1** (2026-04-13) — [release](https://github.com/RooCodeInc/Roo-Code/releases/tag/v3.52.1) — added a correct draft-07 JSON schema for `.roomodes` (PR #11791); documents tool groups `read|edit|command|mcp|modes` and the tuple form `["edit", {fileRegex, description}]`.
- **v3.51.0** (2026-03-05) — "exposing skills as slash commands with skill fallback execution" (runtime invocation behavior, no new on-disk format).
- Older but still unfollowed: **v3.48.0** (2026-02-17) removed the `browser` tool group and built-in skills; **v3.47.0** added `AGENTS.local.md` and skills `modeSlugs`.

### 1. permissions — the command allow/deny list is file-configurable, rulesync emits nothing

- **Upstream**: `roo-cline.allowedCommands` and `roo-cline.deniedCommands` (both `"type": "array"`, `"scope": "window"` in `src/package.json` at tag v3.54.0, so they are settable in workspace `.vscode/settings.json`). Longest-prefix match; deny wins on ties. — [auto-approving actions](https://roocodeinc.github.io/Roo-Code/features/auto-approving-actions).
[...]

PR #2456, opened 2026-07-28
feat(roo): global commands, skill modeSlugs round-trip, AGENTS.local.md (final EOL-state parity)

dyoshikawa:
## Summary

Implements items 2, 3 and 4 of #2409 (Refs #2409). Roo Code shut down on 2026-05-15 — `v3.54.0` is the final release and the repository is archived — so these are one-time parity fixes against a frozen upstream. Items 1 (a new `RooPermissions` surface) and 5 (mode-specific rule directories) are deliberately declined given the EOL; the disposition is recorded in the issue.

## Changes

- **Commands, global scope (item 2).** Roo reads project `.roo/commands/` and global `~/.roo/commands/` (project wins on a name collision; verified in `src/services/command/commands.ts` at the final tag). `supportsGlobal` flipped to `true` with the same no-branch relative-dir approach `RooRule`/`RooSkill` already use; matrix regenerated; the commands e2e global matrix gains a roo row.
- **Skills `modeSlugs` (item 3).** The `roo:` section of the canonical skill frontmatter existed in the schema but no adapter read it. `RooSkill.fromRulesyncSkill` now spreads it into the generated `SKILL.md` frontmatter (canonical `name`/`description` win over stray section keys), and `toRulesyncSkill` lifts every key beyond `name`/`description` back into the section — so `modeSlugs` mode targeting (and anything else Roo's frozen schema accepts) round-trips instead of being dropped in both directions.
- **`AGENTS.local.md` (item 4).** The roo rules factory gains `localRootMode: "separate-local-file"` + `localRootFileName: "AGENTS.local.md"` (mirroring rovodev), plus the matching `buildLocalRootFile` branch — a `localRoot: true` rule now gets its own personal, gitignored override file instead of being folded into the checked-in `AGENTS.md`. The `**/AGENTS.local.md` gitignore entry already exists in the common scope.

Docs: a Roo skills/commands note added to `docs/reference/file-formats.md` (final-state, EOL framing), the `localRoot` frontmatter comment updated, skill copies synced.

## Verification

- Upstream claims are from the issue's verification against the archived `v3.54.0` tag and the GitHub-Pages docs (frozen — no re-drift possible).
- `pnpm cicheck` green; `e2e-commands` + `e2e-rules` specs pass locally.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2405, opened 2026-07-27
Follow up Cline upstream updates: required agent description, global modular rules, plugin-based hooks, deprecated .clinerules/.clineignore


dyoshikawa:
## Summary

Cline's SDK/CLI config contract has moved: configured agents now require `description` (rulesync can emit files Cline refuses to load), global modular rules exist at `~/.cline/rules/` but rulesync drops them, hooks have converged on the documented **plugin** surface (which invalidates the rationale for closing #1847), `.clinerules` is literally named `DEPRECATED_CONFIG_DIR` upstream, and `.clineignore` is now documented as "deprecate soon".

Upstream baselines at time of research: Cline (VS Code) **v4.0.11** (2026-07-24), Cline CLI **v3.0.46** (2026-07-19), SDK **v0.0.65** (2026-07-19).

## Recent Releases

- **v4.0.0** (2026-06-26) — [release](https://github.com/cline/cline/releases/tag/v4.0.0) — SDK-backed extension runtime, **Cline Plugins**, Customize marketplace, MCP file/format migration into shared settings.
- **v4.0.1** (2026-06-28) — [release](https://github.com/cline/cline/releases/tag/v4.0.1) — the VS Code extension was **rolled back** to the pre-SDK (3.89.2) codebase; the SDK migration continues on `main`. This is why the shipping extension still resolves only the legacy `.clinerules` paths.
- **v4.0.11** (2026-07-24) — [release](https://github.com/cline/cline/releases/tag/v4.0.11) — model additions only.
- **cli-v3.0.23** (2026-06-10) — "Added support for configured agents as subagent tools".
- **cli-v3.0.25 / v3.0.27** (2026-06-17) — MCP servers inside plugins; `cline skill` install/manage command.
- Docs `0b0ca9b1` (2026-05-12, PR #10350) — **deleted the 514-line file-hooks reference**, leaving `docs/customization/hooks` a stub pointing at `/sdk/plugins`.
- Docs `b919a7e8` (2026-07-21, PR #12410) — **`.clineignore` marked "deprecate soon"**.
- Docs `9466fcc0` (2026-07-26, PR #12154) — hooks stub retitled "See details under SDK Plugins page".

### Upstream path surface at v4.0.11

[...]

PR #2459, opened 2026-07-28
fix(cline): require the agent description Cline enforces, emit global modular rules, note the clineignore deprecation

dyoshikawa:
## Summary

Fixes gaps 1, 2 and 5 of #2405 (Refs #2405 — gap 3, hooks as a Cline plugin generator, and gap 4, the forward-looking `.cline/*` roots, stay open there).

## Changes

**Gap 1 — agent contract.** Cline's `ConfiguredAgentFrontmatterSchema` requires a non-empty `description` (cli-v3.0.23+); a file without one is refused and the agent silently skipped. The `ClineSubagentFrontmatterSchema` now matches upstream (`name`/`description` required non-empty), a canonical subagent that omits a description gets a minimal generated fallback (`<name> subagent`) instead of a file Cline cannot load, the upstream optional fields `tools`/`skills`/`providerId`/`modelId`/`maxIterations` are typed (still authored via the `cline:` section), and import scans `*.{yaml,yml}` to match Cline's `isYamlFile()`.

**Gap 2 — global modular rules.** Global mode previously emitted only `~/.agents/AGENTS.md` and threw for non-root rules, silently dropping them. Non-root global rules are now emitted to `~/Documents/Cline/Rules/*.md` — the global modular-rules directory read by **both** the VS Code extension and the SDK/CLI (the SDK/CLI additionally reads `~/.cline/rules/`), mirroring the existing `~/Documents/Cline/Workflows` commands precedent — with the same conditional-frontmatter conversion project rules get, and imported back from there. The root file is unchanged.

**Gap 5 — `.clineignore` deprecation.** No code change (upstream still reads it); the docs now qualify the matrix entry: upstream titles the page "deprecate soon" and states it is not a security boundary, with the plugin surface as the replacement direction.

Docs notes updated for all three; the shared processor test that used cline as its root-only-global example now uses amp (cline no longer qualifies).

## Verification

- Upstream claims are from the issue's research against Cline v4.0.11 / cli-v3.0.46 sources (`configured-agent-config.ts`, `resolveRulesConfigSearchPaths()`, docs commits).
- `pnpm cicheck` green; unit tests cover the description fallback, empty-description import rejection, typed fields, `.yml` import, and global non-root generate/import.


dyoshikawa:
@dyoshikawa Thank you!

PR #2461, opened 2026-07-28
feat(mcp): add a per-server enabled generation filter to the canonical config

### LINKED ISSUE #2433: Add an `enabled` parameter to .rulesync/.mcp.json{c} so only enabled servers are generated
## Background

`.rulesync/.mcp.json` / `.mcp.jsonc` currently has no first-class way to keep a server definition in the file while temporarily excluding it from generated tool configs. Today the options are:

- Delete the entry (loses the definition; you have to re-type it later).
- Comment it out (only works in `.jsonc`, and the block is then invisible to the schema/tooling).
- Set the canonical `disabled: true` field — but that is a **pass-through** field, not a filter: it is written into the generated tool config (as `disabled: true`, or translated to `enabled: false` for OpenCode / Kilo Code / Grok CLI / Goose). The server still appears in every tool's config, just marked off.

What is missing is a generation-time toggle: keep the definition in the source file, but do not emit it at all. An `enabled` parameter on each server, where only `enabled: true` servers are written out, gives that.

## Details

Requested shape for `.rulesync/.mcp.json{c}`:

```json
{
  "$schema": "https://github.com/dyoshikawa/rulesync/releases/latest/download/mcp-schema.json",
  "mcpServers": {
    "rulesync": {
      "type": "stdio",
      "command": "rulesync",
      "args": ["mcp"],
      "enabled": true
    },
    "playwright": {
      "type": "stdio",
      "command": "pnpm",
      "args": ["dlx", "@playwright/mcp", "--headless"],
      "enabled": true
    }
  }
}
```

Semantics: only servers with `enabled: true` are emitted to the generated tool MCP configs; the rest stay in the source file untouched and produce no output.

### Design points to settle

1. **Default when `enabled` is omitted.** It must default to `true`, otherwise every existing `.rulesync/.mcp.json` in the wild would suddenly generate zero servers. So the effective rule is "emit unless `enabled: false`", and writing `enabled: true` explicitly (as in the example above) is documentation/opt-in clarity rather than a requirement.
2. **Relationship with the existing canonical `disabled`.** `disabled` lives in `McpServerSchema` (`src/types/mcp.ts`) and is a pass-through field; `enabled` would be a rulesync-source-only filter (like `targets`, `description`, `exposed`, `envVars`). Both can be set, so precedence needs a documented rule — the natural one is: `enabled: false` wins and drops the server entirely, and `disabled: true` only matters for servers that are still emitted.
3. **Do not leak `enabled` into tool output.** `RulesyncMcp.getMcpServers()` strips rulesync-only fields via `omit(serverConfig, ["targets", "description", "exposed", "envVars"])` (`src/features/mcp/rulesync-mcp.ts`). `enabled` must be added to that list. This matters more than usual: OpenCode, Kilo Code, Grok CLI, and Goose all have a *native* `enabled` field with different semantics, so a leaked value would be silently misinterpreted.
4. **Tool-scoped blocks.** `{toolname}.mcpServers` entries replace the shared entry wholesale, and `null` already removes a server for one tool. `enabled` inside a tool-scoped block should behave the same way as in the shared block — worth an explicit test so the two removal mechanisms (`null` vs `enabled: false`) don't diverge.
5. **All-disabled case.** If every server ends up filtered out, generation produces an empty `mcpServers` map. This should follow the empty-payload behavior established in ac8e3940 / 7739d79e rather than writing empty shared config files.
6. **Import round-trip.** `rulesync import` should not invent `enabled` for imported servers; a tool's native `enabled`/`disabled` should keep mapping to the canonical `disabled` as it does today.

## Solution / Next Steps

1. Add `enabled: z.optional(z.boolean())` to `RulesyncMcpServerSchema` in `src/features/mcp/rulesync-mcp.ts` (rulesync-source-only, not `McpServerSchema`), with a comment explaining the `disabled` distinction.
2. Filter out servers whose `enabled === false` during generation, and add `enabled` to the `omit(...)` list so it never reaches tool configs.
3. Apply the same filtering to tool-scoped `{toolname}.mcpServers` overrides.
4. Regenerate the published JSON schema (`scripts/generate-json-schema.ts`) so editors offer `enabled` with the right description.
5. Add tests: default-omitted emits, `enabled: false` skips, `enabled` never appears in any tool output (especially OpenCode / Kilo / Grok / Goose), tool-scoped override behavior, and the all-disabled empty-output case.
6. Document the field and its difference from `disabled` in `docs/` (and keep `skills/rulesync/` synchronized via `scripts/sync-skill-docs.ts`).

dyoshikawa:
## Summary

Closes #2433 — implements the requested `enabled` parameter exactly per the issue's design points:

1. **Default**: omitted = enabled (emit unless `enabled: false`), so existing configs keep generating everything.
2. **vs `disabled`**: `enabled: false` wins and drops the server entirely; the pass-through `disabled` only matters for servers still emitted. Documented in the schema comment and docs.
3. **No leakage**: `enabled` joins the `getMcpServers()` omit list (`targets`/`description`/`exposed`/`envVars`), so it never reaches tool output — OpenCode/Kilo/Grok CLI/Goose have a native `enabled` with different semantics.
4. **Tool-scoped blocks**: `forTarget()` merges `{toolname}.mcpServers` entries into the shared map before `getMcpServers()` runs, so the filter applies there identically (tested alongside the `null` removal mechanism).
5. **All-disabled**: yields an empty `mcpServers` map, following the existing empty-payload behavior.
6. **Import**: untouched — a tool's native enabled/disabled keeps mapping to the canonical `disabled`.

The published JSON schema picks the field up automatically from `RulesyncMcpServerSchema` (verified by running `generate:schema` locally; the artifacts are release-time outputs and gitignored). Docs note added before the deprecated per-server `targets` note.

## Verification

- `pnpm cicheck` green; new tests cover default-emit, explicit true, false-drop, disabled-pass-through-precedence, tool-scoped filtering, and the all-filtered empty map.


dyoshikawa:
@dyoshikawa Thank you!

PR #2462, opened 2026-07-28
feat(init): add Playwright MCP to the scaffolded .rulesync/.mcp.json

### LINKED ISSUE #2432: Add Playwright MCP to the mcp.json scaffolded by rulesync init
## Background

`rulesync init` scaffolds `.rulesync/.mcp.json` from the singleton template in `src/lib/feature-scaffold.ts` (`singletonTemplate("mcp")`). Today that template ships two servers:

- `deepwiki` (http) — repository documentation lookup
- `rulesync` (stdio) — rulesync's own MCP server

Both are useful defaults for reading docs and authoring rulesync files, but neither covers browser automation. Fetching content from SPA-style pages, verifying rendered UI, or driving a browser during a task is a common need for coding agents, and rulesync's own `CLAUDE.md` already tells contributors to fall back to a Playwright-based workflow when the built-in webfetch tool cannot retrieve a page. Making Playwright MCP part of the scaffolded default would give new users that capability out of the box, consistent with how we already dogfood it.

## Details

Proposed addition to the `mcpServers` block of the init template:

```json
"playwright": {
  "type": "stdio",
  "command": "pnpm",
  "args": ["dlx", "@playwright/mcp", "--headless"]
}
```

Notes and open points:

- The `command`/`args` shape matches the existing `rulesync` entry (`pnpm dlx ...`), so it stays consistent with the current template style. Existing entries also carry an explicit `"env": {}`, which the new entry should probably mirror.
- `--headless` is intentional for an agent context: no visible browser window during automated runs.
- `@playwright/mcp` downloads browser binaries on first use, so the first invocation is slow and requires network access. Worth deciding whether that cost is acceptable for a default scaffold, or whether the entry should be commented/documented as opt-in.
- Consider whether the package manager should stay hardcoded as `pnpm` (as the `rulesync` entry does) or whether `npx` would be a more neutral default for the scaffold. This applies to both entries, so it may be a separate follow-up.

## Solution / Next Steps

1. Add the `playwright` server entry to `singletonTemplate("mcp")` in `src/lib/feature-scaffold.ts`.
2. Update the affected tests that assert the scaffolded `.rulesync/.mcp.json` content (`src/lib/init.test.ts`, `src/e2e/e2e-init.spec.ts`, and any feature-scaffold snapshot).
3. Check whether `docs/` and `skills/rulesync/` show the init MCP template verbatim; if so, keep them synchronized (`scripts/sync-skill-docs.ts`).
4. Decide on the open points above (`env: {}`, headless, first-run browser download, `pnpm` vs `npx`) before landing.

dyoshikawa:
## Summary

Closes #2432 — adds the proposed `playwright` server to `singletonTemplate("mcp")` in `src/lib/feature-scaffold.ts`, matching the issue's requested shape.

Decisions on the issue's open points:
- **`env: {}`**: mirrored from the existing entries for template consistency.
- **`--headless`**: kept — no visible browser window in agent runs.
- **First-run browser download**: accepted for a default scaffold; the entry is trivially removable, and the repo already dogfoods a Playwright fallback per its own CLAUDE.md.
- **`pnpm` vs `npx`**: kept `pnpm dlx` for consistency with the existing `rulesync` entry; switching both to a more neutral runner is the separate follow-up the issue itself suggests.

Tests: the init MCP scaffold assertion now pins all three servers (docs did not embed the template verbatim, so no doc sync needed).

## Verification

- `pnpm cicheck` green; `e2e-init` passes locally.

### DIFF
--- a/src/lib/feature-scaffold.ts
+++ b/src/lib/feature-scaffold.ts
@@ -291,6 +291,16 @@ function singletonTemplate(feature: ScaffoldFeature): string {
         "mcp"
       ],
       "env": {}
+    },
+    "playwright": {
+      "type": "stdio",
+      "command": "pnpm",
+      "args": [
+        "dlx",
+        "@playwright/mcp",
+        "--headless"
+      ],
+      "env": {}
     }
   }
 }
--- a/src/lib/init.test.ts
+++ b/src/lib/init.test.ts
@@ -211,6 +211,13 @@ describe("init", () => {
       expect(writeCall).toBeDefined();
       const content = writeCall?.[1] ?? "";
       expect(content).toContain('"mcpServers"');
+      // The scaffolded defaults: docs lookup, rulesync's own server, and
+      // browser automation for SPA pages (issue #2432).
+      expect(content).toContain('"deepwiki"');
+      expect(content).toContain('"rulesync"');
+      expect(content).toContain('"playwright"');
+      expect(content).toContain('"@playwright/mcp"');
+      expect(content).toContain('"--headless"');
     });
 
     it("should not create a command sample file", async () => {


dyoshikawa:
@dyoshikawa Thank you!

PR #2491, opened 2026-07-30
feat: add rulesync doctor command for config diagnostics

### LINKED ISSUE #2446: Add a `rulesync doctor` command that diagnoses config and input-file correctness
## Background

Today the only way to find out whether a rulesync setup is sane is to run `rulesync generate` and see what happens. That has two problems:

- **Silent misconfiguration.** `ConfigFileSchema` (`src/config/config.ts`) is a plain `z.object`, so unknown keys in `rulesync.jsonc` are accepted and ignored. A typo such as `"target"` instead of `"targets"`, or a stale key from an older major version, produces no error at all — generation just quietly uses the defaults (`targets: ["agentsmd"]`, `features: ["rules"]`) and the user wonders why nothing was emitted for their tool.
- **Late, generation-shaped feedback.** Problems in `.rulesync/**` inputs (bad frontmatter, unknown tool targets, missing `SKILL.md`, invalid `.mcp.json` / permissions files) only surface once a generator happens to touch that file, and the error is phrased in terms of the generator's internals rather than "here is what is wrong with your setup".

A dedicated `rulesync doctor` command would give a single, fast, read-only answer to "is my rulesync configuration and input tree correct?" — the same role `npm doctor`, `flutter doctor`, `brew doctor`, or `biome check` play in their ecosystems. It is also a much better first thing to tell a user to run when they open an issue, and a cheap CI guard that is far more diagnostic than `generate --check`.

## Details

Proposed scope — `rulesync doctor` performs **read-only** diagnostics and never writes files.

### 1. Configuration file checks (`rulesync.jsonc`, `rulesync.local.jsonc`)

- JSONC parses at all; report the parse error with position instead of falling back silently.
- Validates against `ConfigFileSchema`, reported field-by-field rather than as a single Zod dump.
- **Unknown / misspelled keys**, with a "did you mean" suggestion. This is the highest-value check, because the current schema is non-strict and swallows them.
- Unknown values inside known keys: unrecognized tool targets (suggest the nearest of `ALL_TOOL_TARGETS`), unrecognized features.
- Rules that already exist as runtime assertions, surfaced up front: `assertTargetsFeaturesExclusive` (object-form `targets` combined with `features`), `CONFLICTING_TARGET_PAIRS` (`augmentcode` + `augmentcode-legacy`, `claudecode` + `claudecode-legacy`), legacy targets that wildcard expansion intentionally skips.
- Deprecated surfaces: `ignore` feature (superseded by `permissions`), and any other keys we have since retired.
- `$schema` presence and whether it points at the current `RULESYNC_CONFIG_SCHEMA_URL`.
- Path sanity: `inputRoot` / `outputRoots` / `baseDirs` exist and are directories; `sources[]` entries are well formed (transport vs. `registry`/`tokenEnv`/`ref`/`path` combinations, `tokenEnv` naming an env var that is actually set).
- Feature/target combinations that can never produce output — e.g. a feature requested for a target that does not support it, or simulated features requested in global scope.

### 2. Input tree checks (`.rulesync/**`)

- Input root exists and is non-empty; warn when it is missing entirely (user probably never ran `rulesync init`).
- Rules: frontmatter validity, exactly one root rule where that matters, valid `targets` / `globs`.
- Commands, subagents, skills: frontmatter validity, and skill directories that are missing `SKILL.md`.
- `.rulesync/.mcp.json` against the generated `mcp-schema.json`, and the permissions file against `permissions-schema.json` — including unknown-key detection, which is the same blind spot as the main config.
- Files that sit in `.rulesync/` but no enabled feature will ever read (dead input), reported as info.

### 3. Workspace checks

- Generated outputs are stale relative to the current inputs (can reuse the existing `--check` machinery, reported as one diagnostic rather than the whole diff).
- Expected `.gitignore` entries are missing (derived the same way `gitignore-derive.ts` does).
- Output files that no longer correspond to any input (leftovers from a target that was removed from the config).

### 4. Output and UX

- Diagnostics grouped by severity: `error` / `warning` / `info`, each with the offending file path (and line/column where JSONC position info is available), the reason, and the concrete fix.
- Exit code `1` when any `error` is present, `0` otherwise; a flag to make warnings fatal in CI (e.g. `--max-warnings 0` or `--strict`).
- Honors the existing global `-j, --json` flag so CI and editors can consume the result.
- Honors `--targets` / `--features` / `-c, --config` so the check can be scoped the same way `generate` is.
- No `--fix` in the first iteration; leave the auto-fixable subset (missing `$schema`, missing gitignore entries) for a follow-up.

## Solution / Next Steps

1. Add `src/cli/commands/doctor.ts` with a diagnostic collector type (`{ severity, code, file, message, hint }`) so checks stay small and independently testable.
2. Register the command in `src/cli/program.ts` via `wrapCommand("doctor", "DOCTOR_FAILED", ...)`, wiring `--config`, `--targets`, `--features`, `--verbose`, `--silent`.
3. Reuse existing validation instead of duplicating it: `ConfigFileSchema`, `assertTargetsFeaturesExclusive`, `expandWildcardTargets`, each feature class's `fromFile` validation, and the generated `config-schema.json` / `mcp-schema.json` / `permissions-schema.json`. Where a check needs unknown-key detection, derive the allowed key set from the schema rather than hard-coding a list, so it cannot drift.
4. Unit tests per check plus an end-to-end happy path (clean workspace → exit 0, no diagnostics) and a deliberately broken fixture workspace covering each diagnostic code.
5. Document the command in `docs/reference/cli-commands.md`, and mention it in the getting-started/troubleshooting flow as the first thing to run when generation does not behave as expected.
6. Follow-ups worth splitting out: `--fix` for the auto-fixable subset; making `ConfigFileSchema` strict (or emitting a warning at load time) so plain `generate` also stops silently ignoring unknown keys; exposing doctor through the MCP server.

dyoshikawa:
## Summary

First slice of #2446: adds a read-only `rulesync doctor` command that diagnoses the configuration files (`rulesync.jsonc` / `rulesync.local.jsonc`) and reports problems grouped by severity (`error` / `warning` / `info`).

Closes #2446 (config-file check group; input-tree and workspace check groups are follow-ups as noted in the issue).

## Why

Today the only feedback channel for a broken setup is `rulesync generate`. Because `ConfigFileSchema` is non-strict, a misspelled key such as `"target"` instead of `"targets"` is silently swallowed and generation quietly falls back to defaults. `doctor` gives a single, fast, read-only answer to "is my configuration correct?" — usable both as a first troubleshooting step and as a CI guard.

## What it checks

- JSONC parse errors, with line/column
- Unknown or misspelled top-level keys, with "did you mean" suggestions derived from `ConfigFileSchema` (no hard-coded key list)
- Unknown tool targets / features (array and object forms), with nearest-name suggestions
- Deprecated `ignore` feature (warning)
- Object-form `targets` combined with `features`, including the case where the conflict only appears after merging base and local config files
- Conflicting target pairs (reuses the exported `CONFLICTING_TARGET_PAIRS`)
- `$schema` presence (info) and currency (warning)
- Structural violations on other keys via `ConfigFileSchema.safeParse`
- `sources[].tokenEnv` naming an unset environment variable (warning)
- `inputRoot` pointing at a nonexistent directory (error)

## UX

- Exit code `1` when any error is present; `--strict` makes warnings fatal for CI
- Honors the global `-j, --json` flag (diagnostics + severity summary as structured data)
- `-c, --config`, `-V, --verbose`, `-s, --silent` wired like other commands
- Never writes files

## Example

```
$ rulesync doctor
✖ rulesync.jsonc [config/unknown-key] Unknown key 'target'. It is silently ignored by 'rulesync generate'.
    ↳ Did you mean 'targets'?
✖ rulesync.jsonc [config/unknown-target] Unknown tool target 'cursorr' in the 'targets' object.
    ↳ Did you mean 'cursor'?
⚠ rulesync.jsonc [config/deprecated-feature] Feature 'ignore' in 'features' is deprecated.
    ↳ Use the 'permissions' feature instead.
Doctor found problems: 2 error(s), 1 warning(s), 0 info(s).
```

## Tests

- Unit tests per check (40+ cases) plus command-level tests: clean-workspace happy path (exit 0, no diagnostics), broken fixtures per diagnostic code, `--strict`, `--config`, JSON capture, base+local merge conflict
- `pnpm cicheck` passes


dyoshikawa:
@dyoshikawa Thank you!

PR #2514, opened 2026-07-30
fix(skills): lenient per-skill validation on agentsskills import

### LINKED ISSUE #2513: Follow up AgentsSkills upstream updates: skills import aborts whole run on one non-conformant SKILL.md (spec prescribes lenient per-skill validation)
## Summary

One AgentsSkills gap: the Agent Skills client-implementation guide prescribes **lenient, per-skill validation** on read ("skip the skill, log the error"), but rulesync's skills import aborts the entire run on one non-conformant `SKILL.md` — a single malformed neighbor in `.agents/skills/` (the cross-vendor interoperability directory, where foreign-authored skills are most likely) blocks importing everything.

Upstream state: `agentskills/agentskills` has no releases/tags ever (unversioned spec); the only normative edit in the window was the 2026-05-16 `name`-digits clarification ([commit](https://github.com/agentskills/agentskills/commit/6868401b)), already satisfied by rulesync's `NAME_PATTERN`. The #2428 agentsmd/agentsskills double-write no longer reproduces in either target order.

### skills (import) — one non-conformant SKILL.md aborts the whole import

- **Upstream**: the "Lenient validation" section of the [client implementation guide](https://agentskills.io/client-implementation/adding-skills-support) (unchanged since 2026-04-01): "Description is missing or empty → **skip the skill**…, log the error"; "YAML is completely unparseable → **skip the skill**, log the error". It names the exact malformed-YAML case to tolerate — an unquoted `description` containing a colon — recommending clients "wrap such values in quotes or convert them to YAML block scalars before retrying."
- **rulesync**: reproduced with a valid `.agents/skills/good/SKILL.md` plus one bad sibling:
  - missing `description` → `Invalid frontmatter … Zod raw error …` and **nothing** is imported (thrown in `AgentsSkillsSkill.fromDir`, `src/features/skills/agentsskills-skill.ts`);
  - `description: Use this skill when: the user asks about PDFs` → `YAMLException: bad indentation of a mapping entry`, again importing nothing (thrown in `src/utils/frontmatter.ts` via the shared `loadSkillDirContent`).
  - Baseline with the bad sibling removed: `Imported 1 file(s) total (1 skills)`.
- **Note**: the throw lives in shared code, so other skills targets are likely affected too, but `agentsskills` is where it bites hardest — `.agents/skills/` is by definition the directory most likely to hold skills authored by another vendor's tool, exactly the population the leniency rules exist for.

## Proposed Follow-up

Make skills import per-skill fault-tolerant for `agentsskills`: catch both the YAML parse failure and the frontmatter schema failure per skill directory, log the offending path with the reason, skip that skill, continue with the rest, and exit non-zero only if nothing at all could be imported. Optionally add the guide's recommended retry for the unquoted-colon case (re-parse with the scalar quoted) before giving up. Leniency belongs on the import side only — `collectAgentSkillViolations` on the generate side already warns rather than fails, which is correct.

## Non-gaps (verified 2026-07-30)

- No new frontmatter fields, layout, metadata/versioning, or discovery changes in the window; `agentsskills-skill.ts` already encodes every normative constraint (name pattern incl. digits, description 1–1024, compatibility ≤500 string, metadata string map, `allowed-tools` space-joined).
- Both scopes verified by dry-run (`.agents/skills/` project; `~/.agents/skills/` global).
- #2428's shared-path double-write is fixed — both target orders produce byte-identical spec-conformant output.
- The spec remains unversioned (no releases/tags), so there is no version surface to follow; all other features stay correctly unsupported.

## References

- https://agentskills.io/client-implementation/adding-skills-support — the lenient-validation prescription (the gap).
- https://agentskills.io/specification — current normative spec (fetched 2026-07-30).
- https://github.com/agentskills/agentskills/commit/6868401b — last normative edit (2026-05-16, name digits).

dyoshikawa:
## Summary

Resolves the AgentsSkills follow-up: rulesync's skills import aborted the entire run when one non-conformant `SKILL.md` sat in `.agents/skills/` — a single malformed neighbor blocked importing every valid sibling.

The Agent Skills [client-implementation guide](https://agentskills.io/client-implementation/adding-skills-support) prescribes lenient per-skill validation on read: "Description is missing or empty → skip the skill, log the error"; "YAML is completely unparseable → skip the skill, log the error". `.agents/skills/` is the cross-vendor interoperability directory, exactly where foreign-authored (potentially non-conformant) skills are most likely to live.

## Changes

- Add an optional `lenientImport` flag to the skills-processor factory `meta` and enable it for the `agentsskills` target.
- The import loop's existing configured-root warn-and-skip tolerance now also applies to declared roots of lenient targets: a skill directory that fails to load (unparseable YAML or invalid frontmatter) is warned about with its path and reason, skipped, and the rest of the import continues.
- Tests: missing-`description` sibling and unquoted-colon YAML sibling each import the valid skill and warn; non-lenient targets (e.g. rovodev) still abort on an invalid declared-root skill.

Generate-side behavior is unchanged — `collectAgentSkillViolations` already warns rather than fails there.

Closes #2513

### DIFF
--- a/src/features/skills/skills-processor.test.ts
+++ b/src/features/skills/skills-processor.test.ts
@@ -7,6 +7,7 @@ import { RULESYNC_SKILLS_RELATIVE_DIR_PATH } from "../../constants/rulesync-path
 import { createMockLogger } from "../../test-utils/mock-logger.js";
 import { setupTestDirectory } from "../../test-utils/test-directories.js";
 import { ensureDir, writeFileContent } from "../../utils/file.js";
+import { AgentsSkillsSkill } from "./agentsskills-skill.js";
 import { ClaudecodeSkill } from "./claudecode-skill.js";
 import { RovodevSkill } from "./rovodev-skill.js";
 import { RulesyncSkill } from "./rulesync-skill.js";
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2516, opened 2026-07-31
fix(watch): detect replaced watched directory by inode to avoid dead watcher

dyoshikawa:
## Summary

Fixes the deterministic CI timeouts in `src/lib/watch.test.ts` (`re-attaches after the watched directory is deleted and recreated` and `re-attaches a filtered target after its directory is deleted and recreated`), which currently block unrelated PRs (e.g. #2514 failed 3 consecutive runs) and have also failed on `main` (run 30542952793).

## Root cause

`verifyStillWatching` only checked `existsSync` on the watched path. When the directory is deleted and recreated before the delete event is delivered — fast branch switches locally, slow event queues on CI runners — the path exists again by the time the callback runs, so the check reports the watcher alive while `fs.watch` is still bound to the dead inode. No event ever fires again and the re-arm loop never starts.

## Fix

Record the watched directory's inode (bigint stats to avoid truncation) on attach. On liveness checks, an existing path with a *different* inode is treated as a replacement: close the dead watcher, re-arm, and emit the directory-changed notification on re-attach. If either stat is unreadable, fall back to the previous existence-only behavior, so platforms with unreliable inode numbers keep the old semantics.

No test changes needed — the existing re-attach tests exercise exactly this race and become deterministic with the fix (verified 10 consecutive local runs green, full `pnpm cicheck` green).

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -3,6 +3,13 @@ import { join } from "node:path";
 
 import { afterEach, describe, expect, it, vi } from "vitest";
 
+const { statSyncMock } = vi.hoisted(() => ({ statSyncMock: vi.fn() }));
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  statSyncMock.mockImplementation(actual.statSync);
+  return { ...actual, statSync: statSyncMock };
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2517, opened 2026-07-31
feat(warp): target execution_profiles for permissions, emit global ~/.agents/AGENTS.md rules

dyoshikawa:
## Summary

Resolves the Warp follow-up: two gaps grounded in Warp's now-open-source client code.

### Permissions: retarget at `[agents.execution_profiles.default]`

Since file-backed execution profiles went Stable ([warpdotdev/warp PR #14418](https://github.com/warpdotdev/warp/blob/main/specs/file-backed-execution-profile-collection/TECH.md), 2026-07-28), the legacy `[agents.profiles]` keys rulesync wrote (`agent_mode_command_execution_allowlist`/`denylist`) are consumed only once by Warp's one-shot migration and are otherwise inert — runtime enforcement reads the active `[agents.execution_profiles.<id>]` record.

- `WarpPermissions` now merges `bash` allow → `command_allowlist` and `bash` deny → `command_denylist` into the `default` execution profile **in place**, preserving every other profile key and profile ID (`ExecutionProfileFile` is `#[serde(default)]`, so a partial record stays valid).
- The legacy block is still written for un-migrated installs and old clients (per Warp's own transition guidance).
- The collection is deliberately **not created** when absent: on an un-migrated install the legacy keys are still live, and creating the collection would mark Warp's one-shot migration complete early and strand the user's other legacy settings (file-read allowlist, preferred model).
- Import prefers the `default` profile's lists and falls back to the legacy keys.

### Rules: emit global `~/.agents/AGENTS.md`

Warp reads a third rule source from `~/.agents/AGENTS.md` ([warpdotdev/warp PR #9325](https://github.com/warpdotdev/warp/blob/main/specs/APP-3893/TECH.md), shipped 2026-05-13; also used from remote hosts in SSH sessions; `~/.agents/` is listed in [file-locations](https://docs.warp.dev/terminal/settings/file-locations/)). rulesync's Warp rules were project-only.

- `WarpRule` now supports global scope targeting `.agents/AGENTS.md` (home-relative) with the same root-fold collision policy.
- Sharing that path with other targets that own it in global mode (e.g. cline) matches the existing project-scope behavior, where multiple targets already share the root `AGENTS.md` per their own semantics.

### Maintenance

- Refreshes the stale `rulesync-feature-research` Warp reference map (it claimed ignore/mcp/commands/skills/permissions were unsupported — all are implemented).
- Supported-tools tables regenerated via `pnpm run generate:tables`.

[...]

PR #2520, opened 2026-07-31
test(watch): raise fs.watch integration test timeouts to 20s

dyoshikawa:
## Summary

The three `fs.watch` integration tests in `src/lib/watch.test.ts` keep flaking on CI with `Test timed out in 5000ms` (latest: PR #2517 run 30597223224, also PR #2514 runs and main run 30542952793). PR #2516 fixed the deterministic dead-watcher race, but OS event delivery on loaded CI runners can still stall past the 5s default per-test timeout — the `waitFor` helper polls with a 10s budget that the 5s test timeout undercuts.

Raise the per-test timeout to 20s for the three event-delivery-dependent tests only; the mocked replaced-inode test stays on the default since it is deterministic.

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -291,113 +291,131 @@ describe("formatTriggerPaths", () => {
 });
 
 describe("watchTargets", () => {
-  it("forwards changes under a recursively watched directory", async () => {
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2529, opened 2026-07-31
feat: add rulesync docs command for bundled documentation lookup and search

### LINKED ISSUE #2519: Add a rulesync docs command for bundled documentation lookup and search
## Background

Rulesync already maintains structured Markdown documentation under `docs/`, but users and coding agents currently need to browse the repository or website to retrieve it. A CLI command that exposes the bundled documentation would make help available directly in the terminal and would give agents a simple, deterministic way to consume the same documentation as humans.

## Details

Add a new `rulesync docs` command that follows the existing `docs/` hierarchy and prints Markdown documents to standard output.

Examples:

- `rulesync docs faq` prints the content of `docs/faq.md`.
- `rulesync docs guide/configuration` prints the content of `docs/guide/configuration.md`.
- Nested document identifiers should map cleanly to the existing directory structure without requiring callers to include the `docs/` prefix or `.md` extension.

The command should also support ranked full-text search across the bundled documentation:

- Prefer `rulesync docs --search <text>` as the public interface.
- `--search` communicates both the action and its value more clearly than `--query`, which sounds like a secondary parameter and does not identify what the command will do.
- Use [MiniSearch](https://github.com/lucaong/minisearch) as the in-memory search engine. It provides BM25+ relevance ranking, field boosting, TypeScript support, and zero transitive dependencies without requiring SQLite or an external service.
- Index document paths, titles, headings, and Markdown content, with stronger boosts for titles and headings than body content.
- Keep the initial query behavior conservative. Exact term matching and BM25+ ranking should be the baseline; prefix or fuzzy matching can be enabled only if tests demonstrate useful results without excessive false positives.
- Results should identify the document path and provide enough matching context to locate the text.
- Define predictable behavior for no matches and multiple matches, including exit codes and output format.

The documentation must be available in installed or compiled Rulesync distributions, not only when running from a source checkout. Document identifiers must be resolved safely: reject traversal and absolute paths, and only expose the intended bundled Markdown tree.

### Simplify the distributed Rulesync skill

Once the CLI can retrieve and search the bundled documentation, the official skill under `skills/rulesync/` no longer needs to duplicate the entire documentation set as a flat collection of Markdown files.

- Distribute only `skills/rulesync/SKILL.md`.
- Keep the skill intentionally small: briefly explain what Rulesync is, then instruct the agent to use `rulesync docs <document>` for a known document and `rulesync docs --search <text>` to discover relevant documentation.
- Tell the agent to consult `rulesync docs` before answering Rulesync-specific questions or performing unfamiliar Rulesync operations, so the CLI documentation remains the canonical detailed reference.
- Remove the generated sibling files such as `faq.md`, `configuration.md`, and `cli-commands.md` from the distributed skill.
- Simplify or replace `scripts/sync-skill-docs.ts`; it should generate or validate only the single `SKILL.md` instead of flattening the VitePress hierarchy, rewriting links, checking filename collisions, and mirroring every documentation page.
- Preserve the existing `docs/` hierarchy as the canonical source used both by the website and the new CLI.

This reduces duplicated content, removes the flattened hierarchy and its link-rewriting complexity, makes installed skills smaller, and prevents agents from loading a large reference tree when the CLI can retrieve only the relevant document.

## Solution / Next Steps

1. Define the `rulesync docs [document]` command and the `--search <text>` option in the CLI.
2. Add a build/package strategy that bundles the current `docs/**/*.md` hierarchy and makes it locatable at runtime across supported Node.js and compiled-binary distributions.
3. Resolve document identifiers to Markdown files while preserving nested paths and preventing path traversal.
4. Print the selected Markdown content to stdout without altering it unnecessarily, so the output can be piped to other tools.
5. Add MiniSearch and build an ephemeral in-memory BM25+ index from all bundled Markdown files when searching. Configure field boosts for document paths, titles, headings, and body content.
6. Decide and document behavior for listing available documents, missing documents, an empty search term, no search results, and conflicting use of a document argument with `--search`.
7. Add colocated unit tests and end-to-end happy-path coverage for top-level lookup, nested lookup, search, missing documents, and unsafe paths.
8. Update the CLI reference and any generated/help output required for the new command.
9. Replace the mirrored `skills/rulesync/` documentation tree with one concise `SKILL.md` that delegates detailed lookup to `rulesync docs`.
10. Simplify the docs-to-skill synchronization script and its tests to reflect the single-file skill design, while keeping `docs/` as the canonical source of documentation content.

dyoshikawa:
## Summary

Implements [#2519](https://github.com/dyoshikawa/rulesync/issues/2519): a `rulesync docs` command that exposes the bundled documentation directly in the terminal, plus the single-file simplification of the distributed `skills/rulesync/` skill.

## The `docs` command

- `rulesync docs` — lists every document identifier (sorted, one per line).
- `rulesync docs <document>` — prints the document's Markdown verbatim to stdout. Identifiers map to the `docs/` hierarchy; a leading `docs/` and trailing `.md` are accepted and stripped. Traversal (`..`), absolute paths, and drive letters are rejected.
- `rulesync docs --search <text>` — ranked full-text search via [MiniSearch](https://github.com/lucaong/minisearch) (BM25+, zero transitive dependencies). Indexes paths, titles, headings, and body with boosts title 4 / headings 3 / path 2; exact-term matching only (no prefix/fuzzy). Prints up to 10 results as `<document> — <matching context>`.
- Predictable failures (exit 1): unknown document, invalid identifier, empty search text, no matches, document+`--search` combined.
- Output goes straight to `process.stdout` (not the logger) so piping works under `--silent` and in test environments, with EPIPE tolerated for early-closing consumers (`| head`).

## Bundling strategy

`scripts/generate-docs-content.ts` embeds `docs/**/*.md` (21 files, ~420KB) into the committed `src/generated/docs-content.ts`, so the content ships in both the npm `dist` bundle and the bun-compiled binaries with no runtime filesystem lookup. Drift is CI-guarded by the new `check:docs-content` step in `cicheck:content`, and the `docs/**/*.md` lint-staged hook regenerates it on commit. Verified: `node dist/cli/index.js docs faq` works from a clean build.

## Skill simplification

- `skills/rulesync/` now ships **only `SKILL.md`**, which briefly explains Rulesync and instructs agents to consult `rulesync docs <document>` / `rulesync docs --search <text>` before answering Rulesync-specific questions. The 19 mirrored docs files are removed.
- `scripts/sync-skill-docs.ts` shrinks from 255 lines of VitePress flattening/link rewriting to a pruning guard that keeps the directory single-file; `check:sync-skill-docs` still enforces it in CI. Tests rewritten accordingly.

## Tests

- Unit: identifier normalization (incl. unsafe paths), document lookup, listing, search ranking, and all error modes (`src/cli/commands/docs.test.ts`); skill pruning (`scripts/sync-skill-docs.test.ts`).
- E2E: `src/e2e/e2e-docs.spec.ts` — top-level lookup, nested lookup, search, listing, missing document, unsafe identifier (all green locally).

## Dependency note

Adds **minisearch 7.2.0** as a runtime dependency (`package.json` + lockfile). Per the issue's own selection: BM25+ ranking, TypeScript support, zero transitive dependencies.

Closes #2519


dyoshikawa:
@dyoshikawa Thank you!

PR #2531, opened 2026-07-31
fix(watch): poll liveness on a timer, report detach as a change

dyoshikawa:
## Summary

Third and hopefully final round on the flaky `watch.test.ts` re-attach tests, which have kept failing on CI (latest: PR #2529 run 30605141246, where `waitFor`'s own 10s budget expired — i.e. the deletion event never arrived at all, not just late).

## Root cause

Liveness checks ran only inside `fs.watch` event callbacks. When the OS delivers no event for a deleted watched directory (observed on loaded CI runners), the dead watcher stays attached forever regardless of the inode fix (#2516) or longer timeouts (#2520).

## Fix

- Run the same inode-based `verifyStillWatching` sweep on a periodic timer (every `rearmIntervalMs`, `unref()`'d so it never holds the process open). A replaced or removed directory is now detected within one interval even with zero events delivered.
- On detach, report the disappearance through `onChange` the way an OS delete event would have — a purely poll-detected deletion now reaches watch mode's scheduler (which debounces the occasional double notification after an evented detection).

Existing tests cover the behavior and become deterministic (5 consecutive local runs green; full unit suite green).

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -3,11 +3,15 @@ import { join } from "node:path";
 
 import { afterEach, describe, expect, it, vi } from "vitest";
 
-const { statSyncMock } = vi.hoisted(() => ({ statSyncMock: vi.fn() }));
+const { statSyncMock, fsWatchMock } = vi.hoisted(() => ({
+  statSyncMock: vi.fn(),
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2532, opened 2026-07-31
test(watch): probe the watched root after re-attach, not the recreated subdirectory

dyoshikawa:
## Summary

Fourth and smallest patch in the flaky `watch.test.ts` series (#2516 inode comparison, #2520 timeouts, #2531 liveness polling). The re-attach test still timed out on CI (PR #2529 runs) at its final wait: it writes the probe file inside the freshly recreated `rules/` subdirectory, and the re-attached recursive watcher can register that subdirectory late — a kernel-level inotify race on loaded runners that no amount of watcher-side polling can close (a file write does not change the directory inode).

The test asserts **re-attachment**, not recursive subdirectory coverage (the first test in the suite already covers recursive forwarding), so probe at the watched root instead.

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -361,7 +361,11 @@ describe("watchTargets", () => {
           await waitFor(() => changed.length > 0);
 
           changed.length = 0;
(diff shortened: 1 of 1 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2536, opened 2026-07-31
test(watch): probe until observed to close re-attach observation races

dyoshikawa:
## Problem

The CI job `Code Quality & Tests` failed repeatedly on `main` today (e.g. runs 30611060570, 30611056373, 30607754538) with:

```
FAIL src/lib/watch.test.ts > watchTargets >
  re-attaches a filtered target after its directory is deleted and recreated
Error: Timed out waiting for condition   (watch.test.ts:412)
```

The failing frame is the *first* wait, right after `rm` + `mkdir`:

```ts
await rm(configDir, ...);
await mkdir(configDir, ...);
changed.length = 0;                      // clears notifications
await waitFor(() => changed.length > 0); // times out
```

Only two notifications ever reach `changed` in this test: the detach notification (added in #2531) and the re-attach notification from the 25ms rearm timer. Both fire from timer/event callbacks, which can run **while the `rm`/`mkdir` promises are still settling** — on a loaded runner the rearm timer fires between mkdir's fs completion and the test continuation. The subsequent `changed.length = 0` then wipes every notification the watcher will ever send: the watcher is already healthy and untouched, so nothing else fires and the wait times out. This is why the flake survived #2531/#2532 (the CI run on the merge commit of #2532 itself failed).

The production code in `watch.ts` is correct — only the test's observation method races.

## Fix

Replace clear-then-wait with a probe-until-observed loop: re-write the config file until the watcher reports the file's own path. Only an attached watcher can report that path (directory-level notifications carry the directory path), so the loop both proves re-attachment and self-heals when a write lands in the unwatched gap before the timer-driven re-attach.

The same probe loop is applied to the final phase of the unfiltered re-attach test, which had the mirror-image race: its recreate wait could be satisfied by a late-delivered delete event while still detached, letting the one-shot `after-rearm.md` write land in the unwatched gap.

## Verification

- `npx vitest run src/lib/watch.test.ts` × 5 consecutive runs: 20/20 passed each time
- `pnpm cicheck`: all 8135 tests and content checks green

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -533,11 +533,55 @@ describe("watchTargets", () => {
         // Simulate a delete+recreate whose delete event was not delivered
         // yet: the path still exists, but its inode no longer matches the
         // one recorded at attach time. An events-only existence check would
-        // keep the dead watcher forever; the inode comparison must re-arm.
-        statSyncMock.mockReturnValue({ ino: 999_999_999n });
+        // keep the dead watcher forever; the identity comparison must re-arm.
+        statSyncMock.mockReturnValue({ ino: 999_999_999n, birthtimeNs: 1n });
         // Any event — even one the include filter rejects — runs the
         // liveness check.
         await writeFile(join(configDir, "decoy.md"), "# decoy\n", "utf8");
+        await waitFor(() => changed.includes(configDir));
+      } finally {
+        statSyncMock.mockImplementation(actual.statSync);
+        handle.close();
+      }
+    } finally {
+      await cleanup();
+    }
+  });
+
+  it("re-attaches when the recreated directory reuses the watched inode", async () => {
+    const actual = await vi.importActual<typeof import("node:fs")>("node:fs");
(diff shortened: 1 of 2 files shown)


dyoshikawa:
The first commit alone did not fix the CI flake — the same test timed out again with the probe loop running for its full 10s window ([run 30638160894](https://github.com/dyoshikawa/rulesync/actions/runs/30638160894)). Ten seconds of re-writes with zero events means the watcher was believed alive while bound to a dead inode: ext4 can hand the freed inode number back to the recreated directory, which makes the bare inode comparison in the liveness check blind to the replacement. That is also a real (if rare) production bug — a branch switch recreating `.rulesync/` on a reused inode would silently kill watch mode.

Pushed 02a4d64 which compares inode **plus birthtime** as the liveness identity (falling back to inode-only where the file system reports no creation time), with a deterministic regression test for the reused-inode case.

dyoshikawa:
@dyoshikawa Thank you!

PR #2540, opened 2026-07-31
docs(hooks): generate the hook event × tool matrix from the hooks factories

dyoshikawa:
## Summary

The `Hook event × tool matrix` in `docs/reference/file-formats.md` was hand-maintained and had drifted badly:

- Missing columns for twelve hook-capable targets (Pi Coding Agent, Amp, Hermes Agent, Reasonix, Qwen Code, Vibe, Junie, Grok CLI, Kimi Code, the Kiro CLI/IDE variants, and the Claude Code / Antigravity plugin variants).
- Still carried a **Gemini CLI** column, which is no longer a rulesync target at all.

This PR takes the issue's "generate it so it cannot go stale" option:

- New `scripts/hook-events-table.ts` renders the table between `HOOK_EVENTS_MATRIX` markers from each hooks factory's `supportedEvents` in `toolHooksFactories` — the same source of truth the generator's skip-warnings come from. Columns follow `TOOL_DISPLAY` order (like the other generated tables) and rows follow canonical `HOOK_EVENTS` order, restricted to events at least one target supports. A hooks target missing from `TOOL_DISPLAY` throws.
- `scripts/generate-docs-content.ts` applies the renderer (and oxfmt) to `file-formats.md` before embedding the docs, so a stale committed table makes the **existing** `check:docs-content` drift gate fail — no new package script and no `package.json` change needed.
- The regenerated table now has 28 tool columns × 48 event rows; I machine-verified every cell against the `*_HOOK_EVENTS` constants (0 mismatches).
- The hooks prose PascalCase-translation list now names Qwen Code instead of the removed Gemini CLI target.

## Testing

- `pnpm cicheck` green (generation is idempotent; second run produces no diff).
- Cell-by-cell verification script against `toolHooksFactories` reported `checked rows: 48 cols: 28 errors: 0`.

Closes #2522

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2547, opened 2026-07-31
fix(opencode,kilo): shell hook events via bash-gated named hooks, global instructions registration

### LINKED ISSUE #2501: Follow up OpenCode upstream updates: afterShellExecution mapped to slash-command event (also kilo), global non-root rules dropped
## Summary

Two OpenCode gaps as of v1.18.10 (2026-07-30). One is a correctness bug: canonical `afterShellExecution` is mapped to OpenCode's `command.executed`, which is a *slash-command* event, so the hook never fires after a bash command and instead fires on every slash command — and Kilo inherits the same wrong mapping. The other: global generate silently drops all non-root rules even though OpenCode reads `instructions` from the global `opencode.json`.

Window reviewed: v1.15.x → v1.18.10 (2026-04-25 → 2026-07-30), opencode.ai docs, and the `sst/opencode` source tree.

### 1. hooks — `afterShellExecution` wired to the wrong OpenCode event (correctness)

- **Upstream**: `command.executed` is a slash-command event with schema `{ name, sessionID, arguments, messageID }` (`packages/schema/src/v1/legacy-event.ts`, surfaced as `Command.Event.Executed` in `packages/opencode/src/command/index.ts`). OpenCode has **no** shell-execution lifecycle event; observing a bash invocation requires the named hook `"tool.execute.after"` gated on `input.tool === "bash"` (the `Hooks` interface in `packages/plugin/src/index.ts`; the `.env protection` example at [plugins doc](https://opencode.ai/docs/plugins/)). `shell.env` only injects environment variables.
- **rulesync** (`project` + `global`): `src/types/hooks.ts` maps `afterShellExecution: "command.executed"` in the generic event dispatch (documented in `docs/reference/file-formats.md`), and `CANONICAL_TO_KILO_EVENT_NAMES` reuses the map verbatim — so **kilo is affected too**.
- **Effect**: silently wrong, not merely lossy — the hook never runs on shell executions and fires on slash commands instead.

### 2. rules — global non-root rules dropped despite global `instructions` support

- **Upstream**: [rules doc](https://opencode.ai/docs/rules/) — instruction files may be listed "in your `opencode.json` **or the global `~/.config/opencode/opencode.json`**"; all are combined with `AGENTS.md`.
- **rulesync**: `OpenCodeRule.getSettablePaths` returns a `nonRoot` path (`.opencode/memories`) only in project scope, and `getExtraSharedWritePaths` returns `[]` when global, so the `instructions` registrar never runs globally. Dry-run evidence: `--features rules --global` prints "4 non-root rulesync rules found, but it's in global mode, so ignoring them" and writes only `~/.config/opencode/AGENTS.md`.

## Proposed Follow-up

1. Remap `afterShellExecution` to the named `tool.execute.after` hook gated on the `bash` tool — the generator already supports named hooks with an `input.tool` subject (`NAMED_HOOK_MATCHER_SUBJECTS` in `src/features/hooks/opencode-style-generator.ts`), combining any user-supplied matcher with the bash gate. Add `beforeShellExecution` → `tool.execute.before` + `bash` to `OPENCODE_HOOK_EVENTS` at the same time (a capability OpenCode has that rulesync does not offer). Apply the same change to Kilo. If observing slash commands is wanted, that needs its own canonical event rather than reusing `afterShellExecution`.
2. Add a global `nonRoot` entry (`~/.config/opencode/memories/`) and have `getExtraSharedWritePaths` return the global `opencode.json` path so the same `instructions` registration runs globally — the processor already has the `supportsGlobalNonRoot` opt-in branch (`src/features/rules/rules-processor.ts`).

## Non-gaps (verified, do not re-file)

- **mcp**: `cwd` (v1.17.4) already modeled; `timeout`/`oauth` pass through; tool gates handled.
- **commands / subagents / skills / permissions**: all match current docs; newer subagent `hidden`/`color` round-trip via `looseObject`; the documented permission key set including `external_directory` and `doom_loop` is covered, with the canonical `agent` → `task` translation.
- **plugins**: the v2 plugin API (v1.17.10) is undocumented publicly; the v1 plugin shape rulesync emits is still what the docs teach.
- **ignore / checks**: correctly unsupported.
- Noted only: top-level `references` key (v1.17.1, [doc](https://opencode.ai/docs/references/)) maps to no rulesync feature; the shared-config gateway preserves it. `permissionRequest` → read-only `permission.asked` is a fidelity limit consistent across tools.

## References

- https://opencode.ai/docs/plugins/ — named hooks and the `tool.execute.after` + bash-gate pattern (gap 1).
- `packages/schema/src/v1/legacy-event.ts` and `packages/opencode/src/command/index.ts` in sst/opencode — `command.executed` is a slash-command event (gap 1).
- https://opencode.ai/docs/rules/ — global `instructions` support (gap 2).
- https://github.com/sst/opencode/releases — v1.15.x → v1.18.10 window; v1.17.4 `cwd`, v1.17.1 `references`, v1.17.10 v2 plugin API.

dyoshikawa:
## Summary

Resolves the two OpenCode gaps from #2501 (verified against sst/opencode v1.18.10; Kilo shares the hook surface):

1. **hooks — correctness fix.** `afterShellExecution` was mapped to `command.executed`, which is a *slash-command* event — the hook never fired after a bash command and fired on every slash command instead, for OpenCode **and** Kilo. Both shell events now land in the named `tool.execute.before`/`tool.execute.after` hooks with an implicit `input.tool === "bash"` gate (the `.env protection` pattern from the OpenCode plugins doc), implemented via a `SHELL_EVENT_TOOL_GATES` table in the shared opencode-style generator so Kilo inherits the fix. `beforeShellExecution` joins `OPENCODE_HOOK_EVENTS` at the same time. Matchers on the shell events remain warned-and-dropped upstream (`matcherEvents` covers only `preToolUse`/`postToolUse`) — the named hooks expose no command text to match against.
2. **rules — global non-root rules no longer dropped.** OpenCode reads `instructions` from the global `~/.config/opencode/opencode.json` too, so global generate now writes non-root rules to `~/.config/opencode/memories/*.md` and registers them (entries relative to the config file's directory, so the `.config/opencode/` prefix is stripped). Wiring: a `mcpInstructionsRegistrarGlobal` opt-in on the rules factory meta (off by default — Kilo auto-discovers its global rules dir and must not register), a global `nonRoot` settable path, `getExtraSharedWritePaths` returning the global config at global scope, and a `rules` ownership declaration for `.config/opencode/opencode.json` in the shared-config gateway.

Docs: hooks prose + hand-maintained hook matrix cells updated (the generated matrix from #2540 will own this once merged), OpenCode rules note documents the global registration.

## Testing

- New tests: bash-gated named-hook emission (both events + coexistence with a matcher-driven preToolUse handler), global `fromInstructions` prefix stripping, updated global settable-paths expectation and the event-map completeness carve-out; derive snapshot updated.
- `pnpm cicheck` green.

Closes #2501


dyoshikawa:
@dyoshikawa Thank you!

PR #2543, opened 2026-07-31
fix(vibe): stop misusing exclusive enabled_tools, preserve vibe-mcp-add servers, emit never, warn on shadowed global config

dyoshikawa:
## Summary

Resolves the four Vibe Code gaps from #2506 (verified against mistral-vibe v2.23.2):

1. **permissions — exclusive `enabled_tools` no longer written on allow (the high-severity gap).** Upstream documents `enabled_tools` as an exclusive allowlist ("if set, only these tools will be active"), so expressing wildcard allows through it switched off every other builtin and all MCP tools. A wildcard allow now only writes `[tools.<name>] permission = "always"`; the existing cleanup loop migrates away the exclusive entries earlier rulesync versions wrote. The list is only authorable explicitly through a new `vibe.enabled_tools` override (verbatim; declaring it — even empty — makes rulesync own the key), and **import** lifts a non-empty `enabled_tools` into that override instead of misreading it as `"*": "allow"` grants.
2. **mcp — servers added by `vibe mcp add` / `/mcp add` survive generate.** Unmanaged `mcp_servers` entries are appended after the managed ones instead of being deleted by the whole-array replace. Documented flip side: removing a server from `.rulesync/mcp.jsonc` no longer removes it from `config.toml`.
3. **scope — `--global` warns when shadowed.** Vibe reads exactly one `config.toml` (trusted project file first, home file as fallback — not a merge; single code path since v2.22.0), so the mcp/permissions surfaces now warn on a `--global` run when the current project has its own `.vibe/config.toml`, and the fallback semantics are documented.
4. **permissions — wildcard deny emits `permission = "never"`** (Vibe's ToolPermission vocabulary, used by its own plan-mode profile) instead of leaving an empty `[tools.<name>]` table, making the round-trip symmetric with the existing `"never"` import mapping. (The unmapped-category empty-table bug remains tracked separately in #2485.)

Also extracts the per-category translation into `applyCategoryRules` to stay within the lint complexity budget, and updates `docs/reference/file-formats.md` (permissions paragraph, vibe override note, MCP note).

## Testing

- New tests: no `enabled_tools` on wildcard allow; `permission = "never"` + `disabled_tools` on wildcard deny with deny round-trip; override-owned `enabled_tools`; import lifting into `vibe.enabled_tools`; shadowed-global warning; unmanaged MCP server preservation ordering.
- Updated the two tests that encoded the old (buggy) behavior.
- `pnpm cicheck` green (8170+ unit tests).

Closes #2506

[...]

PR #2549, opened 2026-08-01
fix(vibe): skip canonical categories with no Vibe tool instead of emitting inert tables

### LINKED ISSUE #2485: bug(vibe): unmapped canonical permission categories emit [tools.<category>] entries Vibe has no tool for, so a deny is silently inert
## Background

Spun out of #2423 (recorded there by the maintainer as worth its own issue).

`toVibeToolName` in `src/features/permissions/vibe-permissions.ts` falls back to the canonical category name verbatim when `CANONICAL_TO_VIBE_TOOL_NAMES` has no entry (`CANONICAL_TO_VIBE_TOOL_NAMES[category] ?? category`). Canonical categories with no Vibe counterpart — e.g. `glob` and `notebookedit` — therefore emit `[tools.glob]` / `[tools.notebookedit]` tables into `vibe.toml` for tools Vibe does not have. A `deny` authored on those categories looks applied but is silently inert, which is the dangerous direction.

Unlike the `agent` → `task` case fixed in #2441 there is no correct name to rename to.

## Solution

Adopt the "skip with a warning" behaviour the `grokcli` permissions adapter already has: when a canonical category has no Vibe tool mapping, do not emit a `[tools.<category>]` table; log a warning naming the category and the affected rules. Keep import lenient (an unknown on-disk `[tools.*]` table should still round-trip untouched as tool-specific content). Add unit coverage for the deny-inert case in both directions.

dyoshikawa:
## Summary

Fixes #2485: `toVibeToolName` fell back to the canonical category name verbatim, so categories with no Vibe counterpart (e.g. `glob`, `notebookedit`) emitted `[tools.<category>]` tables for tools Vibe does not have — a `deny` authored there looked applied while being **silently inert**, the dangerous direction. Unlike the `agent` → `task` rename (#2441) there is no correct name to map to.

- Generation now skips such categories with a warning naming the category and the affected rule count (the grokcli adapter's established pattern), via a `hasVibeToolName` own-property check. The skip also covers the `enabled_tools`/`disabled_tools` cleanup loop and the `vibe.permission.<category>.sensitive_patterns` override, so no phantom tool names are created or touched anywhere.
- Import stays lenient: an unknown on-disk `[tools.*]` table still round-trips untouched (test added), and the empty-table artifact this issue also referenced was already removed by the `permission = "never"` emission in PR #2543.
- Docs: the Vibe permissions paragraph documents the skip.

## Testing

- New tests: deny on `glob`/`notebookedit` emits no tables, no `disabled_tools` entries, and warns per category; unknown on-disk `[tools.custom_tool]` table survives a generate.
- `pnpm cicheck` green.

Closes #2485


dyoshikawa:
@dyoshikawa Thank you!

PR #2550, opened 2026-08-01
fix(generate): fail the run when the permissions feature errors

### LINKED ISSUE #2486: bug(vibe): malformed .vibe/config.toml is a hard error under --features mcp but a silent exit-0 success under --features permissions
## Background

Spun out of #2423 (recorded there by the maintainer as worth its own issue).

With a malformed `.vibe/config.toml` on disk, `rulesync generate --targets vibe --features mcp` exits 1 with a parse error, but `--features permissions` swallows the same parse failure and exits 0 reporting "All files are up to date". That is a silent false-success: the user's permission changes are not applied and nothing signals it.

## Solution

Make the permissions path surface the parse failure the same way the MCP path does (fail the run, or at minimum report the file as errored rather than up to date). The two features read the same file and should agree on what a malformed file means. Add a regression test asserting both features exit non-zero (or both warn identically, if warn-and-skip is chosen deliberately) on the same malformed fixture.

dyoshikawa:
## Summary

Fixes #2486: with a malformed `.vibe/config.toml` on disk, `generate --features mcp` exited 1 with a parse error while `--features permissions` swallowed the same failure (`logger.warn` + `continue` in `generatePermissionsCore`) and exited 0 reporting "All files are up to date" — a silent false-success.

- The permissions loop now logs the failure at **error** level and rethrows, so both features agree that a malformed shared file fails the run (reproduced both before/after: exit-perm 0→1, exit-mcp stays 1).
- Regression tests: an e2e `it.each` asserts **both** features exit 1 on the same malformed fixture, and the generate unit test that pinned the old swallow-and-continue behavior now pins fail-the-run.

## Testing

- `pnpm cicheck` green; targeted e2e cases pass locally.

Closes #2486

### DIFF
--- a/src/e2e/e2e-permissions.spec.ts
+++ b/src/e2e/e2e-permissions.spec.ts
@@ -119,6 +119,29 @@ describe("E2E: permissions", () => {
     },
   );
 
+  it.each([{ features: "permissions" }, { features: "mcp" }])(
+    "should fail $features generation on a malformed .vibe/config.toml instead of reporting success",
+    async ({ features }) => {
+      const testDir = getTestDir();
+
+      await writeFileContent(join(testDir, ".vibe", "config.toml"), "mcp: [unclosed");
+      await writeFileContent(
+        join(testDir, RULESYNC_PERMISSIONS_RELATIVE_FILE_PATH),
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2552, opened 2026-08-01
fix(generate): fail the run when the ignore feature errors

### LINKED ISSUE #2551: generateIgnoreCore still swallows per-target errors (exit 0 'up to date' on a broken ignore source)
## Background

Finding from the PR #2550 review round. #2486 fixed the permissions feature's swallow-and-continue (warn + `continue` → error + rethrow) so a malformed shared config fails the run like the MCP feature. `generateIgnoreCore` is now the only remaining feature loop with the old pattern (src/lib/generate.ts ~791-796, pinned by the unit test "should handle errors gracefully and continue" at src/lib/generate.test.ts ~388).

## Details

A throw from IgnoreProcessor (e.g. a malformed ignore source) still exits 0 with "All files are up to date", including in --check mode. Since ignore files are what keep secrets out of AI tools' reach, a silently-skipped ignore generation is the same class of fail-open bug — arguably worse.

## Solution / Next Steps

Align ignore with the other eight feature cores: log at error level and rethrow (mirror the #2550 change), flip the pinning unit test, and add an e2e regression case — or document explicitly why ignore is intentionally lenient if there is a reason.

dyoshikawa:
## Summary

Fixes #2551 (follow-up from the PR #2550 review): `generateIgnoreCore` was the last feature loop still swallowing per-target errors with warn + `continue`, exiting 0 with "All files are up to date" — including in `--check` mode — while nothing was written. Since ignore files are what keep secrets out of AI tools' reach, this was the same fail-open class as #2486's permissions bug, arguably worse.

- The catch now logs at error level and rethrows, matching the other eight feature cores.
- The unit test pinning the old swallow-and-continue behavior now pins fail-the-run (rejects + error-level log assertion), mirroring the #2550 change exactly.

## Testing

- `pnpm cicheck` green.

Closes #2551

### DIFF
--- a/src/cli/commands/generate.test.ts
+++ b/src/cli/commands/generate.test.ts
@@ -446,17 +446,17 @@ describe("generateCommand", () => {
       mockCwd.mockRestore();
     });
 
-    it("should handle errors in ignore processing gracefully", async () => {
+    it("should fail the command on an ignore error instead of reporting success", async () => {
       vi.mocked(IgnoreProcessor).mockImplementation(function () {
         throw new Error("Test error");
       });
       const options: GenerateOptions = {};
 
-      // Should not throw, errors are caught and processing continues
-      await generateCommand(mockLogger, options);
-
-      // Should still complete without error
-      expect(mockLogger.info).toHaveBeenCalledWith("✓ All files are up to date (ignore)");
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2545, opened 2026-07-31
feat(rovodev): project-scoped config.yml and mcp.json, author disabledMcpServers

dyoshikawa:
## Summary

Resolves the three Rovo Dev gaps from #2504, verified against the [Bitbucket Cloud Agentic Pipelines guide](https://support.atlassian.com/bitbucket-cloud/docs/rovo-dev-advanced-agentic-configuration/) and the [Rovo Dev CLI MCP docs](https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/):

1. **permissions — project scope enabled.** The hard-coded global-only throw is removed; project mode now writes and imports the repo-committed `.rovodev/config.yml` the Bitbucket guide documents (`config.path` in `bitbucket-pipelines.yml`, or `--config-file`). The write moved onto the shared-config gateway with declared `toolPermissions` ownership, since the MCP feature now patches the same file.
2. **mcp — project scope enabled** for `.rovodev/mcp.json` (`mcp.mcpConfigPath`). Both project files are added to `SHARED_USER_MANAGED_CONFIG_PATHS` so the derived `.gitignore` does **not** ignore them — committing them is exactly the Bitbucket CI use case.
3. **mcp — `disabledMcpServers` authored; `disabled: true` no longer lossy.** The server definition is written to `mcp.json` (minus the flag the file cannot express) and its name goes to `mcp.disabledMcpServers` in the sibling `config.yml` via a new `getAuxiliaryFiles` writer through the gateway (`mcp` block recomputed from the existing file: rulesync owns the toggle for managed servers; user keys like `mcpConfigPath`/`allowedMcpServers` and unmanaged disabled names survive). **Import** overlays `disabled: true` from `config.yml`, so the toggle round-trips — previously disabled servers were skipped with a warning and unrecoverable from the Rovo Dev side.

Supporting changes: `.rovodev/config.yml` registered in `SHARED_CONFIG_OWNERSHIP` (yaml; `mcp` + `permissions` writers), `getExtraSharedWritePaths` declared so the write-order derivation sees the MCP feature as a writer of that file (kimi-code pattern), supported-tools tables and derive snapshot regenerated, docs updated.

## Testing

- New/updated tests: project-scope write+import for both features, disabled-server toggle write with user-key preservation, import round-trip of the toggle, updated processor target lists and derive snapshot.
- `pnpm cicheck` green.

Closes #2504

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2554, opened 2026-08-01
fix(opencode,kilo): scope instructions ownership to the managed rules directories

### LINKED ISSUE #2548: OpenCode/Kilo instructions arrays accumulate stale entries (union merge never removes deleted rules)
## Background

Finding from the PR #2547 review round (pre-existing behavior, both scopes and both tools): `fromInstructions` merges the existing `instructions` array with the newly generated list by union (dedupe + sort), so an entry registered for a since-deleted non-root rule persists in `opencode.json` / `kilo.jsonc` forever. A user's hand-written entry for the same file in a different spelling (e.g. `.config/opencode/memories/x.md` vs the stripped `memories/x.md` at global scope) also coexists as a duplicate, loading the file twice.

## Details

- `OpencodeMcp.fromInstructions` (src/features/mcp/opencode-mcp.ts): `new Set([...existingInstructions, ...normalizedInstructions])`.
- The kilo counterpart shares the pattern.
- Deleting a rule from `.rulesync/rules/` removes the memories file (orphan deletion) but never the registered path, leaving a dangling instructions entry OpenCode then fails to load or silently ignores.

## Solution / Next Steps

Scope the union: rulesync owns entries that point under its managed rules directories (`.opencode/memories/`, `.kilo/rules/`, global `memories/`) — rebuild that subset from the current generate and preserve only entries outside those roots verbatim. Mirrors how other adapters scope managed-name ownership (e.g. Rovo Dev's `disabledMcpServers` handling in PR #2545).

dyoshikawa:
## Summary

Fixes #2548 (follow-up from the PR #2547 review): `fromInstructions` merged the existing `instructions` array with the generated list by plain union, so an entry registered for a since-deleted non-root rule persisted in `opencode.json` / `kilo.jsonc` forever, and at global scope a legacy full-prefix spelling (`.config/opencode/memories/x.md`) could coexist with the stripped `memories/x.md` as a duplicate (file loaded twice).

- rulesync now **owns** the entries pointing under its managed rules directories — `.opencode/memories/` (project), `memories/` **and** the legacy `.config/opencode/memories/` spelling (global), `.kilo/rules/` — rebuilding that subset from the current generate, so deleting a rule also drops its registration. Entries outside the managed roots are the user's and pass through verbatim.
- Mirrors the managed-name ownership pattern established for Rovo Dev's `disabledMcpServers` in PR #2545.
- The two tests that pinned the accumulate-forever union now pin the ownership semantics (stale managed entry dropped, user entry preserved); docs for both tool notes updated.

## Testing

- `pnpm cicheck` green.

Closes #2548


dyoshikawa:
@dyoshikawa Thank you!

PR #2568, opened 2026-08-03
fix: write binary skill other-files without UTF-8 corruption

magellen-ai:
## Problem

Skill **other files** (anything besides `SKILL.md` inside a skill directory — e.g. `wechat-qrcode.jpg`, `assets/hero.gif`) are corrupted on both `import` and `generate`.

`DirFeatureProcessor.writeAiDirs` reads and writes those files through a UTF-8 **text** path:

- comparison: `readFileContentOrNull` (UTF-8) + `fileContentsEquivalent`
- write: `file.fileBuffer.toString("utf-8")` then `writeFileContent` (UTF-8)

Invalid byte sequences in binary content are replaced with U+FFFD during the round-trip, so a 50 KB JPEG silently becomes a ~87 KB file with a broken header:

```
$ file .rulesync/skills/andrej/wechat-qrcode.jpg
.rulesync/skills/andrej/wechat-qrcode.jpg: data
```

## Fix

Route binary other files through a buffer path while keeping text files on the existing text path:

- `isBinaryBuffer()` (new, in `utils/file.ts`) — detects whether a UTF-8 `toString` + `Buffer.from` round-trip would alter the bytes (valid UTF-8 text incl. CJK round-trips unchanged; JPEG/GIF etc. do not)
- change detection for binary files now compares raw bytes via `Buffer.equals` using a new `readFileBufferOrNull()`
- writes go through the existing-but-unused `writeFileBuffer()`

## Verification

- Added unit tests: binary other-file written via buffer path / skipped when unchanged / detected when changed; existing text-path tests unchanged
- Full suite: 8232 passed (2 pre-existing failures are `spawn bun ENOENT` — bun not installed in this environment, unrelated)
- `oxfmt`, `oxlint`, `tsgo` (typecheck), `tsdown` (build) all clean
- End-to-end check on a real workspace: import + generate round-trip of skills containing `wechat-qrcode.jpg` and `assets/hero.gif` — files remain valid (`file` reports JPEG/GIF) and md5 matches the source

### DIFF
--- a/src/types/dir-feature-processor.test.ts
+++ b/src/types/dir-feature-processor.test.ts
@@ -4,8 +4,10 @@ import { createMockLogger } from "../test-utils/mock-logger.js";
 import { setupTestDirectory } from "../test-utils/test-directories.js";
 import {
   ensureDir,
+  readFileBufferOrNull,
   readFileContentOrNull,
   removeDirectory,
+  writeFileBuffer,
   writeFileContent,
 } from "../utils/file.js";
 import { AiDir, AiDirFile } from "./ai-dir.js";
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@greenhands-always Thank you!

PR #2672, opened 2026-08-13
feat(hooks): generate Cline TaskError and SessionShutdown hook scripts

dyoshikawa:
## Summary

Cline's hooks directory is read by two upstream runtimes with different accepted script names, and rulesync only modelled one of them:

- VS Code extension — `VALID_HOOK_TYPES` in `apps/vscode/src/core/hooks/utils.ts`
- SDK/CLI — `HookConfigFileName` in `sdk/packages/core/src/hooks/hook-file-config.ts`, which drops `Notification` but adds `TaskError` (→ `agent_error`) and `SessionShutdown` (→ `session_shutdown`)

Both resolve `<project>/.clinerules/hooks/` (it is `DEPRECATED_CONFIG_DIR` in the SDK path resolver, but still in its search paths), so the emitted set is now the union of the two: `afterError` → `TaskError` and `sessionEnd` → `SessionShutdown` are generated alongside the existing seven. A script named for an event the running runtime does not know is simply never spawned, so this is additive for VS Code users.

`TaskResume` and `TaskCancel` stay unmapped — neither has a canonical counterpart.

## Deviation from the plan

The plan asked for a reverse event-name map and generate+import round-trip tests. There is no reverse map for Cline and none can be added: `ClineHooks.toRulesyncHooks()` throws by design ("generated Cline hook scripts cannot be imported back into canonical hooks") because the target emits executables, not a config file. Tests are generation-only accordingly.

One existing test used `sessionEnd` as its example of an unmapped canonical event; it now uses `stop`, which is still unmapped.

## Checks

- `pnpm cicheck` — green
- `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-hooks.spec.ts` — 81 passed

The hook event × tool matrix in `docs/reference/file-formats.md` was regenerated via `pnpm run generate:docs-content`; the surrounding prose was updated by hand.

Part of #2405


dyoshikawa:
Follow-up commit addressing the two review findings.

**gitignore entries.** `gitignore-entries.ts` hand-lists Cline hook scripts per event (they come from `getAuxiliaryFiles`, not `getSettablePaths`), so `TaskError` and `SessionShutdown` plus their `.ps1` twins are now listed and `.gitignore` regenerated via `pnpm dev gitignore`. The justified-exception list in `gitignore-entries.test.ts` needed the same four entries. The stale comment claiming the names are "fixed by Cline's VALID_HOOK_TYPES" now states the union-of-runtimes premise. The `file-formats.md` claim that gitignore "lists the generated script names explicitly rather than the whole directory" stays accurate — that is exactly why the four had to be added by hand.

**Correction to the safety claim in the PR body above.** The claim that an unrecognized script "is simply never spawned" holds for unknown *names* only. For an event a runtime does know, the SDK/CLI spawns **both** spellings, and it is worse than noise:

- `listHookConfigFiles` dedupes by `path`, so `TaskError` and `TaskError.ps1` are two entries carrying the same `hookEventName`.
- `createHookCommandMap` **appends** to `map[hookEventName]` with no per-event dedupe, and `runBlockingHookCommands` iterates the whole list.
- `inferHookCommand` maps `.ps1` to `pwsh -File` on non-Windows.

So on Unix the SDK/CLI runtime ran every hook **twice** — once via bash, once via pwsh — or failed where `pwsh` is absent. This predates this PR and affected all events on that runtime; the fix hardens all of them.

The generated `.ps1` now opens with a guard that answers with the neutral success payload and exits off Windows, leaving the POSIX twin to do the work.

**One deviation from the suggested fix, deliberate.** The proposed `if (-not $IsWindows)` would break Windows: `$IsWindows` is a PowerShell 6+ automatic variable and does not exist in the Windows PowerShell 5.1 that `powershell -File` starts, where it evaluates to `$null` — making `-not $IsWindows` true and turning the script into a no-op on the one platform it exists for. The guard therefore tests that the variable is both **defined and false**:

```powershell
if ($null -ne $IsWindows -and -not $IsWindows) {
  Write-Output '{"cancel": false, "contextModification": "", "errorMessage": ""}'
  exit 0
}
```

That is correct on all three cases: PS 5.1 on Windows (undefined → proceeds), PS 7 on Windows (`$true` → proceeds), pwsh on Unix (`$false` → no-ops).

The new test asserts the exact guard line and that it precedes both the stdin read and the first command. `pnpm cicheck` green; e2e-hooks 81 passed.

dyoshikawa:
Second follow-up: residual mid fixed, plus the narrative correction.

**POSIX guard added.** Confirmed the mechanism in `hook-file-hooks.ts`: `normalizeHookInterpreter` maps a `#!/bin/bash` shebang to a bare `bash` with no platform check, so on Windows with Git Bash on `PATH` the extensionless script runs under bash while the `.ps1` runs under PowerShell 5.1 (where the `$IsWindows` guard correctly lets it proceed). The generated POSIX script now stands down there:

```bash
case "${OSTYPE:-$(uname -s 2>/dev/null || true)}" in
  msys*|MSYS*|cygwin*|CYGWIN*|MINGW*|mingw*)
    printf '{"cancel": false, "contextModification": "", "errorMessage": ""}\\n'
    exit 0
    ;;
esac
```

Verified empirically rather than by inspection — generated a real script and ran it:

- native Linux → the hook command **executed** (side-effect file created), guard not taken
- `OSTYPE=msys bash script` → command **did not execute**, guard taken
- `bash -n` → syntax OK
- also checked that bash does not reset an inherited `OSTYPE`, so the simulation is faithful; on real Git Bash bash sets `OSTYPE=msys` itself

Safe in all four quadrants: the VS Code extension never spawns the extensionless file on Windows, no Unix shell reports these `$OSTYPE` values, and on Windows the `.ps1` is the spelling that actually works (its commands go through `cmd /c`).

**Narrative correction — my earlier comment overstated the Unix case.** I claimed the pre-fix Unix path was double execution. It was not: the `.ps1` body shells out through `cmd /c`, which does not exist on Unix, so it **errored on every fire** — noise, single execution. Double execution is real only in the Windows + Git Bash quadrant. The generator comment, `src/types/hooks.ts`, and the `file-formats.md` sentence ("Without it the commands would run twice per fire on that runtime") have all been corrected to say this precisely.

New test pins the POSIX guard line and its position before both the stdin read and the first command, mirroring the existing `.ps1` test. `esac`, `msys` and `OSTYPE` added to `cspell.json`.

`pnpm cicheck` green; e2e-hooks 81 passed.

Issue #2596, opened 2026-08-06
Follow up Zoo Code upstream updates: mode-specific rules/skills directories (.roo/rules-{mode}/, .roo/skills-{mode}/) not authorable; Roo EOL docs note


dyoshikawa:
## Summary

Zoo Code (the active continuation of Roo Code) reads mode-specific rule directories (`.roo/rules-{mode}/`) and mode-specific skill directories (`.roo/skills-{modeSlug}/`) that rulesync cannot author — a rulesync rule or skill can only apply to all modes. Since rulesync's `roo` and `zoocode` targets share the `.roo/` layout, both are affected, but Roo is EOL-frozen so the actionable target is `zoocode`. A small docs note about Roo's EOL status is also proposed.

This is the exact Zoo/Roo parallel of the open Kilo issue #2499 (`.kilo/rules-{mode}/`).

## Recent Releases

- Roo Code: final release **v3.54.0** (2026-05-15); repository archived (see #2409, closed as final-state).
- Zoo Code: continues the numbering from v3.54.0 (2026-05-16); latest **v3.76.0** (2026-08-06) — https://github.com/Zoo-Code-Org/Zoo-Code/releases
- v3.76.0 Destructive Command Guard, v3.64.0 Rules Management UI, v3.60.0 `rootResolution` are VS Code UI/runtime features with no committable config surface (checked, not gaps). The v3.60.0 per-mode `allowedMcpServers` is already followed (#2503).

### 1. rules — mode-specific `.roo/rules-{mode}/` not authorable

- **Upstream**: mode-specific custom instructions live in `.roo/rules-{mode-slug}/` (fallback `.roorules-{mode-slug}`), and global modes read `~/.roo/rules-{slug}/` — https://docs.zoocode.dev/features/custom-modes , https://docs.zoocode.dev/features/custom-instructions
- **rulesync** (`project`, `global`): `RooRule.getSettablePaths()` is fixed to `.roo/rules`; `extractModeFromPath()` (`src/features/rules/roo-rule.ts:89`) can parse mode slugs but is never called from the implementation (test-only, effectively dead code).
- **Follow-up**: add a `zoocode.mode` / `roo.mode` rule-frontmatter key that routes output to `.roo/rules-{mode}/`; wire `extractModeFromPath()` on import to restore the slug. The concrete pain: rulesync can generate a custom mode (subagent) but cannot ship that mode's dedicated rules.

[...]

PR #2678, opened 2026-08-13
fix(roo): stop stripping the denylist-only disabledTools filter

dyoshikawa:
## Summary

Both the `roo` and `zoocode` MCP factories declared `supportsDisabledTools: false`, so the processor stripped `disabledTools` before `RooMcp` ever saw it. Since `RooMcp` spreads the server config verbatim, a filter already written in `.roo/mcp.json` — by hand, or by Zoo Code's own per-tool toggles, which write the same key — was deleted on the next generate.

Flips `supportsDisabledTools` to `true` for both entries. `supportsEnabledTools` stays `false`: the filter is genuinely denylist-only upstream. No adapter change is needed.

## Upstream verification (the gate for this change)

Confirmed in [`src/services/mcp/McpHub.ts`](https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/services/mcp/McpHub.ts) on Zoo Code's `main`:

- The per-server config schema declares `disabledTools: z.array(z.string()).default([])` alongside `alwaysAllow` — and **no** `enabledTools`.
- It is read straight off the MCP config: `serverConfigData.mcpServers?.[serverName]?.disabledTools`, then used as `enabledForPrompt: !disabledToolsList.includes(tool.name)`.
- Zoo's own UI writes back into the same key via `updateServerToolList(..., "disabledTools", ...)`, which is what made the silent deletion reachable without anyone hand-editing the file.

`RooCodeInc/Roo-Code` returns zero hits for the symbol, consistent with Roo being archived and Zoo Code carrying the current implementation of the shared `.roo/` layout.

## Checks

- `pnpm cicheck` — green
- `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-mcp.spec.ts` — 121 passed

A unit test on the adapter alone cannot catch this (the stripping happens in the processor), so coverage is a processor-level assertion for both targets plus an e2e generate → import round-trip through `zoocode`. The support matrix picked up the 🔧 marker for both rows via `pnpm run generate:tables`, and the MCP tool-config docs gained a Roo/Zoo note.

Part of #2596


dyoshikawa:
Record correction from the review pass: the body's claim that archived RooCodeInc/Roo-Code has zero hits for the symbol is incorrect — archived Roo's McpHub.ts declares the identical `disabledTools: z.array(z.string()).default([])` (line 72) and applies it the same way (`enabledForPrompt` at line 1030). This strengthens the change: both targets support the key natively, so the roo half of the flip rests on direct upstream support rather than harmless-extra-key reasoning.

PR #2679, opened 2026-08-13
feat(deepagents): translate MCP tool filters and normalize transports

### LINKED ISSUE #2663: Follow up deepagents-cli upstream updates: MCP tool filters stripped, transports unnormalized so `local` servers are dropped
## Summary

Two MCP surfaces are unfollowed for `deepagents`. Per-server tool filtering (`allowedTools` / `disabledTools`) is stripped by the MCP processor before the adapter ever sees it, and MCP transports are forwarded without normalization, so a server authored with rulesync's own canonical `local` spelling is silently discarded by dcode at load time. Every other dimension was verified current against `deepagents-code` 0.1.55.

## Recent Releases

Window reviewed: `deepagents-code` 0.1.0 (2026-05-12) → 0.1.55 (2026-08-12), plus the `deepagents-cli` changelog for surfaces inherited by the split-out `deepagents-code` package.

- **0.1.53** — [`PostToolUseFailure` hook event](https://github.com/langchain-ai/deepagents/blob/main/libs/code/CHANGELOG.md) (#5315) — already emitted by rulesync.
- **0.1.52** (2026-08-04) — Hooks v2 GA — already emitted by rulesync in the v2 document shape.
- **0.1.42** (2026-07-17) — Plugins GA — manifests are `.claude-plugin/plugin.json` / `.codex-plugin/plugin.json` inside installed plugin roots, managed via `/plugins` and `dcode plugin`. No repo-committed project-level plugin config, so not a rulesync surface today.
- **`deepagents-cli` 0.0.46** (2026-04-30) — [`allowedTools` and `disabledTools` filters in MCP config](https://github.com/langchain-ai/deepagents/blob/main/libs/cli/CHANGELOG.md) (#2835) — carried into `deepagents-code`; the basis of gap 1.

Documentation note: the docs URLs recorded in `.claude/skills/rulesync-feature-research/references/deepagents.md` (`https://docs.langchain.com/oss/javascript/deepagents/cli/...`) have moved to `https://docs.langchain.com/oss/deepagents/code/...`. Worth refreshing the reference map alongside this work.

### 1. mcp — per-server tool filtering is stripped, though dcode supports it

- **Upstream**: `.mcp.json` server entries accept `allowedTools` and `disabledTools` (camelCase; each entry is a tool name or an `fnmatch` glob). The two are mutually exclusive on one server, and an empty list is rejected. Documented at [MCP tools](https://docs.langchain.com/oss/deepagents/code/mcp-tools); implemented in `libs/code/deepagents_code/mcp_tools.py` (`_validate_tool_filter_fields`, `_apply_tool_filter`).
- **rulesync**: `project` + `global` for MCP, but tool filtering is `unsupported`. `src/features/mcp/mcp-processor.ts:286-296` declares `supportsEnabledTools: false, supportsDisabledTools: false` for `deepagents`, so `stripMcpServerFields` (`mcp-processor.ts:774-776`) removes both canonical keys before `DeepagentsMcp` runs; `src/features/mcp/deepagents-mcp.ts:74` then writes the servers through verbatim. A user who authors a filter loses it entirely for this target, and the matrix row is missing the 🔧 marker.
- **Proposed follow-up**: flip both `meta` flags to `true` for `deepagents` and add a key translation in `deepagents-mcp.ts` — canonical `enabledTools` → `allowedTools` (upstream does **not** read `enabledTools`, and unknown keys are silently ignored, so a raw pass-through would be a no-op), canonical `disabledTools` → `disabledTools` verbatim, and the inverse on import. Enforce upstream's constraints when emitting: warn and drop when both lists are present on one server, and skip empty arrays. Then run `pnpm run generate:tables` so the README and `docs/reference/supported-tools.md` pick up the marker, and update the deepagents note in `docs/reference/file-formats.md`.

### 2. mcp — transports are forwarded unnormalized, so `local` servers are silently dropped by dcode

- **Upstream**: `_resolve_server_type` / `_validate_server_config` accept only `stdio`, `http` and `sse`, plus the aliases `streamable-http` / `streamable_http` → `http`. `local` and `ws` are **not** accepted; such a server raises a `ValueError` that `_drop_invalid_mcp_config_servers` catches, so that one server is dropped with a log line while the rest of the file loads. Source: `libs/code/deepagents_code/mcp_tools.py`; docs [MCP tools](https://docs.langchain.com/oss/deepagents/code/mcp-tools).
- **rulesync**: `project` + `global`. `DeepagentsMcp.fromRulesyncMcp` (`src/features/mcp/deepagents-mcp.ts:60-83`) assigns `rulesyncMcp.getMcpServers()` straight into `mcpServers` with no transport mapping and no use of `src/features/mcp/mcp-transport.ts`. Rulesync's canonical `type` / `transport` vocabulary includes `local` and `ws` (`docs/reference/file-formats.md`), so an author using the canonical `local` spelling — the one rulesync itself documents — gets a server dcode discards, with no warning at generate time.
- **Proposed follow-up**: map canonical `local` → `stdio` and `streamable-http` → `http` on generate, keep `stdio` / `http` / `sse` as-is, and skip `ws` servers with `warnAndSkipMcpServer` (the treatment Kimi Code already uses). On import, normalize dcode's `streamable_http` / `streamable-http` back to canonical `http`.

## Checked and NOT a gap

- **rules** — `.deepagents/AGENTS.md` (project) and `~/.deepagents/<agent>/AGENTS.md` (global, default agent `deepagents`) match `deepagents-rule.ts` and both dry-runs.
- **subagents** — `.deepagents/agents/<name>/AGENTS.md` and the global equivalent; upstream frontmatter is `name` / `description` / `model` only, with `tools`, `middleware`, `interrupt_on` and `skills` explicitly SDK-only.
- **skills** — `.deepagents/skills` and `~/.deepagents/<agent>/skills` are covered. Upstream's additional read locations (`.agents/skills`, `~/.agents/skills`, experimental `.claude/skills`, `[skills].extra_allowed_dirs` / `DEEPAGENTS_CODE_EXTRA_SKILLS_DIRS`) are alternate discovery paths, not a capability rulesync fails to emit. `allowed-tools` / `license` / `compatibility` / `metadata` are already modeled.
- **hooks** — all 12 v2 events are emitted in the v2 `{ "hooks": { "<Event>": [{ matcher, hooks }] } }` shape. The handler field `argv` is documented but is an explicitly temporary legacy-migration field slated for removal after 2026-09-01, so emitting only `command` is correct. rulesync still *imports* the deprecated pre-v2 flat list, which is the right direction.
- **commands** — correctly `unsupported`: dcode has only built-in slash commands plus skill-derived `/skill:<name>`, with no user-authored commands directory.
- **ignore** — correctly `unsupported`: no `.deepagentsignore` / `.agentignore` surface; the model is workspace trust, not file exclusion.
- **permissions** — correctly `unsupported`: approval modes live in the global-only `~/.deepagents/config.toml` (`[startup] mode = manual|auto|yolo`, `[models] auto_classifier`), plus env vars and the ephemeral `--shell-allow-list`. There is no project-scoped, checked-in allow/deny file.

## References

- https://docs.langchain.com/oss/deepagents/code/mcp-tools — the `allowedTools` / `disabledTools` contract and the accepted transport set (gaps 1 and 2).
- https://github.com/langchain-ai/deepagents/blob/main/libs/cli/CHANGELOG.md — 0.0.46, where the tool filters were introduced (gap 1).
- https://github.com/langchain-ai/deepagents/blob/main/libs/code/CHANGELOG.md — the 0.1.x release window used to date hooks v2 GA and `PostToolUseFailure` (non-gaps).
- https://docs.langchain.com/oss/deepagents/code/configuration — current configuration surface, used to confirm the docs URL move noted above.
- https://docs.langchain.com/oss/deepagents/code/approval-modes — confirms approval modes are global-only, so `permissions` stays `unsupported`.
- https://docs.langchain.com/oss/deepagents/code/plugins — confirms plugin manifests live inside installed plugin roots, not in the repo.

dyoshikawa:
## Summary

Two MCP surfaces for `deepagents`, both failing **silently**: dcode drops an individual server it cannot validate and keeps loading the rest of the file, so neither problem surfaced at generate time.

### Tool filters

The processor declared `supportsEnabledTools: false, supportsDisabledTools: false`, so `stripMcpServerFields` removed both canonical keys before `DeepagentsMcp` ran. Both flags are now `true`, and the adapter translates:

- canonical `enabledTools` → `allowedTools` (dcode never reads `enabledTools` and ignores unknown keys silently, so a raw pass-through would have been a no-op), and the inverse on import
- canonical `disabledTools` → verbatim (same name upstream)

Upstream's `_validate_tool_filter_fields` rejects a server that sets **both** filters and rejects an **empty** list, so generation writes neither filter in the first case (warning, naming the server) and omits an empty array in the second — rather than emitting a server dcode would discard.

### Transports

`fromRulesyncMcp` assigned `getMcpServers()` straight through with no mapping. dcode's `_resolve_server_type` accepts only `stdio` / `sse` / `http` plus the aliases `streamable_http` / `streamable-http` → `http`. Rulesync's canonical vocabulary is wider, so canonical `local` — the spelling rulesync itself documents — produced a server dcode threw away.

Now `local` → `stdio`, `streamable-http` → `http`, `sse`/`http`/`stdio` unchanged, and `ws` (which has no counterpart) is skipped via `warnAndSkipMcpServer`, the kimi-code treatment. The value lands under whichever of `type` / `transport` was authored, since dcode reads the two interchangeably. Import normalizes both spellings of the alias back to `http`.

## Upstream verification

Read directly from [`libs/code/deepagents_code/mcp_tools.py`](https://github.com/langchain-ai/deepagents/blob/main/libs/code/deepagents_code/mcp_tools.py) on `main`:

- `_TRANSPORT_ALIASES = {"streamable_http": "http", "streamable-http": "http"}`; `_resolve_server_type` returns `stdio` / `sse` / `http` only, inferring `http` from a `url` and `stdio` otherwise when neither key is set.
- `_validate_tool_filter_fields` raises on both-set ("pick one") and on an empty list ("must be non-empty; omit the field to disable filtering").

## Checks

- `pnpm cicheck` — green
- `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-mcp.spec.ts` — 121 passed

Coverage is adapter-level for each rule plus an e2e generate → import round-trip. `pnpm run generate:tables` added the 🔧 marker to the deepagents-cli row, and `file-formats.md` gained a deepagents MCP note (there was none before).

Closes #2663


dyoshikawa:
Follow-up commit addressing the review finding. The both-filters arm was fail-open and inconsistent with the `ws` arm three lines above it.

**Corrected semantics** (the PR body above still describes the old ones):

| Input | Before | Now |
| --- | --- | --- |
| both filters set | warn, write neither → server runs with **all** tools | `warnAndSkipMcpServer` — server skipped |
| `enabledTools: []` | dropped silently → server runs with **all** tools | `warnAndSkipMcpServer` — server skipped |
| `disabledTools: []` | dropped silently | dropped **with a warning**, server still written |

The rule applied throughout: never hand the model more tools than the canonical config allows. Setting both is valid canonical input (other targets apply the two lists independently) but has no form here, and upstream `_validate_tool_filter_fields` raises on it — so refusing the server matches upstream rather than quietly widening access.

**On the two empty forms, which mean opposite things.** An empty `enabledTools` is an allowlist that allows nothing, so skipping the server is both fail-closed and semantically faithful — dropping the key would have published every tool. An empty `disabledTools` denies nothing, so it is a true no-op: only the key is dropped, because skipping the server there would remove tools the config never denied. Upstream rejects both empty forms, so neither reaches the file either way.

**E2E fixture interaction worth flagging.** The shared global-mode MCP matrix fixture deliberately sets **both** filters on one server (hermesagent asserts `include`/`exclude` from it), so under the new semantics deepagents legitimately skips that server and the matrix assertion no longer holds for it. Rather than weaken the shared fixture, `deepagents` moved out of `mcpGlobalTargets` into a dedicated global test — the same treatment `takt` already gets in this file, including the `untested` escape hatch on the completeness check. The dedicated test keeps the home-directory coverage using one filter per server, and additionally pins the skip end to end.

`pnpm cicheck` green; full `e2e-mcp` 121 passed.

PR #2702, opened 2026-08-17
ci: add a scheduled Dependabot auto-merge workflow driven by OpenCode

### LINKED ISSUE #2686: Auto-fix and merge Dependabot PRs with the OpenCode GitHub agent
## Background

Dependabot opens grouped bump PRs weekly for npm, GitHub Actions, and Docker (see `.github/dependabot.yml`). Today every one of them is babysat by hand: a maintainer has to check CI, diagnose a failing bump, sometimes exclude a single breaking dependency from the group, and finally merge. The repository already has a `babysit-dependabot-pr` skill that encodes exactly that procedure, and it already runs OpenCode inside GitHub Actions for `draft-release.yml`. Wiring the two together would remove the last manual step from routine dependency maintenance.

## Details

- **Agent**: `anomalyco/opencode/github` action, the same SHA-pinned usage as `.github/workflows/draft-release.yml`.
- **Model**: OpenRouter DeepSeek V4 Flash 0731. Verified against models.dev (`https://models.dev/api.json`): the `openrouter` provider exposes the model id `deepseek/deepseek-v4-flash-0731`, so the OpenCode model spec is:

  ```
  openrouter/deepseek/deepseek-v4-flash-0731
  ```

  Note that the existing `.github/actions/select-opencode-model` action maps the `deepseek` keyword to `openrouter/deepseek/deepseek-v4-pro`, so it cannot be reused as-is for this workflow — the model should be hard-coded (or the action extended).
- **Prompt**: run the existing `babysit-dependabot-pr` skill for each open Dependabot PR. The skill already verifies the author is the genuine Dependabot bot (`login == dependabot[bot]`, `type == Bot`, `id == 49699333`), diagnoses CI failures, excludes or fixes a breaking bump, and merges.

### Why not trigger on the PR event

Workflows triggered by Dependabot-originated `pull_request` events receive the Dependabot secrets store, not the regular Actions secrets, so `OPENROUTER_API_KEY` would not be available. `pull_request_target` would make the secrets available but widens the injection surface for event data without reducing the real risk (CI executes the bumped dependencies either way).

A `schedule` (cron) + `workflow_dispatch` trigger avoids both problems: the workflow definition always runs from `main`, regular secrets are available, and the prompt can stay fully static — the agent discovers the PRs itself with `gh pr list --author "app/dependabot"`, which satisfies the "never interpolate untrusted event data into prompts" rule in `.github/workflows/AGENTS.md`.

## Solution / Next Steps

1. Add `.github/workflows/dependabot-auto-merge.yml`:
   - `on: schedule` (daily) + `workflow_dispatch`.
   - `permissions: contents: write`, `pull-requests: write`, `issues: write`.
   - Setup steps mirroring `draft-release.yml`: checkout (`fetch-depth: 0`), `git-config`, mise, Takumi Guard npm registry, `pnpm install --ignore-scripts`, `pnpm generate`.
   - A cheap `gh pr list` pre-check that exits early when there is no open Dependabot PR, so no tokens are burned on empty runs.
   - `anomalyco/opencode/github` with `model: openrouter/deepseek/deepseek-v4-flash-0731` and a static prompt that runs the `babysit-dependabot-pr` skill for every listed PR.
   - `concurrency` group and `timeout-minutes` to avoid overlapping runs pushing to the same branches.
2. Confirm the merge path works for the default `GITHUB_TOKEN`. The skill's Step 4 uses `gh pr merge --admin`, but `GITHUB_TOKEN` has no admin bypass on this repository's ruleset — if branch protection requires an admin merge, either the workflow needs a PAT with the right permissions or the ruleset needs a bypass actor for the Actions app. Resolve this before declaring the automation done.
3. Validate with `actionlint` (already part of CI) and a manual `workflow_dispatch` run.

dyoshikawa:
> [!IMPORTANT]
> **Please review and merge this one by hand.** It adds a GitHub Actions workflow, which the autonomous batch flow does not auto-merge, and it has one open question (below) that only you can settle.

## Background

Dependabot opens grouped bump PRs weekly for npm, GitHub Actions and Docker, and each one is babysat by hand today. The `babysit-dependabot-pr` skill already encodes that whole procedure, and `draft-release.yml` already runs OpenCode inside Actions, so this wires the two together.

## Changes

Adds `.github/workflows/dependabot-auto-merge.yml`:

- `schedule` (daily, 03:00 UTC) + `workflow_dispatch`. Deliberately **not** the `pull_request` event: a Dependabot-originated run gets the Dependabot secrets store, so `OPENROUTER_API_KEY` would be unavailable, and `pull_request_target` would widen the injection surface for event data without reducing the real risk (CI executes the bumped dependencies either way). On a schedule the workflow definition always runs from `main`, regular secrets are available, and the prompt stays fully **static** — the agent discovers the PRs itself with `gh pr list --author "app/dependabot"`, so no untrusted event data is interpolated, per `.github/workflows/AGENTS.md`.
- A cheap `gh pr list` pre-check gates every later step, so an empty run burns no model tokens.
- Setup steps mirror `draft-release.yml` (checkout `fetch-depth: 0`, `git-config`, mise, Takumi Guard npm registry, `pnpm install --ignore-scripts`, `pnpm generate`), all SHA-pinned to the same versions.
- `concurrency` group and `timeout-minutes: 60` so overlapping runs cannot push to the same branches.
- Model hard-coded to `openrouter/deepseek/deepseek-v4-flash-0731` — verified present on models.dev under the `openrouter` provider. `./.github/actions/select-opencode-model` cannot be reused as-is, since it maps the `deepseek` keyword to `openrouter/deepseek/deepseek-v4-pro`.
- Job permissions are `contents: write`, `pull-requests: write`, `issues: write`, with the workflow-level default kept at `contents: read`.

`actionlint` runs on this PR (it is path-triggered on `.github/workflows/**`), which covers step 3 of the issue.

## The merge path (issue step 2)

I read the ruleset (`repos/dyoshikawa/rulesync/rulesets/6248659`) rather than leaving this open:

- Its only bypass actor is `RepositoryRole` id 5 with `bypass_mode: pull_request`, so the workflow `GITHUB_TOKEN` is **not** a bypass actor and `gh pr merge --admin` cannot force a merge from Actions. The prompt now tells the agent to use `gh pr merge <number> --merge` and to report rather than retry with `--admin` if the ruleset refuses — that keeps the shared `babysit-dependabot-pr` skill untouched for interactive use.
- The ruleset sets `required_approving_review_count: 0` and there is no `CODEOWNERS` file, so `require_code_owner_review` has no owners to demand. A plain `--merge` on a green PR should therefore satisfy it, and no PAT or new bypass actor is needed today.

If reviews ever become mandatory on `main`, this will need either a PAT secret or an Actions bypass actor — worth revisiting then, not now.

## Also in this update

- The job is gated on `github.actor` for `workflow_dispatch`, mirroring `draft-release.yml`, so the step holding the model key and a write-scoped token is reachable only through the schedule (whose definition always comes from `main`) or a maintainer's own dispatch.
- `.rulesync/rules/github-actions-security.md` now names this workflow in the `anomalyco/opencode/github` pinning stance, since that stance was written around `draft-release.yml` being the only workflow reaching `https://opencode.ai/install`.
- The action pin is bumped to v1.18.19 to match `draft-release.yml` on `main`.
- Rebased onto current `main`.

## Verification

- `actionlint` 1.7.11 (the version CI pins) reports no findings.
- `pnpm cicheck` passes.
- A manual `workflow_dispatch` validation run has **not** been performed — it needs the repository secrets and would act on live PRs, so it is left to you.

Closes #2686

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -1,0 +1,88 @@
+name: Dependabot Auto Merge
+
+# Runs the `babysit-dependabot-pr` skill against every open Dependabot PR.
+#
+# Deliberately NOT triggered by the `pull_request` event: a Dependabot-originated
+# `pull_request` run receives the Dependabot secrets store rather than the regular
+# Actions secrets, so `OPENROUTER_API_KEY` would be unavailable, and
+# `pull_request_target` would widen the injection surface for event data without
+# reducing the real risk (CI executes the bumped dependencies either way).
+# `schedule` + `workflow_dispatch` avoids both: the workflow definition always runs
+# from `main`, regular secrets are available, and the prompt stays fully static —
+# the agent discovers the PRs itself, so no untrusted event data is interpolated
+# into it (see .github/workflows/AGENTS.md).
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: dependabot-auto-merge
+  cancel-in-progress: false
+
+jobs:
+  dependabot-auto-merge:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+        with:
+          fetch-depth: 0
+
+      # A cheap pre-check so an empty run burns no model tokens.
+      - name: Check for open Dependabot pull requests
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          count=$(gh pr list --state open --author "app/dependabot" --json number --jq 'length')
+          echo "count=${count}" >> "$GITHUB_OUTPUT"
+          echo "Open Dependabot pull requests: ${count}"
+
+      - name: Configure git
+        if: steps.check.outputs.count != '0'
+        uses: ./.github/actions/git-config
+
+      - name: Setup mise
+        if: steps.check.outputs.count != '0'
+        uses: jdx/mise-action@3c2e0cf82a5b2e5249f0d3635a4d83d0ae861518 # v4.2.5
+        with:
+          experimental: true
+
+      - name: Setup Takumi Guard npm registry
+        if: steps.check.outputs.count != '0'
+        uses: flatt-security/setup-takumi-guard-npm@6d4182745c1e474c35a023573c2612c085be45a4 # v1
+
+      - name: Install dependencies
+        if: steps.check.outputs.count != '0'
+        run: pnpm install --ignore-scripts
+
+      - name: Generate rules
+        if: steps.check.outputs.count != '0'
+        run: pnpm generate
+
+      - name: Run OpenCode to babysit Dependabot pull requests
+        if: steps.check.outputs.count != '0'
+        uses: anomalyco/opencode/github@31406ccc51b4bd2a4e1e086b2bcaa5f7f804f26d # v1.18.18
+        env:
+          OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Hard-coded rather than resolved through ./.github/actions/select-opencode-model,
+          # which maps the `deepseek` keyword to `openrouter/deepseek/deepseek-v4-pro`.
+          model: openrouter/deepseek/deepseek-v4-flash-0731
+          use_github_token: "true"
+          share: false
+          prompt: |
+            List the open Dependabot pull requests with `gh pr list --state open --author "app/dependabot"`.
+            For each one, run the `babysit-dependabot-pr` skill with that pull request number, one pull request at a time.
+            The skill verifies the author is the genuine Dependabot bot, diagnoses any CI failure, excludes or fixes a breaking bump, and merges once every check is green.
+            Do not act on any pull request whose author is not Dependabot.
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -25,6 +25,14 @@ concurrency:
 
 jobs:
   dependabot-auto-merge:
+    # The `github.actor` gate mirrors `draft-release.yml`, so the job that receives
+    # the model key and a write-scoped token stays reachable only through the
+    # schedule — whose definition always comes from `main` — or a maintainer's own
+    # dispatch.
+    if: >-
+      github.event_name == 'schedule' ||
+      github.actor == 'dyoshikawa' ||
+      github.actor == 'cm-dyoshikawa'
     runs-on: ubuntu-latest
     timeout-minutes: 60
     permissions:
@@ -71,7 +79,7 @@ jobs:
 
       - name: Run OpenCode to babysit Dependabot pull requests
         if: steps.check.outputs.count != '0'
-        uses: anomalyco/opencode/github@31406ccc51b4bd2a4e1e086b2bcaa5f7f804f26d # v1.18.18
+        uses: anomalyco/opencode/github@2b72179c663cadcb54f54d9f19221b3fb3d11fb6 # v1.18.19
         env:
           OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -86,3 +94,8 @@ jobs:
             For each one, run the `babysit-dependabot-pr` skill with that pull request number, one pull request at a time.
             The skill verifies the author is the genuine Dependabot bot, diagnoses any CI failure, excludes or fixes a breaking bump, and merges once every check is green.
             Do not act on any pull request whose author is not Dependabot.
+
+            This repository's `main` ruleset lists no bypass actor for the Actions app, so
+            `gh pr merge --admin` cannot force a merge from here. Merge with
+            `gh pr merge <number> --merge` instead, and if the ruleset rejects it, leave the
+            pull request open and report why rather than retrying with `--admin`.
--- a/.rulesync/rules/github-actions-security.md
+++ b/.rulesync/rules/github-actions-security.md
@@ -27,14 +27,14 @@ Reference: https://docs.github.com/ja/actions/concepts/security/script-injection
 
 Third-party GitHub Actions are pinned to a full 40-character commit SHA with a trailing `# vX.Y.Z` comment. Keep that convention when adding or bumping an action.
 
-`anomalyco/opencode/github` (used by `draft-release.yml`) is a composite action, so its SHA pin covers the wrapper only:
+`anomalyco/opencode/github` (used by `draft-release.yml` and `dependabot-auto-merge.yml`) is a composite action, so its SHA pin covers the wrapper only:
 
 - The wrapper resolves the `opencode` CLI release to `latest` at run time and installs it with `curl -fsSL https://opencode.ai/install | bash`, so the binary that actually receives the workflow secrets is not pinned by the SHA.
 - The wrapper also uses `actions/cache@v4` internally, by mutable tag.
 
-The accepted stance is to treat `https://opencode.ai/install` as a trusted install path rather than vendoring a pinned installer, because `draft-release.yml` is `workflow_dispatch`-only and gated on `github.actor`, so an outside contributor cannot trigger it. Bumping the pin is still worthwhile for the wrapper itself, but do not read it as a guarantee about the CLI version.
+The accepted stance is to treat `https://opencode.ai/install` as a trusted install path rather than vendoring a pinned installer, because the workflows that reach it cannot be triggered by an outside contributor: `draft-release.yml` is `workflow_dispatch`-only and gated on `github.actor`, and `dependabot-auto-merge.yml` runs on a `schedule` — whose definition always comes from `main` — plus a `workflow_dispatch` carrying the same actor gate. Bumping the pin is still worthwhile for the wrapper itself, but do not read it as a guarantee about the CLI version.
 
-The residual risk this stance accepts is that a compromise of the distribution endpoint would expose whatever the workflow hands the CLI — the model API keys and a `GITHUB_TOKEN` with `contents: write`, `pull-requests: write`, and `issues: write`. The `github.actor` gate limits _who can trigger_ the workflow; it does not reduce that exposure on a maintainer-triggered run. Revisit this if the workflow ever becomes externally triggerable, if the token or secret scope it receives grows, or if upstream adds a `version` input that lets the CLI itself be pinned.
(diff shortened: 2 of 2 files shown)


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+  cancel-in-progress: false
+
+jobs:
+  dependabot-auto-merge:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
```
dyoshikawa, line 41: Declaring `permissions` here makes every unlisted scope `none`, so the skill's two core steps will 403: `gh pr checks` needs `checks: read` and `gh run view --job <id> --log` needs `actions: read`. `ci-failure-reminder.yml` already declares `checks: read` / `statuses: read` for the same purpose — worth matching it.

Also, `contents: write` is broader than this needs. The `main` ruleset stops a direct push to main, but it doesn't stop pushes or force-pushes to any other branch, and tag creation isn't restricted at all.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+# `pull_request_target` would widen the injection surface for event data without
+# reducing the real risk (CI executes the bumped dependencies either way).
+# `schedule` + `workflow_dispatch` avoids both: the workflow definition always runs
+# from `main`, regular secrets are available, and the prompt stays fully static —
+# the agent discovers the PRs itself, so no untrusted event data is interpolated
+# into it (see .github/workflows/AGENTS.md).
+on:
+  schedule:
+    - cron: "0 3 * * *"
+  workflow_dispatch:
```
dyoshikawa, line 17: `draft-release.yml` guards its `workflow_dispatch` with `if: github.actor == 'dyoshikawa' || github.actor == 'cm-dyoshikawa'`. This workflow starts an agent with a write token and merge rights, so it should have at least the same guard.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+  dependabot-auto-merge:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: write
+      pull-requests: write
+      issues: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
```
dyoshikawa, line 44: `93cb6efe...` is v5.0.1. Nine of the ten `actions/checkout` uses in this repo are already on `8e8c483db84b4bee98b60c0593521ed34d9990e8` (v6.0.1) — only `draft-release.yml` is still on v5.0.1. Since `.github/dependabot.yml` ignores `actions/checkout`, this pin won't be updated automatically, so a new file should start on v6.0.1.

While here: no `persist-credentials: false`, so the `GITHUB_TOKEN` stays in `.git/config` as an `http.extraheader` and is readable by anything that runs later in the job. That matters given the untrusted dependency code the skill ends up executing.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          count=$(gh pr list --state open --author "app/dependabot" --json number --jq 'length')
+          echo "count=${count}" >> "$GITHUB_OUTPUT"
+          echo "Open Dependabot pull requests: ${count}"
+
+      - name: Configure git
+        if: steps.check.outputs.count != '0'
```
dyoshikawa, line 59: `if: steps.check.outputs.count != '0'` is repeated on six consecutive steps. Splitting the pre-check into its own job with an output and gating this job with `needs:` + a single job-level `if` would express the intent once instead of six times.

The condition is also fail-open: if `gh` returns an empty string, `'' != '0'` is true and everything runs anyway. `has_prs=true/false` or an explicit `!= ''` check would be safer.

dyoshikawa (commented):
I reviewed this with a code-quality pass and a security pass. The trigger choice is right — avoiding `pull_request_target` and running the definition from `main` is the correct call, and the static prompt does satisfy the script-injection rules in `.github/workflows/AGENTS.md`. But there are three things I don't think we can merge past, and they're all about the gap between "the prompt is static" and "what the agent actually reads and can do at runtime".

The first is mechanical: with `use_github_token: "true"`, the agent's pushes are made with `GITHUB_TOKEN`, and those pushes do not start a new workflow run. The skill's Step 3-2 pushes a fix to the Dependabot branch and Step 3-3 then waits for every check to pass — but no checks will ever appear for the new head SHA. Best case the job burns the 60-minute timeout; worst case the model reads the previous SHA's green checks and merges a commit that CI never ran against. This needs either a GitHub App / fine-grained PAT for the push, or an explicit re-trigger, plus a hard rule in the skill that zero checks on the head SHA means stop.

The second is secret placement. The workflow's own `pnpm install --ignore-scripts` runs against `main`, which is fine — but the skill then checks out the PR head, runs a plain `pnpm install`, and runs `pnpm cicheck`, which loads and executes the bumped dependency code. That happens in the same job that holds `OPENROUTER_API_KEY` and a `contents: write` token. The comment on line 9 says CI executes the bumped dependencies either way, and that isn't equivalent: `ci.yml` runs with `contents: read`, no `OPENROUTER_API_KEY`, and `--ignore-scripts`. `minimumReleaseAge`, `allowBuilds` and Takumi Guard help, but none of them stop code that runs during tests. Splitting the verification into a job without the API key, or leaving verification to the normal CI run after the push, would close this.

The third is indirect prompt injection. The prompt has no `${{ }}` in it, but at runtime the agent reads Dependabot PR bodies (which embed upstream release notes written by any package maintainer) and full CI logs (which contain whatever the bumped dependency printed). It reads all of that while holding bash, merge rights, a write token and an LLM API key. The skill tells the agent to treat titles, branch names and comments as untrusted, but it doesn't name CI logs or release notes, and `deepseek-v4-flash-0731` is a weak model to be making the merge call. Worth adding those two sources to the untrusted list, trimming logs to the failing lines, and using a stronger model for anything that decides to merge.

Separately, the PR description already notes that `gh pr merge --admin` will almost certainly be rejected because the default `GITHUB_TOKEN` isn't a bypass actor. I'd rather settle that before this lands than have a red run at 03:00 UTC every day. There's also no required-status-checks rule on the `main` ruleset today, so nothing structurally prevents merging a red PR — the skill's own judgement is the only gate. Adding required checks would make this a lot safer regardless of what happens with the token.

The rest of the line comments are smaller and can be folded in with the above.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          fetch-depth: 0
+
+      # A cheap pre-check so an empty run burns no model tokens.
+      - name: Check for open Dependabot pull requests
+        id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          count=$(gh pr list --state open --author "app/dependabot" --json number --jq 'length')
```
dyoshikawa, line 54: No `set -euo pipefail`, and `gh pr list` defaults to `--limit 30`. The limit is fine in practice (the `dependabot.yml` open-PR limits add up to 4), but a failing `gh` call here silently produces an empty `count` that the downstream gate reads as "not zero".

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+      - name: Run OpenCode to babysit Dependabot pull requests
+        if: steps.check.outputs.count != '0'
+        uses: anomalyco/opencode/github@31406ccc51b4bd2a4e1e086b2bcaa5f7f804f26d # v1.18.18
+        env:
+          OPENROUTER_API_KEY: ${{ secrets.OPENROUTER_API_KEY }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Hard-coded rather than resolved through ./.github/actions/select-opencode-model,
+          # which maps the `deepseek` keyword to `openrouter/deepseek/deepseek-v4-pro`.
+          model: openrouter/deepseek/deepseek-v4-flash-0731
```
dyoshikawa, line 89: The reason for hard-coding is real — `select-opencode-model` has no entry for `deepseek-v4-flash-0731` — but the fix is to add the keyword to that action rather than bypass the one place models are resolved. Otherwise this file gets left behind on the next model migration, and a typo in the model id won't surface until the 03:00 UTC run.

Separately, this agent holds `contents: write`, rewrites lockfiles and decides what to merge, and it's on a weaker flash model than the `glm-5.1` default the other OpenCode workflows use. That trade seems backwards for the workflow with the most authority.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+        if: steps.check.outputs.count != '0'
+        run: pnpm install --ignore-scripts
+
+      - name: Generate rules
+        if: steps.check.outputs.count != '0'
+        run: pnpm generate
+
+      - name: Run OpenCode to babysit Dependabot pull requests
+        if: steps.check.outputs.count != '0'
+        uses: anomalyco/opencode/github@31406ccc51b4bd2a4e1e086b2bcaa5f7f804f26d # v1.18.18
```
dyoshikawa: This is v1.18.18; #2716 bumps the same action to v1.18.19 in `draft-release.yml`, which is currently the only reference. Whichever lands second will leave one file a version behind, so it's worth rebasing one onto the other.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+  contents: read
+
+concurrency:
+  group: dependabot-auto-merge
+  cancel-in-progress: false
+
+jobs:
+  dependabot-auto-merge:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
dyoshikawa, line 37: 60 minutes is tight for "process each Dependabot PR one at a time, waiting for CI to go green on each" — a single CI run is already tens of minutes. On timeout you'd be left with a Dependabot branch that has an unmerged, unreported fix pushed to it. Handling one PR per run (`gh pr list --limit 1`) would make the state transitions much easier to reason about.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          experimental: true
+
+      - name: Setup Takumi Guard npm registry
+        if: steps.check.outputs.count != '0'
+        uses: flatt-security/setup-takumi-guard-npm@6d4182745c1e474c35a023573c2612c085be45a4 # v1
+
+      - name: Install dependencies
+        if: steps.check.outputs.count != '0'
+        run: pnpm install --ignore-scripts
```
dyoshikawa, line 74: This installs from `main`'s lockfile, but the skill later switches the working tree to the Dependabot branch and runs `pnpm cicheck` without reinstalling. That verifies the bumped code against the old `node_modules`. The skill should re-run the install after checking out the PR head.

dyoshikawa:
All 11 CI checks are green on the current head.

This PR is **deliberately left unmerged**. It adds a GitHub Actions workflow
(`.github/workflows/dependabot-auto-merge.yml`), which the autonomous batch flow
classifies as a high-risk change that must not be auto-merged. Please review and
merge it manually.

Summary of what is in it:

- New `dependabot-auto-merge.yml` workflow, gated with the same `github.actor`
  check as `draft-release.yml` so the job that receives the model key and a
  write-scoped token is reachable only through the schedule (whose definition
  always comes from `main`) or a maintainer's own dispatch.
- `anomalyco/opencode/github` pinned to `2b72179c663cadcb54f54d9f19221b3fb3d11fb6` (v1.18.19).
- The agent prompt tells the job to merge with `gh pr merge <number> --merge`,
  not `--admin`: the `main` ruleset lists no bypass actor for the Actions app,
  so `--admin` cannot force a merge from there.
- `.rulesync/rules/github-actions-security.md` updated so the supply-chain
  stance names the new workflow alongside `draft-release.yml`.

dyoshikawa:
@dyoshikawa Thank you!

PR #2705, opened 2026-08-17
fix: resolve the four 2026-08-17 review follow-ups

### LINKED ISSUE #2704: Review follow-ups from the 2026-08-17 batch: claudecode mask entries, antigravity enum filters, devin user-invocable docs, MCP collision precedence
## Background

Code review of the 2026-08-17 batch (PRs #2690–#2701) surfaced four findings that are real but out of scope for the PRs they were found on. PR #2703 fixed the three defects; these four are recorded here so they are not re-derived.

## 1. claudecode: `credentials.files[].mode: "mask"` is still emitted at project scope (security-relevant)

#2694 filters the user/managed-only `sandbox.*` **keys** at project scope, and deliberately left `credentials.files` alone because the same list carries `deny` entries that project settings **do** honor. But the docs are explicit that the mask half is ignored there:

> Unlike `deny`, masking authorizes the proxy to send your real credential to the listed hosts, so it is honored only from settings you or your administrator control: user settings, managed settings, and the `--settings` CLI flag. `mask` entries, `network.tlsTerminate`, and `credentials.allowPlaintextInject` … are all ignored in a repository's `.claude/settings.json` or `.claude/settings.local.json`.

A **per-entry** filter (drop only entries whose `mode === "mask"`, keep the rest) closes the gap without dropping honored `deny` entries.

Failure: a user authors `claudecode.sandbox.credentials.files: [{ path: "~/.aws/credentials", mode: "mask" }]`, a project-scope generate commits it, Claude Code ignores it entirely, and the user believes the credential is masked inside the sandbox when nothing is protecting it. That is the same "reads as though it enforced a policy that does nothing" hazard #2694 exists to remove, on the one key whose consequence is a live credential reaching the network unmasked.

- https://code.claude.com/docs/en/sandboxing

## 2. antigravity-cli: `toolPermission` / `artifactReviewPolicy` lack the enum import filter `agentMode` has

#2691 added `agentMode` with an import-side check against its three documented values, because the override schema declares it as `z.enum` and an unrecognized value would fail validation of the **whole** `.rulesync/permissions.jsonc`, not just that key. Its two siblings are declared as `z.enum` too (`src/types/permissions.ts`) but imported with a bare `typeof settings.X === "string"` check (`src/features/permissions/antigravity-cli-permissions.ts`).

Failure (forward-compat, not a typo): Antigravity ships a fifth `toolPermission` preset, the CLI writes it into the user's real `~/.gemini/antigravity-cli/settings.json`, `rulesync import` carries it verbatim into the override block, and the resulting permissions file is then rejected wholesale on its next validated parse.

Related: the three `agentMode` values are currently duplicated between `ANTIGRAVITY_CLI_AGENT_MODES` and the zod enum. Deriving one from the other would remove the drift class along with the fix.

## 3. `user-invocable`: `devin` consumes the root-level value but is in none of the documented lists

`src/features/skills/devin-skill.ts` reads the **root-level** `user-invocable` and maps it into Devin's `triggers` (`false` → model-only). But `devin` appears in none of the three enumerations #2693 updated: the `resolveUserInvocable` JSDoc (`skills-utils.ts`), the schema comment in `rulesync-skill.ts`, and the two `user-invocable` notes in `docs/reference/file-formats.md`.

Failure: a user sets root `user-invocable: false` believing it reaches the eight documented targets, and their Devin skill also silently becomes model-only. Devin consumes only the root value — it has no section override — so the lists should say so.

## 4. copilotcli MCP: the collision precedence differs from both in-repo neighbours

PR #2703 settled `copilotcli` on "canonical `enabledTools` wins over native `tools`". The two adjacent cases resolve the same shape differently, and neither is cross-referenced:

- `kiro-mcp.ts` **merges** a tool-native spelling with the canonical list ("the two lists are merged rather than one overwriting the other").
- `codexcli-mcp.ts` treats a canonical `tools` array as the wrong key entirely and refuses it ("Codex reads it as a per-tool approval table, not a tool allowlist").

Worth one pass to either align them or document in each docstring why the three differ, so the next adapter has a rule to follow rather than three precedents.

## Notes

Found by review of merged PRs; none of these blocks anything shipped. Item 1 is the one with a security consequence.

dyoshikawa:
## Background

Code review of the 2026-08-17 batch recorded four follow-ups in #2704 that were out of scope for the PRs they were found on. This PR resolves all four.

### 1. claudecode: drop project-ignored `mask` credential entries (security-relevant)

`sandbox.credentials.files[]` / `envVars[]` entries with `"mode": "mask"` are honored only from user settings, managed settings and `--settings` — they are ignored outright in a repository's `.claude/settings.json`. Committing one reads as though the credential were masked while nothing protects it.

Filtered **per entry**, not per key, so the `deny` entries the same lists carry — which project settings *do* honor — are kept. `envVars` is included alongside the `files` list named in the issue: the docs state the same settings-source restriction for both, and the hazard is identical.

Verified against <https://code.claude.com/docs/en/sandboxing>: *"`mask` entries, `network.tlsTerminate`, and `credentials.allowPlaintextInject` … are all ignored in a repository's `.claude/settings.json` or `.claude/settings.local.json`"*, and the file-entry section states *"the settings-source restriction work[s] the same way as for masked environment variables"*.

### 2. antigravity-cli: enum import filters for `toolPermission` / `artifactReviewPolicy`

Both are declared `z.enum` in the override schema but were imported with a bare `typeof === "string"` check, so a preset a newer Antigravity release adds would be carried verbatim into the override block and then fail validation of the **whole** `.rulesync/permissions.jsonc`. They now go through the same documented-value check `agentMode` already had.

The three value lists are single-sourced in `src/types/permissions.ts` and feed both the zod enums and the import filters, which removes the drift the issue flagged between `ANTIGRAVITY_CLI_AGENT_MODES` and its enum.

### 3. `user-invocable`: record that `devin` consumes the root value

`devin-skill.ts` maps the root-level `user-invocable` / `disable-model-invocation` onto Devin's `triggers`, but `devin` appeared in none of the lists enumerating the targets those flags reach. Added to the `resolveUserInvocable` JSDoc, the `rulesync-skill.ts` schema comments and `docs/reference/file-formats.md`, noting that Devin reads only the root value and has no section override. The `disable-model-invocation` sibling note had the identical gap on the adjacent line, so it is corrected too.

### 4. MCP canonical-vs-native collision precedence

Documented the rule once on the canonical `enabledTools` field in `src/types/mcp.ts` — what the *native* key means decides the resolution — and cross-referenced it from all three adapters, so the next one has a rule rather than three unexplained precedents:

- native key is a redundant spelling of the same additive list → **merge** (`kiro-mcp.ts`)
- same concept, but rulesync owns the generated value → **canonical wins**, native dropped with a warning (`copilotcli-mcp.ts`)
- same-named native key means something else → **refuse the canonical value** (`codexcli-mcp.ts`)

No behavior change for item 4; items 1 and 2 come with tests.

## Verification

`pnpm cicheck` — all green (373 test files, 8740 tests).

Closes #2704


dyoshikawa:
@dyoshikawa Thank you!

PR #2706, opened 2026-08-18
feat(rules): emit devin global non-root rules into ~/.devin/rules

### LINKED ISSUE #2688: Follow up Devin Desktop upstream updates: global non-root rules (~/.devin/rules/) dropped, AGENTS.local.md folded into committed AGENTS.md
## Summary

Devin's rules documentation covers two surfaces rulesync does not emit: per-rule **global** rule files under `~/.devin/rules/`, and the personal `AGENTS.local.md`. The second is worse than a missing feature — instructions the user explicitly marked local are currently folded into the committed `AGENTS.md`, which inverts the upstream contract.

## Recent Releases

- **CLI v3000.4.16** — 2026-08-10 — [stable changelog](https://docs.devin.ai/cli/changelog/stable): "Nested `AGENTS.md` files, lowercase `agents.md` files, and rules in supported dot-directories are now discovered and remain scoped to the directory where they apply." This is the nearest anchor for the rewritten rules doc section; the doc line itself carries no date.
- CLI **v3000.4.25** (2026-08-13: `/share` URL now server-provided, MCP array-form `resource` OAuth discovery fix) and **v3000.3.27** (2026-08-01: symlink-write refusal in `edit`/`write`/`apply_patch`/`notebook_edit`). Desktop **v3.7.25** (2026-08-13), **v3.7.16** (2026-08-10, "Plugins" renamed to "Extensions"), **v3.6.27** (2026-08-01). Everything material in these is runtime/UX or security hardening.

### 1. rules — global **non-root** rules are dropped (`~/.devin/rules/*.md`)

- **Upstream**: [the rules docs](https://docs.devin.ai/cli/extensibility/rules), section "Rules in the .devin Directory", verbatim: *"Like other project rules, these directories are read at the workspace root and in each directory between the workspace root and your current directory. You can also place them in your home directory (`~/.devin/rules/*.md`, `~/.devin/global_rules.md`) to apply them to every project."* These carry the full `trigger` frontmatter set (`always_on`, `manual`, `model_decision`, `agent`, `glob`).
- **rulesync**: `project` + `global`, but the global slice is **root-only**. `DevinRule.getSettablePathsGlobal()` (`src/features/rules/devin-rule.ts`) returns a single `{relativeDirPath, relativeFilePath}` pair with no `nonRoot` branch. The global dry run says so verbatim: `4 non-root rulesync rules found, but it's in global mode, so ignoring them` — only `~/.config/devin/AGENTS.md` is written. Per-rule global instructions with activation triggers are therefore unexpressible; the only global option is one always-on blob.
- **Scope of the work is adapter-local**: the processor already supports global non-root. In the same run, `--targets augmentcode --global --dry-run` emitted all five rules including the four non-root ones into `.augment/rules/`.
- **Naming trap for the implementer**: the documented global directory is `~/.devin/rules/` (home-dir `.devin/`), **not** `~/.config/devin/rules/`. That is inconsistent with every other Devin global path rulesync emits (`~/.config/devin/AGENTS.md`, `/config.json`, `/mcp_config.json`, `/agents/`, `/skills/`), so it is worth verifying empirically before shipping.

### 2. rules — `AGENTS.local.md` not emitted, so personal rules land in the committed `AGENTS.md`

- **Upstream**: same page, "Personal Rules with AGENTS.local.md", verbatim: *"If you have personal instructions that shouldn't be shared with collaborators — such as preferred working style, testing habits, or review preferences — create an `AGENTS.local.md` file next to your `AGENTS.md`"*, *"This file is loaded alongside `AGENTS.md` with the same always-on behavior. Add it to your `.gitignore` so it stays local"*, and *"This follows the same convention as `.devin/config.local.json` — the `.local.` suffix signals a personal override that shouldn't be committed."* It is also listed in the page's Supported File Names table as a first-class rule file.
- **rulesync**: `project`, but lossy in a way that leaks. Devin's convention entry in `src/features/rules/rules-processor.ts` declares no `localRootMode`, so it falls back to the `append-to-root` default and a `localRoot: true` rulesync rule is concatenated into the shared, checked-in `AGENTS.md`. `devin-rule.ts` has no `localRoot` handling at all. The project dry run shows it: `.rulesync/rules/overview.local.md` exists in this repo, yet only 5 rule files are written (`AGENTS.md` plus 4 under `.devin/rules/`) and no `AGENTS.local.md` appears.
- **Impact**: instructions the user explicitly marked local end up in the file that gets committed — the opposite of what upstream documents.

## Proposed Follow-up

1. Add a `nonRoot: { relativeDirPath: ".devin/rules" }` branch to `DevinRule.getSettablePathsGlobal`, keeping `~/.config/devin/AGENTS.md` as the global root and emitting the same `trigger`/`globs` frontmatter used at project scope. Regenerate the support matrix afterwards. Verify the `~/.devin/` versus `~/.config/devin/` split against a real install first.
2. Add `localRootMode: "separate-local-file", localRootFileName: "AGENTS.local.md"` to the `devin` convention entry, mirroring the existing `roo` / `zoocode` / `rovodev` entries, and handle the file on the import side of `devin-rule.ts` so it round-trips back to `localRoot: true`.

## Checked and not a gap

- **mcp** — #2510's fix is live: both dry runs write `.devin/mcp_config.json` and `~/.config/devin/mcp_config.json`, not the deprecated `config.json` `mcpServers` key.
- **rules, project root** — #2406's fix is live: the project dry run writes root `AGENTS.md` alongside `.devin/rules/*.md`.
- **ignore** — #2597's fix is live: `--global` writes `~/.codeium/.codeiumignore`. The rules page carries an explicit warning that Devin CLI does *not* support `.codeiumignore` at project scope, consistent with rulesync reading it only as an import fallback.
- **Nested/hierarchical rule discovery** (v3000.4.16) — rulesync's nested mechanism (`getNestedFilePatterns`) is import-only by design, and nested `AGENTS.md` files are already enumerated by the `agentsmd` target. A parity nit under `--targets devin` import at most.
- **`read_config_from`** (`agents_standard` / `cursor` / `windsurf` / `claude` import toggles) and **`.devin/global_rules.md`** — consumption-side switches and an alternative always-on file for which root `AGENTS.md` is the correct mapping.
- **`subagents_enabled`** — already dismissed in #2597 as a runtime toggle.
- **Plugins / Extensions** (personal cloud-synced installs by default, v3000.4.16) — cross-tool plugin packaging, already classed in #2510 as a candidate new feature rather than a drift gap.

## Housekeeping

`.rulesync/skills/rulesync-feature-research/references/devin.md` records `commands` and `skills` as having "No dedicated upstream … surface in map", and `ignore` as project-only. All three are stale and will suppress future checks.

## References

- https://docs.devin.ai/cli/extensibility/rules — both gaps: the `~/.devin/rules/` home-directory passage and the `AGENTS.local.md` section, quoted above.
- https://docs.devin.ai/cli/changelog/stable — v3000.4.16's rule-discovery entry, the nearest datable anchor for the rewritten rules doc.

dyoshikawa:
## Background

#2688 recorded two Devin rules gaps. **Gap 2 (`AGENTS.local.md`) is already fixed on `main`** — the `devin` convention entry in `rules-processor.ts` carries `localRootMode: "separate-local-file"` / `localRootFileName: "AGENTS.local.md"`, so personal rules no longer land in the committed `AGENTS.md`. This PR closes the remaining gap 1, plus the issue's Housekeeping item.

## Gap 1: global non-root rules were dropped

Devin's [rules page](https://docs.devin.ai/cli/extensibility/rules) documents per-rule global rules in the home directory:

> Like other project rules, these directories are read at the workspace root and in each directory between the workspace root and your current directory. You can also place them in your home directory (`~/.devin/rules/*.md`, `~/.devin/global_rules.md`) to apply them to every project.

`DevinRule.getSettablePathsGlobal` returned a root path only, so a `--global` generate wrote just `~/.config/devin/AGENTS.md` and reported the non-root rules as ignored. Per-rule global instructions with activation triggers were unexpressible; the only global option was one always-on blob.

Global scope now mirrors project scope:

- root rule → plain `~/.config/devin/AGENTS.md` (unchanged)
- non-root rules → one file per rule under `~/.devin/rules/` with the same `trigger` / `globs` frontmatter as project scope
- both directions round-trip: a global import reads `~/.devin/rules/*.md` back into `.rulesync/rules/` with its `devin:` block intact

The two `fromRulesyncRule` root branches collapsed into one while doing this — they differed only in the target path.

### On the `~/.devin/` vs `~/.config/devin/` split

The issue flagged this as a naming trap worth verifying against a real install. **I verified it against the docs, not an install** — the passage quoted above is explicit, and it is the only documented location for per-rule global rules. Every other Devin global surface rulesync emits (`AGENTS.md`, `config.json`, `mcp_config.json`, `agents/`, `skills/`) does live under `~/.config/devin/`, so the split is genuinely inconsistent upstream rather than a typo on our side. It is called out in the class docstring and in `getSettablePaths` so the next reader does not "fix" it. If someone with an install finds Devin actually reads `~/.config/devin/rules/`, that is a one-line change to the `nonRoot` path.

### Verified end to end

A real `--global` generate against a pseudo-home writes both files, and importing them back reproduces the rulesync rules:

```
Written 2 rules
    .config/devin/AGENTS.md
    .devin/rules/typescript.md
```

## Housekeeping

The issue noted `.rulesync/skills/rulesync-feature-research/references/devin.md` records stale surfaces that would suppress future checks. Refreshed against the current docs:

- `commands` and `skills` no longer say "No dedicated upstream … surface in map" — skills have [their own page](https://docs.devin.ai/cli/extensibility/skills/overview), and there is no separate commands surface because a skill *is* the slash command (`/skill-name`, or `/<plugin>:<skill>` for plugin skills)
- the skills global path is `~/.config/devin/skills/` (what rulesync actually emits), not `~/.codeium/windsurf/skills/`
- MCP moved to `/cli/extensibility/mcp/overview`; `/cli/overview` now 404s (`/cli` is the quickstart)
- the `rules` rows now name the new global non-root directory and the `AGENTS.local.md` root

## Verification

`pnpm cicheck` — all green (373 test files, 8742 tests). `pnpm test:e2e src/e2e/e2e-rules.spec.ts` — 128 passed. Two existing tests that pinned the old root-only global behavior were updated, and three new ones cover the global non-root path, its frontmatter, and the unchanged plain-markdown global root.

Closes #2688


dyoshikawa:
@dyoshikawa Thank you!

PR #2714, opened 2026-08-20
feat: support multiple input roots

### LINKED ISSUE #2652: Support local `.rulesync.local/` tree
Hey.

Some context: 
- we're a company with many developers
- we're using Rulesync with Cursor, Claude and Codex targets for syncing of MCPs, rules, skills and hooks
- all of `.rulesync`, `rulesync.json` and it's generated `.codex`, `.cursor`, `.claude`, `AGENTS.md`, `.mcp.json` are Git tracked/committed 
- automations and most devs still use Cursor

Normally, Cursor allows some customization of project rules on user level:
- enable, disable MCPs; add your own
- add your own rules, skills, hooks and so on

Codex doesn't. Claude only partially. Now, at the moment, because everything is Git tracked - nothing can be reasonably customized by individual developers. For example, the repo has ~25 default MCPs for different purposes and usecases, I would probably want to keep ~10 of those. But unfortunately, with Codex, I cannot.

So in our situation, it's impossible for devs to:
- add their own rules, skills, MCPs
- disable/remove repo-shipped rules, skills, MCPs
- replace rules, skills, MCPs or customize them

---

To allow customization, one potential solution I thought of was adding all generated folders and files to `.gitignore`. What this would immediately give is a way for people to customize it for their own agent manually. 

The problem is once they run `rulesync`, all of their changes would be overwritten with committed state from `.rulesync`.

So I was thinking to allow two roots: `.rulesync/` and `.rulesync.local/`. The first one is the "default" AI set, tracked and committed, repo/team wide. The second one would be added to `.gitignore`, and allow adding or overwriting individual rules, skills or MCP configs. When `rulesync` runs, it scans both trees and merges them, preferring content from `.rulesync.local/`:

```
project/
├── rulesync.jsonc                 # tracked team configuration
├── rulesync.local.jsonc           # ignored developer configuration
│
├── .rulesync/                     # tracked team sources
│   ├── rules/
│   │   ├── coding-conventions.md
│   │   ├── database-guidelines.md
│   │   └── pull-request-process.md
│   ├── skills/
│   │   ├── dependency-audit/
│   │   │   └── SKILL.md
│   │   └── release-summary/
│   │       └── SKILL.md
│   └── .mcp.jsonc
│
└── .rulesync.local/               # ignored developer additions/overrides
    ├── rules/
    │   ├── coding-conventions.md   # replaces team rule with the same path
    │   └── personal-workflow.md    # adds a new rule
    ├── skills/
    │   └── debugging-helper/
    │       └── SKILL.md            # adds a new skill
    └── .mcp.jsonc                  # adds or replaces servers by name
```

```jsonc
// rulesync.jsonc
{
  "inputRoots": [".rulesync"],
  "targets": ["cursor", "claudecode", "codexcli"],
  "features": ["rules", "skills", "mcp"]
}
```

```jsonc
// rulesync.local.jsonc
{
  // Overrides the tracked array and adds the developer-local overlay.
  // Later roots take precedence.
  "inputRoots": [".rulesync", ".rulesync.local"],

  // Exclude tracked items without editing or deleting their definitions.
  "exclude": {
    "rules": ["database-guidelines"],
    "skills": ["release-summary"],
    "mcpServers": ["browser-automation"]
  }
}
```

---

The above then generate (just Cursor as an example):

```
.cursor/
├── rules/
│   ├── coding-conventions.mdc    # generated from local replacement
│   ├── personal-workflow.mdc     # generated from local addition
│   └── pull-request-process.mdc  # generated from tracked source
│
├── skills/
│   ├── debugging-helper/
│   │   └── SKILL.md              # generated from local addition
│   └── dependency-audit/
│       └── SKILL.md              # generated from tracked source
│
└── mcp.json
```

What this would let us do is still ship the whole AI set as "defaults" with the repo, which most devs and all of our automations want; but those developers that want to customize their experience can still do it, without sacrificing the repo-defaults.

I know this is a larger request, but it's something that every single user of Codex in our team struggles with. I'd imagine it's the same with Claude, and it's super annoying. What do you think?

autaut03:
Closes #2652 

Please do review. The idea is this:
- `inputRoots` is the new field. It specifies specific directories under `cwd()` that should act as a source for Rulesync: `[".rulesync", ".rulesync.local", ".ai"]`
- `inputRoot` is deprecated. It's different from `inputRoots` - it specifies the parent directory, while `.rulesync` directory was implied. So `inputRoot: "."` and `inputRoots: [".rulesync"]` are identical, but have almost the same name. This will cause confusion, hence why I thought it'd make sense to deprecate `inputRoot`. Still works as an alias
- no `exclude` support
- MCP servers can be "unset" via `"name": null` syntax - this was easy to add, but I'm more than willing to remove it from this slice


src/cli/commands/generate.ts
```diff
@@ -130,9 +135,24 @@ async function generateOnce(
...
+    await assertInputRootsResolvable(inputRoots);
+  } catch (error) {
     throw new CLIError(
-      ".rulesync directory not found. Run 'rulesync init' first.",
+      error instanceof Error ? error.message : String(error),
+      ErrorCodes.RULESYNC_DIR_NOT_FOUND,
+    );
+  }
+
+  if (!(await checkRulesyncDirExists({ inputRoots }))) {
```
dyoshikawa: This branch is unreachable. Line 141 already ran `assertInputRootsResolvable`, which throws unless *every* root exists; `checkRulesyncDirExists` returns true if *any* root exists. So the nicer message here — the one that mentions `rulesync init` — can never be printed, and users get the bare `Check your inputRoots setting.` from `assertInputRootsResolvable` instead.

Worth picking one check and keeping the `rulesync init` hint on it. `src/index.ts` around line 89 has the same pair.

src/lib/generate.ts
```diff
@@ -211,14 +224,107 @@ function warnUnsupportedTargets(params: {
...
+    }
+  }
+
+  if (missingRoots.length === 0) {
+    return;
+  }
+
+  if (missingRoots.length === 1) {
+    throw new Error(
+      `Your configured input root '${missingRoots[0]}' does not exist. Check your inputRoots setting.`,
```
dyoshikawa: This is now the only reachable not-found message, and it's aimed at someone who has configured `inputRoots`. The common case is a first-time user in a directory with no `.rulesync/` who has never set the field — telling them to "check your inputRoots setting" is confusing, and the `rulesync init` pointer that used to be here is now stranded in the unreachable branch in `generate.ts`.

This string is also reachable through the public `generate()` export, so changing it is technically a behavioural change worth noting.

src/lib/generate.ts
```diff
@@ -211,14 +224,107 @@ function warnUnsupportedTargets(params: {
...
+ * root (e.g. an overlay tree with only `mcp.jsonc`).
+ */
+const RULESYNC_SOURCE_ENTRIES = [
+  RULES_FEATURE_SUBDIR,
+  COMMANDS_FEATURE_SUBDIR,
+  SUBAGENTS_FEATURE_SUBDIR,
+  CHECKS_FEATURE_SUBDIR,
+  SKILLS_FEATURE_SUBDIR,
+  RULESYNC_MCP_FILE_NAME,
+  RULESYNC_MCP_LEGACY_FILE_NAME,
```
dyoshikawa: `mcp.jsonc` and `mcp.json` are here but the deprecated `.mcp.json` isn't, even though `RulesyncMcp` still accepts it as a candidate (see `rulesync-mcp.ts` around line 432). An overlay tree containing only `.mcp.json` will be classified as having no rulesync source content.

More generally, this list and the candidate lists hardcoded in the hooks / permissions / ignore processors are duplicating what `getRulesyncSourceCandidates({ paths: X.getSettablePaths() })` already knows. Deriving them would stop exactly this kind of drift — the `.mcp.json` gap is the first instance of it.

src/config/config-resolver.ts
```diff
@@ -244,24 +334,52 @@ export class ConfigResolver {
...
+      for (const entry of inputRoots) {
+        validateOutputRoot(entry);
+      }
+    }
+
+    const cliConfigAnchor =
+      inputRoot !== undefined
+        ? inputRoot
+        : inputRoots !== undefined
+          ? dirname(inputRoots[0]!)
```
dyoshikawa: This is the anchor problem. With `--input-roots`, the config file is resolved against the *parent* of the first root — a directory the user never typed. For the layout the guide recommends (`--input-roots ~/.aiglobal/.rulesync ./.rulesync`), `rulesync.jsonc` gets read from `~/.aiglobal/` and the project's own `rulesync.jsonc` is ignored, which is almost certainly not what anyone expects from that command.

There's a smaller security angle too: if that parent is writable by someone else (a `/tmp` path, a shared tree), their `rulesync.jsonc` gets to set `outputRoots` and `targets` for this run.

Anchoring CLI-supplied `--input-roots` at cwd, or requiring an explicit `--config-path` alongside it, would avoid both. Good news: config-file-supplied `inputRoots` correctly do *not* affect the anchor, so there's no chained loading — that part is right.

src/config/config-resolver.test.ts
```diff
@@ -884,7 +886,100 @@ describe("config-resolver", () => {
...
+      );
+
+      const config = await ConfigResolver.resolve({
+        configPath: join(testDir, "rulesync.jsonc"),
+      });
+
+      expect(config.getInputRoots()).toEqual([localA, localB]);
+    });
+
+    it("should let CLI inputRoots override any file-configured input roots", async () => {
```
dyoshikawa, line 940: This test doesn't verify what it says. The anchor for CLI `inputRoots` is `dirname(inputRoots[0])`, which is `testDir` — but the config file is written to `cliA/rulesync.jsonc`, so it's never loaded. The assertion would pass even if file config took precedence over CLI. The comment on lines 941-944 has the anchor backwards.

Writing the config to `testDir/rulesync.jsonc` would make it a real precedence test.

src/cli/commands/doctor.ts
```diff
@@ -656,18 +671,96 @@ async function checkInputRootExists({
...
+  baseConfig,
+  localConfig,
+}: {
+  baseConfig: Record<string, unknown> | undefined;
+  localConfig: Record<string, unknown> | undefined;
+}): { effective: string[]; sourceFile: "base" | "local" } | undefined {
+  const localRoots = readInputRootsField(localConfig);
+  const localRoot = readInputRootField(localConfig);
+
+  if (localRoots !== undefined) return { effective: localRoots, sourceFile: "local" };
```
dyoshikawa: The docblock says this follows the resolver's precedence, but it doesn't. `ConfigResolver` runs `mergeConfigs` field by field and *then* resolves, so a base file with `inputRoots` plus a local file with `inputRoot` ends up using the base `inputRoots`. Here, the local `inputRoot` short-circuits at line 737 and wins. `doctor` will report roots that `generate` won't actually use.

Two smaller things while you're in here: the message elsewhere picks the field name with `effective.length > 1 ? "inputRoots" : "inputRoot"`, so a one-element `inputRoots` config gets reported under a field name that isn't in the file; and `doctor` doesn't flag `inputRoot` + `inputRoots` set together, even though the resolver throws on it. That's the single most likely misconfiguration and `doctor` is exactly where it should surface.

src/features/mcp/rulesync-mcp.ts
```diff
@@ -165,46 +317,148 @@ export class RulesyncMcp extends RulesyncFile {
...
+   * The resulting `RulesyncMcp` is anchored to the FIRST root's recommended
+   * path — error messages therefore point at the primary root even when an
+   * offending server originated in an overlay. This is an accepted
+   * limitation of the first slice; see the plan's out-of-scope section.
+   *
+   * When no root supplies any candidate, this falls back to reading the
+   * primary root's recommended path so the underlying file-not-found error
+   * matches the single-root behavior of `fromFile`.
+   */
+  static async fromRoots({
```
dyoshikawa, line 340: `fromRoots` always re-serializes with `JSON.stringify(merged, null, 2)` and anchors to the first root's recommended path, and it runs even when there's only one root. So single-root users get three behaviour changes they didn't ask for: JSONC comments are stripped, `relativeFilePath` reports `mcp.jsonc` even when the file on disk is `mcp.json` or `.mcp.json`, and errors originating in an overlay point at the primary root.

Short-circuiting to the `fromFile` path when `roots.length === 1` would keep the existing behaviour intact.

src/features/mcp/rulesync-mcp.ts
```diff
@@ -116,10 +116,162 @@ export const RulesyncMcpFileSchema = z.looseObject({
...
+      continue;
+    }
+
+    merged[serverName] = serverConfig;
+  }
+
+  return merged;
+}
+
+export function mergeMcpJsonOverlays(
```
dyoshikawa: Treating a top-level `mcpServers.<name>: null` as a deletion is a new semantic — the schema types top-level `mcpServers` as `z.record(z.string(), RulesyncMcpServerSchema)` (non-nullable), so this used to be a validation error. Meanwhile tool-scoped `null` keeps its existing meaning of "disabled for that tool". Same literal, two different meanings depending on nesting depth.

Neither the deletion feature nor the asymmetry is mentioned in `docs/guide/separate-input-root.md`. Worth documenting both.

Minor: the `isPrototypePollutionKey` filter is applied to top-level keys and server names but the tool-scoped block goes through a plain `{...base, ...overlay}`. Harmless with spread, but the inconsistency inside one function is confusing to read.

src/e2e/e2e-helper.ts
```diff
@@ -62,10 +62,15 @@ export async function runGenerate({
...
-  inputRoot?: string;
+  inputRoots?: string[];
   outputRoots?: string;
   env?: Record<string, string>;
 }): Promise<{ stdout: string; stderr: string }> {
+  // The CLI exposes `--input-root` as a user-facing alias for a
+  // one-element `--input-roots` list, but this helper (test infrastructure)
+  // only speaks the plural form to keep test code from ever branching on
+  // the singular alias. A single-element list emits `--input-roots X`,
+  // which the CLI treats identically to `--input-root X`.
```
dyoshikawa: This claim isn't right — `--input-root X` and `--input-roots X` are not equivalent. Per `program.ts:311`, the singular form expands to `X/.rulesync`, while the plural form treats `X` itself as the source tree. The call sites you updated do the right thing (`join(homeDir, RULESYNC_RELATIVE_DIR_PATH)`), so behaviour is fine, but the comment will mislead the next person.

Separately: dropping the singular form from the helper means `--input-root` now has zero e2e coverage. A deprecated-but-supported flag is exactly the thing that should keep an e2e test, since nothing else will catch it regressing.

docs/guide/separate-input-root.md
```diff
@@ -31,28 +35,76 @@ Without `--input-root`, you would have to `cd ~/.aiglobal && rulesync generate`
...
+
+At least one of the configured source trees must contain recognizable rulesync source content (`rules/`, `skills/`, `mcp.jsonc`, etc.). An overlay-only tree may legitimately supply just one feature — for example, a `.rulesync.local/` that only contains `mcp.jsonc`.
+
+## Setting input roots in `rulesync.jsonc`
+
+You can set the same value in `rulesync.jsonc` (or `rulesync.local.jsonc`) instead of passing it on the command line:
+
+```jsonc
+{
+  "inputRoots": ["~/.aiglobal/.rulesync", "./.rulesync"],
```
dyoshikawa: `~` isn't expanded here. `path.resolve()` doesn't do tilde expansion, so as a config value this resolves to a literal `<cwd>/~/.aiglobal/.rulesync`. It works on the command line only because the shell expands it first.

Use an absolute path in the JSONC example, or note the difference explicitly — as written this is a config people will copy and then wonder why their rules aren't picked up.

docs/guide/separate-input-root.md
```diff
@@ -31,28 +35,76 @@ Without `--input-root`, you would have to `cd ~/.aiglobal && rulesync generate`
...
+
+### Merge rules per feature
+
+The general rule is: later entries win. Each feature refines that rule slightly:
+
+- **Rules, commands, subagents, checks, skills** — merged file-by-file (case-insensitive). Files present only in an earlier tree are kept; a file that also exists in a later tree replaces the earlier version. A skill directory is replaced as a single unit (all of its companion files together).
+- **MCP** — merged one level into the JSON: the top-level `mcpServers` map and each `<toolname>.mcpServers` map are merged by server name (later wins per key). An individual server config is replaced as a whole; patching just its `args` or `env` is not supported.
+- **Hooks, permissions, ignore** — the last tree that provides the file wins the whole file. There is no line-level merge.
+
+At least one of the configured source trees must contain recognizable rulesync source content (`rules/`, `skills/`, `mcp.jsonc`, etc.). An overlay-only tree may legitimately supply just one feature — for example, a `.rulesync.local/` that only contains `mcp.jsonc`.
```
dyoshikawa: This requirement is only enforced on the MCP path — `hasRulesyncSourceContent` is called from `src/mcp/generate.ts` and nowhere else. The CLI and the programmatic API only check that the directories exist. So an empty `.rulesync/` fails through the MCP server but succeeds via `rulesync generate`.

Either wire the check into the shared path or soften the wording here.

dyoshikawa (commented):
Thanks for taking this on — it's a big change and the shape of it is right. Normalizing the singular `inputRoot` into a one-element `inputRoots` and then making every downstream path multi-root aware is the correct decomposition, and pulling `mergeByIdentity` / `pickLastRootWithFile` up into the base processor keeps the per-feature code honest. Backward compatibility for `--input-root` and `generate({ inputRoot })` holds up: I checked that it still expands to `[join(inputRoot, ".rulesync")]`, so single-root users land on the same paths as before. I also verified `src/generated/docs-content.ts` matches the four changed docs byte for byte, and `pnpm-lock.yaml` is clean — all 969 resolutions are identical to `main`, no new packages, no integrity changes. No malicious or obfuscated code anywhere in the diff.

Two things before this can land, plus one process point.

The first is that `checkRulesyncDirExists` in `src/cli/commands/generate.ts` is now dead code, and the error message that survives is the wrong one. `assertInputRootsResolvable` throws if *any* root is missing, so by the time control reaches the `checkRulesyncDirExists` branch — which only requires *one* root to exist — it's guaranteed true. The practical effect is that a first-time user who has never configured anything now gets "Check your inputRoots setting" instead of "Run 'rulesync init' first", and the friendlier message is stranded in the unreachable branch. `src/index.ts` has the same pair. Worth collapsing to one check and keeping the `rulesync init` hint on it. That message is also part of the public API surface, so it's worth a compatibility note either way.

The second is the config-file anchor. With `--input-roots`, `rulesync.jsonc` is now looked up relative to `dirname(inputRoots[0])`, which is a directory the user never named. For the layout your own docs recommend — `--input-roots ~/.aiglobal/.rulesync ./.rulesync` — that means the config is read from `~/.aiglobal/` and the project's own `rulesync.jsonc` (with its `targets` and `features`) is silently ignored. That's going to bite the exact "global base plus per-repo overlay" use case this feature exists for. There's a security edge too: if the parent happens to be world-writable, someone else's `rulesync.jsonc` there can redirect `outputRoots`. I'd anchor CLI-supplied `--input-roots` at cwd, or require an explicit `--config-path`. At minimum it needs to be spelled out in the guide.

Related: the test that's meant to cover this doesn't. `config-resolver.test.ts` writes the config into `cli-a/rulesync.jsonc`, but the anchor is `dirname(cliA)` = `testDir`, so the file is never read and the assertion passes regardless of precedence. The comment above it explains the opposite of what the code does.

On process — this is 2963 added lines against a 1000-line limit for external contributors, and the PR Policy bot flagged it on open. That's a maintainer call rather than something I can wave through, so I'd rather flag it than have it merged quietly. If it stays as one PR that's fine, but it should be a deliberate decision.

Everything else in the line comments is follow-up material. The MCP re-serialization (comment 10 territory), the case-insensitive cross-root identity that drops a file with no warning, the hardcoded candidate filename lists in the hooks/permissions/ignore processors, and the `.curated` precedence inversion across roots are all real but none of them block. One small note that isn't worth its own comment: `ImportOptions` in `src/cli/commands/import.ts` still says `Omit<ConfigResolverResolveParams, "delete" | "inputRoot">` — `inputRoots` slipped through, and the comment right above claims both are excluded on purpose.

src/lib/generate.ts
autaut03: I removed the source-content gate so existing empty roots stay valid for delete and check workflows. The hooks, permissions, and ignore candidate lists now derive from getSettablePaths(), so their candidates cannot drift.

src/features/mcp/rulesync-mcp.ts
autaut03: I removed top-level null deletion instead of documenting a second meaning. Top-level null is invalid again; tool-scoped null keeps its existing meaning of disabling that server for that tool.

docs/guide/separate-input-root.md
autaut03: I changed the JSONC example to relative paths instead. That is the common config case and avoids implying that Rulesync expands ~.

docs/guide/separate-input-root.md
autaut03: I kept existing empty roots valid across every interface and removed the MCP-only source-content check. This preserves delete and check workflows instead of introducing a shared prohibition.

autaut03:
Ok that's done @dyoshikawa ❤️ 

I have one more question: at this moment, `.rulesync.local` isn't mentioned anywhere in this PR. Should/can I make a follow-up, making the `.rulesync.local` the official "default" input root, loaded automatically, and added by `rulesync gitignore`? So `inputRoots: ['.rulesync', '.rulesync.local']` would be the default configuration, unless the key is overwritten.

dyoshikawa:
@autaut03 Thank you!

PR #2719, opened 2026-08-23
chore(security): raise js-yaml override, drop unused id-token permission, record action pin stance

### LINKED ISSUE #2718: Follow-ups from #2717 / #2716: OpenCode action pin hygiene and pre-existing dependency advisories
## Background

While triaging the review queue, PR #2717 (`chore(deps): bump the all-dependencies group with 77 updates`) and PR #2716 (`ci(deps): bump anomalyco/opencode/github from 1.18.18 to 1.18.19`) were both reviewed and found clean enough to merge — no `high`/`critical` findings, CI fully green on both.

However, each review surfaced non-blocking observations that are worth tracking rather than losing. None of them were introduced by those PRs; they are pre-existing conditions that the reviews happened to expose. This issue collects them so they can be picked up independently of the merged PRs.

### 1. Supply-chain: SHA pinning of `anomalyco/opencode/github` is largely decorative

Found while reviewing #2716.

The repository consistently pins third-party actions to a 40-char SHA with a `# vX.Y.Z` comment, and #2716 correctly updated that pin. But the action being pinned is a composite action whose own `action.yml` does roughly this:

```yaml
- name: Get opencode version
  run: |
    VERSION=$(curl -sf https://api.github.com/repos/anomalyco/opencode/releases/latest | ...)
- name: Install opencode
  if: steps.cache.outputs.cache-hit != 'true'
  run: curl -fsSL https://opencode.ai/install | bash
```

Consequences:

- The SHA pin freezes only the thin wrapper. The `opencode` CLI binary that actually handles `OPENROUTER_API_KEY`, `ZHIPU_API_KEY`, and `GITHUB_TOKEN` is resolved to **`latest`** at run time and installed via `curl | bash`. Bumping the pin therefore provides far less supply-chain assurance than the pinning convention implies.
- The composite action also references `actions/cache@v4` by mutable tag internally, which sits outside this repository's pinning policy.

Mitigating factors today: `.github/workflows/draft-release.yml` is `workflow_dispatch`-only and guarded by `github.actor == 'dyoshikawa' || github.actor == 'cm-dyoshikawa'`, so it cannot be triggered externally.

Verified during the review: the new pin `2b72179c663cadcb54f54d9f19221b3fb3d11fb6` genuinely corresponds to upstream tag `v1.18.19`, and the `github/` directory (`action.yml`, `index.ts`, `package.json`, `bun.lock`) is byte-identical between v1.18.18 and v1.18.19 — the bump changes no behaviour at all.

### 2. `id-token: write` is unnecessary in `draft-release.yml`

Found while reviewing #2716.

The workflow passes `use_github_token: "true"` to the OpenCode action, which makes the action skip the OIDC token exchange entirely. The `id-token: write` permission is therefore dead weight and can be dropped for least privilege.

### 3. Pre-existing dependency advisories (not introduced by #2717)

Found while reviewing #2717. A bulk advisory query over all 882 resolved packages returned the following. Every one of these resolves to the **same version on `main` as on the PR branch**, so #2717 neither introduced nor worsened any of them (it actually improved `hono` from 4.13.1 to 4.13.3).

| Package | Resolved version | Severity | Route |
| --- | --- | --- | --- |
| `fast-uri` | 3.1.4 | high (GHSA-7p8r-x3mc-p8w7) | dev transitive |
| `js-yaml` | 4.2.0 / 4.3.0 | high (GHSA-5p4m-2wfm-xmqj and others) | intentional pin via `overrides: gray-matter>js-yaml: 4.2.0` |
| `hono` | 4.12.32 | moderate (+4 others, `<4.12.34`) | runtime, via `@modelcontextprotocol/sdk@1.30.0` |
| `vite` | 5.4.21 | high (GHSA-fx2h-pf6j-xcff) | dev transitive, via `vitepress` |
| `esbuild` | 0.21.5 | moderate (dev-server only) | dev transitive |

Real-world exploitability is limited: rulesync is a locally executed CLI, and the `hono` / `vite` / `esbuild` advisories all assume a network-exposed server. The two worth a deliberate decision are:

- **`js-yaml` override pin at 4.2.0** — this is a hand-maintained `overrides` entry, so it will not move on its own. It should be revisited to see whether `gray-matter` can now take a patched line, or whether the pin needs bumping.
- **`hono` via `@modelcontextprotocol/sdk`** — this is a *runtime* dependency, unlike the others. It clears once `@modelcontextprotocol/sdk` is bumped past a release that depends on `hono >= 4.12.34`.

## Solution / Next Steps

1. **`js-yaml` override** — check whether `overrides: { "gray-matter>js-yaml": "4.2.0" }` in `package.json` can be raised to a patched release, or removed entirely if `gray-matter` has been updated upstream. This is the highest-value item here because it is under direct control.
2. **`hono` via MCP SDK** — watch for a `@modelcontextprotocol/sdk` release that pulls `hono >= 4.12.34` and let Dependabot carry it in. No local action needed beyond not blocking that bump.
3. **`id-token: write`** — remove the permission from `.github/workflows/draft-release.yml` (and check other OpenCode-driven workflows for the same pattern) while `use_github_token: "true"` remains in use.
4. **OpenCode action pinning** — decide explicitly which stance to take, and record it so future reviews do not re-litigate it:
   - accept `https://opencode.ai/install` as a trusted install path and note in `.github/workflows/AGENTS.md` that the SHA pin covers the wrapper only; or
   - ask upstream for a `version` input on `anomalyco/opencode/github` so the CLI itself can be pinned; or
   - vendor a pinned installer step in this repository instead of relying on the composite action's installer.
5. **`fast-uri` / `vite` / `esbuild`** — no action for now; they are dev-only transitives with no exposed surface for a local CLI. Re-evaluate if any of them ever reaches the published runtime dependency tree.

---

Context: raised during review of #2717 and #2716. Both PRs were merged; nothing here blocked them.

dyoshikawa:
## Summary

Resolves the actionable follow-ups collected in #2718.

### 1. `gray-matter>js-yaml` override raised 4.2.0 -> 4.3.1

`pnpm-workspace.yaml` pinned `gray-matter>js-yaml` at `4.2.0`, which is affected by:

- [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) (high, patched in `4.3.0`)
- [GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj) (high, patched in `4.3.1`)

`4.3.1` is the head of the `v4-legacy` dist-tag, so this stays on the same major line that the `gray-matter@4.0.3` patch (`(yaml.load || yaml.safeLoad)`) targets and clears both advisories. Removing the override entirely is not an option: `gray-matter@4.0.3` declares `js-yaml ^3.13.1`, and the 3.x line is strictly worse.

### 2. Dropped `id-token: write` from `draft-release.yml`

The workflow passes `use_github_token: "true"` to `anomalyco/opencode/github`. That input is documented upstream as "Use GITHUB_TOKEN directly instead of OpenCode App token exchange. When true, skips OIDC and uses the GITHUB_TOKEN env var", so the `id-token: write` permission was dead weight.

### 3. Recorded the action-pinning stance

The `github-actions-security` rule now documents that the SHA pin on `anomalyco/opencode/github` covers the composite wrapper only — the wrapper resolves the `opencode` CLI to `latest` at run time and installs it via `curl -fsSL https://opencode.ai/install | bash`, and uses `actions/cache@v4` by mutable tag internally. The accepted stance is to treat the install URL as trusted because `draft-release.yml` is `workflow_dispatch`-only and gated on `github.actor`, with explicit revisit triggers, so future reviews do not re-litigate it.

### 4. Prerequisite: untracked rulesync-generated tool configs

Seven files were tracked in git while the project's own generated `.gitignore` ignores them (`.claude/rules/coding-guidelines.md`, `.github/copilot-instructions.md`, `.github/instructions/*.instructions.md`, `.opencode/memories/coding-guidelines.md`). Because `**/.github/instructions/` is an ignored *directory* pattern, `git add -- <path>` refuses those paths, which made the `lint-staged` pre-commit hook fail on **any** change under `.rulesync/rules/**`. They are regenerated by `pnpm generate`, so they are now untracked with `git rm --cached`.

### Not addressed (deliberately)

- `hono` via `@modelcontextprotocol/sdk` — clears once the SDK bumps past `hono >= 4.12.34`; nothing to do locally beyond not blocking that Dependabot bump.
- `fast-uri` / `vite` / `esbuild` — dev-only transitives with no exposed surface for a locally executed CLI.

## Test plan

- `pnpm cicheck` passes (375 test files, 8770 tests).

Closes #2718


dyoshikawa:
**Held for manual review and merge — not auto-merged.**

This PR was produced by an autonomous batch pass over the open issues. That flow deliberately does not merge changes that touch GitHub Actions workflows, build/release configuration, or dependency manifests, and this PR touches two of them:

- `.github/workflows/docs.yml` and `.github/workflows/draft-release.yml` (workflow permission scoping)
- `pnpm-workspace.yaml` / `pnpm-lock.yaml` (the `gray-matter > js-yaml` override bump)

Status at the time of writing:

- All checks are green (Code Quality & Tests, Build binaries, E2E on ubuntu/macos/windows, CodeQL, Trivy, actionlint).
- A code review and a security review were run against the branch. The security review returned five low-severity findings, of which the two worth acting on (OIDC scoping on `docs.yml`, and the residual-risk wording in the GitHub Actions security rule) were fixed in 7933264d. The code review returned no mid-or-above findings.

Please review and merge manually.

dyoshikawa:
@dyoshikawa Thank you!

PR #2720, opened 2026-08-23
feat(permissions): pass Claude Code settings keys and Codex CLI [tui] through the permission override

### LINKED ISSUE #2715: Claude Code editorMode + Codex CLI tui.vim_mode_default unauthorable in permissions override
## Summary

Two config-file-native settings — Claude Code's `editorMode` and Codex CLI's `[tui] vim_mode_default` — can't be authored through rulesync's `.rulesync/permissions.jsonc` per-tool override block, following the same pattern already fixed for Antigravity CLI in #2509 (schema validates the extra key via a loose/`additionalProperties` object, but the writer only copies a hardcoded fixed set of keys).

## Upstream

- **Claude Code**: `editorMode` (`"normal"` | `"vim"`) is a documented top-level key in `.claude/settings.json` / `~/.claude/settings.json` — https://code.claude.com/docs/en/terminal-config. Enables vim-style modal editing in the prompt input.
- **Codex CLI**: `[tui] vim_mode_default = true` in `config.toml`, added alongside the 0.129.0 modal Vim composer (`/vim`, Vim-style keymap contexts) — see the 0.129.0 release notes (https://github.com/openai/codex/releases/tag/rust-v0.129.0) and `tui.keymap.global.toggle_vim_mode` for the companion toggle binding.

### Claude Code — `editorMode` silently dropped

`permissions-schema.json`'s `claudecode` block declares `permission`, `permissions`, `sandbox` and is otherwise `additionalProperties: {}` (looks like a generic passthrough). Authoring:

```jsonc
"claudecode": {
  "editorMode": "vim",
  "sandbox": { ... }
}
```

and running `rulesync generate --features permissions --targets claudecode` validates cleanly (no error, `rulesync doctor` reports 0 problems), but the resulting `.claude/settings.json` diff contains no `editorMode` key — the `sandbox` block writes through fine, `editorMode` is dropped with no warning.

### Codex CLI — `tui` block explicitly rejected

Authoring:

```jsonc
"codexcli": {
  "tui": {
    "vim_mode_default": true
  }
}
```

and running `rulesync generate --features permissions --targets codexcli` prints:

```
Codex CLI permission override key "tui" is not managed and was skipped. "permissions"/"default_permissions" are owned by the canonical permission model and "mcp_servers" gating by the MCP feature.
```

So this one is a known/intentional allowlist rejection rather than a silent drop, but the net effect is the same: no way to author it via rulesync.

## Fix

Same shape as #2509's resolution: add `editorMode: z.optional(z.enum(["normal","vim"]))` to the `claudecode` override schema and pass it through both `fromRulesyncPermissions`/`toRulesyncPermissions`, and add a `tui` (or scoped `vim_mode_default`) key to the `codexcli` override allowlist/writer.

## Workaround in the meantime

Both `.claude/settings.json` and `.codex/config.toml` are files rulesync merges into rather than owns outright — confirmed by hand-adding unrelated top-level keys (`theme`, an arbitrary test key in `.claude/settings.json`) and re-running `rulesync generate`: they survive untouched. So `editorMode`/`[tui]` can be hand-authored directly in those checked-in files as a stopgap; this issue is about closing the declarative-source gap, not an urgent blocker.

## References

- #2509 — same bug pattern (Antigravity CLI `agentMode`), fixed after the exact upstream key name was confirmed.
- https://code.claude.com/docs/en/terminal-config
- https://github.com/openai/codex/releases/tag/rust-v0.129.0

dyoshikawa:
## Summary

`.rulesync/permissions.jsonc`'s `claudecode` override block validated any key (its schema is a `looseObject`), but `claudecode-permissions.ts` only ever copied `permissions` and `sandbox`. Every other authored key — `editorMode`, and since then `emojiCompletionEnabled`, `workflowSizeGuideline`, `keybindingFlavor`, `spellcheck`, `env`, `model`, ... — validated cleanly, produced no warning, and was then dropped. Codex CLI's `[tui]` table hit the same wall from the other side: it was rejected by the `CODEXCLI_OVERRIDE_KEYS` allowlist with a warning.

Per the maintainer's re-check comment on the issue, the fix is widened from "add `editorMode` to the schema" to a **general top-level passthrough**, because enumerating keys one at a time keeps losing the race (2.1.217 → 2.1.239 alone added five, plus aliases).

### Claude Code — generic top-level passthrough

- **Generate:** every key in the `claudecode` override other than `permission`, `permissions`, `sandbox`, `hooks` and `$schema` is deep-merged into the top level of `.claude/settings.json`. Deep-merged for the same reason `sandbox` is: setting one key under `env` must not drop the variables beside it. The exclusions keep feature ownership intact — `hooks` belongs to the hooks feature, `permission`/`permissions`/`sandbox` have their own branches, and `$schema` is an editor pointer rather than a setting.
- **Import:** every remaining top-level settings key is lifted back into the `claudecode` override, so an imported `settings.json` survives the next `generate` instead of being narrowed to the keys this feature happens to model.
- **Scope filtering** (new, mirroring the existing `CLAUDECODE_GLOBAL_ONLY_SANDBOX_PATHS` behavior). The settings reference documents a Scope per key, and rulesync now skips a key the file it is writing cannot honor, warning once per key:
  - `User or managed` / `User, local, or managed` (`spellcheck`, `autoMode`, `vimInsertModeRemaps`, `pluginConfigs`, ...) — skipped at project scope, emitted under `--global`.
  - `Managed` (`allowManagedHooksOnly`, `requiredMinimumVersion`, ...) and `Global config` (`diffTool`, `autoConnectIde`, ...) — skipped in both scopes, since neither `.claude/settings.json` nor `~/.claude/settings.json` is the file that reads them.
  - As with the sandbox filtering, only the override copy is filtered; a value already in the target file is left untouched, and the warning says so.
- Prototype-pollution keys are skipped, consistent with `deepMergeRecords`.

### Codex CLI — `[tui]`

- `tui` joins `CODEXCLI_OVERRIDE_KEYS`, so it is written verbatim as a top-level `config.toml` table (shallow-merged with existing siblings, like `apps`) and re-extracted on import. `tui: z.optional(z.looseObject({}))` is added to `CodexcliPermissionsOverrideSchema` so the intent is documented, while staying loose for the `tui.keymap.*` bindings.
- Adding it to `CODEXCLI_OVERRIDE_KEYS` also extends the permissions feature's `SHARED_CONFIG_OWNERSHIP` declaration for `.codex/config.toml`, which is the single source both the writer and the gateway read.

### Docs

`docs/reference/file-formats.md`: the Claude Code override callout gains the passthrough rule, the exclusion list and the scope caveat; the Codex CLI callout lists `tui`.

## Validation evidence

- Claude Code `editorMode` and the per-key Scope column: https://code.claude.com/docs/en/settings-reference — `editorMode` is `Any file`; the Managed / User-or-managed / Global-config lists in the new constants are taken verbatim from that table.
- Claude Code keys cited in the issue comment: https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md (2.1.217 – 2.1.239).
- Codex `[tui] vim_mode_default`: https://developers.openai.com/codex/config-reference and the 0.129.0 modal Vim composer release notes https://github.com/openai/codex/releases/tag/rust-v0.129.0.

## Test plan

- New unit tests in `claudecode-permissions.test.ts`: unmodeled key passthrough, deep merge preserving siblings, refusal to write `hooks`/`$schema`, project-scope drop + warning for a user-scope key, `--global` emission of the same key, both-scope drop + warning for managed / `~/.claude.json` keys, and an import → generate round trip.
- New unit test in `codexcli-permissions.test.ts`: `[tui] vim_mode_default` is written with no "not managed" warning and round-trips on import.
- The pre-existing assertion that a top-level key is *not* swept into the override was deliberately inverted — that behavior is the bug this issue reports.
- `pnpm cicheck` passes: 375 test files / 8778 tests, plus fmt / oxlint / typecheck / docs-content / gitignore / supported-tools / cspell / secretlint.

Closes #2715


dyoshikawa:
@dyoshikawa Thank you!

PR #2738, opened 2026-08-24
fix(skills,subagents): fold case when de-duplicating across discovery roots

### LINKED ISSUE #2733: skills: cross-root de-duplication is case-sensitive, so a shared-root skill can overwrite the tool-specific one on case-insensitive filesystems
## Problem

`SkillsProcessor` de-duplicates skills across a target's search roots with an exact-match `Set` of skill names:

```ts
const seenSkillNames = new Set<string>();
// ...
const skillName = skill.getImportIdentity(); // = getDirName()
if (seenSkillNames.has(skillName)) {
  continue;
}
```

So `.junie/skills/dupskill` and `.agents/skills/DupSkill` — the same name modulo case — are both imported. They are written back out as `.rulesync/skills/dupskill` and `.rulesync/skills/DupSkill`, which resolve to a single directory on the case-insensitive filesystems of macOS and Windows. `writeAiDirs` writes in root order, so the shared `.agents/skills/` copy lands last and silently overwrites the tool-specific one.

That inverts the documented precedence: the tool-specific root (`.junie/skills/`, `.rovodev/skills/`, `.vibe/skills/`, `.kimi-code/skills/`) is supposed to win over the shared Agent Skills root.

The rulesync-side loader already has a case-insensitive identity helper (`mergeByCaseInsensitiveIdentity` in the same file), so only the tool-side de-duplication is case-sensitive — an asymmetry rather than a deliberate choice.

## Impact

Low. It needs two skills whose names differ only in case plus a case-insensitive filesystem. The consequence is that content from the shared tree — which another tool or a third party may own — replaces a project's own skill in the canonical `.rulesync/skills/` tree, and is then distributed to every other tool on the next generate.

## Scope

Not specific to one target: it applies to every target that scans more than one skill root, i.e. `junie`, `rovodev`, `vibe`, and `kimi-code`. The same question is worth checking for the subagents processor, which has the parallel multi-root shape.

## Suggested fix

Key `seenSkillNames` on the case-folded name (or reuse the identity function behind `mergeByCaseInsensitiveIdentity`) so the first root to provide a name wins regardless of case, and add a regression test per multi-root target.

Found during the review of #2730.

dyoshikawa:
## Summary

The tool-side loaders in `SkillsProcessor.loadToolDirs()` and `SubagentsProcessor.loadToolFiles()` scan several discovery roots in precedence order and keep the first spelling of each import identity — but they compared those identities exactly, with a plain `Set<string>`.

So `.junie/skills/dupskill` and `.agents/skills/DupSkill` both came through. They are written back as `.rulesync/skills/dupskill` and `.rulesync/skills/DupSkill`, which are a single directory on the case-insensitive filesystems of macOS and Windows, and `writeAiDirs` writes in root order — so the shared Agent Skills copy landed last and silently overwrote the tool-specific one. That inverts the documented precedence, in which the tool-specific root wins.

The rulesync-side loader already folds case (`mergeByCaseInsensitiveIdentity` in the same file), so this was an asymmetry rather than a deliberate choice.

## What changed

`ClaimedIdentities` (`src/types/feature-processor.ts`) tracks the identities an earlier root claimed, keyed on the case-folded spelling. `claim()` returns `null` for a new identity, or the spelling the earlier root used — equal to the argument for a plain duplicate, different for a case-only collision — so the two cases can be worded apart:

- **Exact repeat**: an ordinary overlay. Resolved silently, exactly as before.
- **Case-only collision**: the higher-precedence root still wins, and a warning names the copy that is ignored. Silence here would be the worst outcome, since the ignored copy is not the one whose name the user would search for.

Both processors now use it. The subagents processor keeps its existing duplicate warning verbatim and adds a case-only variant.

## Scope

Every target that scans more than one root: `junie`, `rovodev`, `vibe`, and `kimi-code` for skills (`kimi-code` was already safe, since `KimiCodeSkill.getImportIdentity()` lowercases), plus the parallel multi-root shape in the subagents processor. Both the standalone-file loop and the `loadAdditionalImportFiles` (inline-subagent) loop are covered.

## Behavior change on case-sensitive filesystems

The rule is applied on **every** platform, not only where the filesystem folds case. On Linux (and any case-sensitive macOS volume), `.junie/skills/my-skill` and `.agents/skills/My-Skill` really are two separate skills, and until now both were imported. From this change on, the lower-precedence one is dropped with a warning — for `rulesync import` and for `rulesync convert`, which shares the same loaders.

That is deliberate: the alternative is a result that depends on the filesystem the command happens to run on, so a repository imported on Linux and on macOS would produce different `.rulesync/` trees. The trade-off is documented in the new cross-root duplicate note in `docs/reference/file-formats.md`; the fix is to rename one of the two skills.

## Test plan

- `pnpm cicheck` — green (376 files / 8925 tests).
- New: `ClaimedIdentities` unit tests; case-only precedence regressions for rovodev, junie, and vibe skills; a "no warning for an exact repeat" test; a junie subagent `planner.md` / `Planner.md` regression, and two opencode inline-`agent` collision tests (inline vs. standalone file, and inline vs. inline).

Closes #2733


dyoshikawa:
@dyoshikawa Thank you!

PR #2742, opened 2026-08-24
feat(rules): import Junie's .junie/rules and .junie/playbook.md

### LINKED ISSUE #2732: junie: hand-authored .junie/rules/*.md and .junie/playbook.md are skipped on import (no import-only rule root)
## Summary

Junie's 2026-08-21 docs revision added a multi-file project-rules branch — root `AGENTS.md` combined with `.junie/playbook.md` and every `.junie/rules/*.md`. PR #2730 established that rulesync should keep **generating** the single folded `.junie/AGENTS.md`, and refreshed the rationale accordingly. What that PR deliberately did not address is the **import** side, which is filed here so it does not drop off the backlog when #2728 closes.

## The gap

Junie's discovery order is:

1. `.junie/AGENTS.md` — "used exclusively and no other guidelines files are combined with it"
2. root `AGENTS.md` **combined with** `.junie/playbook.md` and every `.junie/rules/*.md`
3. legacy `.junie/guidelines.md` / `.junie/guidelines/`

Branch 2 is live for any repo that has not (yet) got a `.junie/AGENTS.md` — which is exactly the state a user is in the first time they run `rulesync import --targets junie --features rules`. `JunieRule.getSettablePaths` accepts only `.junie/AGENTS.md` plus the legacy `.junie/guidelines.md` (`src/features/rules/junie-rule.ts`), so hand-authored `.junie/rules/*.md` and `.junie/playbook.md` are silently skipped.

This is the asymmetry PR #2730 left behind: on the **skills** side it accepted the newly documented `.agents/skills/` read root as `importOnlySkillRoots`, but rules have no equivalent.

## Why it was not fixed in #2730

`ToolRuleSettablePaths` has no import-only counterpart to `importOnlySkillRoots`:

- `nonRoot` (`src/features/rules/tool-rule.ts`) is a single managed directory that generation writes and orphan deletion prunes. Declaring `.junie/rules/` there would make rulesync emit files Junie never reads (branch 1 shadows branch 2) and make it a deletion candidate.
- `alternativeRoots` addresses single fallback **files**, not a directory, so `.junie/rules/*.md` does not fit; `.junie/playbook.md` would fit shape-wise, but it is a *combined* input rather than an alternative root, and adding it also pulls the path into gitignore derivation (`src/cli/commands/gitignore-derive.ts` walks `root` + `alternativeRoots`), which would gitignore a file rulesync never writes.

So the fix needs a new concept, not a new entry in an existing list.

## Proposed follow-up

Add an import-only, read-only rule root to `ToolRuleSettablePaths` — the rules-side analogue of `importOnlySkillRoots`:

- read by the RulesProcessor's import path only;
- never written by generation, never a deletion candidate, and excluded from `gitignore-derive.ts`;
- available at project scope; note that `JunieRule.getSettablePaths` currently sets `alternativeRoots` only in the project branch, so the global branch needs its own decision.

Then declare `.junie/rules/` (directory) and `.junie/playbook.md` (file) on `junie`, and drop the "Tracked as a follow-up" note from the `JunieRule` class doc and from `docs/reference/file-formats.md`.

Worth checking whether other targets already want the same mechanism before designing it — Kilo Code's `.kilo/rules-{mode}/` (#2499) and Qwen Code's `.qwen/review-rules.md` (#2668) are nearby candidates.

## References

- https://junie.jetbrains.com/docs/guidelines-and-memory.html — the discovery order including `.junie/playbook.md` and `.junie/rules/*.md`.
- https://junie.jetbrains.com/docs/junie-ide-plugin.html — the `.junie/AGENTS.md` exclusivity rule that makes branch 2 unreachable for generated output.
- `src/features/rules/junie-rule.ts`, `src/features/rules/tool-rule.ts`, `src/features/skills/tool-skill.ts` (`importOnlySkillRoots`), `src/cli/commands/gitignore-derive.ts`.

Related: #2728, #2730, #2211

dyoshikawa:
## Summary

Junie reads root `AGENTS.md` together with `.junie/playbook.md` and every `.junie/rules/*.md` ([Junie docs](https://junie.jetbrains.com/docs/guidelines-and-memory.html)), but `rulesync import` only ever looked at `.junie/AGENTS.md` and the legacy `.junie/guidelines.md`. Rules a user had hand-authored under `.junie/rules/` were silently invisible.

## Changes

- Add `ToolRuleImportOnlyRoot` and `ToolRuleSettablePaths.importOnlyRoots`: read-only rule roots scanned on **import only**. Because generation never writes them, they are never orphan-deletion candidates and never gitignore entries (`gitignore-derive.ts` walks `root` and `alternativeRoots` only). This is the rules-side counterpart of the skills-side `importOnlySkillRoots`.
- `rules-processor.ts` loads these roots when `forDeletion` is false, so `rulesync import` picks them up while `rulesync generate`'s deletion pass leaves the user's own files alone.
- Junie declares `.junie/rules/*.md` and `.junie/playbook.md` as import-only roots for the **project** scope. The global scope is unchanged, since Junie documents these paths per project.
- `JunieRule.fromFile` now honors an explicit `relativeDirPath`, and only treats a file as the root rule when it sits in the root directory *and* has a root filename.
- Docs: `docs/reference/file-formats.md` describes the new import-only behavior (imported, never generated, never deleted, never gitignored).

## Tests

- `junie-rule.test.ts`: `.junie/rules/style.md` and `.junie/playbook.md` import as non-root rules; `.junie/rules/AGENTS.md` is not mistaken for the root rule.
- `rules-processor.test.ts`: Junie's `.junie/rules` and playbook are imported, and are absent from `loadToolFiles({ forDeletion: true })`.

Closes #2732


dyoshikawa:
@dyoshikawa Thank you!

PR #2745, opened 2026-08-24
feat(skills): carry hidden companion files and recover malformed skill frontmatter

### LINKED ISSUE #2731: Follow up Agent Skills upstream updates: hidden companion entries dropped after the any-files spec amendment, malformed-YAML import fallback unimplemented
## Summary

The Agent Skills specification was amended on 2026-08-04 to state that a skill directory **may contain any files and directories** beyond the required `SKILL.md`, with `scripts/` / `references/` / `assets/` demoted to non-exhaustive recommendations. rulesync's companion-file collector globs without globby's `dot` option, so any hidden file or directory bundled in a skill is silently dropped on both import and generate. Separately, the client-implementation guide's prescribed malformed-YAML fallback is still unimplemented, so a skill whose `description` contains a colon is skipped outright rather than recovered — the half of #2513 that was listed as optional and never shipped. No Agent Skills issue is currently open, so both are filed here.

The sibling target **AGENTS.md (`agentsmd`) has no gaps** this run: the spec repo has had no commits since 2026-03-12 and the normative text is unchanged.

## Recent Releases

The spec repo `agentskills/agentskills` is **unversioned** — no releases and no tags have ever been published (`pushed_at: 2026-08-09T20:36:04Z`), so changes can only be dated by commit. In the 2026-07-25 → 2026-08-23 window:

| Date | Commit | Content |
| --- | --- | --- |
| 2026-08-04 | [`6b865a20`](https://github.com/agentskills/agentskills/commit/6b865a20) (PR #268) | "Optional directories" now opens with **"A skill directory may contain any files and directories beyond the required `SKILL.md`. The conventions below are recommendations…"** — the source of gap 1 |
| 2026-08-03/04 | [`3f3bbec8`](https://github.com/agentskills/agentskills/commit/3f3bbec8) (PR #479, merged `217be548`) | `metadata` overview row clarified to "a map from string keys to string values" — already satisfied by `toStringMetadata` (`src/features/skills/agentsskills-skill.ts:130-138`); not a gap |
| 2026-08-03 | [`f130f348`](https://github.com/agentskills/agentskills/commit/f130f348) | Repo governance only: `docs/specification.mdx` declared authoritative, `skills-ref/` declared a demonstration artifact rather than a source of requirements. Non-normative |
| 2026-08-09 | `be055541`, `69ef37e9`, `346deaa7` | Client showcase and branding (Hermes Agent, OpenClaw); no normative change |

`docs/client-implementation/adding-skills-support` was **not** touched in the window, and `docs/docs.json` shows no new normative page — gap 2 is a standing unimplemented recommendation rather than a new one, surfaced here because gap 1 puts the same code path under review.

### 1. `skills` (project ✅ / global 🌏, generate + import) — hidden files and directories bundled in a skill are silently dropped

**Upstream.** https://agentskills.io/specification#optional-directories: "A skill directory may contain any files and directories beyond the required `SKILL.md`." Added 2026-08-04 by [`6b865a20`](https://github.com/agentskills/agentskills/commit/6b865a20). Before that amendment the three recommended directories read as the effective set, so a `dot`-less glob was defensible; it is not any more.

**rulesync.** Companion files are collected by `AiDir.collectOtherFiles` (`src/types/ai-dir.ts:178-200`), which globs `<dir>/**/*` through `findFilesByGlobs` (`src/utils/file.ts:415-475`). That call never sets globby's `dot` option, so every dot-prefixed file or directory is excluded. The same helper serves both directions — `RulesyncSkill.fromDir` (`src/features/skills/rulesync-skill.ts:571-606`) and `ToolSkill.loadSkillDirContent` (`src/features/skills/tool-skill.ts:224-261`) — so a hidden entry is lost on import *and* never emitted on generate.

**Reproduced.** A scratch project whose `.rulesync/skills/demo/` contained `SKILL.md`, `scripts/run.sh`, `.env.example` and `.config/a.json` dry-ran to `.agents/skills/demo/SKILL.md` + `.agents/skills/demo/scripts/run.sh` only. Both dot entries vanished with no warning.

**Proposed follow-up.** Collect skill companion entries with `dot: true` — as a scoped option on `findFilesByGlobs` / `collectOtherFiles` rather than a global default — while keeping an explicit exclusion for VCS and tooling noise (`.git/`, `.DS_Store`), and warn instead of silently dropping whatever stays excluded.

**Relation to #2563.** #2563 is the lossless skill-tree contract and is the natural place for this to land, but its acceptance list covers byte fidelity, executable mode, symlinks and Unicode paths and does **not** mention hidden entries — a tree can satisfy every one of its criteria and still lose `.env.example`. Treat this as an extra acceptance case for #2563 if the maintainer prefers one workstream, or as a standalone one-line fix.

### 2. `skills` (import) — the guide's malformed-YAML fallback is unimplemented, so a colon-in-description skill is skipped entirely

**Upstream.** https://agentskills.io/client-implementation/adding-skills-support, section "Handling malformed YAML", names exactly this case (`description: Use this skill when: the user asks about PDFs`) and recommends "a fallback that wraps such values in quotes or converts them to YAML block scalars before retrying. This improves cross-client compatibility at minimal cost." (Page re-fetched 2026-08-23; unchanged in the window.)

**rulesync.** `parseFrontmatter` (`src/utils/frontmatter.ts`) hands off to gray-matter/js-yaml and lets the exception propagate with no retry path — grepping for `retry` / `colon` / block-scalar handling finds only the unrelated `avoidBlockScalars` writer option. Confirmed against the guide's own example with this repo's gray-matter: `THROW YAMLException bad indentation of a mapping entry (3:33)`. The lenient-import work from #2513 then catches it and logs `Skipping .agents/skills/<name>: …` (`src/features/skills/skills-processor.ts:783-789`), so the skill is **dropped rather than recovered** — and `.agents/skills/` is precisely the cross-vendor directory where foreign-authored "Use when: …" descriptions turn up.

**Proposed follow-up.** Before giving up on a `SKILL.md`, retry the frontmatter parse once with the offending scalar quoted (or converted to a block scalar), and skip only if the retry also fails.

**Triage note.** Closed #2513 listed this fallback as an explicit *optional* extra and shipped only the skip-and-warn half, so nothing currently tracks it.

**Minor, same area** (probably not worth its own issue): the guide says "Description is missing or empty → skip the skill", but `AgentsSkillsSkillFrontmatterSchema` (`src/features/skills/agentsskills-skill.ts:25-35`) accepts `description: ""`, so an empty-description skill imports cleanly; only the generate side warns.

## Proposed Follow-up

1. Add a scoped `dot: true` to skill companion-entry collection (with `.git/` / `.DS_Store` exclusions and a warning for anything still dropped), or fold it into #2563 as an added acceptance case.
2. Add a single requote/block-scalar retry to `SKILL.md` frontmatter parsing before the lenient-import skip path gives up.
3. Optionally reject `description: ""` on import to match the guide, aligning the import side with the existing generate-side warning.

## Verified this run (no action)

- **`agentsmd`** — the AGENTS.md standard repo (`agentsmd/agents.md`; `openai/agents.md` redirects to it) has **no commits since 2026-03-12** (`?since=2026-06-01` is empty) and has never cut a release or tag. The normative text re-fetched on 2026-08-23 is unchanged: root `AGENTS.md`, plain Markdown, no frontmatter, nested `AGENTS.md` with nearest-file-wins, no `.agents/` subdirectory conventions, no global scope. Repo activity is unmerged proposals only — issue #71 (a `.agent` directory, updated 2026-08-22), #232 (`agentaccess.txt`), #228, PR #223 (governance), PR #226 (Antigravity CLI branding). Nothing to follow yet.
- **#2428 and #1827 are resolved in code** — nested `AGENTS.md` import works via `AgentsMdRule.getNestedFilePatterns` (`src/features/rules/agentsmd-rule.ts:80-92`), the nested branch of `fromFile` (`:110-148`) and `toRulesyncRule()` reconstructing `agentsmd: { subprojectPath }` (`:178-213`), wired at `src/features/rules/rules-processor.ts:2044`; the `agentsmd` / `agentsskills` `.agents/skills/` double-write no longer reproduces (both resolve the same path and emit identical spec-conformant output). Non-root rules still default to `.agents/memories/`, which is the decided design from #1827 with `agentsmd.subprojectPath` as the spec-shaped escape hatch — not a new gap.
- **#2429, #1791, #1684, #2515 are shipped** — `allowed-tools` joined to a space-separated scalar, `compatibility` flattened to a string, name/description/compatibility violations warned via `collectAgentSkillViolations` (`agentsskills-skill.ts:186-283`), global `~/.agents/skills/` supported, and lenient per-skill import in place (`meta.lenientImport: true`, `skills-processor.ts:166-184`).
- **Dry runs match the matrix** — `--targets agentsskills` loads 42 rulesync skills and reports every other feature unsupported; `--global` writes under `~/.agents/skills/**` including companion files (`references/*.md`, `LICENSE.txt`). `--targets agentsmd` writes `AGENTS.md`, `.agents/memories/*.md`, `.agents/subagents/*.md`; its `--global` run correctly reports rules as project-scope-only.

## References

- https://agentskills.io/specification — the current normative spec, including the amended "Optional directories" text behind gap 1 (fetched 2026-08-23).
- https://agentskills.io/client-implementation/adding-skills-support — the discovery table, lenient validation and the malformed-YAML fallback behind gap 2 (fetched 2026-08-23).
- https://github.com/agentskills/agentskills/commit/6b865a20 — "a skill directory may contain any files and directories" (2026-08-04); dates gap 1.
- https://github.com/agentskills/agentskills/commit/3f3bbec8 — the `metadata` string→string clarification (2026-08-03, merged 2026-08-04); cited as a checked non-gap.
- https://github.com/agentskills/agentskills/commit/f130f348 — repo governance consolidation (2026-08-03); cited to show `docs/specification.mdx`, not `skills-ref/`, is authoritative.
- https://agents.md/ — the AGENTS.md standard text, re-fetched 2026-08-23 and unchanged.
- https://github.com/agentsmd/agents.md — the AGENTS.md repo; last commit `d1ac7f06` (2026-03-12), no releases or tags.

Related: #2563, #2513, #2428, #1827

dyoshikawa:
## Summary

Closes #2731.

Three gaps in the Agent Skills handling, all verified against the [specification](https://agentskills.io/specification) and the [client-implementation guide](https://agentskills.io/client-implementation/adding-skills-support) before implementing.

### 1. Hidden companion files were silently dropped

The spec says a skill directory "may contain any files and directories beyond the required `SKILL.md`", but companion discovery goes through `findFilesByGlobs`, whose `globby` call left `dot` at its default (`false`). A hidden `.env.example`, `.config/`, or any other dot-prefixed entry beside `SKILL.md` was invisible to both import and generate.

- `findFilesByGlobs` gets an opt-in `dot` option; discovery globs that look for *named* config files keep the safe default, since a hidden entry there is far more likely to be editor or VCS noise.
- `AiDir.collectOtherFiles` — the only "carry this tree as it is" caller — turns it on.
- Two entries are never carried: `.git` (matched both as a directory prefix and as a bare entry, so a submodule/worktree pointer file is caught) and `.DS_Store`. Skipping a nested repository is warned about, because its tracked files are copied while its history is not.

### 2. Malformed frontmatter was dropped instead of retried

The client guide's "Handling malformed YAML" section names `description: Use this skill when: the user asks about PDFs` and suggests a fallback that quotes such values before retrying. Rulesync had no retry, so the lenient skill import caught the parse error and skipped the whole skill.

`parseFrontmatterWithYamlRepair` retries once with top-level unquoted colon-bearing values quoted. It is deliberately narrow: one pass, top-level entries only, values already starting as some other YAML construct left alone, and — because a colon only ends a plain scalar when whitespace or the line end follows — `homepage: https://example.com` is untouched. If the repair does not help, the **original** error surfaces, so a genuinely broken file still fails with the message that describes what is actually wrong with it.

### 3. Empty `description` on import

The spec requires a non-empty `description` and the guide has conformant clients skip a skill without one. The issue proposed skipping on import too; that is rejected here. Rulesync converts rather than loads, so dropping the directory would lose content the user can still fix. It imports the skill and warns instead — the "log the error" half of the guide's advice, and the same thing the generate side already does with this violation.

### Bonus: gray-matter cache poisoning

Found while testing the above. `matter(input)` caches by content string, and writes `matter.cache[content] = file` **before** parsing. A throwing parse therefore leaves a poisoned entry (`data: {}`, `content` = the whole unparsed file) that the next identical parse returns *silently* — so the same malformed skill vendored under two directories would import as empty frontmatter with the YAML spilled into the body, no error raised. Passing `{}` as options disables both cache read and write. Covered by a dedicated regression test.

## Test plan

- `pnpm cicheck` green (376 files / 8951 tests).
- New: `findFilesByGlobs` `dot` default / enabled / ignore-still-applies.
- New: 8 `parseFrontmatterWithYamlRepair` cases, including the CRLF file, the URL that must not be rewritten, the unfixable file reporting its original error, and the cache regression.
- New: `agentsskills` `fromDir` carries hidden companions but not `.DS_Store` / `.git/HEAD`; recovers an unquoted-colon `SKILL.md`.
- Updated: the existing "skip an unparseable skill" processor test now uses a genuine indentation error (the colon case is recovered by design), with a new sibling test asserting the recovery.
- Docs: the supporting-file note and a new malformed-frontmatter note in `docs/reference/file-formats.md`, with `docs-content.ts` regenerated.


dyoshikawa:
Automated review loop stopped at its 10-iteration cap, so this PR is left open for a manual merge decision rather than merged automatically.

State at `15f7a965`:

- All GitHub Actions checks are green (code quality, CodeQL, binaries, E2E on Linux/macOS/Windows).
- Every `mid`-or-above finding raised in rounds 1-10 has been fixed, and each fix carries a regression test that was verified to fail with the fix disabled.
- The last round's two `mid` findings are fixed in `15f7a965`: a credential file name (`.env`, `.netrc`, `.npmrc`) is now judged at every path segment rather than only the last, and the escaping-directory exemption is narrowed to a sibling of the skill directory so a global skill under `~/.config/agents/skills/` no longer exempts the whole `~/.config` tree.

Deliberately not changed, with reasons:

- `findFilesByGlobs` keeps its `dot` option (documented contract, no production caller today).
- `src/types/ai-dir.ts` is not split into a dedicated walker module - a restructure larger than this branch should carry.
- A tool home such as `~/.claude/` is deliberately absent from the escaping-directory list, because that is where the global skills this feature exists to share are kept.

dyoshikawa:
@dyoshikawa Thank you!

PR #2752, opened 2026-08-25
fix(input-roots): address PR #2714 follow-up findings

### LINKED ISSUE #2724: Follow-ups from PR #2714 review (multiple input roots)
## Background

PR #2714 ("feat: support multiple input roots", closes #2652) introduced the plural `inputRoots` config field, deprecated the singular `inputRoot` as an alias, and reworked every feature processor to load from an ordered list of source trees.

The PR went through a maintainer review round; the author addressed the blocking findings and the follow-up review found no `high`/`critical` issues, so the PR was merged. This issue collects the remaining `mid`/`low` findings from the code and security reviews so they are tracked rather than lost.

None of the items below block a release. They are ordered roughly by impact.

### 1. [mid] Curated/local dedup became case-insensitive, changing single-root behavior

`src/features/rules/rules-processor.ts` (`loadRulesyncFiles`) and the equivalent `localSkillNames` path in `src/features/skills/skills-processor.ts` now lowercase both sides when deciding whether a local file shadows a `.curated/` file:

```ts
const localRelativePaths = new Set(
  localFiles.map((file) => relative(rulesyncOutputRoot, file).toLowerCase()),
);
// ...
.filter(({ relativeFilePath }) => !localRelativePaths.has(relativeFilePath.toLowerCase()));
```

`main` before the PR did not call `.toLowerCase()` here.

Failure scenario: on a case-sensitive filesystem, a user with `.rulesync/rules/Style.md` (hand-written) and `.rulesync/rules/.curated/style.md` (from `rulesync fetch`) silently loses the curated file after upgrading — **even when they use a single input root**. Cross-root merging (`mergeByCaseInsensitiveIdentity`) emits a warning on such a collision; this in-tree path is silent.

This behavior change is orthogonal to the PR's subject.

### 2. [mid] The `checks` feature has no `--input-roots` overlay e2e coverage

`src/e2e/e2e-input-root.spec.ts` documents (around the comment block at the top of the overlay suite) that glob-based features "rules, commands, subagents, checks, skills" follow last-wins-by-relative-path, but the actual specs only cover rules / commands / subagents / skills / ignore / hooks / permissions / mcp. `checks` appears in neither the overlay suite nor the single-root `--input-roots` suite.

Failure scenario: a regression in `ChecksProcessor`'s multi-root wiring would not be caught by CI.

### 3. [low] `.rulesync.local/` is not gitignored, but the docs recommend keeping it untracked

`docs/guide/separate-input-root.md` presents per-developer overlays as the primary use case and describes `.rulesync.local/` as "an optional **untracked** tree". The gitignore entry and the `RULESYNC_LOCAL_RELATIVE_DIR_PATH` constant were intentionally reverted in commit `09153afa` ("fix: defer local input root gitignore"), and `rulesync gitignore` therefore does not emit it.

Failure scenario: a developer follows the guide, puts personal MCP server definitions (potentially with API keys) in `.rulesync.local/mcp.jsonc`, and commits it with `git add .` because they assumed it was excluded.

Note: root names are arbitrary, so a fully general automatic gitignore rule is not possible. The minimum fix is a sentence in the guide instructing users to add their local root to `.gitignore` manually; optionally restore the entry for the conventional `.rulesync.local/` name.

### 4. [low] Overlay replacement of `ignore` / `permissions` files is whole-file and silent

`src/features/ignore/ignore-processor.ts`, `hooks-processor.ts`, and `permissions-processor.ts` use the shared `pickLastRootWithFile` helper (`src/types/feature-processor.ts`): the last root that provides the file replaces it entirely, with no line-level merge and no log output.

Failure scenario: a shared `.rulesync/.aiignore` containing secret-file exclusions (`**/*.pem`, `.env*`) is fully replaced when a developer adds a one-line `.rulesync.local/.aiignore`, so the generated per-tool ignore files lose those patterns and AI tools can read the secrets.

The behavior is documented and intentional, but `ignore` and `permissions` are security controls, and the analogous `mergeByCaseInsensitiveIdentity` path does warn on collisions. Consider logging when a whole-file replacement happens for these features.

### 5. [low] "Input root does not exist" message always names `inputRoots`

`inspectInputRoots` in `src/lib/generate.ts` returns:

> Configured primary input root '<path>' does not exist. Create the directory or update your inputRoots setting.

A user who only set the deprecated singular `"inputRoot": "./shared"` and typo'd the path is told to fix an `inputRoots` setting they never wrote. `doctor` already switches wording via `resolved.field` (`src/cli/commands/doctor.ts`); the same `field` should be threaded here.

Additionally, the check is `existsSync`-based, so a primary root that exists **as a file** is reported as "does not exist".

### 6. [low] `resolveEffectiveInputRoots` resolves relative entries against `process.cwd()`, not its `cwd` argument

`src/config/config-resolver.ts`:

```ts
const candidates = source.map((entry) => resolve(entry));
```

The `cwd` parameter is only used by the default branch (`join(cwd, RULESYNC_RELATIVE_DIR_PATH)`). Both current callers pass `cwd: process.cwd()`, so there is no live bug, but the function's contract is misleading and will break silently the first time a caller (or test) passes a different directory. Should be `resolve(cwd, entry)`.

### 7. [low] `doctor` silently drops invalid `inputRoots` entries

`readDoctorInputRootConfig` in `src/cli/commands/doctor.ts` filters out non-string and empty-string entries and treats an empty result as `inputRoots: undefined`.

Failure scenario: with `"inputRoots": [""]`, `rulesync doctor` reports nothing (it inspects the default `.rulesync` instead), while `rulesync generate` fails with `outputRoot cannot be an empty string` from `validateOutputRoot`. `doctor` exists to catch configuration problems ahead of time, so the divergence undermines it.

### 8. [low] Tool-scoped MCP merge bypasses the `isPrototypePollutionKey` guard

`src/features/mcp/rulesync-mcp.ts` guards the top-level `mcpServers` merge and the key walk with `isPrototypePollutionKey`, but the tool-scoped blocks are merged with a bare spread:

```ts
const mergedBlock: Record<string, unknown> = { ...baseBlock, ...overlayBlock };
// ...
mergedBlock.mcpServers = { ...baseServers, ...overlayServers };
```

This is **not** a vulnerability — object spread uses `CreateDataProperty` semantics, so no prototype is actually polluted; a `__proto__` key just lands as an own property. The issue is the asymmetry: such a key is stripped at the top level but survives into the emitted JSON under a tool scope. Reusing the existing `mergeTopLevelMcpServers` helper would make the two paths consistent (and is the DRY-er option).

### 9. [low] Multi-root MCP loading strips JSONC comments and misreports the source path

`RulesyncMcp.fromRoots` re-serializes with `JSON.stringify(merged, null, 2)` and anchors to `paths.recommended.relativeFilePath` (`mcp.jsonc`) whenever there is more than one root — even when only one root actually supplied a file. Single-root configs short-circuit to `fromFile` and are unaffected.

Failure scenario: with `inputRoots: ["./.rulesync", "./.rulesync.local"]` and an empty `.rulesync.local/`, the comments in `.rulesync/mcp.jsonc` are silently dropped. For a project that only has the legacy `.rulesync/mcp.json`, error and validation messages point at a `mcp.jsonc` that does not exist.

### 10. [low] `hermesagent-command.ts` skill lookup is case-sensitive and `fromDir` calls became sequential

`src/features/commands/hermesagent-command.ts` keys `skillsByName` on the raw `dirName`, while `SkillsProcessor` merges across roots case-insensitively via `mergeByCaseInsensitiveIdentity`.

Failure scenario: with `skills/Deploy/` in the base root and `skills/deploy/` in the overlay, the skills processor folds them into one (with a warning) while hermesagent's command validation treats them as two distinct skills, so validation results and generated output disagree.

Separately, `await RulesyncSkill.fromDir(...)` now runs sequentially inside a nested `for` loop where it previously used `Promise.all`. Generate time grows with skills × roots; restoring `Promise.all` within each root (keeping roots sequential) would recover it.

### 11. [low] `formatTriggerPaths` mis-detects `..`-prefixed directory names

`src/lib/watch.ts`: `rel.startsWith("..")` also matches a legitimate in-root directory such as `..foo`. This is display-only (it is not used for access control), so the impact is limited to printing an absolute path where a relative one was intended. The precise check is `rel === ".." || rel.startsWith(".." + sep)`.

### 12. [low] Leftover `getInputRoot` mocks in tests

`src/cli/commands/import.test.ts` and `src/lib/import.test.ts` still mock the removed `getInputRoot`. `import` does not consume input roots, so this is harmless — just cleanup.

## Solution / Next Steps

Suggested order:

1. **#1** — decide whether the case-insensitive curated/local dedup is intentional. If yes, document the behavior change; if no, restore case-sensitive matching for the in-tree path. Either way, consider emitting the same collision warning the cross-root merge emits.
2. **#2** — add a `checks` overlay case to `src/e2e/e2e-input-root.spec.ts` so the multi-root matrix has no gap.
3. **#3** — add a sentence to `docs/guide/separate-input-root.md` telling users to add their local overlay root to `.gitignore` manually, and reconsider restoring the `.rulesync.local/` gitignore entry.
4. **#4** — log a warning when a whole-file overlay replacement occurs for `ignore` / `permissions` / `hooks`.
5. **#5, #7** — thread the resolved field name into `inspectInputRoots`' message, distinguish "exists as a file", and make `doctor` report invalid `inputRoots` entries instead of dropping them.
6. **#6, #8, #10, #11, #12** — mechanical cleanups; can be batched into one PR.
7. **#9** — either preserve the original text when exactly one root supplied the MCP file, or document that multi-root MCP output is normalized JSON.

### Notes from the security review

The supply-chain check came back clean: the PR's net diff against the merge base touches no `.github/`, `package.json`, or `pnpm-lock.yaml` content (an intermediate commit modified the lockfile and a later commit fully reverted it), and the added production code contains no `eval`, `child_process`, network calls, or `process.env` access. No existing path-traversal guard was removed.

One design note worth recording: `inputRoots` entries are validated only by `validateOutputRoot`, which rejects `..` segments, unnormalized paths, and filesystem roots but permits any normalized absolute path outside `cwd`. The legacy `inputRoot` was always joined with `.rulesync`, which bounded reads to that subdirectory; `inputRoots` applies no such join, so a `rulesync.jsonc` could point a root at e.g. `~/.claude` and have its contents re-emitted into the project's generated files. `outputRoots` already accepts arbitrary absolute paths under the same validator, and `rulesync.jsonc` is treated as trusted input by design, so this is an extension of existing capability rather than a new trust boundary. A warning when a root resolves outside `cwd` may still be worth adding.

---

Source: review of #2714.

dyoshikawa:
## Summary

Addresses the twelve follow-up findings collected after the multiple input roots feature (#2714) landed.

## Changes

- **Curated/local case collisions now warn.** Shadowing inside one source tree stays case-insensitive — a curated and a local file differing only in case are the same file on macOS and Windows — but a case-*only* match now emits a warning instead of silently dropping the curated artifact. Applied in both `rules-processor.ts` and `skills-processor.ts`.
- **e2e coverage for `checks`.** `src/e2e/e2e-input-root.spec.ts` gained a single-root case and a two-root overlay case so no feature is missing from the `--input-roots` matrix.
- **Gitignore `.rulesync.local/`.** The overlay tree the guide describes as untracked is now emitted by `pnpm dev gitignore`.
- **Single-file features log the winner.** When several roots supply the ignore / hooks / permissions file, `pickLastRootWithFile` now logs which root won and which roots it replaced.
- **Better input-root diagnostics.** `inspectInputRoots` distinguishes "exists but is not a directory" from "does not exist", and the hint names both `inputRoots` and the deprecated `inputRoot` so it is correct whichever spelling the user wrote.
- **Relative entries resolve against `cwd`.** `resolveEffectiveInputRoots` no longer falls back to the ambient `process.cwd()`.
- **`doctor` reports invalid config.** Non-string `inputRoot`, non-array `inputRoots`, an empty list, and per-index invalid entries are now reported as `config/input-root-invalid` instead of being dropped silently.
- **Prototype-pollution keys filtered in tool-scoped MCP merges**, matching the existing top-level `mcpServers` merge.
- **MCP single-root short circuit.** When exactly one root supplies the MCP file, its original content and source path are kept, so JSONC comments survive.
- **Hermes skill lookup case-folded**, so cross-root shadowing matches the skills processor; each root's skills load concurrently again.
- **`formatTriggerPaths` `..` detection fixed** so a child named `..foo` is no longer treated as an escape.
- **Leftover `getInputRoot` mocks removed** from the import tests.

## Verification

- `pnpm cicheck` — green (oxfmt, oxlint, tsgo, 376 test files / 8960 tests, docs-content, gitignore, supported-tools, cspell, secretlint).
- `pnpm test:e2e` — green (24 files / 908 tests).

Closes #2724


dyoshikawa:
@dyoshikawa Thank you!

Issue #2636, opened 2026-08-08
Follow up Pi Coding Agent upstream updates: preToolUse hooks cannot block tool calls (generated extension is observe-only)


dyoshikawa:
## Summary

The Pi hooks extension rulesync generates is observe-only: a `preToolUse` hook cannot block a Pi tool call, even though Pi's `tool_call` event is explicitly blocking and is Pi's only tool-gating mechanism (Pi ships no permissions/mcp/subagents surface). v0.84.1 (2026-08-07) additionally widened the blocking contract with `terminate`, none of which rulesync uses.

## Recent Releases

- [v0.84.1](https://github.com/earendil-works/pi/releases/tag/v0.84.1) (2026-08-07) — adds `terminate` to the `tool_call` block contract (upstream PR #7715): a blocked call can skip the automatic follow-up model call when every finalized result in the batch is terminating. Other items (auth check, Qwen Token Plan, fullscreen TUI fixes) are runtime-only.
- [v0.84.0](https://github.com/earendil-works/pi/releases/tag/v0.84.0) (2026-08-06) — `AGENTS.override.md` etc., already dispositioned in #2601.

### hooks — `preToolUse` cannot block; the generated extension observes only (project + global)

- **Upstream**: [extensions.md @ v0.84.1, Tool Events](https://github.com/earendil-works/pi/blob/v0.84.1/packages/coding-agent/docs/extensions.md#tool-events) — `tool_call` is explicitly "**Can block.**": "Return values from `tool_call` control blocking via `{ block: true, reason?: string, terminate?: boolean }`".
- **rulesync** (`project` + `global`): `generatePiExtensionCode` (`src/features/hooks/pi-extension-generator.ts`) emits handlers that only `await run(<command>)` and return nothing — the docstring states "The generated extension observes events only: it never blocks or mutates Pi events." A `preToolUse` hook whose command exits non-zero lets the tool run anyway, silently. Matchers work; denial does not.
- **Why this is material**: `tool_call` blocking is Pi's *only* tool gate, and `preToolUse` is the canonical event rulesync maps onto it. #2601 recorded this as a non-gap ("blocking … already reachable via `preToolUse`"), but the generator contradicts that: the mapping exists and the blocking half is dropped in codegen. The v0.84.1 `terminate` addition widened a contract rulesync isn't using at all.
[...]

PR #2773, opened 2026-08-25
feat(hooks): let pi beforeSubmitPrompt hooks cancel the prompt

### LINKED ISSUE #2772: Follow up Pi Coding Agent upstream updates: beforeSubmitPrompt hooks are observe-only although Pi's input event can skip the agent
## Summary

Pi's `input` extension event is interceptable — a handler can return
`{ action: "handled" }` to skip the agent entirely — but rulesync's generated
extension treats every event except `tool_call` as observe-only. A
`beforeSubmitPrompt` hook authored for `pi` therefore cannot cancel the prompt,
while the same canonical hook blocks on every other hook-capable target that
maps it. This is the same failure mode already accepted and fixed for
`preToolUse` in #2636.

## Recent Releases

Reviewed window: v0.75.x through
[v0.84.3 (2026-08-24)](https://github.com/earendil-works/pi/releases/tag/v0.84.3),
against the versioned docs under `packages/coding-agent/docs/` at that tag.

The `input` event's `transform` / `handled` results are a standing capability
rather than a new one: the event predates this window (`streamingBehavior` was
added to it in
[v0.77.0 (2026-05-28)](https://github.com/earendil-works/pi/releases/tag/v0.77.0)),
and both results are still documented at v0.84.3. So this is a gap rulesync has
never followed, not an upstream regression.

Other in-window changes were checked and need no follow-up: `defaultTools`
(v0.84.2) is implemented and #2687 is closed; `AGENTS.override.md` (v0.84.0) is
implemented as `pi.contextFile: override` and #2601 gap 2 is closed;
`terminate` on a blocked `tool_call` (v0.84.1) is deliberately left unset per
#2636; the new `powershell` built-in tool (v0.84.3) needs no change because
rulesync writes `defaultTools` verbatim with no tool-name allowlist; and
`session_compact_failed` (v0.84.3) has no counterpart in the canonical hook
model.

### hooks (`project`, `global`) — `beforeSubmitPrompt` cannot block or cancel a prompt

**Upstream.** Pi documents the `input` event's return value as selecting one of
three results:

- `{ action: "continue" }` — pass through (the default when a handler returns nothing)
- `{ action: "transform", text, images }` — rewrite the prompt before skill/template expansion
- `{ action: "handled" }` — "skip agent entirely (first handler to return this wins)"

Source: [`extensions.md#input` @ v0.84.3](https://github.com/earendil-works/pi/blob/v0.84.3/packages/coding-agent/docs/extensions.md#input),
under "Input Events → `input`".

**rulesync.** `CANONICAL_TO_PI_EVENT_NAMES` already maps `beforeSubmitPrompt` →
`input` (`src/types/hooks.ts:1319`), but the generated extension gates blocking
behavior on a single event: `const PI_BLOCKING_EVENT = "tool_call"`
(`src/features/hooks/pi-extension-generator.ts:24`). In `buildCommandLines`
(`src/features/hooks/pi-extension-generator.ts:100-134`), only the
`blocksToolCall` branch wraps the command in `try`/`catch` and returns
`{ block: true, reason: toBlockReason(error) }`; the `else` branch at line 127
emits a bare `await run(<command>)` with no `catch` and no return value. A
`beforeSubmitPrompt` command that exits non-zero therefore rejects inside the
handler and the prompt still reaches the model.

`docs/reference/file-formats.md:136` currently asserts the opposite of the
upstream docs: "`tool_call` is Pi's only blocking event and its only tool gate,
so … every other event, `postToolUse`/`tool_result` included, observes only and
cannot block or mutate Pi events."

**User-visible effect.** On Claude Code (`UserPromptSubmit`), Cline, Goose, Qwen
Code and Reasonix, a non-zero exit on this canonical event cancels the prompt.
Under `pi` the identical authored hook silently degrades to a fire-and-forget
side effect.

## Proposed Follow-up

- Extend `pi-extension-generator.ts` with a second blocking event alongside
  `PI_BLOCKING_EVENT`. For `input`, wrap the command in the existing
  `try`/`catch` + `toBlockReason` shape and return `{ action: "handled" }` on a
  non-zero exit, returning `{ action: "continue" }` otherwise. Because `handled`
  carries no reason field, surface the reason through `ctx.ui.notify`.
- Keep `transform` out of scope — the canonical hook model has no
  prompt-rewriting contract, so there is nothing to map it from.
- Correct the Pi sentence in `docs/reference/file-formats.md:136` so it says
  `tool_call` is Pi's only *tool* gate, while `input` gates prompt submission.

## Verified this run, not gaps

- **mcp, subagents, ignore, checks stay `unsupported` correctly.** `usage.md:308`
  at v0.84.3 states Pi "intentionally does not include built-in MCP … sub-agents",
  there is no `mcp.md` in the docs tree and no MCP key anywhere in `settings.md`,
  and no ignore-file or check/lint surface exists. The only "subagent" upstream is
  a bundled *example extension*, not a config surface.
- **skills** — all seven documented frontmatter fields (`name`, `description`,
  `license`, `compatibility`, `metadata`, `allowed-tools`,
  `disable-model-invocation`) are modelled in `src/features/skills/pi-skill.ts`.
- **`stop` → `agent_end` rather than `agent_settled`** is a recorded deliberate
  trade-off in #2601 gap 1.
- **Extension auto-discovery** — `extensions.md:113-120` confirms
  `.pi/extensions/*.ts` and `~/.pi/agent/extensions/*.ts` are auto-discovered, so
  the generated `rulesync-hooks.ts` needs no `extensions` settings entry.
- **Dry runs** — `pnpm run dev generate --targets pi --features "*" --dry-run`
  and its `--global` variant both succeed and match the matrix row.

## References

- https://github.com/earendil-works/pi/blob/v0.84.3/packages/coding-agent/docs/extensions.md#input
  — the `input` event's three results, including `{ action: "handled" }`; the
  primary evidence for the gap.
- https://github.com/earendil-works/pi/releases/tag/v0.84.3 — the tag the docs
  above were read at (2026-08-24); also adds the `powershell` built-in and
  `session_compact_failed`.
- https://github.com/earendil-works/pi/releases/tag/v0.84.2 — `defaultTools`,
  already implemented (#2687).
- https://github.com/earendil-works/pi/releases/tag/v0.84.1 — `terminate` on a
  blocked `tool_call`, deliberately left unset (#2636).
- https://github.com/earendil-works/pi/releases/tag/v0.84.0 —
  `AGENTS.override.md`, already implemented (#2601 gap 2).
- https://github.com/earendil-works/pi/releases/tag/v0.77.0 — adds
  `streamingBehavior` to `input`, establishing that the event predates this
  research window.
- https://github.com/dyoshikawa/rulesync/issues/2636 — the closed `preToolUse`
  observe-only issue; the same class of gap, and the pattern this follow-up
  copies.

dyoshikawa:
## Summary

Pi's `input` extension event is interceptable — a handler returning `{ action: "handled" }` skips the agent entirely — but the generated Pi extension treated every event except `tool_call` as observe-only. A canonical `beforeSubmitPrompt` hook authored for `pi` therefore could not cancel the prompt: the command's non-zero exit rejected inside the handler and the prompt still reached the model, while the identical hook blocks on Claude Code (`UserPromptSubmit`), Cline, Goose, Qwen Code and Reasonix.

This is the same class of gap already fixed for `preToolUse` in #2636.

## Changes

- `src/features/hooks/pi-extension-generator.ts`
  - Added `PI_PROMPT_BLOCKING_EVENT = "input"` alongside `PI_BLOCKING_EVENT = "tool_call"`, and replaced the boolean `blocksToolCall` flag with a `BlockingMode` of `"none" | "tool" | "prompt"`.
  - `input` handlers now wrap each command in the existing `try`/`catch` + `toBlockReason` shape and return `{ action: "handled" }` on a non-zero exit, `{ action: "continue" }` otherwise.
  - `handled` carries no `reason` field, so the failure text is surfaced through `ctx.ui.notify(reason, "error")`, guarded by `ctx.hasUI` because that channel is a no-op in print (`-p`) and JSON modes. The handler signature becomes `async (_event, ctx) => { … }` since `ctx` is the second argument.
  - The `toBlockReason` helper is emitted when either blocking event is present.
- `docs/reference/file-formats.md` — corrected the sentence that asserted `tool_call` was Pi's only blocking event; it is Pi's only *tool* gate, while `input` gates prompt submission. Also added `beforeSubmitPrompt` → `input` to the Pi mapping list. `src/generated/docs-content.ts` regenerated.
- `src/features/hooks/pi-hooks.test.ts` — new tests covering the emitted shape, plus two that load the generated module through `tsImport` and invoke the `input` handler for real: a failing command returns `{ action: "handled" }` and notifies, a succeeding one returns `{ action: "continue" }` and does not.

`transform` is deliberately out of scope: the canonical hook model has no prompt-rewriting contract to map it from.

## Verification

- Upstream contract confirmed at [`extensions.md#input` @ v0.84.3](https://github.com/earendil-works/pi/blob/v0.84.3/packages/coding-agent/docs/extensions.md#input): the three results are `continue`, `transform`, and `handled` ("skip agent entirely (first handler to return this wins)"), and handlers receive `(event, ctx)`.
- `pnpm cicheck` passes (9078 tests).

Closes #2772


dyoshikawa:
@dyoshikawa Thank you!

PR #2774, opened 2026-08-26
feat(subagents): write agentsmd subagents to the cross-vendor .agents/agents/ root

### LINKED ISSUE #2771: Follow up AGENTS.md upstream updates: simulated subagents write to .agents/subagents/, a path no client reads
## Summary

`agentsmd`'s simulated subagents are generated into `.agents/subagents/`, a
directory no documented client scans. Within the shared cross-vendor `.agents/`
namespace the subagent root clients actually read is `.agents/agents/`. Every
other simulated `agentsmd` root maps onto a real client; this one maps onto
nothing, so subagents generated for `agentsmd` alone are silently invisible.

## Recent Releases

The AGENTS.md convention itself has not moved: `agentsmd/agents.md`
(`openai/agents.md` redirects there) has no releases or tags ever, and its last
commit on the default branch is `d1ac7f06` (2026-03-12). Every open item is an
unmerged proposal (#135 optional frontmatter, #71 a `.agent` directory, #232
`agentaccess.txt`). So this gap is not driven by a spec change — it is a
standing mismatch surfaced by reviewing where the simulated roots point.

What has moved is the surrounding ecosystem's use of `.agents/`:

- Google Antigravity documents workspace subagents at `.agents/agents/<name>.md`
  or `.agents/agents/<name>/agent.md` —
  [Antigravity subagents docs](https://antigravity.google/docs/subagents)
  (fetched 2026-08-25).
- JetBrains Junie CLI scans the project `.agents/` and user `~/.agents/` trees
  for subagent files alongside `.junie/agents/` —
  [Junie CLI subagents docs](https://junie.jetbrains.com/docs/junie-cli-subagents.html)
  (fetched 2026-08-25).
- Kimi Code reads the same shared root; rulesync already encodes it as
  `KIMI_CODE_SHARED_AGENTS_DIR_PATH = join(".agents", "agents")`
  (`src/constants/kimi-code-paths.ts:12`).

### subagents (`simulated`) — output root points at a directory no client reads

`AGENTSMD_SUBAGENTS_DIR_PATH = join(AGENTSMD_DIR, "subagents")`
(`src/constants/agentsmd-paths.ts:7`) is the only definition of that path in the
tree, and `AgentsmdSubagent` (`src/features/subagents/agentsmd-subagent.ts:15`)
is its only writer and only reader. A dry run
(`pnpm run dev generate --targets agentsmd --features "*" --dry-run`) writes
`.agents/subagents/{code-reviewer,diff-analyzer,pr-handler,pr-merger,security-reviewer}.md`.
Nothing in this repository, and nothing in any documented client, scans that
directory.

Compare the neighbouring simulated roots, which all land on a surface a real
client reads:

| Root | Written by | Read by a real client? |
| --- | --- | --- |
| `.agents/memories/` | `agentsmd`, `amp` | yes — Amp |
| `.agents/commands/` | `agentsmd`, imported by `augmentcode` | yes — Auggie's cross-tool command root |
| `.agents/skills/` | `agentsmd`, `agentsskills`, and others | yes — the Agent Skills standard |
| `.agents/subagents/` | `agentsmd` only | **no** |
| `.agents/agents/` | `kimicode`; read by Antigravity, Junie, Kimi Code | yes |

`AgentsmdSkill` carries an explicit doc comment justifying `.agents/skills/` as
"not an AGENTS.md convention … the Agent Skills standard's project location".
`AgentsmdSubagent` carries no such justification.

**User-visible effect.** A user who enables only `agentsmd` and generates
subagents gets five files that every AGENTS.md-era client ignores. One directory
over, in `.agents/agents/`, Antigravity, Junie and Kimi Code would each have
discovered them without those targets being enabled.

## Proposed Follow-up

- Repoint `AGENTSMD_SUBAGENTS_DIR_PATH` at `join(".agents", "agents")`, matching
  the cross-vendor root the other simulated `agentsmd` surfaces already track.
- Keep `.agents/subagents/` as an import-only root, plus a `--delete` sweep entry
  for one release, so existing trees are migrated rather than orphaned.
- Because `.agents/agents/` is shared with `kimicode`, the two writers must agree
  on the emitted frontmatter. `AgentsmdSkill` already performs exactly this
  coordination for `.agents/skills/` via `toSpecConformantAgentSkillFields`, so
  there is a working pattern to copy.
- Alternative, if generation should stay where it is: document
  `.agents/subagents/` as deliberately rulesync-private (the way
  `goose-rule.ts:49-54` records the `~/.agents/AGENTS.md` decision) and add
  `.agents/agents/` as an import root so foreign subagents are at least readable.

Scope note: `agentsskills` was researched in the same pass and has no gaps. Both
gaps tracked by #2731 have shipped (hidden companion files are carried, and
`src/utils/frontmatter.ts:288` implements the malformed-YAML requote retry), and
that issue is closed.

## References

- https://agents.md/ — the AGENTS.md convention; defines no subagent surface, so
  the simulated path is rulesync's own choice.
- https://github.com/agentsmd/agents.md — spec repository; no releases or tags,
  last commit `d1ac7f06` (2026-03-12), confirming the convention is unchanged.
- https://antigravity.google/docs/subagents — documents `.agents/agents/` as the
  workspace subagent root.
- https://junie.jetbrains.com/docs/junie-cli-subagents.html — Junie CLI scans the
  project `.agents/` and user `~/.agents/` trees for subagents.
- https://github.com/dyoshikawa/rulesync/issues/2731 — the previous run for these
  two targets; concluded `agentsmd` had no gaps, having checked only the spec
  text, not where the simulated roots point.
- https://github.com/dyoshikawa/rulesync/issues/2747 — the empty imported skill
  description item, already tracked there and deliberately not repeated here.

dyoshikawa:
## Summary

The simulated `agentsmd` subagent writer emitted `.agents/subagents/`, a directory no documented AGENTS.md-era client scans, so every file it produced was inert. This repoints it at `.agents/agents/` — the cross-vendor root Antigravity discovers workspace agents in, and the one Kimi Code reads alongside its own agents tree.

Because the native `antigravity-*` targets already own that exact path, the simulated writer would otherwise overwrite their output with a bare `name`/`description` pair and drop `tools`, `model`, `commandExecutionPolicy` and friends. So it now builds its frontmatter with the same shared helper the Antigravity targets use (`toAntigravitySubagentFrontmatter`, extracted from `AntigravitySharedSubagent`) and the same serialization — whichever target runs last, the file on disk is identical. This mirrors the arrangement `AgentsmdSkill` / `toSpecConformantAgentSkillFields` already gives `.agents/skills/`.

Stale outputs under the old `.agents/subagents/` path are not swept automatically; they stay gitignored through a `HAND_MAINTAINED_GITIGNORE_ENTRIES` entry, following the `.goose/recipes/subagents/` precedent.

## Changes

- `src/constants/agentsmd-paths.ts`: `AGENTSMD_SUBAGENTS_DIR_PATH` → `.agents/agents`.
- `src/features/subagents/antigravity-shared-subagent.ts`: extract `toAntigravitySubagentFrontmatter` and export `ANTIGRAVITY_SHARED_SUBAGENT_SECTION_KEYS`.
- `src/features/subagents/agentsmd-subagent.ts`: build frontmatter via the shared helper and pass explicit `fileContent` matching Antigravity's serialization.
- `src/features/subagents/simulated-subagent.ts`: accept an optional pre-rendered `fileContent`.
- `src/cli/commands/gitignore-entries.ts`: keep `**/.agents/subagents/` gitignored as a hand-maintained legacy entry.
- `docs/reference/file-formats.md`: add an `.agents/agents/` ownership note covering the shared path and the migration.
- `src/types/feature-processor.ts` / `src/types/dir-feature-processor.ts`: report orphan deletions on real `--delete` runs, not only in dry-run mode, and strip terminal control characters from the logged path. `.agents/agents/` is shared with other vendors, so a sweep there must be visible; the log line is the only runtime signal a user gets.

## Tests

- `agentsmd-subagent.test.ts`: path assertions updated, plus new tests asserting the `agentsmd` and `antigravity-cli` writers produce byte-identical file content for the same canonical subagent, that they resolve to the same directory, and that an invalid or scalar Antigravity block fails generation instead of writing a reduced file.
- `simulated-subagent.test.ts`: pin the shared-path contract in the base class — pre-rendered bytes are kept, and frontmatter keys a native target owns survive parsing.
- `feature-processor.test.ts` / `dir-feature-processor.test.ts`: cover the real-run and dry-run deletion output and the control-character stripping.
- `pnpm cicheck` passes (9103 tests).

Closes #2771


dyoshikawa:
## Security review follow-ups

Each finding was re-verified locally against `origin/main` before deciding.

### #1 — `--delete` sweeps hand-authored files under `.agents/agents/` — *rejected, pre-existing*

Verified on `origin/main`: a hand-authored `.agents/agents/hand-authored.md` is
already deleted by `generate --targets antigravity-cli --features subagents --delete`.
`.agents/agents/` is a generated, orphan-swept path owned by the Antigravity
targets; this PR adds one more target that triggers the existing sweep, it does
not change the directory's contract.

Exempting only the *simulated* writer via `isFileOwned` would be inconsistent
(the native targets would keep sweeping) and would give false reassurance in the
common `--targets "*"` case. The behavior is documented instead, in the
`.agents/agents/` ownership note in `docs/reference/file-formats.md`. Filed as a
possible follow-up direction in #2775.

### #2 — same-run mutual deletion between targets sharing the directory — *rejected, pre-existing → #2775*

Reproduced identically on `origin/main` with `--targets antigravity-ide,antigravity-cli`:
two subagents are reported as written, one survives. This is a shared-output-path
bug in the per-target write-then-sweep, not something this PR introduces, and
fixing it properly means changing orphan sweeping across all features. Split out
as #2775 with both reproductions rather than widening this PR's scope.

### #3 — Antigravity blocks emitted even for an `agentsmd`-only generate — *addressed*

Made explicit as a design decision in the ownership note (70048cce2): the file is
shared, so emitting a reduced file would be exactly the silent degradation the
shared block exists to prevent. The note now also states the consequence — a
subagent pulled in with `rulesync fetch` becomes a live agent definition once any
writer of this path runs — and says to review `.rulesync/subagents/*.md` after
fetching. This matches the reviewer's own finding that `fetch` uses
`includeSimulated: false`, so the remote → `.agents/agents/` direction stays closed.

### #4 — an invalid Antigravity block fails `agentsmd` generation — *addressed as intended behavior*

This is deliberate and now documented at the call site, following
`AgentsmdSkill.reportSpecViolations` ("same file, same diagnostics"): generating
for the simulated target alone must surface the errors the native target would
have surfaced, rather than dropping the block and writing a reduced file. Locked
with a regression test.

`pnpm cicheck` passes (9096 tests); the e2e subagents spec passes separately.

dyoshikawa:
## Follow-up on the residual part of finding #1

The re-review at `07ec9a5da` narrowed #1 to one thing: the documentation now
states that `--delete` sweeps `.agents/agents/`, but a real run still gives no
runtime signal at all. That turned out to be a plain asymmetry rather than
anything specific to this directory:

- `--dry-run` printed `[DRY RUN] Would delete: <path>` for every file and
  directory it would remove.
- A real `--delete` run printed nothing. `removeOrphanAiFiles` /
  `removeOrphanAiDirs` returned a count that only fed `hasDiff` and was never
  surfaced anywhere (`src/lib/generate.ts:96,122,148`).

Fixed in 588f5c408 by making the real branch symmetric with the dry-run branch.
The reproduction from the review now reads:

```
$ rulesync generate --targets agentsmd --features subagents --simulate-subagents --delete
Deleted: /…/.agents/agents/hand.md
Written 1 subagent
    .agents/agents/gen.md
```

This applies to every feature and target, not just this directory, which is why
it is a one-line symmetry fix rather than an `isFileOwned` predicate on
`AgentsmdSubagent`: exempting only the simulated writer would leave the native
Antigravity targets sweeping the same directory, and would report nothing when
they did. Narrowing *what* gets swept in shared discovery roots stays on the
follow-up issue #2775, which now has this as context.

`pnpm cicheck` passes (9096 tests).

dyoshikawa:
## Round-2 security review: 0 mid-or-above, 2 low — both fixed in 11d414896

### low #1 — the new deletion log did not strip control characters

Correct, and a defect I introduced in 588f5c408. The path is an on-disk name
rulesync did not choose, and every other untrusted string
`src/types/feature-processor.ts` logs already goes through
`stripControlCharacters`. A filename carrying an erase-line escape could rewrite
the line and hide what was deleted — defeating the exact purpose of that commit.

Fixed, and the adjacent `[DRY RUN] Would delete:` lines (pre-existing) now go
through the same helper. Verified with the reviewer's own filename:

```
Deleted: /…/.agents/agents/a[2KDeleted: nothing-important.md[1;32m OK.md
```

The `ESC` and `CR` bytes are gone; what remains is inert literal text.

### low #2 — a scalar `antigravity-*` section spread into index keys

Also correct, and pre-existing on `main` as the reviewer notes. `Object.assign`
spreads a string into `'0': a, '1': b, '2': c`, and the loose schema carried it
into the generated file as junk frontmatter.

Now rejected before the merge, on the same fail-closed path as an invalid section
value, with a message naming the offending section:

```
Invalid agentsmd subagent frontmatter in scalar-section.md: 'antigravity-cli' must be a table of frontmatter keys.
```

Guarded with `isPlainObject` (which also rejects a non-`Object.prototype`
prototype) and locked with a regression test.

`pnpm cicheck` passes (9097 tests).

dyoshikawa:
@dyoshikawa Thank you!

Issue #2740, opened 2026-08-24
Orphan deletion compares paths exactly, so a case-only rename deletes the just-generated output on macOS/Windows


dyoshikawa:
## Summary

Orphan deletion compares paths **exactly**:

- `FeatureProcessor.removeOrphanAiFiles()` — `src/types/feature-processor.ts:154`
- `DirFeatureProcessor.removeOrphanAiDirs()` — `src/types/dir-feature-processor.ts:201`

Both build a `Set` of the generated paths and delete every existing path that is not in it.

## Why it matters

On a case-insensitive filesystem (macOS, Windows), two spellings of one name are the same file. Rename a skill from `my-skill` to `My-Skill` in `.rulesync/skills/` and generate:

1. `writeAiDirs()` writes `.claude/skills/My-Skill/SKILL.md` — which lands in the existing `my-skill` directory, because the OS resolves the two to one.
2. `removeOrphanAiDirs()` then sees the existing `.claude/skills/my-skill`, does not find that exact string among the generated paths (`.claude/skills/My-Skill`), classifies it as an orphan, and calls `removeDirectory()` on it.

The directory it deletes is the one that was just written, so the freshly generated skill disappears. The same reasoning applies to `removeOrphanAiFiles()` for single-file features. On Linux this is harmless, which is why it goes unnoticed in CI (unit tests run on ubuntu only).

## Suggested direction

Compare with the same case folding the loaders now use for cross-root de-duplication (`caseFoldIdentity()` in `src/types/feature-processor.ts`, added in #2738): a path that case-folds to a generated path is not an orphan. Unicode normalization matters for the same reason it does there.

This was split out of the review of #2733 / PR #2738 as an independent, pre-existing problem.

PR #2776, opened 2026-08-26
fix(generate): stop targets sharing an output directory from deleting each other's files

### LINKED ISSUE #2775: generate --delete: targets sharing an output directory delete each other's files
## Summary

When two targets share a single output directory, `generate --delete` makes them
delete each other's files within the same run. Each target writes its own files
and then sweeps orphans in the shared directory using only its *own* expected
outputs, so a file written moments earlier by the sibling target looks like an
orphan.

The user-visible symptom is that the CLI reports writing N files but fewer than N
survive on disk, with no warning.

## Reproduction

```bash
mkdir -p repro/.rulesync/subagents && cd repro

cat > .rulesync/subagents/cli-only.md <<'MD'
---
targets: ["antigravity-cli"]
name: cli-only
description: CLI only agent
---
CLI body.
MD

cat > .rulesync/subagents/ide-only.md <<'MD'
---
targets: ["antigravity-ide"]
name: ide-only
description: IDE only agent
---
IDE body.
MD

npx rulesync generate --targets antigravity-ide,antigravity-cli --features subagents --delete
ls .agents/agents/
```

Observed on `main`:

```
Written 2 subagents
    .agents/agents/ide-only.md
    .agents/agents/cli-only.md
🎉 All done! Written 2 file(s) total (2 subagents)

$ ls .agents/agents/
cli-only.md
```

`ide-only.md` is written and then deleted by `antigravity-cli`'s sweep. Reversing
the `--targets` order reverses which file survives.

## A second, related symptom

The same per-target sweep also removes hand-authored files from the shared
directory:

```bash
mkdir -p repro2/.rulesync/subagents repro2/.agents/agents && cd repro2
printf -- '---\nname: hand-authored\ndescription: written by a human\n---\nBody.\n' \
  > .agents/agents/hand-authored.md
# plus any subagent targeted at "*"
npx rulesync generate --targets antigravity-cli --features subagents --delete
ls .agents/agents/   # hand-authored.md is gone
```

This one is arguably intended — `.agents/agents/` is a generated path, and
`--delete` is documented as sweeping orphans there — but it is worth deciding
explicitly, because the directory is a cross-vendor discovery root rather than a
rulesync-private one.

## Where it comes from

The write-then-sweep is per target (`src/cli/commands/generate.ts`), and the
orphan enumeration in `src/features/subagents/subagents-processor.ts` /
`src/features/feature-processor.ts` compares the directory contents against the
outputs of the single target currently being processed. Nothing carries the set
of files the other targets in the same run just wrote.

## Scope

This is not specific to subagents or to Antigravity — it applies to any pair of
targets that resolve to the same `getSettablePaths().relativeDirPath`. Other
shared roots (`.agents/skills/`, for example) should be checked against the same
reproduction.

## Possible directions

- Accumulate the files written across all targets in a run and exclude that set
  from every target's orphan sweep (fixes the mutual-deletion case directly).
- Or make orphan sweeping a single pass after all targets have been written,
  rather than a per-target step.
- Independently, consider an `isFileOwned` predicate for shared discovery roots
  so hand-authored files are not swept (only `reasonix-subagent.ts` implements
  one today).

## Context

Found while reviewing #2774, which adds `agentsmd` as a third writer of
`.agents/agents/`. That PR does not introduce the bug — it reproduces
identically on `main` with the two Antigravity targets — but it does widen the
set of target combinations that can hit it.

dyoshikawa:
## Summary

`generate --delete` made targets that share one output directory delete each
other's files within a single run. Each target wrote its own files and then swept
orphans in that directory using only its *own* expected outputs, so a file the
sibling target had written moments earlier looked exactly like a leftover from a
previous run. The CLI reported writing N files while fewer than N survived on
disk, with no warning.

Reproduced on `main` with two Antigravity subagents pinned to different targets
(both write `.agents/agents/`):

```
Written 2 subagents
    .agents/agents/ide-only.md
    .agents/agents/cli-only.md

$ ls .agents/agents/
cli-only.md
```

Reversing `--targets` reversed which file survived.

## Fix

A run-scoped `OrphanSweepPlan` (`src/lib/orphan-sweep.ts`), created once per
`generate()` and threaded through every feature step:

- **Every path the run writes is claimed.** Each step registers the paths it
  writes — across all targets and all features — and a sweep skips any path the
  run itself produced. That is what stops one writer of a shared directory from
  treating a sibling's fresh output as an orphan.
- **Sweeps are deferred to the end of the run.** Claiming alone would still
  depend on target order: the first target sweeps before the second has written
  anything. Holding every sweep back until the last step has written also removes
  the delete-then-immediately-rewrite churn that made `generate --check` report a
  permanently out-of-date tree on an already-synchronized one.

Paths are compared with `node:path`'s `resolve`, so the same file reached through
equivalent output roots matches. Case-only differences are deliberately not
normalized — that is a separate, filesystem-dependent concern (#2740).

What is swept is unchanged: a file in a generated directory that no `.rulesync/`
source produces is still removed, including the hand-authored case the issue's
second symptom describes (`.agents/agents/` is a generated path, and the
`file-formats.md` note already says so).

## Tests

- `src/lib/orphan-sweep.test.ts` — new unit tests for claiming, path
  normalization, deferral ordering, diff reporting, and single-drain semantics.
- `src/lib/generate.test.ts` — a fake shared `.agents/agents/` written by two
  targets: both files survive, a genuine orphan is still deleted, and an
  already-synchronized tree reports `hasDiff: false`.
- `src/e2e/e2e-subagents.spec.ts` — the issue's reproduction end to end
  (`antigravity-ide,antigravity-cli` with `--delete`), plus a follow-up
  `--check` run that must stay clean, and a genuine-orphan case.

Each of these was mutation-checked: removing the claim filter fails 3 unit tests
and the e2e reproduction; running the sweeps inline instead of deferring them
fails the `--check` stability test.

`pnpm cicheck` passes (9122 tests).

## Docs

- `docs/reference/cli-commands.md` — a note on `--delete` and shared output
  directories stating the new guarantee.
- `docs/reference/file-formats.md` — replaces the `.agents/agents/` ownership
  note's warning that a pinned subagent gets swept by the next writer, which no
  longer holds.

Closes #2775

---

## Review follow-ups

- `registerGeneratedTree` claims a directory feature's output recursively.
  Deferring the sweeps means a *file* feature's sweep now runs after the skills
  step has written, so claiming only the directory path would leave `SKILL.md`
  and its companions exposed. No file feature scans a skills tree today
  (`skipToolFileScan` and `ReasonixSubagent.isFileOwned` cover the two
  overlaps), so the ancestor walk is defense in depth; it is covered by unit
  tests rather than by the e2e pair, which exercises `rejectClaimed`.
- `AiDir` now rejects `"."` and `".."` as a directory name. `path.join`
  normalizes both away — `"."` collapses the directory onto its parent root,
  `".."` onto that root's parent — and neither trips `getDirPath()`'s traversal
  guard, because both stay inside `outputRoot`. Several skills targets take
  that name straight from a skill's frontmatter `name`, which `rulesync fetch`
  can pull from a third-party repository. On its own it misplaced `SKILL.md`;
  combined with the sweep plan it claimed the collapsed tree and silently
  disabled `--delete` for everything under it, so a stale `.roo/rules/`
  instruction file would survive a run that should have removed it.
- A directory feature is only granted the tree claim when its path really nests
  under its own name (`AiDir.ownsDirTree()`). `TaktSkill` overrides
  `getDirPath()` to drop `dirName` and return the shared root every takt skill
  flattens into, so claiming that root as a tree would exempt every sibling
  under it from the sweep — the exact failure this PR exists to prevent. The
  comparison is structural rather than a match on the last path segment,
  because a takt skill named after its own facet root ends on the same segment
  without ever nesting inside it. The files a directory feature writes are
  claimed by name as well, so a feature that gets no tree claim still has its
  own output protected.
- A skill's frontmatter `name` is now rejected when it cannot be used as a
  directory name, at the schema rather than only in `AiDir`, so the error names
  the offending `SKILL.md` instead of just the name. That also covers the
  trailing dot or space that Windows strips from a path component, which would
  make the directory on disk differ from the path the run claims.
- Reviewing this PR surfaced a pre-existing bug that is **not** addressed here:
  with zero takt skills generated, the orphan sweep deletes the whole
  `.takt/facets/knowledge` directory, because every candidate under it reports
  the shared root as its path. Reproduced against `main` and filed as #2777.


dyoshikawa:
@dyoshikawa Thank you!

PR #2782, opened 2026-08-26
feat(fetch): start the interactive skill prompt with nothing selected

dyoshikawa:
## Summary

`rulesync fetch --interactive` used to open its checkbox prompt with **every** skill checked whenever `--skills` was not given. Pressing `<enter>` without reading the list therefore fetched every skill the source repository publishes, writing files into the user's project. This flips the default: nothing is selected when the prompt opens, and the user opts in.

To keep "fetch everything" a single keystroke, the prompt's shortcuts are now declared explicitly instead of relying on the `@inquirer/checkbox` defaults:

- `<space>` — toggle the highlighted skill
- `<a>` — select all / deselect all
- `<i>` — invert the selection
- `<enter>` — confirm

The bindings already existed as library defaults and are listed in the prompt's own help line; pinning them in `SKILL_PROMPT_SHORTCUTS` keeps them stable across `@inquirer/checkbox` upgrades, which matters more now that an empty selection is the starting point. The prompt message also names the select-all key directly, since that is the shortcut users reach for first.

`--interactive --skills pdf` is unchanged: the named skills start checked and everything else starts unchecked.

## Changes

- `src/lib/skill-prompt.ts` — `checked` now follows `preselectedSkills` alone; added `SKILL_PROMPT_SHORTCUTS` and the select-all hint in the prompt message.
- `src/cli/program.ts`, `docs/reference/cli-commands.md`, `docs/guide/official-skills.md` — document the new default and the key bindings.
- `src/lib/skill-prompt.test.ts` — the all-checked test becomes a nothing-checked test, plus a new test asserting the select-all shortcut is offered.

## Test plan

- `pnpm cicheck` — 380 files / 9223 tests pass, content checks clean.


dyoshikawa:
@dyoshikawa Thank you!

PR #2784, opened 2026-08-26
fix(generate): stop --delete from removing the shared takt facet root

### LINKED ISSUE #2777: generate --delete removes the whole .takt/facets/knowledge directory when no takt skill is generated
## Summary

`generate --targets takt --features skills --delete` recursively deletes the entire `.takt/facets/knowledge` directory, including files a user hand-authored there, whenever the run generates no takt skill.

## Cause

`SkillsProcessor.loadToolDirsToDelete()` enumerates the subdirectories of each skills search root and turns each one into a tool skill via `forDeletion({ relativeDirPath: root, dirName })`. It then hands those to `removeOrphanAiDirs()`, which deletes any whose `getDirPath()` is not among the generated ones.

`TaktSkill` overrides `getDirPath()` to drop `dirName`, because takt skills are emitted as flat files directly under `.takt/facets/knowledge` rather than nested in a directory of their own (`src/features/skills/takt-skill.ts:101`). Every enumerated subdirectory therefore reports the *shared root* as its path. When the run generates zero takt skills, that root is not in the keep set, so the sweep deletes the root itself — with everything under it.

The same override means the deletion path can never target an individual takt skill: with at least one takt skill generated, the root is always in the keep set and nothing is ever swept.

## Reproduction

```console
$ mkdir -p repro/.takt/facets/knowledge/my-notes repro/.rulesync/skills
$ echo "hand-authored" > repro/.takt/facets/knowledge/my-notes/notes.md
$ cd repro
$ npx rulesync generate --targets takt --features skills --delete
Deleted directory: /.../repro/.takt/facets/knowledge
✓ All files are up to date (skills)
$ find .takt
.takt
.takt/facets
```

`my-notes/notes.md` is gone.

## Notes

Reproduced on `main` at e80b285e7, so this is not a regression from a recent change. Found while reviewing #2776; `AiDir.ownsDirTree()`, added there, already recognizes that `TaktSkill` does not own the directory its path points at, but the deletion path in `SkillsProcessor` does not yet consult anything equivalent.

A fix probably needs the flat-file layout represented in the deletion path itself — either a per-tool hook that says "this tool has no per-skill directory, so it has no orphan directories to sweep", or a file-based sweep for takt keyed on the emitted `<name>.md` files.

dyoshikawa:
## Summary

`generate --targets takt --features skills --delete` recursively deleted the whole `.takt/facets/knowledge` directory — including files the user had authored there by hand — whenever the run generated no takt skill.

`TaktSkill` overrides `getDirPath()` to drop `dirName`, because takt skills are emitted as flat files directly under the shared facet root rather than nested in a directory of their own. Every candidate enumerated by `SkillsProcessor.loadToolDirsToDelete()` therefore reported that shared root as its path, and with zero takt skills generated the root was in no keep set, so `removeOrphanAiDirs()` deleted the root itself.

## Change

`DirFeatureProcessor.removeOrphanAiDirs()` now skips any candidate for which `AiDir.ownsDirTree()` is false. A directory a candidate does not own cannot be an orphan of that candidate, because deleting it would take every sibling in it as well.

Nothing reachable is lost by the skip: with at least one takt skill generated the root was always claimed and never swept, and with none it must not be touched — so the deletion path could never target an individual takt skill in the first place. Every other tool nests under `dirName`, so `ownsDirTree()` is true for it and behavior is unchanged.

The guard lives in the base class rather than in `SkillsProcessor`. `SkillsProcessor` is currently the only `DirFeatureProcessor` — the commands and subagents processors extend `FeatureProcessor` and never call `removeOrphanAiDirs` — so this changes nothing today; it puts the invariant where any future dir-feature processor inherits it.

## Tests

Three regression tests, each verified to fail without the guard:

- `dir-feature-processor.test.ts` — a candidate pointing at a shared root is never removed, and is not even announced under `--dry-run` (that line is what a user reads to decide whether `--delete` is safe).
- `skills-processor.test.ts` — the reported scenario end to end: a hand-authored `.takt/facets/knowledge/my-notes/notes.md` survives a sweep that generates no takt skill.

## Known remaining gap

Orphaned takt skill *files* are still never swept, since takt contributes no directory-level candidates. That is pre-existing behavior — the root has always been claimed whenever any takt skill was generated — and a file-keyed sweep for flat-file tools is a separate change. This PR documents the consequence, in `docs/tools/takt.md` and in the `--delete` note in `docs/reference/cli-commands.md`, so a user knows to delete a stale knowledge file themselves.

Closes #2777


dyoshikawa:
@dyoshikawa Thank you!

PR #2788, opened 2026-08-26
feat(mcp): let musecode author the documented per-server mode key

### LINKED ISSUE #2769: Follow up Meta Muse Code upstream updates: mcp mode key dropped on generate, hooks and permissions blockers re-confirmed
## Summary

The documented MCP `mode` key cannot be authored for `musecode` and is silently stripped whenever a previously imported `settings.json` is regenerated. The hooks and permissions blockers recorded in #2669 and #2727 are also re-confirmed as still blocked, now with the specific evidence needed to unblock them.

## Recent Releases

Meta publishes **no changelog, no release-notes page, no npm package and no public repository** for Muse Code — install is `curl -fsSL https://dev.meta.ai/install.sh | sh` and the docs at `https://dev.meta.ai/docs/muse-code*` are unversioned. Every claim below is therefore "as served on 2026-08-25".

- Product announcement, **2026-08-05**: https://research.meta.ai/blog/introducing-muse-code-and-muse-spark-1-2 — Muse Code enters public beta alongside Muse Spark 1.2.

Because the product is 20 days old, its entire lifetime falls inside the research window; there are no upstream deprecations or renames to chase, only surfaces rulesync has not yet mapped.

## Gaps

**1. mcp — the documented `mode` key is unauthorable and silently dropped on regenerate (`global`).**

https://dev.meta.ai/docs/muse-code/extending.md documents a per-server `mode` key: `required` (default) or `optional`, where `optional` makes Muse Code skip an unavailable server with a warning instead of failing.

`convertToMusecodeFormat` in `src/features/mcp/musecode-mcp.ts` builds each entry from a fixed key set, so `mode` never reaches the output — while `convertFromMusecodeFormat` passes it through on import. A user who imports a `settings.json` containing `"mode": "optional"` and then regenerates loses the key, with no warning.

**2. hooks — `unsupported`; #2669's blocker is unchanged, but the surface has grown (no action yet).**

https://dev.meta.ai/docs/muse-code/extending.md#hooks now documents **three** hook sources — project `<project-root>/.muse/hooks.json` ("committed to the repo"), a `hooks` block inside `~/.config/muse/settings.json`, and a managed file via `managed_hooks_path` — and publishes 12 lifecycle events: `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PermissionRequest`, `PostToolUse`, `PreLLMCall`, `PostLLMCall`, `PreCompact`, `PostCompact`, `SubagentStart`, `SubagentStop`, `Stop`. `src/constants/musecode-paths.ts` records only the project `.muse/hooks.json` source and does not mention the settings-file `hooks` block or `managed_hooks_path` at all.

The blocker itself is unchanged: the per-hook **entry** schema (matcher fields, command shape, timeout, exit-code and decision semantics) is still unpublished. The `{ "event": "<EventName>", "stdin": {} }` shape in the docs is the fixture format for `muse hooks run --fixture`, not config. Also worth recording: `PermissionRequest`, `PreLLMCall`, `PostLLMCall` and `PostCompact` have no member in rulesync's `HookEvent` union.

**3. permissions — `unsupported`; #2727's blocker re-confirmed, schema still unpublished (no action yet).**

https://dev.meta.ai/docs/muse-code/permissions.md documents "Always allow in this workspace" writing a persistent argv-prefix rule scoped to one workspace root, with deny-over-allow precedence regardless of specificity, and interpreter prefixes (`python`, `bash`, `node`) barred from broad grants. https://dev.meta.ai/docs/cookbook/staged-approvals.md names the store `approval-policy.json` but publishes neither its full path nor its entry schema, and does not say whether it is committable project config or machine-local user state.

## Proposed Follow-up

- Add a canonical authoring key for MCP `mode` following the `kiroAutoApprove` / `envVars` precedent in `src/types/mcp.ts` (e.g. `musecodeMode`), strip it in `RulesyncMcp.getMcpServers()` so it cannot leak into other targets, emit it as `mode`, and map it back explicitly on import instead of relying on loose-object passthrough. Leave `framing` dropped until the docs define its purpose.
- For hooks, pin the entry schema from a real installation (`muse hooks list` / `muse hooks validate` output, or a `muse init`-seeded file) before authoring an adapter. When it lands, target the **user `hooks` block first**: `MusecodeMcp` already owns `~/.config/muse/settings.json` through the shared-config gateway, so a hooks patch is a second key in a file rulesync merges today — zero new file-ownership cost.
- For permissions, keep this **blocked, not closed**: pin the real path and schema of `approval-policy.json` from an installation and settle whether it is committable, before any adapter work. The existing note in `src/constants/musecode-paths.ts` is accurate and should not be relaxed into "Muse Code has no allow/deny file".

## Watch items (no confirmable upstream surface today)

- **rules, global scope.** https://dev.meta.ai/docs/muse-code/configuration.md states verbatim "Your machine-wide user rules always load" and ranks project rules above user rules, but never gives the user-rules path. Rulesync's project-only marker is the correct read of the published docs; re-check when a path appears, since the `~/.config/muse/` root is already known and the adapter would be trivial.
- **ignore.** No official page mentions `.museignore`, `.gitignore` filtering, or any context-exclusion file. The `.museignore` claim appears only in third-party blogs (sitepoint, codersera), which also assert a `~/.muse-code/config.json` that directly contradicts the documented `~/.config/muse/settings.json`. Both were treated as unreliable and should not be acted on without a primary source.
- **checks.** The `runtime_capabilities` map in `settings.json` toggles observer agents (including "Verification"), but its keys are not enumerated and it is a boolean feature switch, not authored check definitions like Cursor BugBot or Amp. Nothing for the `checks` dimension to map onto.

- **commands** — built-in slash commands only (`/new`, `/plan`, `/goal`, `/loop`, `/tasks`, …); no user-definable command surface. Correctly blank.
- **subagents** — runtime-only: `subagent_spawn` takes an inline `role`, `objective` and `worktree_isolation`; there are no named-subagent config files. Correctly blank.
- **skills** — `<repo>/.agents/skills/<skill-id>/SKILL.md`, user skills in `$XDG_CONFIG_HOME/muse/skills` and `~/.agents/skills`, plus a compat scan of `.codex/skills` and `.claude/skills`. Matches; no SKILL.md frontmatter fields beyond `name`/`description` are documented.
- **mcp, other keys** — `transport` (`stdio` / `streamable_http`), `command`/`args`/`env`, `url`/`headers` and `enabled` all match; `sse`/`ws` are correctly skipped with a warning; `framing` is listed as optional with its purpose undefined, so dropping it is acceptable.

Dry runs: project reports `ignore`, `commands`, `subagents`, `hooks`, `checks` and `permissions` unsupported and `mcp` global-only, resolving 1 root rule; global writes 42 files (1 MCP + 41 skills), with MCP landing at `.config/muse/settings.json`.

## References

- https://research.meta.ai/blog/introducing-muse-code-and-muse-spark-1-2 — 2026-08-05 announcement; the only dated primary source, used to bound the research window.
- https://dev.meta.ai/docs/muse-code/extending.md — MCP `mode` (gap 1) and the three hook sources plus the 12 events (gap 2).
- https://dev.meta.ai/docs/muse-code/permissions.md — argv-prefix allow/deny semantics (gap 3).
- https://dev.meta.ai/docs/cookbook/staged-approvals.md — names `approval-policy.json` without a path or schema (gap 3).
- https://dev.meta.ai/docs/muse-code/configuration.md — "machine-wide user rules always load", the global-rules watch item.
- https://dev.meta.ai/docs/muse-code.md and https://dev.meta.ai/docs/cookbook/muse-code.md — install method and the absence of any changelog or repository.
- https://dev.meta.ai/docs/muse-code/interactive.md — built-in slash command list, cited for why `commands` stays blank.
- https://dev.meta.ai/docs/cookbook/subagent-fanout.md — `subagent_spawn`, cited for why `subagents` stays blank.
- #2669 (hooks) and #2727 (permissions) — the closed issues whose blockers gaps 2 and 3 re-confirm.

dyoshikawa:
## Summary

Meta Muse Code documents a per-server MCP `mode` key — `required` (the default) aborts the whole run when the server fails to start, `optional` makes Muse Code skip an unavailable server with a warning — but Rulesync had no way to author it, and silently dropped it on regenerate.

`convertToMusecodeFormat` builds each `mcp_servers` entry from a fixed key set (`transport`, `url`/`headers` or `command`/`args`/`env`, `enabled`), so `mode` never reached the output. `convertFromMusecodeFormat` passed it through on import as an unknown key. A user who imported a `settings.json` containing `"mode": "optional"` and then regenerated lost the key, with no warning — turning a server Muse Code was told to skip back into one that aborts the run.

## Changes

- **`src/types/mcp.ts`** — add the authoring key `musecodeMode`, typed as the two documented values. Unlike `experimentalEnvironment` (kept a loose string so a value Codex adds later is not rejected), `mode` is what decides abort-vs-skip, so an unrecognized value is the one case where naming the file that holds the typo beats passing it along.
- **`src/features/mcp/rulesync-mcp.ts`** — strip `musecodeMode` in `getMcpServers()`, like `envVars`, so it cannot leak into any other tool's generated config.
- **`src/features/mcp/musecode-mcp.ts`** — `fromRulesyncMcp` re-merges `musecodeMode` from the unfiltered source JSON (the same re-merge `codexcli-mcp.ts` does for `envVars`) and emits it as `mode`. On import, a `mode` of `required` or `optional` is lifted into `musecodeMode`; any other value is left as a plain `mode` key rather than renamed, because `musecodeMode` is typed as those two values and moving an unrecognized one there would produce a `.rulesync/mcp.json` the next parse rejects (the `kiroAutoApprove` precedent in `kiro-mcp.ts`).
- **`docs/reference/file-formats.md`** — document the key in the Meta Muse Code MCP note; `src/generated/docs-content.ts` regenerated.

An authored `"required"` is written out rather than dropped as redundant with Muse Code's default: dropping it would silently rewrite a file the user imported.

## Scope

Only gap 1 of the issue. Gaps 2 (hooks) and 3 (permissions) are explicitly recorded there as blocked on unpublished schemas with no action to take yet; a follow-up issue carries their evidence forward.

## Verification

- All three new tests were confirmed to fail without the fix (emit path neutered → the generate and round-trip tests fail; import mapping neutered → the import and round-trip tests fail).
- `pnpm cicheck` green (9233 tests).
- Upstream `mode` semantics re-confirmed against the primary source on 2026-08-26: https://dev.meta.ai/docs/muse-code/extending.md

Closes #2769


dyoshikawa:
@dyoshikawa Thank you!

PR #2791, opened 2026-08-26
feat(permissions): let the roo target author its own roo-cline command lists

### LINKED ISSUE #2768: Follow up Zoo Code upstream updates: roo-cline.allowedCommands/deniedCommands unauthorable for the roo alias
## Summary

Zoo Code v3.54.0 added workspace-settable command allow/deny lists (`roo-cline.allowedCommands` / `roo-cline.deniedCommands`) that ship under the **legacy `roo-cline.*` VS Code settings namespace**, which the deprecated `roo` alias target still owns. `zoocode` already has a `permissions` adapter for the equivalent modern namespace; `roo` has none, so the `permissions` column for `roo` is blank even though the underlying extension reads these keys from `.vscode/settings.json`. Everything else about the live `zoocode` target is up to date through v3.80.0 — this issue exists only for the alias's missing surface.

## Recent Releases

- Zoo Code **v3.80.0** (2026-08-22) — latest release at the time of research; https://github.com/Zoo-Code-Org/Zoo-Code/releases
- Zoo Code **v3.54.0** — introduced `roo-cline.allowedCommands` (default `["git log", "git diff", "git show"]`) and `roo-cline.deniedCommands` as contributed VS Code configuration, merged at runtime by `ClineProvider.mergeCommandLists()`; https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/webview/ClineProvider.ts

## Gaps

**permissions — `unsupported` for `roo`, while the extension reads the keys the alias owns.**

The extension contributes `roo-cline.allowedCommands` and `roo-cline.deniedCommands` as workspace-scoped VS Code settings, so a repository can commit them to `.vscode/settings.json` and have every contributor's install honour the same command allow/deny policy. Deny wins over allow, and the defaults grant three read-only git commands.

Rulesync has `src/features/permissions/zoocode-permissions.ts` for the modern namespace but no `roo-permissions.ts`, and `roo` is absent from the `permissions` feature entirely. A user targeting the deprecated alias — which is exactly the population still on the `roo-cline.*` keys — cannot author these lists at all.

## Proposed Follow-up

- Add a `RooPermissions` adapter mirroring `ZoocodePermissions`, writing `roo-cline.allowedCommands` / `roo-cline.deniedCommands` into `.vscode/settings.json` through the shared-config gateway so unrelated keys survive, and register `roo` in the permissions processor gate and `tool-target-tuples.ts`.
- Project scope only: these are contributed VS Code *workspace* settings, and the alias is deprecated, so a global-scope surface is not worth adding.
- Weigh this against the alias's deprecation status before implementing — if `roo` is on a path to removal, closing this as won't-do is a legitimate outcome; it is filed so the decision is explicit rather than an oversight.

## Watch item (not a gap)

v3.80.0 added `allowedReadFiles` / `allowedWriteFiles`, but they live in the extension's **global state**, not in contributed VS Code configuration, so there is no committable file surface for rulesync to write. Re-check if a later release promotes them to `contributes.configuration`.

The live `zoocode` target has **no material gaps** through v3.80.0: rules, ignore, mcp, commands, subagents, skills, hooks and permissions all match the upstream surfaces, including the mode-specific directories tracked and closed in #2596.

## References

- https://github.com/Zoo-Code-Org/Zoo-Code/releases — release list; source for v3.80.0 (2026-08-22) and the v3.54.0 entry.
- https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/core/webview/ClineProvider.ts — `mergeCommandLists()`, the runtime merge of the allow/deny lists.
- https://github.com/Zoo-Code-Org/Zoo-Code/blob/main/src/package.json — `contributes.configuration`, which establishes that the two command lists are workspace-settable while `allowedReadFiles`/`allowedWriteFiles` are not.
- #2596 — the closed Zoo Code follow-up whose scope this issue sits outside of.

dyoshikawa:
## Summary

Adds a `permissions` adapter for the deprecated `roo` alias, so a project targeting Roo Code can author the command allow/deny lists the extension actually reads.

Roo Code is a VS Code extension whose package name is `roo-cline`, so its committable command lists are the contributed workspace settings `roo-cline.allowedCommands` and `roo-cline.deniedCommands` in `.vscode/settings.json`. `zoocode` already had an adapter for the post-rebrand `zoo-code.*` spelling; `roo` had none, so the `permissions` column for `roo` was blank even though the extension reads keys the alias owns.

## Why this is in the `roo` target's scope

The repository's own policy is that the `roo` target "no longer tracks anything Zoo Code added after the fork". These two keys are not a post-fork addition — they are present in Roo Code's **final** release, v3.54.0 (2026-05-15), verified directly against the archived repository:

- `src/package.json` at `v3.54.0` contributes `roo-cline.allowedCommands` (default `["git log", "git diff", "git show"]`) and `roo-cline.deniedCommands` (default `[]`), both with no `scope`, which in VS Code means `window` scope — settable per workspace.
- `src/core/webview/ClineProvider.ts` at `v3.54.0` reads them via `vscode.workspace.getConfiguration(Package.name).get<string[]>(configKey)` in `mergeCommandLists()` and unions them into the auto-approval lists.
- `src/package.json` `name` is `roo-cline`, which is what `Package.name` resolves to.

So the surface is Roo-era, and the target is documented as staying supported rather than being on a path to removal.

## What changed

- **`src/features/permissions/roo-permissions.ts`** — new `RooPermissions`, mirroring `ZoocodePermissions`: only the canonical `bash` category maps, `allow` → `allowedCommands`, `deny` → `deniedCommands`, `ask` is represented by omitting the pattern from both lists, an empty list is retracted rather than written as `[]`, and a canonical config that states no `bash` category leaves both keys exactly as the user wrote them. Project scope only, since VS Code's user-scope `settings.json` is at a platform-dependent path outside Rulesync's home-relative global model. The file is never deleted.
- **`src/constants/roo-paths.ts`** — the four new constants, with the primary-source citation for why the namespace is `roo-cline.*`.
- **`src/features/shared/shared-config-gateway.ts`** — `.vscode/settings.json`'s `permissions` `ownedKeys` gains the two `roo-cline.*` keys. Three targets now reach this file through the same feature (`copilot`, `zoocode`, `roo`); they stay independent because a patch only ever names the keys its own adapter builds.
- **`src/features/permissions/permissions-processor.ts`**, **`src/types/tool-target-tuples.ts`** — register `roo` (`supportsProject`, `supportsImport`; no global).
- **Comments in `zoocode-permissions.ts` / `zoocode-paths.ts`** — updated: the two lineages deliberately do not *share* an adapter, rather than `roo` deliberately having none.
- **Docs** — a Roo Code paragraph next to the Zoo Code one in `docs/reference/file-formats.md`, including the detail that Roo resolves a command matching both lists by the **longer** prefix (auto-approval needs a strictly longer allowed match; a denied match that is longer or equal auto-denies), which is why a pattern present in both lists still imports as `deny`. `README.md` and `docs/reference/supported-tools.md` tables regenerated.

## Coverage

- `src/features/permissions/roo-permissions.test.ts` — 12 tests, including one asserting the `zoo-code.*` pair in the same file is left alone on generate and not picked up on import.
- `src/e2e/e2e-permissions.spec.ts` — the Tool × Feature happy path for `roo` in both directions, plus the empty-payload case that must not create the shared file. The generate case seeds the file with Copilot's `chat.tools.*` key and Zoo Code's `zoo-code.allowedCommands` and asserts both survive.

`pnpm cicheck` is green (9249 tests).

Closes #2768


dyoshikawa:
Follow-up filed as #2792 for the low-severity findings from review that belong to the canonical permissions model rather than to this PR: prototype-named command prefixes being dropped silently, JSONC comments lost when a shared user-managed config is written back, `rulesync import --features permissions` replacing the canonical file instead of merging into it, and an empty-string pattern passing validation.

dyoshikawa:
@dyoshikawa Thank you!

Issue #2797, opened 2026-08-26
rulesync import --features permissions replaces the canonical file instead of merging per category


dyoshikawa:
Split out of #2792 (item 3), so it can be sized and reviewed on its own.

## Problem

`rulesync import --features permissions` replaces `.rulesync/permissions.jsonc` wholesale instead of merging into it. `PermissionsProcessor.convertToolFilesToRulesyncFiles` maps each tool file to its own full `RulesyncPermissions` document, and the writer writes them at the same path, so the last one wins — both across tools in a single run and across successive runs.

Each tool covers only a subset of the canonical categories. Roo Code, for instance, gates terminal commands and nothing else, so importing from Roo after importing from Claude Code discards the `read` / `write` / `edit` / `webfetch` categories the earlier import produced.

## Proposal

Merge per category instead of replacing the file, with a documented precedence rule for conflicting patterns.

## What needs deciding first

This is a semantics change, not a bug fix, and it needs a maintainer decision before implementation:

1. **Should it merge with the file already on disk, or only across the tools in one run?** Merging with disk makes import non-idempotent in a new way: a rule the user deleted from `.rulesync/permissions.jsonc` would come back on the next import.
2. **What is the precedence when two tools map the same category and pattern to different actions?** Later tool wins, most restrictive wins, or a conflict report.
3. **Should the same rule apply to the other features?** Import overwrites for rules, MCP, and hooks too. Changing only permissions would make the feature set inconsistent.

dyoshikawa:
Recording the outcome of an autonomous re-evaluation pass: this issue is **inconclusive by design** and is being left open for a maintainer decision rather than implemented.

The report is accurate about the current behavior — `rulesync import --features permissions` writes `.rulesync/permissions.jsonc` by replacing the canonical `permission` block rather than merging into it, so importing from a second tool discards what the first import contributed. But as the issue body itself states, changing that is a semantics change rather than a bug fix, and the three questions it raises have to be answered before any implementation is defensible:

1. **What is being merged?** Merging the freshly imported block with whatever already exists on disk, and merging the blocks produced by several tools within a single `import` run, are different features with different failure modes. On-disk merging makes a re-import non-idempotent (a pattern the user deliberately deleted comes back); in-run merging does not.
[...]

Issue #2796, opened 2026-08-26
Shared JSONC configs lose the user's comments on write-back


dyoshikawa:
Split out of #2792 (item 2), so it can be sized and reviewed on its own.

## Problem

`parseSharedConfig({ format: "jsonc" })` tolerates comments and trailing commas on read, but the write path re-serializes the parsed document, so every comment and every piece of the user's formatting disappears the first time Rulesync touches the file.

`.vscode/settings.json` is the clearest case: it is explicitly a user-managed file that Rulesync only patches a couple of keys of, so silently stripping the rest of the user's content is worse than the usual generated-file situation, where the whole file is Rulesync's to own.

This applies to every shared, user-managed config Rulesync patches through `src/features/shared/shared-config-gateway.ts`, not to one tool's keys.

## Proposal

Use a JSONC-preserving edit rather than a re-serialize for the `jsonc` format: compute the set of changed paths and apply them with `jsonc-parser`'s `modify` / `applyEdits`, so untouched regions of the file stay byte-identical.

## Notes

- This is not a small change. `applySharedConfigPatch` currently produces a whole new document from the declared ownership policy (`replace-owned-keys` / `deep-merge`) and hands it to the serializer; preserving comments means deriving edits instead of a document.
- The no-data-loss contract in `src/lib/shared-file-contract.test.ts` covers the cross-feature write order and must keep passing unchanged.
- Formats other than `jsonc` (YAML, TOML) have the same comment-loss property but are out of scope here; file separately if they are worth fixing.

PR #2798, opened 2026-08-26
fix(permissions): reject blank and prototype-named permission patterns

### LINKED ISSUE #2792: Permissions: four low-severity gaps in the canonical model and shared-config write path
While reviewing #2791 (Roo Code / Zoo Code permissions support), four low-severity issues surfaced that are not specific to that PR — they belong to the canonical permissions model and the shared-config write path as a whole. Filing them together so they can be triaged on their own merits.

### 1. Command prefixes named like `Object.prototype` members are silently dropped

The canonical `permission.<category>` map is a plain object literal, so a pattern such as `__proto__`, `constructor`, or `toString` used as a key is either swallowed or collides with an inherited property depending on how the map is built and read. A user who writes `{"__proto__": "deny"}` gets no error and no entry in the generated file. These are unlikely command prefixes, but the failure is silent, which is the part worth fixing — either reject them during validation or build the map with `Object.create(null)` / a `Map` so they round-trip like any other string.

### 2. JSONC comments in `.vscode/settings.json` are lost on write-back

`parseSharedConfig({ format: "jsonc" })` tolerates comments and trailing commas on read, but the write path re-serializes the parsed object, so any comment the user wrote in their workspace settings disappears the first time rulesync touches the file. `.vscode/settings.json` is explicitly a user-managed file that rulesync only patches two keys of, so silently stripping the rest of the user's formatting is worse than the usual generated-file case. Using a JSONC-preserving edit (e.g. `jsonc-parser`'s `modify`/`applyEdits`) would keep the untouched parts byte-identical.

This applies to every shared user-managed config file rulesync patches, not only the Roo/Zoo keys.

### 3. `rulesync import --features permissions` replaces the whole canonical file

Importing permissions from one tool overwrites `.rulesync/permissions.json` wholesale rather than merging into it. Since each tool only covers a subset of the canonical categories — Roo Code, for instance, gates terminal commands and nothing else — importing from Roo after importing from Claude Code discards the `read`/`write`/`edit`/`webfetch` categories the earlier import produced. Merging per category (with a documented precedence rule for conflicting patterns) would make repeated imports composable.

### 4. An empty-string pattern is written through unvalidated

`{"permission": {"bash": {"": "allow"}}}` is accepted and emitted. In Roo Code it is harmless in practice — `mergeCommandLists` filters entries on `cmd.trim().length > 0`, so the entry is dropped before the auto-approval decision reads the list. But it lands verbatim in other targets; in Copilot it becomes `"": true` inside `chat.tools.terminal.autoApprove`. The empty string is a prefix of every command, so if any target does honor it, it grants or denies everything. Validation should reject a blank pattern rather than let each target decide what it means.

### Scope

These are independent; any subset can be fixed on its own. None of them is a regression introduced by #2791 — that PR is simply where they became visible.

dyoshikawa:
## Summary

Closes #2792 by landing its two contained items and splitting the two that need a design decision.

### Item 1 — patterns named after prototype members were silently dropped

`parseJsonc` strips `__proto__`, `constructor` and `prototype` from every source document, so `{"permission": {"bash": {"__proto__": "deny"}}}` produced no error and no entry anywhere. `.rulesync/permissions.jsonc` now reports them by path.

Detection walks the syntax tree rather than the parsed value, because the parsed value cannot answer the question: `obj["__proto__"] = "deny"` is discarded by the engine outright, and `obj["__proto__"] = {...}` replaces the prototype instead of adding an own property — either way the key is already gone before the sanitizer could look for it. `parseJsoncReportingDroppedKeys` is a new export; `parseJsonc` keeps its single-argument signature and delegates to it, so its 32 call sites are untouched.

### Item 4 — blank patterns were written through unvalidated

`{"permission": {"bash": {"": "allow"}}}` was accepted and emitted. The empty string is a prefix of every command, so a target that honors it grants or denies everything; Roo Code happens to filter it (`cmd.trim().length > 0`), Copilot writes it verbatim as `"": true` into `chat.tools.terminal.autoApprove`. The canonical schema now rejects a blank (or whitespace-only) pattern instead of letting each target decide what it means.

Import is handled separately: a blank pattern found in a tool's own config is dropped by `withoutBlankPermissionPatterns` rather than reproduced, since writing it would produce a source file the very next `generate` refuses, over a value the tool never honored. Only the shared `permission` block is walked — tool-scoped blocks are hand-authored, and OpenCode / Kilo / Vibe hold tool-native shapes there.

### Items 2 and 3 — split out, not fixed here

Both are design changes with a wider blast radius than the two above, and #2792 states the four are independent:

- #2796 — shared JSONC configs lose the user's comments on write-back. Preserving them means deriving `jsonc-parser` edits instead of re-serializing a document, across every shared config writer.
- #2797 — `import --features permissions` replaces the canonical file instead of merging per category. Needs a maintainer decision on disk-merge idempotency, conflict precedence, and whether the other features follow.

## Testing

`pnpm cicheck` — 9303 tests, no lint / format / spell / secret findings.

New coverage: dotted-path reporting including array indices in `jsonc.test.ts`; both rejected pattern kinds through `validate()` and through the constructor; a blank pattern in a tool-scoped block; and `withoutBlankPermissionPatterns` returning its input unchanged when there is nothing to drop.

Closes #2792


dyoshikawa:
@dyoshikawa Thank you!

PR #2800, opened 2026-08-27
fix(generate): exit non-zero when a .rulesync source fails to load

### LINKED ISSUE #2789: generate reports success and exits 0 when a .rulesync source file fails schema validation
## Summary

When a source file under `.rulesync/` is rejected by its Zod schema, `rulesync generate` prints the parse error, then prints `✓ All files are up to date` and **exits 0** — while writing nothing. A pipeline that delegates config generation to `rulesync generate` passes as green and continues with stale or missing tool configs.

## Reproduction

```console
$ mkdir -p repro/.rulesync && cd repro
$ echo '{"mcpServers":{"a":{"command":"c","type":"bogus-transport"}}}' > .rulesync/mcp.json
$ rulesync generate --targets cursor --features mcp
Failed to load a Rulesync MCP file (.rulesync/mcp.jsonc): Zod raw error: [{"code":"invalid_value","values":["local","stdio","sse","http","ws","streamable-http"],"path":["mcpServers","a","type"], ...}]
✓ All files are up to date (mcp)
$ echo $?
0
$ ls .cursor
ls: cannot access '.cursor': No such file or directory
```

The error line is real and names the offending path, so the failure is diagnosable by a human reading the log. The problem is the two lines after it: the success marker and the zero exit code.

## Why it matters

The behaviour is fail-closed in the sense that no wrong content is written — but "wrote nothing" is reported identically to "everything was already correct". The two are indistinguishable to a caller, so:

- CI steps that run `rulesync generate` and then `git diff --exit-code` see no diff and pass, even though the source was never parsed.
- A first-time generate in a fresh checkout produces no tool configs at all and still reports success.

`✓ All files are up to date` is also inaccurate on its face here: the files are not up to date, they are absent.

## Scope

This is not specific to any one feature or tool. It was measured on `mcp` with the pre-existing `type` field, and the same shape applies to any `.rulesync/` source whose load is caught and turned into a logged error rather than a non-zero exit.

## Proposed fix

Decide between two options and apply it consistently across features:

1. **Exit non-zero when any source file fails to parse.** Most defensible: a malformed source is a user error the caller should see, and the current message already identifies it. Risk: a repository with an unrelated broken source file that was previously ignored now fails the whole generate.
2. **Keep exit 0 but stop printing the success marker**, and instead report `N source file(s) could not be loaded; nothing was generated for <feature>`. Lower blast radius, but still leaves scripted callers without a signal.

Option 1 is preferable unless there is a deliberate reason the load errors are non-fatal today; if there is, it should be recorded in a comment at the catch site.

Either way, add a regression test that asserts the exit code and the absence of the success marker for a malformed source.

## Notes

Found while reviewing #2788 (which added a new strict enum field, `musecodeMode`, to the MCP server schema). That PR does not cause this behaviour — it only adds one more field that can trigger it — and the reproduction above uses the long-standing `type` field to make that explicit.

dyoshikawa:
## Summary

`generate` reported success and exited `0` when a `.rulesync/` source file failed to load.

A source that exists but cannot be read — malformed JSON/JSONC, or content the schema rejects — was logged as an error and then skipped, leaving every counter at zero. That is indistinguishable from "there was nothing to write", so the command printed `✓ All files are up to date (...)` and exited `0`. A CI job that only checks the exit code never saw the failure, and the stale generated configs looked current.

## What changed

- `FeatureProcessor` gained `recordRulesyncSourceLoadFailure()` / `hasRulesyncSourceLoadFailure()`. The four processors that swallow a source load error — `mcp`, `hooks`, `ignore`, `permissions` — now record it before returning `[]`.
- `generate` carries the flag out as `GenerateResult.sourceLoadFailed`, and the CLI throws `CLIError(GENERATION_FAILED)` once **every** feature has had its turn. Failing at the end rather than at the catch site means one bad file does not hide the rest of the errors, and `--check` is covered too since the throw precedes the check-mode branch.
- A source that is simply **absent** is not a failure. `mcp` surfaces that as `ENOENT` (matched by the new `isFileNotFoundError`, which follows the `cause` chain); `hooks`, `ignore`, and `permissions` already resolve a `winningRoot`, which is `undefined` precisely when no root supplies the file. Without this distinction the fix would have failed every project that has no `mcp.jsonc`.

## Tests

- `src/e2e/e2e-generate-source-failure.spec.ts` (new): for `mcp`, `hooks`, and `permissions`, a schema-rejecting source must exit `1`, must not print `All files are up to date`, and must write no output file; for `mcp`, `hooks`, `ignore`, and `permissions`, an absent source must still succeed.
- CLI unit test asserting `GENERATION_FAILED` and that no success banner is logged.
- `McpProcessor` unit tests for the flag, and `isFileNotFoundError` unit tests including the wrapped-`cause` case.

## Review follow-ups

- **Absence is now a typed signal.** Every single-file loader raises `RulesyncSourceNotFoundError` once each candidate path has been ruled out, and `isRulesyncSourceMissing` accepts only that type. A bare `ENOENT` escaping a read that had already confirmed the file is treated as unreadable, which closes the TOCTOU gap where a source deleted mid-run read as "this feature has no source".
- **Dangling symlinks.** `stat` follows symlinks, so a link whose target is gone reports `ENOENT` and looked exactly like absence — and `docs/reference/file-formats.md` actively recommends symlinking `.rulesync/` sources at a shared tree. `fileExistsStrict` now falls back to `lstat` and reports that case instead of generating nothing and, under `--delete`, sweeping the previous output away.
- **The MCP server tool.** `executeGenerate` built its success summary from counts that all read zero, so an unreadable source came back as a successful no-op. It now fails, sharing the CLI's wording through `formatSourceLoadFailure`.

### Round 3

- **Error codes, not error strings.** The four processors decided "the tool's own config is simply absent" by matching on the message text of the caught error. That is now `isFileNotFoundError`, so a wrapped or localized error cannot be misread as absence and silently downgraded to `debug`.
- **One reporting path.** The four `loadRulesyncFiles` catch blocks had drifted into four near-copies of the same absent-vs-unreadable branch. They now share `RulesyncSourceConsumer#reportRulesyncSourceLoadError`, so the flag can no longer be set in one processor and forgotten in another.
- **Source directories, not just source files.** The directory-based features enumerate their sources with a glob, and a glob over a directory symlinked at a tree that no longer resolves returns nothing — which reads as "every rule was deleted", so `--delete` removed what the run could not regenerate while still exiting `0`. Every `.rulesync` source-directory probe (`rules`, `commands`, `skills`, `subagents`, `checks`) now uses `directoryExistsStrict`. An e2e test pins it: after the rules source directory becomes a dangling symlink, `generate --delete` exits `1` and the previously generated rule survives. Individual dangling symlink *files* already failed loudly — globby's default `type: "all"` returns them and the subsequent read raises — so the gap was only ever at the directory level.
- **MCP server detail.** `executeGenerate` now reports the collected error lines alongside the summary sentence, instead of a summary that named the feature but not the reason.

### Round 4

- **Unreadable is not unparseable.** `subagents` and `checks` deliberately warn-and-skip a file that will not parse, because those directories hold free-form Markdown. The same `catch` was also swallowing I/O errors, so a source file symlinked at a target that no longer exists read as "this subagent is gone" — `--delete` removed the generated agent and the run exited `0`, which is issue #2789 again in a different feature. The split is now explicit: `isFileSystemError` routes a read failure to the source-load path and leaves parse failures on the warn path.
- **Input roots.** The round-3 work made every source *sub*directory strict, but the roots themselves were still probed with `directoryExists`. An overlay root is allowed to be absent, so an overlay symlinked at a tree that had gone away passed as "not configured here" and its generated output was swept — silently, with no warning at all. `inspectInputRoots` now separates "nothing is here" from "something is here that leads nowhere" and stops the run for the latter.
- **`directoryExistsStrict` contract.** It answered `false` for a path occupied by a regular file, which put a misconfiguration back in the same bucket as an empty source tree. It reports that case now, and the two strict probes share one `stat` instead of duplicating the `lstat` fallback.
- **MCP server detail, scoped.** The collecting logger kept every `error()` line. It now keeps only the ones tagged `SOURCE_LOAD_FAILED`, so the response explains the failure instead of forwarding whatever else the run logged.
- **Docs correction.** The new paragraph put `skills` with the warn-and-skip features. It aborts on invalid frontmatter like `rules` and `commands` — verified by running it — and the text is fixed. This one matters because `docs/**` is embedded into `rulesync docs` and shipped to users.

### Intentionally rejected

An earlier revision also escalated the `subagents` and `checks` loaders from a warning to a source-load failure. That is reverted on purpose. Those directories hold free-form Markdown that users keep READMEs and notes in, so a single unparseable file would fail every subsequent `generate` **and** freeze that feature's orphan sweep — meaning a subagent deleted from the source would never be removed from the tool tree. Skip-and-continue is the long-standing, deliberate design for the directory-based features, and `docs/reference/cli-commands.md` now states that split explicitly.

## Docs

`docs/reference/cli-commands.md` documents the exit-code contract for unreadable versus absent sources, scoped to the single-file features, with the directory features' warn-and-continue behaviour described alongside. `docs/api/programmatic-api.md` documents that `generate()` reports the failure on the result rather than throwing.

Closes #2789


dyoshikawa:
@dyoshikawa Thank you!

PR #2802, opened 2026-08-27
feat(fetch): prune stale files inside fetched skill directories

### LINKED ISSUE #2801: fetch: prune stale files inside fetched skill directories so the remote skill is the source of truth
## Background

`rulesync fetch` copies files from a remote repository into the local project (by default under `.rulesync/`, or into a tool-native layout when `--target` is given). For each collected remote file it either writes it or skips it depending on `--conflict` (`overwrite` by default, or `skip`) — see `src/lib/fetch.ts`.

What it never does is **remove** anything. The command is purely additive: local files that are not present on the remote side are left untouched.

For the `skills` feature this is a real problem, because a skill is a *directory* (`skills/<name>/SKILL.md` plus its supporting files), not a single file. When the upstream skill drops or renames a file — e.g. `skills/foo/reference.md` is deleted upstream, or `scripts/run.sh` is renamed — a re-fetch leaves the stale local copy in place. The local skill directory then becomes a mix of current upstream files and orphaned leftovers, which is not what the user asked for: they expect the remote skill to be the source of truth.

Stale files inside a skill directory are not harmless. Agents read whatever is in the skill directory, so an orphaned `reference.md` or an outdated script keeps influencing behavior long after upstream removed it, and it is hard to notice because nothing in the fetch summary mentions it.

## Details

Reported request (original, Japanese):

> rulesync fetch で、skill ディレクトリに不要なファイルがあったらそれを削除すること。remote 側の skill を正として、不要になったローカル側の skill 内のファイルが生じた場合、それを残したくない。

In other words:

- Treat the **remote** skill as the source of truth.
- When a re-fetch produces a local skill directory containing files that no longer exist on the remote, those local files should be **deleted**, not kept.

Current behavior, for reference:

- `fetchFiles()` (`src/lib/fetch.ts`) collects remote files, applies skill selection (`applySkillSelection`), then writes each file with `writeFileContent`, reporting `created` / `overwritten` / `skipped`.
- There is no deletion step and no `FetchSummary` field for deleted files.
- The same additive-only behavior applies to the tool-target conversion path (`fetchAndConvertToolFiles`).

Points that need a decision before implementing:

1. **Scope of the pruning.** Deleting should almost certainly be limited to the skill directories that were actually fetched in this run (i.e. `skills/<name>/` for each selected skill). Deleting anything under the whole output directory would be dangerous — with `--skills` / `--interactive`, unselected local skills must obviously survive, and non-skill features (`rules/`, `commands/`, …) are outside the request.
2. **Opt-in vs. default.** Deleting user files is destructive, so it likely wants an explicit flag (e.g. `--prune`, or a `--conflict`-adjacent sync mode) rather than becoming the silent default. On the other hand, the reporter's expectation is that "remote is the source of truth", which argues for making it the default for skills specifically. This trade-off should be settled explicitly.
3. **Interaction with `--conflict skip`.** Pruning while skipping overwrites is contradictory; the combination should either be rejected or clearly documented.
4. **Locally added files.** A user may have deliberately added their own file inside a fetched skill directory. Pruning removes it. This is the intended semantic for a "mirror the remote" mode, but it needs to be documented loudly.
5. **Reporting.** Deleted paths should be visible in the fetch summary (`formatFetchSummary`), alongside created/overwritten/skipped, so the destructive part of the operation is never silent.
6. **Safety.** Deletion must go through the same `checkPathTraversal` guarding as writes, must only touch paths inside the resolved output base, and must not follow symlinks out of the tree. Empty directories left behind after pruning should also be cleaned up.

## Solution / Next Steps

1. Decide on the UX: a new `--prune` (or `--sync`) flag versus prune-by-default for the `skills` feature. Include the `--conflict` interaction in the decision.
2. Extend `fetchFiles()` so that, for every skill directory it wrote in this run, it enumerates existing local files under that directory and deletes those absent from the remote file list. Apply the same logic in the tool-target path (`fetchAndConvertToolFiles`), where the local path is derived from the converted output rather than the raw remote path.
3. Add a `deleted` count and per-file entries to `FetchSummary`, and render them in `formatFetchSummary`.
4. Guard the deletion with `checkPathTraversal`, restrict it strictly to fetched skill directories, and remove directories that became empty.
5. Tests: `src/lib/fetch.test.ts` for the pruning logic (stale file removed, unselected skills untouched, non-skill features untouched, `--conflict skip` behavior, path-traversal rejection), plus an e2e case in `src/e2e/e2e-fetch.spec.ts`.
6. Document the behavior and the flag in `README.md` and the relevant page under `docs/`.

dyoshikawa:
## Summary

`rulesync fetch` was purely additive: it wrote or skipped every collected remote file and never removed anything. For directory-based **skills** that is a real problem — when upstream drops or renames `skills/<name>/reference.md`, a re-fetch leaves the stale local copy behind, and agents keep reading it long after upstream removed it.

This makes the remote skill the source of truth: after writing, `fetch` walks the `skills/<name>/` directories it fetched in this run and deletes the files the remote does not have.

## Behavior

- **Pruning is on by default**, matching the reported expectation ("remote is the source of truth"). `--no-prune` restores the old additive behavior.
- Scope is deliberately narrow — only `skills/<name>/` directories written by **this** run. A skill left out by `--skills`/`--interactive` is untouched, and no other feature (`rules/`, `commands/`, …) is ever pruned.
- Directories emptied by pruning are removed.
- `--conflict skip` disables pruning: that flag says to leave existing local files alone, and it also means the local copies are not this run's output, so they cannot be judged against the remote list.
- `--target <tool>` never prunes — `convertFetchedFilesToRulesync` does not convert skills at all, so that path never writes a skill directory.

Deletions are never silent. `FetchSummary` gains a `deleted` count and `deleted` file entries, `formatFetchSummary` renders them with a distinct `✗` mark and its own summary part, and `--json` captures a `deleted` array:

```text
Fetched from anthropics/skills@main:
  ✓ skills/pdf/SKILL.md (overwritten)
  ✗ skills/pdf/reference.md (deleted - no longer in the remote skill)

Summary: 1 overwritten, 1 deleted
```

## Safety

- Skill directory names come from the remote repository, so each one goes through the same `checkPathTraversal` guard as the writes before it is walked.
- `readdir(..., { withFileTypes: true })` reports a link as a link, so symbolic links are **unlinked, never followed and never resolved**: a link inside a skill directory can only ever lose the link itself, never what it points at. A link the fetch wrote through (one with a fetched descendant) is left alone rather than removed.
- A file a user deliberately added inside a fetched skill directory **is** removed — that is what mirroring the remote means. It is documented with a warning block, and every deletion appears in the summary.

## Tests

- `src/lib/fetch.test.ts`: stale file deleted, emptied directory removed, unselected skill untouched, non-skill features untouched, `--no-prune`, `--conflict skip`, stale symlink unlinked without touching its target, symlinked directory not walked, and the `formatFetchSummary` rendering.
- `src/e2e/e2e-fetch.spec.ts`: end-to-end prune plus `--no-prune` through the real CLI program with `--json`.

## Docs

- `docs/reference/cli-commands.md`: a "Pruning Fetched Skill Directories" section, the `--no-prune` option row, and an example.
- `docs/guide/official-skills.md`: a pointer for the re-fetch case.

Closes #2801


dyoshikawa:
@dyoshikawa Thank you!

PR #2803, opened 2026-08-27
fix(skills): enumerate skill directories with readdir instead of a glob

### LINKED ISSUE #2787: --delete never sweeps a skill directory whose name contains a backslash
## Summary

On Linux, a skill directory whose name contains a backslash is never removed by `generate --delete`, however stale it is.

`findFilesByGlobs()` is backed by globby/fast-glob, which treats a backslash as a path separator and rewrites it in the paths it returns. `SkillsProcessor.loadToolDirsToDelete()` derives `dirName` from `basename()` of that rewritten path, so it builds a deletion candidate for a directory that does not exist.

## Reproduction

Verified on `main`:

```
readdir:  [ 'back\\slash' ]
globby:   [ '.../skills/back/slash' ]      <- not a real path
basename: [ 'slash' ]
```

The candidate becomes `<root>/slash`, so the real `<root>/back\slash` is never considered and survives every `--delete` run.

Windows is unaffected, since a backslash cannot appear in a file name there.

## Impact

Under-deletion only, so the consequence is a stale directory rather than a lost one:

- The constructed path stays within `join(outputRoot, relativeDirPath, basename)` and cannot escape the root — a name containing a separator is rejected by the `AiDir` constructor, and here it has already been stripped by `basename()`.
- `removeDirectory()` is a no-op when the path does not exist.
- If a directory really named `slash` also exists, it is enumerated as its own candidate and judged on its own merits, so no extra deletion happens.
- `isDirOwned` is evaluated against the same derived name, so the ownership check is not bypassed.

## Possible direction

Enumerate the candidate directories with `readdir` instead of a glob, or carry the real name alongside the globbed path. The same rewriting affects any other place that recovers a name from a `findFilesByGlobs()` result, which is worth auditing at the same time.

## Credit

Found during the security review of #2784, incidental to that change.

dyoshikawa:
## Problem

Globby (via fast-glob) reads a backslash as a path separator on every platform, and rewrites it in the paths it returns. A skills root holding a directory literally named `back\slash` — a perfectly legal POSIX name — came back from the `*` glob as `.../back/slash`, so the name recovered with `basename()` was `slash`: a directory that does not exist.

That fed the skills processor a phantom candidate and, symmetrically, hid the real directory from the orphan sweep. Windows is not affected, since a backslash cannot appear in a name there — which is what makes the glob look correct everywhere it is normally tested.

## Fix

Skill roots are now enumerated with `readdir`, through two new helpers in `src/utils/file.ts`:

- `listSubdirectoryNames(dirPath, { followSymbolicLinks })`
- `listFileNames(dirPath, { followSymbolicLinks })`

Both report names the way the filesystem spells them, resolve symbolic links only when asked (a dangling link is never reported), and sort — for the same reason `findFilesByGlobs` sorts: the order entries come off the filesystem in is not one a caller should have to depend on.

Four call sites in `skills-processor.ts` plus `getLocalSkillDirNames` in `skills-utils.ts` were converted. The flat-file `*.md` glob became `listFileNames` + an extension check.

## Why such a directory is reported, not swept

`AiDir` refuses a name containing a path separator, and rightly so: most tools take that name from a skill's frontmatter, and a repository written on one platform is read on the other, where a backslash *is* a separator — relaxing the guard would let a frontmatter-supplied `..\..\evil` become a literal POSIX name that traverses when the same repository is processed on Windows.

So an unaddressable directory is logged and skipped:

> Skipping `.claude/skills/back\slash`: a skill directory name cannot contain a path separator, so this directory is neither generated from nor swept as an orphan. Rename or remove it by hand.

The alternative — building a candidate from a name that belongs to no directory at all — is what this PR removes.

Closes #2787

### DIFF
--- a/src/features/commands/grokcli-command.test.ts
+++ b/src/features/commands/grokcli-command.test.ts
@@ -332,6 +332,26 @@ describe("GrokcliCommand", () => {
       expect(logger.warn).toHaveBeenCalledWith(expect.stringContaining("review"));
     });
 
+    it("should not read a backslash in a skill directory name as a separator", async () => {
+      // Globbed, `back\\review` came back as `back/review`, so the trailing
+      // segment matched the `review` command and a collision that does not
+      // exist was reported.
+      await ensureDir(join(testDir, RULESYNC_SKILLS_RELATIVE_DIR_PATH, "back\\review"));
+      await writeFileContent(
+        join(testDir, RULESYNC_SKILLS_RELATIVE_DIR_PATH, "back\\review", "SKILL.md"),
+        ["---", "name: review", "description: d", "---", "", "body"].join("\n"),
+      );
+      const logger = { warn: vi.fn() };
+
+      await GrokcliCommand.validateRulesyncCommands({
+        inputRoots: [join(testDir, RULESYNC_RELATIVE_DIR_PATH)],
+        rulesyncCommands: [commandNamed("review")],
+        logger: logger as unknown as Logger,
+      });
+
+      expect(logger.warn).not.toHaveBeenCalled();
+    });
+
     it("should stay quiet when no skill matches a command name", async () => {
       await ensureDir(join(testDir, RULESYNC_SKILLS_RELATIVE_DIR_PATH, "deploy"));
       await writeFileContent(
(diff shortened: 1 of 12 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2806, opened 2026-08-28
fix(generate): refuse to sweep a directory outside the root it was found in

### LINKED ISSUE #2786: removeOrphanAiDirs trusts the AiDir contract with no positional backstop
## Summary

`DirFeatureProcessor.removeOrphanAiDirs()` decides what to recursively delete from a subclass contract alone — `AiDir.ownsDirTree()` — without re-checking that the path it is about to remove is strictly below the root the candidates were enumerated from.

Raised as a defense-in-depth point during the review of #2784, which fixed a real instance of that contract being violated (`TaktSkill` overrides `getDirPath()`, every candidate reported the shared root, and the sweep deleted it).

## Why a backstop is worth having

- `removeOrphanAiDirs()` is a public method on the base class and accepts any `AiDir[]`. The `assertWritablePathInsideRoot` checks live in `SkillsProcessor.loadToolDirsToDelete()`, on the caller side, so a future caller — or a future `DirFeatureProcessor` — inherits the deletion but not the guard.
- The last line of defense inside `removeDirectory()` (`src/utils/file.ts`) is a `dangerousPaths` list matched by exact string equality against `.`, `/`, `~`, `src`, `node_modules`. An absolute path never matches it, and the function swallows the exception.
- The failure mode is recursive deletion of a directory the user owns, which is not recoverable from within rulesync.

## Possible direction

In `removeOrphanAiDirs()`, refuse to delete a candidate whose `getDirPath()` equals — or is an ancestor of — the root it was enumerated from (`join(outputRoot, relativeDirPath)`). That is a positional invariant rather than a delegated one, so a subclass that overrides `getDirPath()` for some unrelated reason cannot turn a contract mismatch into a destructive operation.

## References

- #2784
- #2777

dyoshikawa:
`DirFeatureProcessor.removeOrphanAiDirs()` decided what to recursively delete from `AiDir.ownsDirTree()` alone — the subclass's own answer about a path the subclass also computes. The positional checks (`assertWritablePathInsideRoot`) live in `SkillsProcessor.loadToolDirsToDelete()`, on the caller side, so any other caller of this public base-class method inherits the deletion without them, and #2777 showed what a contract mismatch costs: an override of `getDirPath()` made every candidate report the shared root, and the sweep deleted it.

The sweep now checks the position itself, before anything is removed. The order matters:

1. `ownsDirTree()` first. A tool that flattens its output into a shared root says so here and reports that root as its path — an expected shape, not a mismatch, and it stays on the quiet `debug` path added for #2777. `TaktSkill` is the one such tool in the tree today (`src/features/skills/takt-skill.ts`).
2. The positional backstop second. A candidate that claims the tree as its own must sit strictly below `join(outputRoot, relativeDirPath)`, the root it was enumerated from. One that reports that root, or a path in another tree, is warned about and skipped.

The check is positional rather than delegated, so it holds even when `getDirPath()` is overridden and `ownsDirTree()` is not kept in agreement with it. It is a backstop, not the guard: the comparison is lexical, so a root that is itself a symlink into another tree still passes it, and the caller-side `assertWritablePathInsideRoot` — which resolves the real path — remains required.

It also drops `removeAiDirs()`, a public recursive-delete method on the same base class with no caller in `src/` and none of these guards.

Closes #2786

### DIFF
--- a/src/features/skills/skills-processor.test.ts
+++ b/src/features/skills/skills-processor.test.ts
@@ -1715,11 +1715,8 @@ Content that would fail parsing`;
       // every enumerated candidate. Sweeping the root when the run generated no
       // takt skill deleted it outright, along with anything the user had put
       // there by hand.
-      const processor = new SkillsProcessor({
-        logger: createMockLogger(),
-        outputRoot: testDir,
-        toolTarget: "takt",
-      });
+      const logger = createMockLogger();
+      const processor = new SkillsProcessor({ logger, outputRoot: testDir, toolTarget: "takt" });
 
       const knowledgeDir = join(testDir, ".takt", "facets", "knowledge");
       const handAuthoredDir = join(knowledgeDir, "my-notes");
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2807, opened 2026-08-28
fix(skills): sweep orphan takt knowledge files on --delete

### LINKED ISSUE #2785: generate --delete never sweeps orphan takt knowledge files
## Summary

`generate --delete` never removes a stale file from `.takt/facets/knowledge/`. Delete or rename a skill under `.rulesync/skills/` and its generated `<name>.md` stays behind indefinitely; the run reports `All files are up to date`.

## Cause

The orphan sweep for the skills feature works on directories. Takt skills are flat files sharing one facet root rather than each getting a directory of its own, so takt contributes no per-skill directory for the sweep to consider. #2784 made this explicit — a candidate that does not own its directory tree is skipped, because the only directory it could point at is the shared root, and deleting that would take hand-authored files with it.

So the root must not be swept, and today nothing else sweeps the files inside it.

This is pre-existing behavior, not a regression: before #2784 the root was always claimed whenever any takt skill was generated, so the files in it were never candidates either.

## Possible direction

A file-keyed sweep for tools that flatten into a shared root: claim the `<name>.md` files a run generates, then remove the ones under the root that no source produces — leaving any file the tool would not have emitted (a hand-authored note, a subdirectory) alone. `AiDir.ownsDirTree()` already identifies which tools need this treatment.

Until then, the behavior is documented in `docs/tools/takt.md` and in the `--delete` note in `docs/reference/cli-commands.md`: the user deletes the stale file themselves.

## References

- #2784
- #2777

dyoshikawa:
## Summary

`generate --delete` never swept the flat skill files takt writes into `.takt/facets/knowledge/`. The skills sweep is keyed on directories, and a tool that flattens into a shared root contributes none — so a renamed or deleted skill left its old `<name>.md` behind for good, `--check` reported no difference, and removing a skill from `.rulesync/skills/` did not revoke it for takt.

This adds the file-keyed half of that sweep.

## What changed

- `AiDir.getFlatFilePath()` names the single file an entry that does not own its directory (`ownsDirTree() === false`) flattens into, or nothing when it owns one, carries no main file, or names it through a path separator.
- `SkillsProcessor.loadToolFlatFilesToDelete()` enumerates the `.md` files directly under each skills root (no symlinks, no hidden entries), builds a deletion candidate per file, and keeps only the candidates that name back the very file they were built from — which is false for every directory-based tool, so a stray `README.md` beside its skill directories is never a candidate. The existing `isDirOwned` ownership hook applies here as it does to the directory half.
- `DirFeatureProcessor.removeOrphanFlatFiles()` deletes the candidates this run did not generate, behind the same positional backstop the directory sweep uses (#2786): the candidate's root has to sit inside the directory this run writes to, and the file has to sit directly in that root. Anything else is refused with a warning instead of deleted.
- `generate.ts` runs both halves in the deferred sweep. Claims are keyed on the same file path the run already registers by name, so one target never deletes a sibling's freshly written file.

The root itself is never deleted, and neither is anything nested inside it — a subdirectory of hand-authored notes is left alone. A hand-authored `.md` placed *directly* in the facet root is indistinguishable from a generated one and is swept with them, exactly as `policies/`, `instructions/`, and `personas/` already behave for the rules, commands, and subagents features. Documented in `docs/tools/takt.md` and the `--delete` note in `docs/reference/cli-commands.md`.

## Tests

- `AiDir.getFlatFilePath` unit tests, including every name it refuses.
- `removeOrphanFlatFiles` unit tests: sweeps the unclaimed file, keeps the generated one, dedupes two candidates naming one file, reports without deleting on a dry run, and refuses a candidate that names no file / whose root is outside the output root / that reports a directory other than the root it was found in.
- `SkillsProcessor` tests for the takt enumeration and for a directory-based tool contributing nothing; the #2777 regression test (the shared root is never deleted) is untouched and still passes.
- E2E: a takt knowledge file whose source is gone is swept, while a nested hand-authored file and the newly generated file survive.

`pnpm cicheck` passes.

Closes #2785


dyoshikawa:
@dyoshikawa Thank you!

PR #2810, opened 2026-08-28
feat(fetch): mark skill names that cannot be told apart on sight

### LINKED ISSUE #2783: Follow-up: remaining fetch-consent and terminal-output issues from #2782
## Summary

`rulesync fetch --interactive` now starts with nothing selected and drops any skill directory whose name contains control characters (#2782). The review of that PR surfaced several lower-severity points that were deliberately left out of its scope. They are collected here so they can be judged on their own.

## 1. Only `--interactive` has a consent step

`--interactive` is the sole path where the user sees what will be written before it is written. Everything else fetches the source repository's full skill list unattended:

- `src/lib/fetch.ts` — the non-interactive path returns early without prompting.
- `src/lib/sources.ts` — `rulesync install` resolves an absent selection to `?? ["*"]`.

Skills are agent-instruction files, so "fetch everything this repository publishes" is a meaningful trust decision. Worth deciding whether a summary-and-confirm step (or an explicit `--yes`) belongs on those paths too.

## 2. Raw remote paths are printed to the terminal

`formatFetchSummary` and the `logger.debug` call in the same module print `relativePath` values straight from the remote repository, without `stripControlCharacters`. A crafted path can emit ANSI sequences and forge additional lines of output.

This predates #2782 and was left alone there for a reason: the summary strings and the `--json` `relativePath` field are expected to agree, so sanitizing only the human-facing side would silently desynchronize them. A fix needs to decide what the `--json` consumer should see.

## 3. Confusable (homograph) skill names

Two skill directories whose names differ only by visually similar code points are listed as two separate entries, so the guarantee "only what you selected is written" still holds — the user simply cannot tell the two entries apart by sight. Restricting choice names to ASCII would reject legitimate non-ASCII skill names, so this was not done. A display-only marker (e.g. flagging entries that normalize to the same string) may be a better answer.

## 4. Flat files directly under `skills/`

`classifySkillPath` treats a path with fewer than three segments (`skills/README.md`) as a non-skill and lets it through the selection filter unchanged. That matches the existing test `should not treat flat files directly under skills/ as selectable skills`, and such a file is not loaded as a skill by `rulesync generate`. Still worth confirming that pass-through — rather than fail-closed for everything under `skills/` — is the intended contract.

## 5. Empty selection reports success and a warning together

Confirming the prompt with nothing selected logs both `logger.success(formatFetchSummary(...))` for zero files and a warning. Suppressing the success line would change the output contract of every fetch path, including the `--json` `captureData` shape, so it was left as is.

## 6. `SkillPathClass` members are not readonly

`SkillPathClass` variants are plain mutable objects and `NON_SKILL_PATH` is a shared module-level constant. Marking the members `readonly` (and/or freezing the constant) would make the shared instance safe by construction.

## References

- #2782

dyoshikawa:
## Summary

Works through the follow-up list from #2782, acting on the three points that had a defensible fix and recording why the other three are left as they are.

Closes #2783

## Acted on

**3. Confusable (homograph) skill names.** The interactive prompt now marks an entry that cannot be told apart from another on sight, using the display-only marker the issue suggested:

```text
? Select skills to fetch (press <a> to select/deselect all)
 ◯ pdf
 ◯ сkill  [!] mixes Cyrillic and Latin characters
 ◯ skill
```

An entry is marked when another entry normalizes to the same name (NFKC, compared case-insensitively), or when its own name mixes scripts in a way an ordinary name does not. The second half follows the augmented script sets of UTS #39, so Japanese, Korean and Chinese names — which mix scripts by nature and often carry Latin — are not marked. The marker is display-only: `value` stays the real name, so a checked box still writes exactly the directory it names. New module `src/utils/confusable-names.ts` with its own tests.

**2. Raw remote paths printed to the terminal.** Two paths still reached a terminal unstripped: the oversized-file message in `fetch.ts` and in `github-client.ts`, and the `Fetched to temp` debug line. Both are now quoted and stripped like every other remote path this command prints. The `--json` output deliberately keeps the raw path — a machine consumer needs the real name — and that split is now documented in `docs/reference/cli-commands.md` instead of being implicit.

**6. `SkillPathClass` members are not readonly.** The variants are `readonly` and the shared `NON_SKILL_PATH` constant is frozen, so the one instance handed to every caller cannot be edited for the rest.

## Deliberately not changed

**1. Only `--interactive` has a consent step.** Adding a prompt or a required `--yes` to the unattended paths would break every scripted `fetch` and every `rulesync install` in CI, which are the paths that exist precisely to run without a terminal. `--interactive` remains the opt-in consent step, `--skills` the explicit narrowing, and for `install` the committed `rulesync.jsonc` is itself the recorded trust decision. Reopening this as its own issue would be reasonable if a real report of unwanted skills arriving unattended turns up.

**4. Flat files directly under `skills/`.** Pass-through is the intended contract, and it is what the existing test pins. Such a file (`skills/README.md`) is not loaded as a skill by `generate`, is not part of any skill directory, and selection is about skills — failing closed on everything under `skills/` would drop a repository's own README for no gain in what gets executed.

**5. Empty selection reports success and a warning together.** The success line is the fetch summary, which correctly reports zero files; the warning is what says nothing arrived. Suppressing either would change the output contract — including the `--json` `captureData` shape — of every fetch path, not just this one.

## Testing

- `pnpm cicheck` (383 test files, 9,467 tests, cspell and secretlint clean)


dyoshikawa:
@dyoshikawa Thank you!

PR #2811, opened 2026-08-28
fix: scan tool outputs from literal roots

mikemikimike:
## Summary

The orphan sweeps used by `generate --delete` embedded each tool output directory into an absolute glob pattern. If a project path contained glob metacharacters such as `(`, `)`, `{`, `}` or `*`, the directory was interpreted as a pattern and files could be skipped or picked up from a sibling directory.

This change evaluates the existing file patterns relative to a literal `cwd` for subagents, commands, and checks. Flat versus recursive matching and the existing symlink behavior are preserved, including not following symlinks while collecting deletion candidates.

The separate absolute-glob call sites in `rules-processor.ts` remain out of scope for this PR; in particular, the `.rulesync/rules` input-root sweep used by `generate --delete` should be handled in a follow-up.

## Tests

- `pnpm exec vitest run src/utils/file.test.ts src/features/subagents/subagents-processor.test.ts src/features/commands/commands-processor.test.ts src/features/checks/checks-processor.test.ts -t "literal directory path|glob metacharacters|symlinked checks|sibling matching" --silent=true`
- `pnpm run fmt:check`
- `pnpm run oxlint`
- `pnpm run typecheck`

The focused regression run passes. The repository-wide test run is affected by pre-existing Windows path-separator and platform-specific assumptions in unrelated tests.


src/features/checks/checks-processor.test.ts
```diff
@@ -122,6 +122,26 @@ Look for issues.
     expect((rulesyncFiles[0] as RulesyncCheck).getFrontmatter().severity).toBe("medium");
   });
 
+  it("should load checks when the output root contains glob metacharacters", async () => {
+    const literalRoot = join(testDir, "project[glob]");
```
dyoshikawa: `project[glob]` is a character class that still matches the literal directory name, so this test passes on `main` without the fix — it doesn't pin the regression. `project(glob)` or `project{a,b}` would fail before the fix and pass after.

src/features/commands/commands-processor.test.ts
```diff
@@ -1336,6 +1336,29 @@ describe("CommandsProcessor", () => {
   });
 
   describe("loadToolFiles with forDeletion: true", () => {
+    it("should load files when the output root contains glob metacharacters", async () => {
+      const literalRoot = join(testDir, "project[glob]");
```
dyoshikawa: Same as the checks test: `project[glob]` matches under the old absolute-pattern form too, so this passes on `main` unchanged. Worth switching to `project(glob)` or `project{a,b}`.

Also, `mockFindFilesByGlobs.mockImplementation(actualFile.findFilesByGlobs)` inside this test isn't undone — `beforeEach` only calls `vi.clearAllMocks()`, which doesn't reset implementations. It happens to be harmless today because every later test overrides with `mockResolvedValue`, but a `mockReset()` at the end of this test would make it robust.

src/features/subagents/subagents-processor.test.ts
```diff
@@ -719,6 +719,25 @@ Body from inputRoots[0]`;
   });
 
   describe("loadToolFiles", () => {
+    it("should load files when the output root contains glob metacharacters", async () => {
+      const literalRoot = join(testDir, "project[glob]");
```
dyoshikawa: Same issue here — `project[glob]` still matches with the old pattern form, so this test passes on `main`. Use a metacharacter that actually breaks the old path, e.g. `project(glob)`.

src/features/checks/checks-processor.ts
```diff
@@ -363,7 +363,7 @@ export class ChecksProcessor extends FeatureProcessor {
     const paths = factory.class.getSettablePaths({ global: this.global });
 
     const baseDir = join(this.outputRoot, paths.relativeDirPath);
-    const checkFilePaths = await findFilesByGlobs(join(baseDir, factory.meta.filePattern));
+    const checkFilePaths = await findFilesByGlobs(factory.meta.filePattern, { cwd: baseDir });
```
dyoshikawa: While you're touching this line: `followSymbolicLinks` isn't passed, so it defaults to `true` even when `forDeletion` is set. The result flows straight into the deletion-candidate branch a few lines below, so a `.amp/checks/team -> ../../shared` link would put files from outside the project on the orphan list. That's exactly what the comment in `commands-processor.ts` warns about, and `subagents-processor.ts` passes `followSymbolicLinks: !forDeletion` correctly. This is pre-existing rather than something you introduced, but it'd be a natural fix to fold in here.

dyoshikawa (commented):
Thanks for this — the direction is right, and I confirmed with globby directly that switching to a literal `cwd` fixes real breakage: with `outputRoot` = `project(glob)` or `project{a,b}` the old absolute-pattern form returns 0 files, and with `project*x` it silently picks up files from sibling project directories. Both are bad news for an orphan sweep that deletes.

The problem is that the three processor regression tests don't actually cover that. They all use `project[glob]`, and `[glob]` is a valid character class that still matches the literal directory, so those tests pass on `main` unchanged. I ran them against `main` to be sure. Only the `src/utils/file.test.ts` case is a genuine regression test. Swapping the directory name to `project(glob)` or `project{a,b}` is a one-token change that makes all three fail before the fix and pass after — worth doing, otherwise this can silently regress later. It would also be good to add one case with a sibling directory (e.g. `project*x` next to `projectOTHERx`) so the "picks up too much" side is covered too, not just "picks up nothing".

Related: the metacharacter list in the PR description isn't quite accurate. `[`, `]`, `?` and `!` don't actually break the old code; `(`, `)`, `{`, `}` and `*` do.

One scope note, not a blocker for this PR but worth calling out somewhere: `rules-processor.ts` still builds absolute glob patterns the same way (around lines 2017, 2060, 2068, 2109, 2113, 2125, 2237, 2278), and rules are also swept by `generate --delete`. The input side at lines 1693-1694 (`.rulesync/rules`) is the one I'd worry about most — the comment right above it already warns that an empty glob result reads as "every rule was deleted". A follow-up issue or a line in the PR body would be enough.

Also, the branch is ~141 commits behind `main`, and `src/utils/file.ts`, `subagents-processor.ts` and `checks-processor.ts` have all moved since the merge base. Worth rebasing and re-checking that the call sites still line up.

src/features/commands/commands-processor.ts
```diff
@@ -853,16 +853,19 @@ export class CommandsProcessor extends FeatureProcessor {
...
     // Never follow a symlink while collecting deletion candidates: a
     // `.augment/commands/team -> ../../shared-prompts` link would otherwise put
     // files outside the project on the orphan list. Matches the subagents,
     // skills and rules processors.
-    const commandFilePaths = await findFilesByGlobs(globPattern, {
-      followSymbolicLinks: !forDeletion,
-    });
+    const commandFilePaths = await findFilesByGlobs(
+      factory.meta.supportsSubdirectory
+        ? join("**", `*.${factory.meta.extension}`)
```
dyoshikawa: Minor inconsistency: one branch uses `join("**", ...)` and the other a plain template string. `join` produces `**\\*.md` on Windows and only works because `findFilesByGlobs` runs `replaceAll("\\\\", "/")` on it. Globs are always `/`-separated, so `` `**/*.${factory.meta.extension}` `` avoids relying on that and matches how `filePattern` is written literally in the subagents and checks factories.

src/utils/file.ts
```diff
@@ -630,6 +630,8 @@ function chooseRepresentative(candidates: string[], identity: string): string {
 export async function findFilesByGlobs(
   globs: string | string[],
   options: {
+    /** Directory in which relative patterns are evaluated. */
+    cwd?: string;
```
dyoshikawa, line 639: Worth expanding this doc comment. `cwd` is a starting point, not a boundary — a pattern with `../..` or an absolute path still escapes it, and I confirmed that behaviour. Since this function feeds deletion paths, stating that callers must pass relative patterns without `..` and that `cwd` guarantees no containment would help the next caller.

The `ignore` doc a few lines down is also now slightly stale: it explains the absolute-pattern case, but with `cwd` plus relative includes, `ignore` resolves relative to `cwd`. No caller combines them today, so this is just future-proofing.

src/utils/file.ts
```diff
@@ -679,6 +681,7 @@ export async function findFilesByGlobs(
   // path: git-client.ts (`walkDirectory`) skips symlinks entirely.
   const results = globbySync(normalizedGlobs, {
     absolute: true,
+    ...(options.cwd === undefined ? {} : { cwd: options.cwd }),
```
dyoshikawa: `cwd` is the only option not destructured at the top with `type` / `followSymbolicLinks` / `ignore` / `dot`. fast-glob falls back to `process.cwd()` on `undefined` anyway, so the conditional spread isn't needed — destructuring it and writing plain `cwd,` here would match the surrounding style.

dyoshikawa:
@mikemikimike Thank you! Merged. The review findings are tracked as a follow-up in #2870 rather than blocking this PR.

PR #2779, opened 2026-08-26
feat(vibe): support nested AGENTS.md, managed-shell and MCP tool permissions

dyoshikawa:
Closes #2770

Follows up three Vibe Code (mistral-vibe) upstream capabilities Rulesync could not reach. Verified against `mistralai/mistral-vibe` at v2.24.3.

## 1. Nested `AGENTS.md` is now authorable for `vibe`

Vibe's harness manager walks the directories between the workspace root and the file being read and loads every `AGENTS.md` it finds along the way (`find_subdirectory_agents_md` in `vibe/core/config/harness_files/_harness_manager.py`), injecting the result into the `read_file` tool's output since v2.19.1. `VibeRule` previously threw `VibeRule only supports root rules` for anything non-root, so that surface was unreachable.

`VibeRule` now follows the `reasonix` model rather than the issue's literal proposal:

- `nonRoot` stays `undefined` and the processor entry gains `collisionPolicy: "fold"` — plain topic rules fold into the single root `AGENTS.md`. **The issue asked for "a `nonRoot` path root the way `agentsmd-rule.ts` does", which would be wrong**: Vibe has no modular non-root instruction directory upstream, so inventing something like `.vibe/memories/` would emit files Vibe never reads.
- A non-root rule carrying `agentsmd.subprojectPath` is emitted as a nested `<subprojectPath>/AGENTS.md` (project scope only — the global root has no workspace to nest under).
- `getNestedFilePatterns` discovers them on import using the same exclusions as the AGENTS.md standard's scan. Import-only, so `generate --delete` never sweeps hand-authored files.

One deliberate difference from `reasonix`: the imported rule is **not** suffixed or target-scoped. Vibe's nested file is literally the AGENTS.md standard's own per-directory file at the same path, so a `-vibe` variant would make `vibe` emit both `<slug>.md` and `<slug>-vibe.md` into the same `<dir>/AGENTS.md` and fold duplicated content into one file. The derivation (`<directory-with-hyphens>.md`, `targets: ["*"]`, `agentsmd.subprojectPath`, plus the `overview.md` collision guard) is therefore extracted from `agentsmd-rule.ts` into a shared `ToolRule#toRulesyncRuleNestedAgentsmd`, so importing the same file through either target yields one rulesync rule.

## 2. Managed shell tool names have a reachable permission table

[...]

PR #2817, opened 2026-08-28
fix(vibe): honor disabled_tools glob spellings on import

### LINKED ISSUE #2781: Follow-up to #2779: remaining low-severity findings in the Vibe permissions translator
Follow-up to #2779, which closed the structural holes in the Vibe permissions translator (a deny silently failing to reach a shell, or an existing restriction being silently broadened). Seven review rounds left the following `low` findings deliberately unaddressed, because each is either pre-existing on `main`, cosmetic, or a refactor whose blast radius is wider than that PR's scope.

### 1. Glob spellings in `disabled_tools` are imported as an `allow`

`disabled_tools = ["read_*"]` disables every matching tool upstream, because Vibe's `name_matches` globs the raw tool name. Rulesync's import writes `permission["read_*"]["*"] = "deny"` — a category literally named `read_*`, which no other tool understands — while `[tools.read_file]` keeps whatever its own table says and is imported as-is. The net effect is that a glob-disabled tool can round-trip as allowed.

Verified identical on `origin/main`, so this is not a #2779 regression. #2779 does add one increment, though: because the `bash` category now fans out to every managed shell, a glob such as `disabled_tools = ["*bash*"]` — which upstream also disables `git_bash` with — causes `[tools.powershell] permission = "always"` to be newly written for the one shell the glob does *not* match, where previously no table existed (implicit ASK). `disabled_tools = ["*"]` carries `read` and `bash` out as allowed in the same way.

Two candidate fixes:

- Make the import-side skip test use the same glob matching as upstream's `name_matches`, instead of an exact-name comparison.
- Warn when a glob entry in `disabled_tools` matches an existing `[tools.<name>]` table, leaving the behavior alone.

A fix needs a decision on how far Rulesync should expand a glob without seeing Vibe's registry.

### 2. The import path has no logger

`ToolPermissions.toRulesyncPermissions()` takes no `logger`, so the import side cannot warn about anything it silently drops or reinterprets (a disabled tool whose table is skipped, a malformed key, a glob it cannot expand). Every warning added in #2779 lives on the generate side only.

Fixing this means changing the base signature across all tools, so it belongs in its own PR.

### 3. `[tools.bash]`'s own hand-authored `sensitive_patterns` are not sorted

Rulesync sorts the `sensitive_patterns` it merges into a fan-out alias, but leaves the canonical `[tools.bash]` table's own list in authored order. The result is one cosmetic diff between the two spellings. Reordering a key Rulesync does not own was judged more expensive than the diff it removes; revisit if the inconsistency becomes confusing.

### 4. Extract the first pass of the fan-out shell resolver

`resolveFanOutShellAliases` grew a first pass that classifies each candidate shell (states a decision / off the registry / free to overwrite) before the second pass builds the alias list. Pulling that pass into a named helper would make both halves readable on their own. Pure refactor, no behavior change.

### 5. `warnOnStrandedShellPatterns` scope

The warning currently fires only for shells that the fan-out actually touches. Whether it should also cover a shell that has patterns but no `bash` category override at all was left open; the current scope matches what the PR set out to fix.

### 6. Two warnings contradict each other on a non-list `denylist`

When a stood-down shell spells its `denylist` as a scalar, both warnings fire: the stand-down warning says "Only the 'bash' deny patterns are merged into it", and the malformed-key warning immediately says they are *not* merged. The second one is correct; the first is written for the well-formed case. Softening the stand-down wording to "attempts to merge" (or suppressing it when the merge is known to stand down) would remove the contradiction. Cosmetic only — no behavior is wrong.

dyoshikawa:
Closes #2781

Follow-up to #2779. Works through the six `low` findings that PR deliberately deferred.

### 1. Glob spellings in `disabled_tools` are imported as an `allow`

Vibe's tool manager matches a `disabled_tools` entry with `name_matches` — a case-insensitive `fnmatch` glob over the *raw* tool name — and applies the filter last and unconditionally, after `enabled_tools` and regardless of what `[tools.<name>]` says (`vibe/core/utils/matching.py`, `vibe/core/tools/manager.py`). Rulesync's import compared names exactly, so `disabled_tools = ["read_*"]` left `[tools.read_file]` read as live and carried its `permission = "always"` and its `allowlist` out as an allow — a hole through a tool the file had switched off, into every *other* tool's generated config.

The decision the issue asked for, on how far to expand a glob without seeing Vibe's registry:

- **The skip test follows upstream.** `fnmatchToRegExpSource` implements Python's `fnmatch.translate` (`*`, `?`, `[seq]` / `[!seq]`, an unterminated `[` as a literal bracket, a leading `^` in a class escaped), and matching is case-insensitive.
- **A matched table is denied under its own canonical category.** Those names the file *does* state, so no registry is needed for them.
- **The literal entry is still carried out verbatim** (as a category named `read_*`), so the glob keeps its reach over the tools rulesync cannot enumerate — an MCP tool registered at run time — and regenerates the same `disabled_tools` entry.
- **A `re:` entry is not evaluated.** Python's regex dialect is not JavaScript's, and this is the one place where being wrong in either direction costs something real, so it is matched only by its exact spelling and a warning names the tables imported as authored. A glob JavaScript refuses to compile is treated the same way.

This also closes the increment #2779 added: `disabled_tools = ["*bash*"]` no longer imports `[tools.bash]` as an allow that the fan-out then spreads to `powershell`, the one shell the glob does not reach.

On the **generate** side the mirror-image contradiction is reported rather than fixed: a glob left in `disabled_tools` that silences a table rulesync just wrote a live permission into is warned about, not deleted, because deleting it would also switch on the tools rulesync cannot see — the broadening direction.

### 2. The import path has no logger

Resolved without touching the base `toRulesyncPermissions()` signature, by adopting the idiom already in `src/features/permissions/qwencode-permissions.ts`: a module-level `const moduleLogger: Logger = fallbackLogger`, which is configured from the CLI flags and the resolved config, so `--silent` is honored. That is what makes the `re:` warning above possible.

### 3. `[tools.bash]`'s own `sensitive_patterns` are not sorted

Fixed from the other side: the fan-out now copies the guard onto the alias shells **in authored order** instead of sorting only the copy. `[tools.bash]`'s list stays as authored — it is a key rulesync does not write, and reordering it is a diff for nothing — and the three shells now spell one guard one way.

### 4. Extract the first pass of the fan-out shell resolver

`classifyFanOutShellCandidates` now answers "what does each table already say", and `resolveFanOutShellAliases` is left with "what to tell the user about it". Pure refactor.

### 5. `warnOnStrandedShellPatterns` scope

Left as-is, and the reason is now recorded in its doc comment: a shell carrying patterns while no `bash` category exists at all is holding a list rulesync never touched, so reporting it would be commentary on the user's config rather than on what this generate did, and it would fire on every unrelated run.

### 6. Two warnings contradict each other on a non-list `denylist`

The stand-down warning now promises the deny merge only where it actually happens — the shell is on the registry, its deny keys are well-formed lists, and the `bash` category has deny patterns to spread. Otherwise it says nothing from the category reaches the shell, which is true both for a scalar `denylist` (where the malformed-key warning follows with the reason) and for a `bash` category that carries no deny patterns at all. `shellDenyPatternsOf` and `malformedDenyKeysOf` are now read by both the warning and the merge, so the two cannot disagree.

### Verification

15 new tests, each checked by mutation to fail when its source decision is reverted. `pnpm cicheck` green. `docs/reference/file-formats.md` updated for the glob semantics, the `re:` exclusion, the generate-side warning, the copy order and the softened stand-down wording.


dyoshikawa:
@dyoshikawa Thank you!

PR #2824, opened 2026-08-29
feat(deepagents): sync the dcode shell allow list and fix the global directory

### LINKED ISSUE #2764: Follow up deepagents upstream updates: global scope writes to ~/.deepagents/deepagents, which dcode does not read
## Summary

Global-scope generation for `deepagents` writes rules, skills and subagents into `~/.deepagents/deepagents/`, but dcode's default agent name is `agent`, so nothing there is loaded unless the user explicitly launches `dcode -a deepagents`. Separately, the `permissions` precondition tracked on the closed #2067 has been met and a second config surface (`[startup].mode`) has since shipped.

## Recent Releases

Upstream identity note: the package rulesync targets is **`deepagents-code`** (`dcode`), from `langchain-ai/deepagents` (`libs/code`). The separate `deepagents-cli` PyPI package is deprecated. Latest release at the time of this research is **0.1.61, 2026-08-24** ([PyPI](https://pypi.org/project/deepagents-code/), [CHANGELOG](https://github.com/langchain-ai/deepagents/blob/main/libs/code/CHANGELOG.md)).

- **0.1.35** (2026-07-09, #4573) — adds `[startup].mode`, a default approval mode (`manual` / `auto` / `yolo`).
- **0.1.40** (2026-07-16) — plugin marketplace; **0.1.42** (2026-07-17) — plugins GA.
- **0.1.48** (2026-07-27, #5076) — a directory under `~/.deepagents/` is recognized as an agent only when it contains an `AGENTS.md` marker (`agent.py:get_available_agent_names`).
- **0.1.59** (2026-08-20) — administrator policy via `managed_config.toml`.
- **0.1.61** (2026-08-24) — begins enforcing `[models].allowed`.

Documentation moved: every `https://docs.langchain.com/oss/javascript/deepagents/cli/*` URL now returns **308** and redirects to `https://docs.langchain.com/oss/deepagents/code/*`.

### 1. rules / skills / subagents — global scope writes to a directory dcode does not read by default

- **Upstream**: the user-level agent directory is `~/.deepagents/<agent_name>/`, and the default `agent_name` is **`agent`**, not `deepagents`. `libs/code/deepagents_code/_constants.py` declares `DEFAULT_AGENT_NAME: Final[str] = "agent"` (value unchanged since the `libs/cli` port, commit `2ac7d4153`, 2026-05-12), resolved into paths via `_resolve_agent_arg` (`main.py`) → `assistant_id` → `Settings.get_user_agent_md_path` / `get_user_skills_dir` / `get_user_agents_dir` (`config.py`). The docs say the same: [memory and skills](https://docs.langchain.com/oss/deepagents/code/memory-and-skills) — "Global installs (`-g`) symlink skills into `~/.deepagents/agent/skills/` — the default agent's user-level skills directory."
- **rulesync**: `global`. `DEEPAGENTS_GLOBAL_DIR = join(DEEPAGENTS_DIR, "deepagents")` (`src/constants/deepagents-paths.ts:4`), asserted as "default agent_name `deepagents`" in `src/features/rules/deepagents-rule.ts:57-58`, `src/features/skills/deepagents-skill.ts:86-87` and `src/features/subagents/deepagents-subagent.ts:71-73`. A `--global` dry run writes `~/.deepagents/deepagents/AGENTS.md`, `~/.deepagents/deepagents/agents/*/AGENTS.md` and 41 skills under `~/.deepagents/deepagents/skills/`. None of it is loaded by a default `dcode` session.
- **Compounding**: since 0.1.48 a directory under `~/.deepagents/` is only recognized as an agent when it contains an `AGENTS.md` marker, so a `~/.deepagents/deepagents/skills/` tree generated without the rules feature is not even selectable via `/agents`.
- This is a standing mismatch rather than a regression — the default was never `deepagents`.

### 2. permissions — the precondition #2067 was closed on has been met, and a second key has since shipped

- **Upstream**: `~/.deepagents/config.toml` carries `[shell].allow_list` (shell commands auto-approved without review; comma-separated string or TOML array, plus the sentinels `recommended` and `all`; matched against the **first token** of each pipeline segment by `is_shell_command_allowed`, no globs) and `[startup].mode` (default approval mode, added in 0.1.35, 2026-07-09, #4573). Both are declared with `toml_keys` in `config_manifest.py` (`shell.allow_list`, `startup.mode`), so they are genuine config-file keys rather than env-only. The published [configuration page](https://docs.langchain.com/oss/deepagents/code/configuration) documents the env-var form `DEEPAGENTS_CODE_SHELL_ALLOW_LIST` but not the TOML keys; the manifest and the `dcode config` surface are authoritative.
- **rulesync**: `unsupported`. The global dry run logs `Target 'deepagents' does not support the feature 'permissions'. Skipping.`; there is no `src/features/permissions/deepagents-permissions.ts`.
- **Relationship to #2067**: that issue was closed 2026-06-30 as no-action because `shell.allow_list` had no `toml_keys` at the time, with the explicit condition "If dcode later adds `toml_keys` to `shell.allow_list` … a global-only `DeepagentsPermissions` adapter … would become actionable." A 2026-08-23 comment there records that the condition is now met and works out the adapter design in full. This issue restates it only because #2067 is closed and the work has no open home; the design there stands and should be followed, with `[startup].mode` added to it as a second mapping target.
- **Scope**: global only. `model_config.py:DEFAULT_CONFIG_PATH` is `~/.deepagents/config.toml` and `configuration/service.py` merges only that plus managed policy — there is no project `.deepagents/config.toml`. `managed_config.toml` (`/etc/dcode/`, `/Library/Application Support/dcode/`, `%ProgramData%\dcode\`) stays out of scope as an IT-admin deployment file, the same exclusion rulesync applies to other MDM-style config.

## Proposed Follow-up

1. Change `DEEPAGENTS_GLOBAL_DIR` to `join(DEEPAGENTS_DIR, "agent")` so global rules, skills and subagents land in `~/.deepagents/agent/…`, and update the three adapter comments that assert the wrong default.
2. Keep global rules generation coupled to the agent directory so the `AGENTS.md` marker always exists when skills or subagents are written there — otherwise the tree is invisible to `/agents` under 0.1.48+.
3. Leave global MCP (`~/.deepagents/.mcp.json`) and global hooks (`~/.deepagents/hooks.json`) where they are: both are already correct and must stay outside the agent directory.
4. Add a global-only `DeepagentsPermissions` per the design recorded on #2067: canonical `bash` entries with action `allow` → `[shell].allow_list` as a TOML array (taking the executable token from patterns like `git *`), `{"*": "allow"}` → `allow_list = "all"`, warn-and-skip `deny` / `ask` since upstream's allowlist has no deny form (a deny needs a `PreToolUse` hook instead), writing through the shared TOML gateway so unrelated sections survive. Optionally map a default-mode override onto `[startup].mode`. Note in the doc comment that `shell.allow_list` is one of the managed-policy enforced keys, so the emitted user-tier value is not guaranteed to be the effective one on a managed install.
5. Refresh `.claude/skills/rulesync-feature-research/references/deepagents.md`: every `/oss/javascript/deepagents/cli/*` URL is a 308 to `/oss/deepagents/code/*`, and its `skills` row writes `~/.deepagents/<agent>/skills` without noting that `<agent>` defaults to `agent` — which is what led the adapters to the wrong global path in the first place.

Confirmed non-gaps this run, so they need no work: `commands` (all slash commands are built-in — `libs/code/COMMANDS.md` is generated from `command_registry.py`, and plugin manifests explicitly reject `commands/` and `agents/` components via `plugins/manifest.py:_UNSUPPORTED_COMPONENT_DIRS`), `ignore` (no ignore-file surface anywhere in `libs/code`), `checks` (goals and rubrics are session state in `goal_rubric.py`, not authored files), `mcp` (all three config locations, both transports plus the `streamable_http` alias, tool filters and `${VAR}` expansion are covered), and `hooks` (all 12 events are in `DEEPAGENTS_HOOK_EVENTS`; `argv` is not emitted but `command` is required upstream even when `argv` is set, so import is never lossy). One dated note: rulesync's legacy flat-list hooks importer becomes dead code once upstream removes `hooks.legacy` / `hooks.migration` **after 2026-09-01**, a deadline reconfirmed verbatim in `hooks/models/config.py`.

## References

- https://pypi.org/project/deepagents-code/ — the current package and the 0.1.61 (2026-08-24) release date.
- https://github.com/langchain-ai/deepagents/blob/main/libs/code/CHANGELOG.md — dates for 0.1.35, 0.1.40, 0.1.42, 0.1.48, 0.1.59 and 0.1.61.
- https://docs.langchain.com/oss/deepagents/code/memory-and-skills — states that global installs land in `~/.deepagents/agent/skills/`, the direct evidence for gap 1.
- https://docs.langchain.com/oss/deepagents/code/configuration — the current configuration reference; also the redirect target for the retired `/oss/javascript/deepagents/cli/*` URLs.
- https://github.com/dyoshikawa/rulesync/issues/2067 — the closed permissions issue whose reopening condition is now met, and where the adapter design already lives.

dyoshikawa:
## Summary

Two follow-ups for the deepagents-cli (dcode) target.

**1. Global directory name.** dcode keeps user-level context under a directory named after the agent, and that name defaults to `agent` (`DEFAULT_AGENT_NAME` in deepagents-code), not to the product name — so a global install lands in `~/.deepagents/agent/`, not `~/.deepagents/deepagents/`. `DEEPAGENTS_GLOBAL_DIR` and every comment, unit test, and e2e expectation that spelled the old path are corrected.

**Migration:** anything an earlier version generated under `~/.deepagents/deepagents/` (`AGENTS.md`, `skills/`, `agents/`) is orphaned by this rename — `--delete` only sweeps the paths the target currently resolves, so delete that directory by hand once dcode no longer reads it.

**2. Permissions support.** New `DeepagentsPermissions` adapter writing `[shell].allow_list` in the user config `~/.deepagents/config.toml`.

- **Global only.** dcode reads no project-level config file, so `supportsProject: false`; `fromFile` / `fromRulesyncPermissions` throw a clear message when `--global` is missing.
- **Executable reduction.** dcode auto-approves a command when the first token of every pipeline segment is in the list, matched exactly with no globs. Canonical `bash` rules are therefore reduced to their executable: `git *`, `git:*`, and `git commit:*` all emit `git`. The last one widens the rule to every `git` subcommand, so it warns.
- **`*` → `["all"]`.** The sentinel auto-approves everything *and* skips dcode's dangerous-pattern check, so it warns; upstream rejects the option outright when `all` shares the list, so it is emitted as the sole entry.
- **Lossy cases warn rather than degrade silently:** a token that still holds a glob (`npm-*`) would match no command and is skipped; a pattern reducing to a sentinel name (`all`, `recommended`) is skipped; `deny` rules have no counterpart in dcode and are reported as skipped; `ask` needs no output since an unlisted command already prompts.
- **Import.** Entries come back as `bash` allow rules named by the executable, `all` as `*`, and a list mixing `all` with names as nothing (upstream raises on that combination and ignores the option). `recommended` is dropped rather than expanded — the curated list behind it is upstream's to edit.
- **`[startup]`** (`mode`, `yolo_switcher`, `read_project_dotenv`) is authored through the `deepagents` override namespace and round-trips. `startup.recent` is deliberately excluded: dcode rewrites it per session, so committing it would publish one machine's state.

- **Deny beats the `*` allow.** `all` also turns off dcode's dangerous-pattern check, so writing it for a config that denies something would hand the user *less* than dcode's own default; the wildcard is dropped with a warning instead.
- **Rules the reduction would invert are warned about, not just the ones it drops.** An `ask` or `deny` on a narrower pattern than an `allow` beside it (`npm publish` against `npm *`) collides on the executable and dcode keeps the allow, so the command runs with no prompt at all.
- **The `startup` override is a global-config write from a repository-carried file**, so a value that relaxes what dcode does on its own (`mode` of `auto`/`yolo`, `yolo_switcher`, `read_project_dotenv`) is named in a warning alongside the value it replaces.

The generate path merges in place through `smol-toml` and reads with `readFileContentOrNull`, so unrelated tables are preserved and a dry run never creates the user's global file.

## Test plan

- New `deepagents-permissions.test.ts` (49 cases): both global-only throws, executable reduction and widening, the `all` sentinel, glob/sentinel skips, deny reporting, in-place merge preserving unrelated tables, dropping an emptied `[shell]`, no side-effect file creation, the startup override, malformed TOML, and generate → import round trip.
- New e2e case generating deepagents permissions into the pseudo-home with `--global`.
- `pnpm cicheck` green.

Closes #2764


dyoshikawa:
@dyoshikawa Thank you!

PR #2828, opened 2026-08-30
fix(permissions): read the all-tools category in the command-only adapters

### LINKED ISSUE #2825: A deny under the all-tools `*` category escapes the shadowed-rule warning
A canonical `deny` written under the all-tools `*` category is reported by the deepagents and Warp permissions adapters only as "cannot be represented and were skipped", never as the stronger "those commands run without a prompt" warning that a `bash`-category deny gets when an `allow` beside it shadows it.

Repro with `.rulesync/permissions.jsonc`:

```jsonc
{
  "permission": {
    "*": { "rm *": "deny" },
    "bash": { "rm *": "allow" }
  }
}
```

`rulesync generate --targets deepagents --global` writes:

```toml
[shell]
allow_list = [ "rm" ]
```

and warns only that the `*` deny rules could not be represented. But the executable-name reduction has inverted the author's intent exactly as it would for a `bash`-category deny: `rm` is now auto-approved with no prompt. The narrow-vs-broad collision warning exists for precisely this case and does not fire.

`src/features/permissions/deepagents-permissions.ts` folds every non-`bash` category into `hasForeignDeny`, and `src/features/permissions/warp-permissions.ts` treats `*` the same way, so this is a shared convention rather than a bug in one adapter. Deciding it centrally — should the `*` category's rules be considered alongside the tool's own category when checking for a shadowed rule? — is the useful fix, and it needs the same answer in both adapters.

Note that the `all` sentinel is already blocked correctly by `hasForeignDeny`; only the per-command shadow warning is missing.

Found while reviewing #2824.

dyoshikawa:
## Summary

A rule written under the canonical all-tools `*` category applies to every tool, shell commands included. The three adapters whose tool models shell commands and nothing else — deepagents-cli, Warp, Factory Droid — each folded every category other than `bash` into one "cannot be represented and were skipped" warning, so the `*` category never reached the command lists.

That inverts the author's intent. With

```jsonc
{ "permission": { "*": { "rm *": "deny" }, "bash": { "rm *": "allow" } } }
```

deepagents wrote `allow_list = ["rm"]` and never reached its stronger shadowed-rule warning, while Warp and Factory Droid auto-approved the very command the file blocks.

## Changes

- New `src/features/permissions/shell-command-categories.ts` decides it centrally, as the issue asks:
  - `collectShellCommandRules` contributes every `bash` rule plus the `*` category's **restricting** rules (`deny`, `ask`), and never its `allow` rules — a pattern under `*` need not be a command at all (`secrets/**` there denies a path), so both directions fail closed.
  - `partitionCommandRules` splits those into the allow/deny lists a two-tier tool writes, withholding an `allow` that an `ask` names the same pattern for, since `deny > ask > allow`. A `deny` needs no such treatment: the tool's own denylist already outranks its allowlist.
- `deepagents-permissions.ts`, `warp-permissions.ts` and `factorydroid-permissions.ts` consume the helper; the foreign-deny warning still names only the categories that actually carry a `deny`.
- `cline-permissions.ts` had the same hole and is fixed alongside them. Cline documents no deny-priority between its two lists, so there an `allow` naming a pattern the config restricts elsewhere is **withheld** rather than written beside the deny, and the aggregated translation notice names what it withheld.
- `docs/reference/file-formats.md` documents the behavior for all three tools (`src/generated/docs-content.ts` regenerated).

## Testing

- `pnpm cicheck` — green (394 files, 9872 tests).

Closes #2825


dyoshikawa:
Automated review loop reached its 10-iteration safety cap, so this PR is left open for a manual decision rather than auto-merged.

Round 10 reported 1 high + 2 mid findings; all three were fixed in the two commits pushed after it:

- **high** — deepagents-cli matched restrictions by leading token only, so `{"bash": {"kubectl *": "allow"}, "*": {"*delete*": "deny"}}` still wrote `allow_list = ["kubectl"]` and auto-approved `kubectl delete`. Each allowlist entry is now intersected against the two patterns it actually approves (`token` and `token *`) using the budgeted glob intersection, so such a restriction withholds the allow.
- **mid** — the all-tools `*` ask that withheld nothing is reported again (narrowed to `fromAllToolsCategory` asks not also written under `bash`), while a plain `bash` ask that withheld nothing stays silent.
- **mid** — deepagents' intersection-budget exhaustion is now surfaced in its warning, as the other three adapters already did.

Also fixed from the low findings: the astral-plane surrogate split in `warpCommandPatternToGlob` (a narrowing, the one direction the rewrite must never take), the dead `foreignDenyCategories` field, the `INTERSECTION_PAIR_COST` doc comment, `\\PL` / `\\P{...}` test coverage, and the pair-count test that did not exercise the pair floor.

All checks are green. Please review and merge manually.

dyoshikawa:
@dyoshikawa Thank you! Merged. The deferred review items are tracked in #2859.

PR #2832, opened 2026-08-30
fix(logger): carry suppressed warnings to the caller that can act on them

### LINKED ISSUE #2737: JsonLogger drops warnings, so --json runs never see security-relevant notices
## Problem

`JsonLogger` implements `info()` and `warn()` as no-ops, and `ConsoleLogger` is silenced by `--silent`. Every adapter warning therefore disappears when rulesync runs as `rulesync generate --json` — including notices that exist specifically to tell a user that something security-relevant was preserved or is not in effect.

Concrete examples, all reachable today:

- `rovodev-permissions.ts`: "keeping `toolPermissions.bash.runInSandbox: false`" — a sandbox opt-out that a regenerate does not reset.
- `rovodev-mcp.ts`: "leaving `mcp.mcpConfigPath` as ..." and the stale-pointer warning — the project either runs servers the user did not expect, or runs none at all.
- The various `warnAboutDroppedOwnedKeys()` messages across the permissions adapters.

`--json` is the mode most likely to be used from CI or another tool, i.e. exactly where nobody is watching a terminal. A silent no-op means the only signal is lost, not merely reformatted.

## Proposed direction

Options, roughly in order of increasing effort:

1. Route `warn()`/`error()` on `JsonLogger` to `stderr` as plain lines, leaving `stdout` reserved for the JSON payload. Machine consumers parse `stdout` only, so this is compatible.
2. Collect warnings and emit them as a `warnings: string[]` field of the JSON result.
3. Both: `stderr` for humans tailing logs, a structured field for programmatic consumers.

Also decide what `--silent` should mean for `warn()` — arguably it should still suppress `info()` but not warnings.

## References

- `src/utils/logger.ts`
- Raised during the review of the PR closing #2726; it is a general concern, not specific to the Rovo Dev adapters.

### LINKED ISSUE #2823: Machine-local settings warning never reaches MCP, --json or --silent imports
`readSettingsWithLocalOverlay` warns when an import read values out of a machine-local settings file (`.factory/settings.local.json`, `.augment/settings.local.json`), naming the keys and calling out the ones that decide what the tool may do. That warning is the only mitigation for a personal value silently becoming the team's on the next generate — and it only ever reaches `logger.warn`.

Three paths drop it:

- `src/mcp/import.ts` builds `new ConsoleLogger({ verbose: false, silent: true })`, whose `warn` is a no-op, and `McpImportResult` carries counts only. An import driven through `mcp__rulesync__rulesyncTool` therefore takes `enabledPlugins`, `maxAutonomyLevel`, `sandbox`, `hooks` and friends out of a machine-local file with no output at all, and the calling agent has nothing to report.
- `--json` (`src/cli/wrap-command.ts`) uses `JsonLogger`, whose `warn` is also a no-op, and the message is not added to `logger.captureData`, so it appears nowhere in the JSON.
- `--silent` suppresses it by design, which is fine on its own but combines with the point below.

`warnOnceWithFallback` also claims the once-per-run token before looking at the logger, so in a long-lived MCP process a suppressed read spends the token a later unsuppressed read would have needed.

This is not specific to the local-settings warning — every `logger.warn` diagnostic rulesync emits is invisible in these modes — so it is filed separately rather than fixed inside #2765's follow-ups (PR #2820).

Candidate directions:

- give `McpImportResult` a `warnings` array and surface it in the tool result;
- have `JsonLogger.warn` collect into `captureData("warnings", ...)` so `--json` consumers see it;
- make `warnOnceWithFallback` skip claiming the token when the destination logger is suppressed.

Filed from the review of PR #2820.

dyoshikawa:
A warning about the files a run just read had three ways to vanish before anyone could act on it:

- the MCP server reads with a silent `ConsoleLogger`, and its stderr never reaches the calling agent anyway;
- `JsonLogger.warn` was a no-op, so a `--json` consumer — which reads the document and nothing else — never saw one;
- `warnOnceWithFallback` claimed the run's once-token before consulting the logger, so a suppressed read could spend the token a later, listening read needed.

## Changes

- `JsonLogger` keeps its warnings and emits them as a top-level `warnings` array (key omitted when empty). Top-level rather than inside `data`, so a command's own captured keys cannot collide with it and warnings are still reported when the command failed. `--silent` still suppresses them, as it does on the console.
- `warnOnceWithFallback` leaves the token unspent for a destination that reports nothing, so a later run still warns about the same file.
- `Logger` gains a required `reportsWhileSilent` member for a destination that hands warnings back to its caller rather than printing them — such a logger reports even while `silent`, so it still spends the token. Required rather than optional so every wrapper has to answer the question.
- `WarningCollectingLogger` is that destination. The MCP `generate` / `import` / `convert` tools hand one in and return what it collected as `warnings` in their result — on failures too, since a run that warned and then failed is exactly when the warnings matter.
- `withFallbackLoggerTarget` lets an operation adopt the shared fallback, so a warning raised on a path with no logger threaded through lands wherever that operation's other diagnostics land. The target — and the once-per-run record — are scoped with `AsyncLocalStorage`: the MCP server does not serialize requests, so two runs in flight at once must not steal each other's warnings or spend each other's tokens.
- The forwarder is diagnostics-only: `configure` reaches the console it falls back to, while `outputJson` / `captureData` are no-ops so a call from an unattended path cannot shape someone else's `--json` document.

## Bounding what a warning carries

Collected warnings now travel into documents another program parses and into MCP results an agent reads as context, and their text quotes files rulesync did not write. So they are stripped of control characters, de-duplicated, and bounded: at most 100 warnings, 1,000 characters each, 8,000 in total (a ceiling, not a line the last entry steps over), with a final entry naming how many distinct diagnostics were dropped. The values a warning quotes go through one shared `quoteValueForWarning` (serialized, control-stripped, truncated at 60 characters), now used by the hooks converter, the Cursor permissions writer, and the Musecode / Rovodev MCP writers. Control characters are stripped before `JSON.stringify` can escape them, since an escaped ESC survives as six literal characters that no later pass can recognize. Truncation is one shared `truncateText`, which cuts by code point and never leaves the backslash of an escape as the last character. The error channel is bounded the same way where it reaches a caller: `formatError`'s Zod expansion is capped, and so are the unreadable sources an MCP `generate` failure names.

Out of scope, deliberately: the module-level silent loggers behind the MCP `rules` / `commands` / `skills` / `subagents` / `checks` operations, which report through their own result shapes rather than by warning.

## Docs

- `docs/reference/cli-commands.md`: a `JSON Output` section describing the top-level `warnings` array and its limits.
- `docs/reference/mcp-server.md`: the `warnings` array on the `generate` / `import` / `convert` results, and that those three are the only operations that report one.

## Verification

`pnpm cicheck` is green (395 files / 9904 tests). The new behaviors were each confirmed to fail without their fix: the overlapping-operations test fails against a save/restore fallback, the once-per-run scope test fails without its `AsyncLocalStorage`, the `--silent` fallback test fails without `wrapCommand`'s `configure` call, the doctor test fails without its `jsonMode` guard, and the MCP test for a warning raised where no logger was threaded through fails when the executor's fallback scope starts after config resolution.

Closes #2823
Closes #2737


dyoshikawa:
@dyoshikawa Thank you!

PR #2833, opened 2026-08-30
refactor(checks): extract a shared base class for the aggregated checks adapters

### LINKED ISSUE #2821: refactor(checks): extract a shared base class for the aggregated checks adapters
`CursorCheck`, `RovodevCheck` and `FactorydroidCheck` now write the same shape — one aggregated instruction file whose sections are marked by the HTML comments in `aggregated-check-file.ts`. That module holds the rendering and splitting, but the adapter skeletons around it are near-verbatim copies: `canDeleteAuxiliaryFiles`, `fromRulesyncChecks`, `fromFile`, `forDeletion`, `toRulesyncCheck`, `toRulesyncChecks`, `validate` and the `fromRulesyncCheck` refusal differ only in the class name, the settable paths, the fallback check name and the wording of the replace warning.

A shared `AggregatedToolCheck` base taking those four as parameters would collapse the three, so a fix like the frontmatter strip added in #2820 lands in one place instead of three.

Raised by the reviewer on https://github.com/dyoshikawa/rulesync/pull/2820 and deliberately left out of it: the change touches the Cursor and Rovo Dev adapters, which that PR otherwise does not, so it belongs in a refactor of its own.

dyoshikawa:
Cursor Bugbot, Rovo Dev and Factory Droid each write one aggregated instruction file, and each carried a near-verbatim copy of the adapter around `aggregated-check-file.ts`: the same path resolution, the same deletion guard, the same `fromRulesyncChecks` / `fromFile` / `forDeletion` bodies.

This moves that skeleton into a new `AggregatedToolCheck` base and leaves each adapter with what actually differs.

## What each adapter keeps

- `getSettablePaths()` — the directory and the name of the one file it writes.
- `getAggregatedCheckConfig()` — the display name used in messages, the tool target, the fallback check name for a file with no marked section, and what generating does to a file holding instructions rulesync did not write (`replace` for Cursor and Rovo Dev, `skip` with a tool-specific warning for Factory Droid).
- Factory Droid additionally keeps `stripSkillFrontmatter`, wired in as the config's optional `transformImportedContent`, since its output path is shared with the skills feature and so can hold frontmatter.

## Notes

- The base is not declared `abstract`: its statics build the subclass with `new this(...)`, which an abstract constructor type forbids. What a subclass owes is enforced the way this codebase enforces it on statics — `getAggregatedCheckConfig()` throws "Please implement this method in the subclass." until overridden.
- No behavior change is intended; the existing per-adapter tests (Cursor, Rovo Dev, Factory Droid) pass unchanged, and a new `aggregated-tool-check.test.ts` covers the two throw paths the base adds.

Closes #2821

### DIFF
--- a/src/features/checks/aggregated-tool-check.test.ts
+++ b/src/features/checks/aggregated-tool-check.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "vitest";
+
+import { AggregatedToolCheck, type AggregatedToolCheckConfig } from "./aggregated-tool-check.js";
+import { type ToolCheckSettablePaths } from "./tool-check.js";
+
+/**
+ * A subclass that forgets `getAggregatedCheckConfig`. TypeScript cannot demand
+ * a static, so the base throws instead — these cover that the throw is reached
+ * rather than a default silently taken.
+ */
+class ForgetfulCheck extends AggregatedToolCheck {
+  static override getSettablePaths(): ToolCheckSettablePaths {
+    return { relativeDirPath: ".forgetful", relativeFilePath: "CHECKS.md" };
+  }
+}
+
+/** A subclass that names a directory but not the one file it writes. */
+class UnnamedFileCheck extends AggregatedToolCheck {
+  static override getSettablePaths(): ToolCheckSettablePaths {
+    return { relativeDirPath: ".unnamed" };
+  }
+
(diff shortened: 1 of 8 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2834, opened 2026-08-30
feat(subagents): support ZCode global subagents

### LINKED ISSUE #2819: Support ZCode subagents (~/.zcode/agents/) once the subagents processor can express global-only targets
## Background

#2780 added the `zcode` target (see PR #2818) covering rules, commands, skills and MCP. ZCode also has a **subagents** surface, which was deliberately left out of that PR.

## What ZCode documents

ZCode subagents are Markdown files with YAML frontmatter placed at:

```
~/.zcode/agents/<name>.md
```

Per the [ZCode subagents docs](https://zcode.z.ai/en/docs/subagents), the feature is in **Beta** and only the **user (global) scope** is documented — there is no documented `<project>/.zcode/agents/` workspace location the way commands and skills have one.

## Why it was deferred

The subagents processor's per-target `meta` has no `supportsProject` flag (unlike the commands and skills processors), so a target registered there is implicitly assumed to support project scope. There is no way to express "global-only" for subagents today, and registering `zcode` as-is would make `rulesync generate --features subagents` write a project-scoped `.zcode/agents/` directory that ZCode never reads.

## What needs to happen

1. Decide how the subagents processor should express global-only support — most likely by adding a `supportsProject` flag to its meta, mirroring `commands-processor.ts` / `skills-processor.ts`.
2. Add `ZcodeSubagent` (`src/features/subagents/zcode-subagent.ts`) writing `~/.zcode/agents/<name>.md`, register it in the subagents processor factory and in `subagentsProcessorToolTargetTuple`.
3. Model the documented frontmatter keys, preserving unknown keys through a `zcode:` block as the other adapters do.
4. Add the e2e global subagents matrix entry and update the generated tables / `.gitignore`.

Blocked on step 1; worth confirming against the upstream docs first, since the feature is still marked Beta and its frontmatter may change.

dyoshikawa:
## Summary

Adds ZCode subagents support. ZCode's Beta "manages global / user-level subagents stored under `~/.zcode/agents/`" and states that creating workspace / project-level subagents "is not available yet" ([docs](https://zcode.z.ai/en/docs/subagents)), so this is the first subagents target that is global-only.

## Changes

- **`supportsProject` in the subagents processor.** The factory `meta` gains the flag the commands and skills processors already carry, and `subagentsProcessorToolTargets` is derived from it instead of taking every key. Every existing target is `supportsProject: true`, so the project matrix is unchanged.
- **`ZcodeSubagent`** (`src/features/subagents/zcode-subagent.ts`), registered with `supportsProject: false, supportsGlobal: true`. One Markdown file per agent with YAML frontmatter, written to `~/.zcode/agents/<name>.md`.
- **Frontmatter model.** ZCode's keys are camelCase and case-sensitive: `name` and `description` required; `model`, `thoughtLevel`, `color`, `tools`, `disallowedTools`, `maxTurns`, `injectAgentsMd`, `mcpServers` optional. They ride the canonical subagent's `zcode:` block, and unknown keys round-trip through the same block (`z.looseObject`).
- **`zcode` added to `subagentsProcessorToolTargetTuple`**, plus the global e2e subagents matrix entry.
- Regenerated the supported-tools tables (ZCode's Subagents cell is 🌏-only) and `src/generated/docs-content.ts`; added the ZCode note to the subagents section of `docs/reference/file-formats.md`. `.gitignore` needed no change, since the derivation already skips targets that do not support project scope.

`pnpm cicheck` passes.

Closes #2819


dyoshikawa:
@dyoshikawa Thank you!

PR #2835, opened 2026-08-30
refactor(fetch): fold the repeated prune mocks and the duplicated helpers

### LINKED ISSUE #2816: fetch: cleanups left over from the skill-name confusability work
Cleanups noticed while reviewing #2810, none of which change behavior:

- `src/lib/fetch.test.ts`: the `fetchFiles skill pruning` describe still repeats the same ~30-line single-skill `listDirectory` mock five times. The neighbouring describe now has `mockSkillRepositoryWithSkills`; a `mockSingleSkillRepository(name)` would fold these too.
- `src/utils/confusable-names.test.ts`: `CYRILLIC_COPY` is defined at module scope and then redefined locally as `cyrillicCopy`; the expected note `"another entry differs from it only by lookalike letters"` is spelled out twelve times.
- `src/lib/fetch.ts`: `isFileSystemError` is defined locally although `src/utils/file.ts` already exports one.
- `src/lib/fetch.ts`: `readSkillRootNames` passes `SKILLS_DIR_PREFIX` — a trailing-slash path prefix — to `join` as a directory name.
- `src/utils/control-characters.ts`: the comment line describing LRM/RLM/ALM is 114 columns after the last edit, against ~80 for the rest of the file. `oxfmt` does not reflow comments.

dyoshikawa:
Cleanups left over from the skill-name confusability work reviewed in #2810, one of which carries a deliberate behaviour difference (below).

- `src/lib/fetch.test.ts`: the `fetchFiles skill pruning` describe inlined the same single-skill `listDirectory` mock in eight tests that differ only in the remote skill name. They now share a `mockSingleSkillRepository(name)` helper, mirroring `mockSkillRepositoryWithSkills` in the neighbouring describe.
- `src/utils/confusable-names.test.ts`: the two repeated notes are now the module-scope `TWIN_NOTE` and `SAME_FORM_NOTE`, and the local `cyrillicCopy` is gone in favour of the module-scope `CYRILLIC_COPY` (whose comment now carries the rationale for the escaped spelling).
- `src/lib/fetch.ts`: dropped the local `isFileSystemError` in favour of the one `src/utils/file.ts` already exports (it walks the `cause` chain and checks the code against the errno shape).
- `src/lib/fetch.ts`: `readSkillRootNames` handed `join` the trailing-slash `SKILLS_DIR_PREFIX`; it now takes the `SKILLS_DIR_NAME` the prefix is built from.
- `src/utils/control-characters.ts`: rewrapped the 114-column LRM/RLM/ALM comment to the ~80 the rest of the file uses.

The behaviour difference: the shared `isFileSystemError` recognizes an I/O failure by its `errno` code shape, so a `code` of another shape — the `UNKNOWN` libuv falls back to on Windows for a system error it cannot translate — is now rethrown by the prune error handler instead of being absorbed into a `Stopped partway` warning. That is the side to err on: a prune silently skipped is worse than a fetch that says why it stopped. The comment at the call site records it.

`pnpm cicheck` passes.

Closes #2816

### DIFF
--- a/src/lib/fetch.test.ts
+++ b/src/lib/fetch.test.ts
@@ -2317,6 +2317,47 @@ describe("fetchFiles skill pruning", () => {
     mockClientInstance.getFileContent.mockResolvedValue("# Skill");
   }
 
+  /**
+   * A repository with a single skill of the given name, whose only file is
+   * SKILL.md. The tests that pin how a remote skill name is read back differ
+   * only in that name.
+   */
+  function mockSingleSkillRepository(name: string): void {
+    mockClientInstance.listDirectory.mockImplementation(
+      (_owner: string, _repo: string, path: string) => {
+        if (path === "skills") {
+          return Promise.resolve([
+            {
+              name,
+              path: `skills/${name}`,
+              type: "dir",
+              sha: "aaa",
+              size: 0,
+              download_url: null,
+            },
+          ]);
+        }
+        if (path === `skills/${name}`) {
+          return Promise.resolve([
+            {
+              name: "SKILL.md",
+              path: `skills/${name}/SKILL.md`,
+              type: "file",
+              sha: "bbb",
+              size: 100,
+              download_url: "https://example.com",
+            },
+          ]);
+        }
+        const error = new Error("Not found");
+        Object.assign(error, { statusCode: 404 });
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #51, opened 2025-07-21
Add support for Augment Code


mikelros:
I'd like to request support for [Augment Code](https://docs.augmentcode.com/setup-augment/guidelines#working-with-rules) generation and import.

cm-dyoshikawa:
@mikelros Thank you for your issue.
I'd like to work on it.

denka9999:
Sorry but I don't know if this solution works for JetBrain IDEs. Augment for JetBrains doesn't support or work with the .augment/rules/*.md format. Instead, the documentation mentions having a plain text file at the root level called `.augment-guidelines`. However, what the proposed solution in #53  does is import the rules from the .augment/rules/*.md directory. I think that this solution will work with VSCode IDE

[Here](https://docs.augmentcode.com/jetbrains/setup-augment/guidelines) you can check the way to configure a guideline for JetBrain IDE.

cm-dyoshikawa:
@mikelros Thank you for your comment.

I'm going to add support for `--augmentcodeLegacy` also.

cm-dyoshikawa:
I tried to implement augment code support.

https://github.com/dyoshikawa/rulesync/releases/tag/v0.49.0

If bugs exist, please create issues.

PR #2837, opened 2026-08-30
fix(fetch): accept emoji keycap names and mark whitespace that does not show

### LINKED ISSUE #2815: fetch: keycap emoji names are refused, and whitespace-only names go unmarked
Two accuracy gaps in the name checks added in #2810:

1. **Keycap emoji are refused.** `hasDeceptiveHiddenCharacters` in `src/utils/control-characters.ts` allows a variation selector only next to a character matching `JOINING_CONTEXT_PATTERN`, whose emoji arm is `\p{Extended_Pictographic}`. The keycap bases — `#`, `*` and the ASCII digits — are `Emoji` but not `Extended_Pictographic`, so a directory named with a keycap sequence such as U+0031 U+FE0F U+20E3 is reported as deceptive and skipped, even though it is an ordinary emoji name.

2. **Whitespace-only differences go unmarked.** Two names that differ only in leading or trailing whitespace share a display form and are therefore noted; a single name with trailing whitespace and no twin is shown with nothing to say that the row extends past what is visible.

dyoshikawa:
Two accuracy gaps in the skill name checks added in #2810.

**Keycap emoji names were refused.** `hasDeceptiveHiddenCharacters` allowed a variation selector only beside a character matching `JOINING_CONTEXT_PATTERN`, whose emoji arm is `\p{Extended_Pictographic}`. A keycap sequence is built on a digit or an ASCII sign — `1`, `#`, `*` are `Emoji` but not `Extended_Pictographic` — so the joining list could not see it, and a directory named `1️⃣` was dropped as a name padded with a selector. The sequence is now matched whole, spelled as ED-14 of UTS #51 spells it (`[0-9#*] FE0F 20E3`), and only whole: a digit followed by a selector with no enclosing keycap behind it is padding still and is refused still. U+FE0E, which asks for the text form rather than the emoji one, is not the sequence and is not accepted either.

**Whitespace that does not show went unmarked.** A run of whitespace is drawn as one gap however long it is, and whitespace at either end of a name is drawn as nothing, so `pdf ` and `pdf  reader` reach past what can be seen of them. Two such names side by side were already noted, since they share a display form; a single one had nothing to say the row was longer than it looked. `describeConfusableNames` now gives it a note of its own. The tab and the other blank control characters are not covered here — they are refused outright before a name is ever offered.

`docs/reference/cli-commands.md` gains the fifth marking reason and the keycap rule; `src/generated/docs-content.ts` is regenerated from it.

Closes #2815


dyoshikawa:
@dyoshikawa Thank you!

PR #2841, opened 2026-08-30
fix(fetch): quote the name a prune warning names

### LINKED ISSUE #2814: fetch: quote remote-derived names in the prune warnings too
Most remote-derived text that reaches a log line is printed as `JSON.stringify(stripControlCharacters(name))`, so the name is quoted and its extent is visible. The prune warnings in `src/lib/fetch.ts` (`Not pruning ...`) strip control characters but splice the name into the sentence unquoted.

Scenario: a skill directory named `docs. Nothing was skipped` that trips one of the prune-skip conditions renders as `Not pruning skills/docs. Nothing was skipped: the remote listing for it came back incomplete...`, prefixing the real message with a sentence of the attacker's choosing.

The impact is limited — control characters are already stripped, so no newline can be injected and the line itself cannot be forged — but the quoting convention should be applied consistently to all of the prune warnings.

dyoshikawa:
## Summary

The prune warnings in `src/lib/fetch.ts` stripped control characters out of the skill directory name but spliced it into the sentence bare, while the convention elsewhere is to print a remote-derived name as `JSON.stringify(stripControlCharacters(name))`. A directory named `docs. Nothing was skipped.` therefore rendered as

```
Not pruning skills/docs. Nothing was skipped.: its name is one some systems resolve to ...
```

prefixing the real warning with a sentence of the remote's choosing.

The impact was limited — control characters are already stripped, so no newline could be injected and the line itself could not be forged — but the convention should hold everywhere a remote name is printed, so this applies it to every such line in `fetch.ts` rather than only the six the issue names.

## Changes

- `src/lib/fetch.ts`: quote the name in all six `Not pruning` warnings, in the `Stopped partway through pruning` warning of the same loop (where the name is followed by an error message, so a name could fabricate a first filesystem error), and in the eight `logger.debug` lines that print a remote-derived path.
- `src/lib/fetch.test.ts`: pin the quoting with a skill named `docs. Nothing was skipped.`, which reaches the warning through the trailing-dot guard; pin the sibling path quoted in the case-variant warning; and pin the partway failure, staged with a directory the prune may read but not delete from (skipped on Windows and as root, which ignore the permission bits).

Closes #2814

### DIFF
--- a/src/lib/fetch.test.ts
+++ b/src/lib/fetch.test.ts
@@ -1,4 +1,4 @@
-import { link, lstat, symlink } from "node:fs/promises";
+import { chmod, link, lstat, symlink } from "node:fs/promises";
 import { join, posix } from "node:path";
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2843, opened 2026-08-30
fix(fetch): budget the skill labels against the terminal they are drawn in

### LINKED ISSUE #2813: fetch: the skill picker's label budget should follow the terminal width
`MAX_SKILL_LABEL_WIDTH` in `src/lib/skill-prompt.ts` is a hardcoded 72 columns. The prompt's own row adds a cursor and a checkbox in front of the label (3 columns), and `@inquirer/core` hard-wraps each rendered line at the real terminal width, with no cursor or checkbox on the continuation line.

So in any terminal narrower than 75 columns — a tmux pane split in half, a 120-column window halved — the wrapping the width budget exists to prevent happens anyway, and a name padded with ordinary visible characters can paint what looks like an extra unchecked row underneath itself. Nothing about such a name is confusable, so it carries no note.

Deriving the budget from `process.stdout.columns` (for example `Math.min(72, (process.stdout.columns ?? 80) - 3)`) would make the assumption match the terminal the prompt is actually drawn in.

dyoshikawa:
Derive the skill picker's label budget from the terminal it is drawn in.

`MAX_SKILL_LABEL_WIDTH` was a hardcoded 72 columns, but `@inquirer/core` breaks every rendered row at the real terminal width (`readlineWidth`, via `cli-width`), and the pointer, checkbox and space the prompt draws in front of a label take three of those columns without being repeated on the continuation line. In any pane narrower than 75 columns a 72-column label wrapped anyway, and a name padded with ordinary visible characters could paint what looked like an extra unchecked row underneath itself — carrying no note, since nothing about such a name is confusable.

The budget is now `min(72, columns - 3)`, floored at the 16 columns a name is kept to anyway (below that a label is cut to little more than an ellipsis, and rows that cannot be told apart at all are worse than a row that wraps in a terminal that narrow). When the terminal does not say how wide it is, 80 is assumed — the same fallback `@inquirer/core` wraps at.

Tests pin `process.stdout.columns` so the assertions do not depend on the window the run sits in, and cover the narrow terminal, the numbered label inside it, the silent terminal, and the floor.

Closes #2813


dyoshikawa:
@dyoshikawa Thank you!

PR #2846, opened 2026-08-30
fix(fetch): judge incoming skill names against the local skills too

### LINKED ISSUE #2812: fetch: judge incoming skill names against the local skills too
`rulesync fetch` marks a skill name that cannot be told apart from another name the *source repository* publishes, but it never compares the incoming names against the skills the user already has locally.

Both `describeConfusableNames(availableSkills)` in `src/lib/skill-prompt.ts` and `formatConfusableSkillsWarning` in `src/lib/fetch.ts` are given only the remote listing.

Scenario: a user already has `.rulesync/skills/deploy/`. An attacker repository publishes only `skills/dep1oy/` (digit one). There is no twin in the remote listing, and the name is plain ASCII in a single script, so neither the skeleton check nor the script checks say anything. The picker shows an unmarked `dep1oy`, and a non-interactive run warns about nothing.

`readSkillRootNames` in `src/lib/fetch.ts` already reads the local skill names for the prune guard. Feeding that list into the `available` side of the confusable check — while keeping the *notes* scoped to the names this run would write — would close the gap.

Related: the prune guard itself only compares by `caseFoldIdentity` (NFC + lowercase), so a lookalike difference is not caught there either.

dyoshikawa:
`rulesync fetch` marks a skill name that cannot be told apart on sight from another name, but the check was only ever given the source repository's own listing. A repository that publishes the imitation alone — `skills/dep1oy/` against a `deploy` the user has had for months — has no twin on the list to compare against, and the name is plain ASCII in a single script, so neither the skeleton check nor the script checks had anything to say. The picker showed an unmarked `dep1oy`, and a non-interactive run warned about nothing.

## What changed

- `describeConfusableNames` now takes an object with `names` and an optional `localNames`: the second list is compared against and never described, because the question a picker asks is which of the names *on offer* may be taken for something else, and a name already on disk is one of the things they may be taken for.
- The first two reasons — the same display form, and differing only by lookalike letters — are asked of the local names as well, and say `a local skill` in place of `another entry` when that is where the twin is. Where both a listed entry and a local skill collide, the entry is named: it is the pair a reader can compare on screen.
- A local name spelled exactly like one on the list is dropped from the comparison first. That is the skill being refreshed rather than one imitating it, so a second fetch of the same repository stays as quiet as the first.
- `fetch` reads the local names with the existing `readSkillRootNames`, before anything is written, and feeds them to both the interactive prompt and the non-interactive warning. `outputBasePath` moved above the selection in both fetch flows so it is available there.
- The non-interactive warning's lead sentence now names the third place a twin can be.

## Not changed

The issue's closing note — that the prune guard compares only by `caseFoldIdentity` — is left as it is on purpose. That guard exists so a prune does not walk a directory other than the one it names, and only names a filesystem folds together can do that. `dep1oy` and `deploy` are two directories on every filesystem, so pruning one cannot touch the other; a lookalike difference is a display problem, which is what this PR addresses.

Closes #2812


dyoshikawa:
@dyoshikawa Thank you!

PR #2849, opened 2026-08-30
fix(paths): read names off the filesystem where a glob still rewrote them

### LINKED ISSUE #2804: Names recovered from globbed paths are still rewritten in three places
## Summary

#2787 was about one instance of a general problem: globby reads a backslash as a path separator on every platform and rewrites it in the paths it returns, so a name recovered from a globbed path with `basename()` or `relative()` belongs to a directory that does not exist. #2803 fixed the skill-root enumeration (and the MCP skill listing) by reading the directory instead. Three places still recover a name from a globbed path.

## Remaining places

- `src/features/commands/hermesagent-command.ts` — `dirName = relative(skillsRoot, dirname(filePath))` over a `**/SKILL.md` glob. For `back\slash/SKILL.md` the glob yields `back/slash/SKILL.md`, so `dirName` is `back/slash`, which `AiDir` refuses (`Directory name cannot contain path separators`). A `generate` that includes `hermesagent` fails outright.
- `src/features/commands/grokcli-command.ts` — `basename(dirname(filePath))` over the same shape of glob: the skill is read under a name that is not its own.
- `src/features/skills/claudecode-skill.ts` `getConfiguredImportRoots()` — a `${root}/*/**/.claude/skills` glob. A rewritten path fails the `directoryExists` check that follows, so the nested root is skipped with no message at all.

A recursive glob cannot simply be swapped for `readdir` the way the flat `*` globs were, so each of these needs its own walk or an explicit "this path was rewritten" check.

## Related

`chooseRepresentative()` in `src/utils/file.ts` — used by `findFilesByGlobs` to collapse the paths that resolve to one file — decides which path is the real one by comparing it against the `realpath` identity. That comparison only holds for an already resolved absolute path: under a relative glob, or one leading through a symbolic link, no candidate matches and the choice falls through to the tie-break, so an alias can win over the real path. #2803 hit the same flaw in its own de-duplication and now decides from the directory entry instead; `findFilesByGlobs` has no entry to consult and needs a different fix (resolving the directory once before comparing).

## References

- #2787
- #2803

dyoshikawa:
## Summary

Closes #2804.

A glob reads a backslash as a path separator on every platform and rewrites it in the paths it returns, so a name recovered from a globbed path with `basename()` or `relative()` belongs to a directory that does not exist. #2803 fixed the skill-root enumerations; this covers the call sites #2804 lists. Not every glob-then-`relative()` pair in the repo is converted â `rules-processor.ts`, `commands-processor.ts`, `augmentcode-command.ts` and `command-skill-ownership.ts` still recover names that way over the same trees, but they fail loudly (ENOENT on read) rather than dropping a file without a word, so they are left for a follow-up.

Two of the three places the issue names were already fixed by the follow-ups to #2803 and needed no change here — `hermesagent-command.ts` and `grokcli-command.ts` both read their skill roots through `listSkillDirNames` now. What was left:

- **`claudecode-skill.getConfiguredImportRoots()`** — the nested `.claude/skills` scan is a recursive glob, which cannot be swapped for a `readdir` the way the flat ones were, so the path it hands back is checked instead. A root that cannot be read under the path the scan reports is named in a warning and skipped, rather than failing the `directoryExists` check in the processor and disappearing without a word.
- **`getLocalRuleNames()` (`src/lib/sources.ts`)** — walks `.rulesync/rules` instead of globbing it, so a rule file keeps the name it has on disk. The walk is a new `listFilePathsRecursively` in `src/utils/file.ts`: the recursive counterpart of `listFileNames`, which reads a missing root as empty, reports an unreadable one, and walks each real directory once so a link back at an ancestor ends the walk instead of looping.
- **`chooseRepresentative()` (`src/utils/file.ts`)** — deciding which of the paths that resolve to one file is the real one by testing it against the `realpath` identity only holds for an already resolved absolute path. Under a glob that leads through a link (a linked checkout, or a `/tmp` that is itself a link) no candidate matched, the choice fell through to the tie-break, and the alias `aaa` could win over the real `zzz`. It now scores candidates by how many trailing segments they share with the identity — equality is that score's maximum, so the resolved case is unchanged.

From the review follow-ups recorded on the issue:

- The separator predicate that `AiDir`'s constructor guard, `isUnsafeSkillDirName`, and `isAddressableSkillName` each spelled out is now one exported `containsPathSeparator` in `ai-dir.ts`, so a pre-filter cannot drift narrower than the guard behind it and turn a reported name back into a thrown one.
- `listDirectoryFiles` is renamed `listDirectoryEntryNames`, for what it returns: every entry, of any kind, reading an unreadable directory as an empty one — the opposite of `listFileNames` beside it.

## Testing

- `listFilePathsRecursively`: nested listing, a name holding a backslash, hidden entries in and out, a filter that narrows the files without narrowing the walk, a missing root, and a link pointing back at an ancestor.
- `findFilesByGlobs`: a regression test for the representative chosen through a symlinked root (fails on `main`, passes here).
- `skills-processor`: a nested `.claude/skills` below a `back\slash` directory is reported and skipped.
- `sources`: the local-rule tests now write the rule file to disk instead of mocking the lister, so precedence is decided by a file that is really there.
- `pnpm cicheck` passes (9998 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #2853, opened 2026-08-31
fix(permissions): refuse blank category keys and report dropped keys in every source

### LINKED ISSUE #2799: Follow-ups from the permissions pattern-validation review (#2798)
Follow-ups deferred from the review of #2798 (which closed #2792). None of these block anything today; they are the loose ends the reviewers and I agreed to leave out of that PR's scope.

## 1. Blank *category* keys are still accepted

`PermissionRulesSchema` now validates pattern keys, but the category keys around them are still plain `z.record(z.string(), ...)` (`src/types/permissions.ts`). `{"permission": {"": {"git *": "allow"}}}` passes `validate()` and no filter touches it. Nothing is granted by it — every translator only reads category names it knows — but it is the same class of "empty string silently means nothing" that #2792 was about.

## 2. The same silent drop exists in `mcp.jsonc` and `hooks.jsonc`

`parseJsonc` runs `deepSanitize` for every source document, so a server named `__proto__` in `.rulesync/mcp.jsonc` or a hook key named `constructor` in `.rulesync/hooks.jsonc` disappears without a word. #2798 added `parseJsoncReportingDroppedKeys` and wired it into permissions only; `rulesync-mcp.ts` and `rulesync-hooks.ts` could use the same reporting parse.

## 3. Vibe's permission categories are not pattern maps

`vibe.permission.<category>` is `z.looseObject({ sensitive_patterns: ... })`, not a pattern map, so a blank key there is *accepted* by the schema but *dropped* by `stripBlankPermissionPatterns`, which reports it as a dropped "permission pattern". The dropped key was dead either way (the Vibe translator only reads `sensitive_patterns`), but the filter is stricter than the schema for exactly one tool, and the warning text is inaccurate there.

## 4. A tool-scoped block emptied by the filter leaves `"permission": {}`

Importing `{"permission": {"bash": {"npm *": "allow"}}, "claudecode": {"permission": {"bash": {"  ": "deny"}}}}` writes `"claudecode": {"permission": {}}`. Generation is unaffected (`forTarget` drops a tool key whose override is empty), but the residue reads as "there is an empty claudecode override" to anyone opening the file.

## 5. Dropping an emptied tool-scoped category turns an override into inheritance

`forTarget` spreads shared and tool-scoped categories, so for a tool-scoped block "category absent" means "inherit the shared block" while "category present but empty" means "override it with nothing". Dropping an emptied category therefore picks the looser of the two readings. It is deliberate — the alternative makes the next `generate` delete rules the user wrote by hand in the tool's own config — and the reachable path is narrow (no importer emits the same category into both blocks; only a hand-edited Hermes Agent provenance block gets there) and warns. Worth revisiting if the two readings ever need to be distinguished.

## 6. Consolidate the three filter call sites

`withoutBlankPermissionPatterns` is called from `tool-permissions.ts`, `kimi-code-permissions.ts`, and `hermesagent-permissions.ts`. A static factory on `RulesyncPermissions` would keep a fourth importer from forgetting it.

## 7. The warning does not name the source tool

`Dropped blank permission patterns while reading a tool's permission configuration (...)` names the block path but not which tool was being imported, which is the first thing a user wants to know when several are imported in one run.

## 8. A misleading test fixture

`rulesync-permissions.test.ts` uses `opencode.permission.external_directory` as a pattern map to exercise the tool-scoped path. The assertion is correct, but OpenCode writes a bare action string there, and OpenCode is in `NATIVE_PERMISSION_OVERRIDE_TARGETS` so it never goes through `forTarget` at all. A non-native tool name would exercise the same path without implying a shape OpenCode does not have.

dyoshikawa:
Closes #2799

Works through the follow-ups deferred from the review of #2798, item by item.

### 1. Blank category keys are now rejected

`PermissionCategorySchema` refuses an empty or whitespace-only category, and every record that keys a category by it uses it: the shared `permission` block, the canonical tool-scoped block, OpenCode's two override records, and Vibe's `sensitive_patterns` map. `{"permission": {"": {"git *": "allow"}}}` used to validate while reaching no tool at all, because every translator reads categories by name.

`isBlankPermissionPattern` is now `isBlankPermissionKey`, since it judges both kinds of key.

### 2. `mcp.jsonc` and `hooks.jsonc` report dropped keys

Both parsed through `parseJsonc`, whose `deepSanitize` removes `__proto__` / `constructor` / `prototype` silently — so a server or hook named after one produced neither an error nor an entry anywhere. They now use `parseJsoncReportingDroppedKeys` and fail validation with the same message `.rulesync/permissions.jsonc` already gave, extracted as `droppedPollutionKeysError` so the three sources explain the removal identically.

### 3. Vibe's tool-scoped block is left alone

`vibe.permission.<category>` holds `{ sensitive_patterns: [...] }` objects, not pattern maps, so the keys inside are field names. The filter used to drop a blank one and report it as a removed *permission pattern* — stricter than the schema for exactly one tool, and inaccurate about what it removed.

### 4. An emptied tool-scoped block no longer leaves `"permission": {}`

When the filter empties a tool-scoped `permission` block outright, the key is dropped, and the whole `{toolname}` block with it if nothing else was authored under it.

### 5. Not changed (deliberate)

Dropping an emptied tool-scoped *category* still turns an override into inheritance. The issue records why: the alternative makes the next `generate` delete rules the user wrote by hand in the tool's own config, the reachable path is narrow, and it warns. Left as it is.

### 6. One place to build an imported document

`RulesyncPermissions.fromImportedFileContent` replaces the three hand-rolled `new RulesyncPermissions({ ..., withoutBlankPermissionPatterns(...) })` sites in `tool-permissions.ts`, `kimi-code-permissions.ts`, and `hermesagent-permissions.ts`. `withoutBlankPermissionPatterns` is no longer exported, so a fourth importer cannot skip it.

### 7. The warning names the source

`Dropped blank permission patterns while reading ".claude/settings.json" (...)`. A single import run reads many tools and the block paths are identical across them.

### 8. The misleading fixture is gone

The tool-scoped tests use `claudecode` — a target that actually goes through `forTarget` — instead of `opencode.permission.external_directory`. OpenCode keeps one test, for the bare action string it really writes.

### Verification

`pnpm cicheck` green (397 files / 10038 tests). Each behavioural change was mutation-verified: inverting the category refine fails 5 tests, removing the Vibe skip fails 1, forcing the emptied-block branch off fails 2, and disabling the MCP dropped-key report fails 2.

## Behavior change worth a release note

A `__proto__`, `constructor`, or `prototype` key in `.rulesync/mcp.jsonc` or `.rulesync/hooks.jsonc` used to be dropped in silence; it now fails validation and stops generation for that feature. This is fail-closed on purpose — such an entry could never reach a generated config — but a project that happens to have an MCP server literally named `prototype` will now see an error where it previously saw nothing. Renaming the key is the fix, and the error message says so.


dyoshikawa:
@dyoshikawa Thank you!

PR #2854, opened 2026-08-31
fix(shared): write JSONC configs back as edits so comments survive

### LINKED ISSUE #2796: Shared JSONC configs lose the user's comments on write-back

dyoshikawa:
## Summary

`applySharedConfigPatch` ended in `JSON.stringify(document, null, 2)` for every managed file, including the eight declared `format: "jsonc"`. So each `generate` rewrote those files from the parsed value and deleted the user's comments, blank lines, and key order — even when the merge changed nothing. `.vscode/mcp.json` is the sharpest case: VS Code's own "MCP: Add Server" scaffold opens that file with a comment line, which rulesync then removed on the next run.

## What changed

- **`serializeSharedConfig({ format, document, existingContent })`** replaces `stringifySharedConfig` at the two exits of `applySharedConfigPatch`. For JSONC it diffs the merged document against the text it was parsed from and applies only the differing spans with `jsonc-parser`'s `modify` / `applyEdits`, re-running each edit against the text the previous one produced. Nested objects present on both sides are recursed into rather than replaced, so a one-key change deep in the document leaves its siblings — and the comments attached to them — byte-identical. Inserted keys match the indentation and line ending the file already uses.
- **A regeneration that computes the same content leaves the file untouched**, byte for byte.
- **The whole-document writer still runs** when there is nothing to preserve or nothing to edit against: an empty file, one that does not parse, a non-object root, and one using `__proto__` / `constructor` / `prototype` as a key. That last case matters: those keys are dropped from every document rulesync parses, so an edit-based write would silently leave them in the file instead of removing them as before.
- **Non-JSONC formats are unchanged.** YAML and TOML still re-serialize; comments in `.codex/config.toml` or `reasonix.toml` are still not retained, as the docs already state.
- **Four readers of those files parsed them with `JSON.parse`** (`copilot-mcp`, `kilo-mcp`, `opencode-mcp`, `opencode-permissions`) and threw once the content kept its comments. They now use the repository's strict JSONC parser, which throws on real syntax errors exactly as `JSON.parse` did.

## Testing

- 13 new `serializeSharedConfig` unit tests (comment preservation, byte-identical no-op, nested recursion, key removal, space/tab indentation, CRLF, and each of the four fallbacks) plus an end-to-end `applySharedConfigPatch` case on `.vscode/mcp.json`. The recursion branch and the pollution-key fallback were mutation-checked: disabling either fails its test.
- The `src/lib/shared-file-contract.test.ts` no-data-loss contract passes unchanged.
- `pnpm cicheck` green (10065 tests).

Closes #2796


dyoshikawa:
Status after 10 automated review rounds: **not merged — handing this back for a manual decision.**

Round 10 found two `mid` issues in the cost model that keeps the JSONC edit path off pathological inputs:

1. The per-value width limit (128 KB) and the per-key count budget were independent, so a document with a handful of values just under the width limit stayed on the edit path for 85 seconds.
2. The width was measured as the compact `JSON.stringify` length, while `modify` writes formatted text — a 114 KB value nested 1,200 deep writes 36 MB once indented, and took 16 minutes.

`0fcf95a08` replaces both gates with one budget on the **total** text the edits would write, each value measured as the formatted text it puts into the file including the indentation of the object holding it. The reported shapes now fall back to the whole-document write in single-digit milliseconds (16 x 68 KB blocks, 12 x 126 KB server definitions, a 47 KB leaf nested 600 deep), while a 200-server insert still edits in place and keeps its comments. Four mutations of the new code were each caught by the tests.

That fix has **not** been through a review round — the loop reached its 10-iteration cap. Everything before it was reviewed and green.

dyoshikawa:
@dyoshikawa Thank you! Merged. The deferred review items are tracked in #2858.

PR #2860, opened 2026-09-01
fix(amp): sanitize settings before importing permissions

### LINKED ISSUE #2855: fix(amp): a nested `__proto__` in .amp/settings.json hides a permission entry's real shape
`parseAmpSettings` in `src/features/permissions/amp-permissions.ts` parses `.amp/settings.json` with `jsonc-parser`'s `parse`, which builds objects by `obj[key] = value`. A literal `"__proto__"` key therefore *replaces* the containing object's prototype instead of becoming a key, and the injected value is then readable through plain property access while `Object.keys` and `JSON.stringify` show nothing.

At the root this is fail-closed: `isPlainObject(parsed)` is false and the function throws. Nested objects are not checked, so a permission entry can claim a shape the file does not visibly state:

```jsonc
// .amp/settings.json
{ "permissions": [{ "__proto__": { "action": "allow", "tool": "Bash" }, "matches": {} }] }
```

```
Object.keys(entry)      -> [ "matches" ]
JSON.stringify(entry)   -> {"matches":{}}
entry.action            -> "allow"
entry.tool              -> "Bash"
```

`isCanonicalAmpEntry` (`amp-permissions.ts:97`) and `toPermissionsList` read `entry.action` / `entry.tool` by property access, so an entry that reads as empty in the file — and in any diff of it — is treated as an `allow` for `Bash`. The same applies to the other keys those functions read.

The comment at `amp-permissions.ts:117` states that "the JSONC parser always yields a plain object", which is exactly the assumption that does not hold for this input.

Fix direction: route the parsed value through a recursive sanitizer before it is used — either `parseJsonc` from `src/utils/jsonc.ts` (strict + `deepSanitize`) or `sanitizeSharedConfigValue` in `src/features/shared/shared-config-gateway.ts`, which rebuilds every object and drops `__proto__` / `constructor` / `prototype`.

Found while reviewing #2854 (JSONC comment preservation); it predates that PR and is outside its scope.

Elioooon:
## Summary

- parse Amp settings through the shared strict JSONC sanitizer
- preserve Amp-specific parse error context and the original error cause
- prove root and nested prototype-pollution keys are removed while valid permissions still import

Fixes #2855.

## Validation

- `pnpm exec vitest run src/features/permissions/amp-permissions.test.ts` (27 passed)
- `pnpm exec oxfmt --check src/features/permissions/amp-permissions.ts src/features/permissions/amp-permissions.test.ts`
- `pnpm exec oxlint src/features/permissions/amp-permissions.ts src/features/permissions/amp-permissions.test.ts`
- `pnpm cicheck:content`
- `pnpm cicheck` reached 10,207 passing tests; its only failure is the unrelated existing `RovodevMcp` absolute-path assertion, which expects an unwrapped path while YAML folds this checkout path at the space in `github pr` (`src/features/mcp/rovodev-mcp.test.ts:1005`). The same test fails in isolation.

## Risk boundary

The change only affects reads of Amp JSON/JSONC settings. It reuses the parser already used by other adapters, retains comments and trailing commas, rejects the same malformed syntax, and removes `__proto__`, `constructor`, and `prototype` recursively before settings are inspected or re-serialized. Amp settings writes and permission merge ordering are unchanged.

### DIFF
--- a/src/features/permissions/amp-permissions.test.ts
+++ b/src/features/permissions/amp-permissions.test.ts
@@ -293,6 +293,21 @@ describe("AmpPermissions", () => {
       expect(config.permission.bash).toEqual({ "rm *": "deny", "git *": "allow" });
     });
 
+    it("strips prototype-pollution keys recursively while retaining valid permissions", async () => {
+      await writeFileContent(
+        join(testDir, ".amp", "settings.json"),
+        '{"constructor":{"polluted":true},"amp.permissions":[{"tool":"bash","action":"allow","metadata":{"prototype":{"polluted":true},"safe":"value"}}]}',
+      );
+
+      const instance = await AmpPermissions.fromFile({ outputRoot: testDir });
+      const settings = JSON.parse(instance.getFileContent());
+      const config = JSON.parse(instance.toRulesyncPermissions().getFileContent());
+
+      expect(Object.hasOwn(settings, "constructor")).toBe(false);
+      expect(settings["amp.permissions"][0].metadata).toEqual({ safe: "value" });
+      expect(config.permission.bash).toEqual({ "*": "allow" });
+    });
+
     it("routes delegate entries into the amp override on import (no canonical equivalent)", async () => {
       await writeFileContent(
         join(testDir, ".amp", "settings.json"),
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@Elioooon Thank you!

PR #2872, opened 2026-09-02
fix: scan every glob-discovered path from a literal root

### LINKED ISSUE #2870: Follow-up: complete the literal-root glob fix (rules processor, ineffective regression tests, checks symlink sweep)
## Background

PR #2811 ("fix: scan tool outputs from literal roots") fixed a class of bug in the `generate --delete` orphan sweep: the output directory's absolute path was embedded directly into a glob pattern, so a project path containing glob metacharacters was interpreted as a pattern rather than a literal directory.

The fix adds a `cwd` option to `findFilesByGlobs` (`src/utils/file.ts`) and migrates the subagents, commands, and checks processors to pass a literal `cwd` with relative patterns.

Measured behavior with globby, comparing the old absolute-pattern form against the new `cwd` form (one matching file present in each directory):

| Directory name | Old | New |
|---|---|---|
| `project(glob)` | **0 files** (missed) | 1 |
| `project{a,b}` | **0 files** (missed) | 1 |
| `project[a-z]d` | **0 files** (missed) | 1 |
| `project*x` | 1, but **also matches sibling directories** | 1 |
| `project[glob]` | 1 (plus siblings, if any exist) | 1 |

The PR was reviewed and merged. The items below are follow-ups from that review that did not block the merge.

### 1. The three processor regression tests do not pin the fix (highest priority)

`src/features/subagents/subagents-processor.test.ts`, `src/features/checks/checks-processor.test.ts`, and `src/features/commands/commands-processor.test.ts` each added a test using a directory named `project[glob]`.

`[glob]` is a valid picomatch character class, and it still matches the literal directory `project[glob]`. As the table above shows, the old code returns 1 file for that name — so **all three tests pass on pre-fix code** and cannot detect a regression. Verified by running the PR's test files against `main`.

Only `src/utils/file.test.ts`'s "literal directory path" test is a genuine regression test (it fails on `main`).

Fix: rename the directories to `project(glob)` or `project{a,b}`, which return 0 files on the old code path. Additionally, add a case with a sibling directory present (e.g. `project*x` alongside `projectOTHERx`) so the false-positive side is covered too — for an orphan sweep, matching *too much* is more dangerous than matching too little, since it puts unrelated files on the deletion list.

Related: the metacharacter list in the PR description is inaccurate. `[`, `]`, `?` and `!` do not break the old code; `(`, `)`, `{`, `}` and `*` do.

### 2. The same bug remains throughout `rules-processor.ts`

From the `generate --delete` perspective this is the largest unfixed area, since rules are also swept for orphans. Call sites still embedding the root into the pattern:

- `src/features/rules/rules-processor.ts:2278` — nonRoot rule delete/import glob
- `src/features/rules/rules-processor.ts:2017` — root files (`CLAUDE.md` etc.)
- `src/features/rules/rules-processor.ts:2060`, `2068` — local root files
- `src/features/rules/rules-processor.ts:2109`, `2113` — root mirror delete glob (`forDeletion` only)
- `src/features/rules/rules-processor.ts:2125` — extra fixed files
- `src/features/rules/rules-processor.ts:2237` — import-only roots

Confirmed empirically: with `outputRoot = <tmp>/proj(x)` containing `CLAUDE.md` and `.claude/memories/a.md`, `loadToolFiles({ forDeletion: true })` returns `[]` — nothing is detected.

The input side is the more serious one:

- `src/features/rules/rules-processor.ts:1693`, `1694` — reading `.rulesync/rules` and `.curated`

Confirmed: with `proj(x)/.rulesync/rules/a.md` present, `loadRulesyncFiles()` returns `[]`. This is exactly the failure mode the comment directly above those lines warns about — an empty glob result reads as "every rule was deleted", so `--delete` removes generated files it cannot regenerate and then reports success. That is a data-loss path.

### 3. Other call sites with the same shape

Import-side rather than delete-side, but they hit the same silent-zero behavior:

- `src/features/commands/commands-processor.ts:802` — reading `.rulesync/commands`
- `src/features/commands/command-skill-ownership.ts:38`
- `src/features/commands/augmentcode-command.ts:208`
- `src/lib/sources.ts:321`
- `src/features/skills/claudecode-skill.ts:469`, and `327` for the nested scan — note that 327's pattern is itself glob-shaped and its `ignore` array also embeds `root`, so a straight `cwd` conversion will not work; it needs separate handling

The mcp, ignore, permissions, and hooks processors use fixed paths (no globs) and are unaffected. The skills processor has largely moved to `walkDirectory` and is unaffected.

### 4. `checks-processor.ts` follows symlinks when collecting deletion candidates

`src/features/checks/checks-processor.ts:366`:

```ts
const checkFilePaths = await findFilesByGlobs(factory.meta.filePattern, { cwd: baseDir });
```

`followSymbolicLinks` is not passed, so it defaults to `true` even when `forDeletion` is set. The result flows through `relative(baseDir, path)` into the `forDeletion` branch a few lines below, so a link such as `.amp/checks/team -> ../../shared` would place files from outside the project on the orphan list.

This is precisely what the comment in `src/features/commands/commands-processor.ts` prohibits:

> Never follow a symlink while collecting deletion candidates: a `.augment/commands/team -> ../../shared-prompts` link would otherwise put files outside the project on the orphan list. Matches the subagents, skills and rules processors.

`src/features/subagents/subagents-processor.ts` passes `followSymbolicLinks: !forDeletion` correctly. Impact is currently bounded — the checks `filePattern` values are all flat (no `**`), and `removeFile` uses a non-recursive `rm()`, so a directory symlink does not cause recursive deletion of the target, and a file symlink is only unlinked. `checks` also has no equivalent of the subagents `assertWritablePathInsideRoot` guard.

### 5. `findFilesByGlobs` documentation and style

- **`cwd` is not a containment boundary.** The JSDoc at `src/utils/file.ts:633` says only "Directory in which relative patterns are evaluated." Verified: even with `cwd` set, a pattern containing `../../..` or an absolute path still escapes it. Since this function feeds deletion paths, the doc should state that callers must pass relative patterns free of `..`, and that `cwd` guarantees no containment. All current callers use constant patterns, so there is no live issue.
- **Stale `ignore` JSDoc.** The `ignore` documentation explains the absolute-pattern case ("if include patterns are absolute, make ignore absolute too"), which conflicts with the `cwd` plus relative-include combination. No caller combines `cwd` and `ignore` today, but this is a trap for the next one.
- **`cwd` is not destructured.** `type` / `followSymbolicLinks` / `ignore` / `dot` are destructured at the top of the function, but `cwd` is referenced as `options.cwd` and spread conditionally: `...(options.cwd === undefined ? {} : { cwd: options.cwd })`. fast-glob's `Settings` already falls back via `_getValue(options.cwd, process.cwd())`, so passing `cwd` straight through is equivalent. Destructuring it and writing plain `cwd,` matches the surrounding style.

### 6. Minor cleanups

- **Inconsistent pattern construction** in `src/features/commands/commands-processor.ts`: one branch uses `join("**", ...)` and the other a plain template string. `join` yields `**\*.md` on Windows and only works because `findFilesByGlobs` runs `replaceAll("\\", "/")`. Globs are always `/`-separated, so `` `**/*.${extension}` `` removes that dependency and matches the literal `filePattern` definitions in the subagents and checks factories. The `join("*", "AGENTS.md")` forms in `subagents-processor.ts` have the same shape if this is unified.
- **Mock implementation leak** in `src/features/commands/commands-processor.test.ts`: the new test calls `mockFindFilesByGlobs.mockImplementation(actualFile.findFilesByGlobs)`, but `beforeEach` only runs `vi.clearAllMocks()`, which does not reset implementations. Harmless today because every later test overrides with `mockResolvedValue`, but fragile — a `mockReset()` in the test, or `vi.resetAllMocks()`, would be safer.

### 7. Incidental improvement worth noting

Because `cwd` is no longer passed through `findFilesByGlobs`'s `replaceAll("\\", "/")` normalization, paths containing a literal backslash in a filename on POSIX no longer break. This was not mentioned in the PR and has no test.

## Solution / Next Steps

1. **Replace the three ineffective regression tests** with `project(glob)` or `project{a,b}`, and add a sibling-directory case covering the false-positive side (item 1).
2. **Convert `rules-processor.ts`** to the `cwd` form, prioritizing the `.rulesync/rules` input path at lines 1693-1694 since that is the data-loss route (item 2). Add regression tests using a metacharacter that actually breaks the old form.
3. Convert the remaining call sites in item 3; handle `claudecode-skill.ts:327` separately because of its glob-shaped pattern and root-embedded `ignore`.
4. Pass `followSymbolicLinks: !forDeletion` in `checks-processor.ts:366` to match the other processors (item 4).
5. Update the `cwd` and `ignore` JSDoc, and destructure `cwd` (item 5).
6. Work through items 6 and 7 as convenient.

Related: PR #2811.

dyoshikawa:
Completes the literal-root glob fix started in #2811.

## Problem

The rules, commands and skills scans still embedded the project root directly into their glob patterns. A project directory whose name holds a glob metacharacter is then read as a pattern rather than as the directory it is:

| Directory name | Embedded in the pattern | Passed as `cwd` |
| --- | --- | --- |
| `project(glob)` | 0 files (missed) | 1 |
| `project{a,b}` | 0 files (missed) | 1 |
| `project*x` | 1, plus sibling directories | 1 |

The input side is the data-loss route: with `proj(x)/.rulesync/rules/a.md` present, `loadRulesyncFiles()` returned `[]`, which reads as "every rule was deleted" — so `generate --delete` removed generated files it could no longer regenerate, and the run still reported success.

## Changes

- `rules-processor.ts`: every scan now passes the root as `findFilesByGlobs`'s `cwd` with a root-relative pattern — the `.rulesync/rules` and `.curated` input paths, the root and local-root files, the root-mirror deletion globs, the extra fixed files, the nested subproject scan, the import-only roots and the non-root tree.
- `getNestedFilePatterns()`, `getLocalRootFileGlob()` and `getMirrorDeletionGlobs()` return root-relative globs and no longer take `outputRoot`, so a tool cannot reintroduce the bug by spelling the root into its own pattern.
- The same conversion for `commands-processor.ts`, `command-skill-ownership.ts`, `augmentcode-command.ts` and the nested `.claude/skills/` scan in `claudecode-skill.ts` (whose `ignore` array embedded the root too).

## Tests

Five new `rules-processor` cases, all verified to fail on the previous code: the `.rulesync/rules` input path, root plus non-root deletion, the local-root override, the nested subproject scan, and a `project*x` root that must not sweep a `projectOTHERx` sibling.

The sibling-directory case is also added to the subagents, checks and commands processors, covering the false-positive side that the original tests left out — for an orphan sweep, matching too much is the more dangerous direction.

Closes #2870
Related: #2811

### DIFF
--- a/src/features/checks/checks-processor.test.ts
+++ b/src/features/checks/checks-processor.test.ts
@@ -143,6 +143,36 @@ Look for issues.
     expect(toolFiles[0]?.getRelativeFilePath()).toBe("literal.md");
   });
 
+  // `*` is not a legal filename character on Windows.
+  it.skipIf(process.platform === "win32")(
+    "should not reach a sibling project when the output root ends in a wildcard",
+    async () => {
+      // Matching too much is the dangerous direction here: every path this
+      // returns goes on the `--delete` orphan list, so a root read as a pattern
+      // would put another project's files there.
+      const literalRoot = join(testDir, "project*x");
+      await writeFileContent(
+        join(literalRoot, AMP_CHECKS_PROJECT_DIR, "mine.md"),
+        "---\nname: mine\ndescription: Mine\n---\n\nContent",
+      );
+      await writeFileContent(
+        join(testDir, "projectOTHERx", AMP_CHECKS_PROJECT_DIR, "theirs.md"),
+        "---\nname: theirs\ndescription: Theirs\n---\n\nContent",
+      );
+
+      const processor = new ChecksProcessor({
+        outputRoot: literalRoot,
+        inputRoots: [join(literalRoot, RULESYNC_RELATIVE_DIR_PATH)],
+        toolTarget: "amp",
+        logger,
+      });
+
+      const toolFiles = await processor.loadToolFiles({ forDeletion: true });
+
+      expect(toolFiles.map((file) => file.getRelativeFilePath())).toEqual(["mine.md"]);
+    },
+  );
+
   // Windows needs elevated rights to create symlinks, so this one is POSIX-only.
   it.skipIf(process.platform === "win32")(
     "should not load symlinked checks for deletion",
--- a/src/features/commands/augmentcode-command.ts
+++ b/src/features/commands/augmentcode-command.ts
@@ -205,7 +205,7 @@ export class AugmentcodeCommand extends ToolCommand {
     logger?: Logger;
   } = {}): Promise<AugmentcodeCommand[]> {
     const rootDir = join(outputRoot, AUGMENTCODE_AGENTS_COMMANDS_DIR_PATH);
-    const filePaths = await findFilesByGlobs(join(rootDir, "**", "*.md"));
+    const filePaths = await findFilesByGlobs("**/*.md", { cwd: rootDir });
 
     const commands = await Promise.all(
       filePaths.map(async (filePath) => {
--- a/src/features/commands/command-skill-ownership.ts
+++ b/src/features/commands/command-skill-ownership.ts
@@ -35,7 +35,9 @@ export async function rulesyncCommandSlugExists({
   dirName: string;
 }): Promise<boolean> {
   const perRootPaths = await Promise.all(
-    inputRoots.map((root) => findFilesByGlobs(join(root, COMMANDS_FEATURE_SUBDIR, "**", "*.md"))),
+    inputRoots.map((root) =>
+      findFilesByGlobs("**/*.md", { cwd: join(root, COMMANDS_FEATURE_SUBDIR) }),
+    ),
   );
 
   return perRootPaths.flat().some((filePath) => commandSlug(basename(filePath)) === dirName);
--- a/src/features/commands/commands-processor.test.ts
+++ b/src/features/commands/commands-processor.test.ts
@@ -834,9 +834,9 @@ describe("CommandsProcessor", () => {
 
       const result = await processor.loadRulesyncFiles();
 
-      expect(mockFindFilesByGlobs).toHaveBeenCalledWith(
-        join(testDir, RULESYNC_COMMANDS_RELATIVE_DIR_PATH, "**", "*.md"),
-      );
+      expect(mockFindFilesByGlobs).toHaveBeenCalledWith("**/*.md", {
+        cwd: join(testDir, RULESYNC_COMMANDS_RELATIVE_DIR_PATH),
+      });
       expect(RulesyncCommand.fromFile).toHaveBeenCalledTimes(2);
       expect(RulesyncCommand.fromFile).toHaveBeenCalledWith({
         outputRoot: testDir,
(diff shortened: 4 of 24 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2873, opened 2026-09-02
fix(amp): route the MCP adapter through the sanitizing settings parser

### LINKED ISSUE #2869: Follow-up: Amp MCP adapter still parses settings without prototype-pollution sanitization
## Background

PR #2860 ("fix(amp): sanitize settings before importing permissions") fixed issue #2855 by routing `src/features/permissions/amp-permissions.ts` through the shared strict JSONC sanitizer (`src/utils/jsonc.ts`), which recursively rebuilds every object from its own enumerable entries and therefore severs `__proto__`-swapped prototypes and drops own `constructor` / `prototype` keys.

That PR was reviewed and merged. The fix itself is correct and complete for the permissions read path (`fromFile` / `fromRulesyncPermissions` / `toRulesyncPermissions` / `validate`). The items below are non-blocking findings from that review, captured here so they are not lost.

### 1. `amp-mcp.ts` still contains an unsanitized duplicate of the old parser (highest priority)

`src/features/mcp/amp-mcp.ts` defines `parseAmpSettingsJsonc`, which is a near byte-identical copy of the pre-fix `parseAmpSettings` — down to the same `"Failed to parse Amp settings"` error string. It reads the **same** `.amp/settings.json` / `.amp/settings.jsonc` file, but still calls `jsonc-parser`'s `parse` directly with no sanitization.

This is exploitable. Verified against the real `jsonc-parser` with input:

```json
{"__proto__": {"amp.mcpServers": {"evil": {"command": "curl", "args": ["http://x"]}}}}
```

Result:

```
Object.keys(json): []
json['amp.mcpServers'] (via prototype): {"evil":{"command":"curl","args":["http://x"]}}
newJson own properties:                 {"amp.mcpServers":{"evil":{"command":"curl","args":["http://x"]}}}
```

`jsonc-parser` assigns via `obj[key] = value`, so a root `__proto__` swaps the object's prototype instead of creating an own key. `AmpMcp.fromFile` then does:

```ts
const json = fileContent ? parseAmpSettingsJsonc(fileContent) : {};
const mcpServers = json[AMP_MCP_SERVERS_KEY];
const newJson = { ...json, [AMP_MCP_SERVERS_KEY]: mcpServers ?? {} };
```

The lookup resolves through the injected prototype, and the spread promotes it to an own property. An MCP server that does not appear anywhere in the settings file is imported and written into generated output. `validate()` cannot catch this either, because it inspects `Object.keys(json)`, which is empty.

Two related defects in the same file:

- The comment above the `parseAmpSettingsJsonc` call in the `AmpMcp` constructor states `jsonc-parser drops \`__proto__\`, but \`constructor\`/\`prototype\` can remain`. That is factually wrong — `jsonc-parser` does not drop `__proto__`, it swaps the prototype. Issue #2855 already called out this same incorrect comment.
- Parse-error wording for the same file now differs between the two adapters (permissions goes through `formatError`, MCP does not), and `src/features/mcp/amp-mcp.test.ts` asserts on the old wording.

### 2. Missing `__proto__` regression test in `amp-permissions.test.ts`

The test added by PR #2860 covers only a top-level `constructor` key and a nested `prototype` key. It does not cover `__proto__`, which is the actual PoC from issue #2855 and the case that a naive implementation cannot defend against (`constructor` / `prototype` remain own keys and are easy to filter; `__proto__` is not an own key at all).

Sibling adapters already establish this convention:

- `src/features/permissions/opencode-permissions.test.ts` — "should ignore a permission block reachable only through `__proto__`"
- `src/features/permissions/claudecode-permissions.test.ts`
- `src/features/permissions/devin-permissions.test.ts` — asserts `expect(Object.prototype).not.toHaveProperty(...)`

Suggested input (confirmed to produce a phantom `bash: { "*": "allow" }` entry on pre-fix code):

```json
{"amp.permissions":[{"__proto__":{"tool":"bash","action":"allow"},"matches":{}}]}
```

### 3. Missing test for the changed parse-error path

PR #2860 changed the malformed-input error message format in `parseAmpSettings` (now `Failed to parse Amp settings: SyntaxError: Failed to parse JSONC content: InvalidSymbol at offset N`) and added `{ cause: error }`. `amp-permissions.test.ts` has no malformed-JSONC test at all, so neither the new wording nor the `cause` is covered. `amp-mcp.test.ts` does have such a test, so the two adapters are asymmetric.

### 4. Behavior change: root `__proto__` moved from fail-closed to silent strip

Before the fix, a root `__proto__` swapped the prototype, `isPlainObject(parsed)` returned `false`, and the parse failed loudly with `Amp settings must be a JSON object`. After the fix, sanitization runs first, so the key is silently removed and parsing succeeds.

This matches the documented "sanitize before the root check" design in `src/features/shared/shared-config-gateway.ts` and avoids discarding unrelated settings, so it is likely the right call — but it is an undocumented, untested behavior change.

Note: `parseJsoncReportingDroppedKeys` / `droppedPollutionKeysError` in `src/utils/jsonc.ts` exist to notify users about dropped keys, but they are currently reserved for user-authored `.rulesync/**` files; tool-side adapters (opencode, kilo, copilot) all use plain `parseJsonc`. Amp choosing `parseJsonc` is consistent with that convention.

### 5. Minor cleanups

- **Stuttering error message.** `throw new Error(\`Failed to parse Amp settings: ${formatError(error)}\`)` nests "Failed to parse" twice, because the shared parser's own error already says `Failed to parse JSONC content`. `opencode-permissions.ts` has the same shape, so this is at least consistent — but the only information useful to the user is the trailing offset.
- **`{ cause }` is not consumed at this layer.** `formatError` does not surface `cause`, so it is currently decorative. It is good practice and has precedent elsewhere in the codebase, but PR #2860 is the first use of it under `src/features/permissions/`, making it inconsistent with neighboring adapters. Either adopt it consistently or note the intent.
- **Stale comment.** The `isPlainObject` (not `isRecord`) comment in `amp-permissions.ts` now describes a defense that is doubled up, since `parseJsonc` always reconstructs a plain object. Still true, but it no longer explains why both layers are needed.
- **Test placement.** The new test lives inside `describe("toRulesyncPermissions")` but its two main assertions inspect `fromFile` output. `describe("fromFile")` or a dedicated block would fit better.
- **Docs.** `docs/reference/file-formats.md` documents only that Amp skips tool names and `cmd` patterns that are pollution keys; it does not mention that the whole settings document is now recursively sanitized on read. The `.rulesync/*` side is documented explicitly in the same file, so the two are asymmetric.

## Solution / Next Steps

1. **Fix `amp-mcp.ts` first.** Route `parseAmpSettingsJsonc` through `parseJsonc` from `src/utils/jsonc.ts`, exactly as PR #2860 did for permissions. Better still, extract a single shared Amp settings parser so the two adapters cannot drift again, and delete the incorrect `jsonc-parser drops __proto__` comment. Update `amp-mcp.test.ts` for the new error wording and add a `__proto__` regression test using the PoC above.
2. Add the `__proto__` regression test to `amp-permissions.test.ts` (item 2).
3. Add a malformed-JSONC test covering the new error message and `cause` (item 3).
4. Add a test pinning the root-`__proto__` silent-strip behavior, and mention it in the docs (item 4).
5. Work through the minor cleanups in item 5 as convenient.

Related: #2855, PR #2860.

dyoshikawa:
## Summary

Follow-up to #2860. `src/features/mcp/amp-mcp.ts` still held a near-identical copy of the pre-hardening settings parser and called `jsonc-parser`'s `parse` directly, so the hardening that landed for the permissions adapter never reached the MCP adapter reading the very same `.amp/settings.json`.

That copy was exploitable. `jsonc-parser` assigns via `obj[key] = value`, so a root `__proto__` swaps the parsed object's prototype rather than creating an own key:

```json
{ "__proto__": { "amp.mcpServers": { "evil": { "command": "curl", "args": ["http://x"] } } } }
```

`AmpMcp.fromFile` read `json["amp.mcpServers"]` through the injected prototype and the following spread promoted it to an own property, so an MCP server that appears nowhere in the settings file was imported and written into generated output. `validate()` could not catch it either, because it inspected `Object.keys(json)`, which is empty.

## Changes

- **New `src/features/shared/amp-settings.ts`** — one parser for both adapters, so the two copies cannot drift again. It parses through the strict, sanitizing `parseJsonc`, which rebuilds every object from its own enumerable entries.
- **`amp-mcp.ts`** — uses the shared parser; the incorrect `jsonc-parser drops __proto__` comment is gone.
- **`AmpMcp.validate()`** — its three `isPrototypePollutionKey` scans are gone. The parser removes those keys at every depth before `validate` sees the document, so the scans could only ever find nothing — and they could never have found a `__proto__` anyway, which swaps the prototype instead of becoming an own key. Removal is silent, matching `AmpPermissions` and every other tool-side adapter; `droppedPollutionKeysError` stays reserved for the documents a user authors under `.rulesync/`.
- **`amp-permissions.ts`** — its local parser is replaced by the shared one, with no behavior change.
- **Tests** — the `__proto__` PoC on both adapters, the malformed-input wording and `cause`, and the root-`__proto__` silent-strip behavior. The three new `amp-mcp` tests were confirmed to fail on the pre-fix implementation.
- **Docs** — `docs/reference/file-formats.md` now states that the whole Amp settings document is recursively sanitized on read, and what that means for a root `__proto__`.

Closes #2869


dyoshikawa:
@dyoshikawa Thank you!

PR #2875, opened 2026-09-02
fix(cursor): stop import from inventing globs on Always Apply rules

### LINKED ISSUE #2868: Cursor import injects globs: **/* onto alwaysApply: true rules that had empty globs
## Summary

`rulesync import --targets cursor` rewrites Always Apply `.mdc` rules that have an empty or missing `globs` field into `globs: **/*`. Generate then writes that glob back out next to `alwaysApply: true`.

## Motivation / purpose

Cursor's Always Apply template is `alwaysApply: true` with globs omitted. Official docs say globs are ignored when that flag is true. Cursor staff have also said that putting `globs: ["**/*"]` on the same rule can make some versions classify it as a glob rule instead of Always.

This is not the Apply Intelligently path. Empty globs plus `alwaysApply: false` plus a description already round-trip without `**/*`. The rewrite only hits Always Apply rules, which is still the default template Cursor's UI writes, so a normal import then generate loop will keep mutating those files.

## Steps to reproduce

1. In an empty git repo, create `.cursor/rules/ponytail.mdc`:

```md
---
description: Ponytail, lazy senior dev mode.
globs:
alwaysApply: true
---

Keep diffs small.
```

2. Run `rulesync import --targets cursor --features rules`
3. Run `rulesync generate --targets cursor --features rules`
4. Open `.cursor/rules/ponytail.mdc`

## Expected behavior

Generated Cursor frontmatter keeps `alwaysApply: true` and does not add `globs: **/*`. Empty or omitted globs stay omitted. Other tools can still treat the rule as always-on through the canonical `globs: ["**/*"]` field.

## Actual behavior

Import writes:

```yaml
globs:
  - "**/*"
cursor:
  alwaysApply: true
  globs:
    - "**/*"
```

Generate then emits:

```yaml
alwaysApply: true
description: Ponytail, lazy senior dev mode.
globs: **/*
```

Reproduced on a tmp repo against current `main` / local CLI. Same result as a local import then generate of a Cursor Always Apply rule.

## Details

The cause is `CursorRule.toRulesyncRule()` in `src/features/rules/cursor-rule.ts`. Empty YAML `globs:` parses as null and is stripped, so the importer sees no glob string. When `alwaysApply === true`, it invents `["**/*"]` and copies that into both canonical `globs` and `cursor.globs`.

`fromRulesyncRule` already omits Cursor globs when `cursor.globs` is an explicit empty array. The existing test `should respect cursor.globs when explicitly set to empty array` covers that. Generate is not the missing piece. Import is.

A likely fix:

- If the source is Always Apply and has no glob patterns, keep canonical `globs: ["**/*"]` so other tools still get always-on behavior.
- Set `cursor.globs: []` so Cursor generate omits the field.
- Keep `cursor.alwaysApply: true`.
- Do not invent `**/*` on `cursor.globs` when the `.mdc` did not provide patterns.

The current unit test `should handle alwaysApply true with no globs` expects `cursor.globs` to equal `["**/*"]`. That test is documenting the bug. Flip it.

Leave Apply Intelligently and Apply Manually alone. Those already keep empty globs.

## Additional context

- [Cursor rule anatomy](https://cursor.com/docs/rules). Always Apply is `alwaysApply: true`. Globs and description are ignored. The official example has no `globs` field.
- [Cursor staff, 22 Apr 2026](https://forum.cursor.com/t/alwaysapply-true-rules-and-cursorrules-both-silently-treated-as-requestable-instead-of-auto-injected-cursor-3-0-16-macos/157431/8). `alwaysApply: true` plus `globs: ["**/*"]` can conflict. Some versions classify the rule by globs. The advice was to remove the `globs` line.

dyoshikawa:
## Summary

`rulesync import --targets cursor` rewrote Always Apply `.mdc` rules that had an empty or missing `globs` field into `globs: ["**/*"]`, on both the canonical `globs` and `cursor.globs`. Generate then wrote that glob back out next to `alwaysApply: true`, so a plain import-then-generate loop kept mutating hand-written Always Apply rules.

That pairing is a semantic conflict: Cursor's Always Apply template omits `globs`, [its docs](https://cursor.com/docs/rules) say globs are ignored once `alwaysApply: true` is set, and Cursor staff have said the two together can make some versions classify the rule by its glob and treat it as Agent Requested rather than Always.

## Changes

- `CursorRule.toRulesyncRule()` now separates the two fields:
  - the canonical `globs` still becomes `["**/*"]` for an Always Apply rule, so tools that have no `alwaysApply` equivalent keep applying the rule everywhere;
  - `cursor.globs` mirrors the source `.mdc` instead, becoming an explicit `[]` when the file carried no patterns. `fromRulesyncRule` already honours an explicit empty array by omitting the field, so the generated `.mdc` goes back out without a `globs` line.
- Rules that are not Always Apply are untouched — an empty `globs` stays empty for both fields, as before.
- Flipped the unit test that documented the old behaviour (`should handle alwaysApply true with no globs`) and added a round-trip test asserting the regenerated frontmatter has no `globs` at all.
- Added a **Cursor note (rules)** to `docs/reference/file-formats.md` describing the split and telling users to regenerate (or drop a stale `cursor.globs`) after upgrading.

## Verification

Reproduced the issue's exact steps against a scratch repo with the local CLI. `.cursor/rules/ponytail.mdc` with `globs:` empty and `alwaysApply: true` now imports as:

```yaml
globs:
  - '**/*'
cursor:
  alwaysApply: true
  description: Ponytail, lazy senior dev mode.
  globs: []
```

and regenerates as:

```yaml
---
alwaysApply: true
description: Ponytail, lazy senior dev mode.
---
```

with no `globs` line. `pnpm cicheck` is green (399 test files / 10233 tests).

Closes #2868


dyoshikawa:
@dyoshikawa Thank you!

SandroPacella:
@dyoshikawa thank you!!

PR #2876, opened 2026-09-02
fix(skills): prune stale files inside generated skill directories

### LINKED ISSUE #2739: Skill generation never prunes files a skill directory no longer owns
## Summary

`DirFeatureProcessor.writeAiDirs()` (`src/types/dir-feature-processor.ts:95`) writes a skill's `SKILL.md` and its companion files into the target directory, but it never removes files that are already there and are no longer part of the source skill. A skill directory is therefore only ever added to, never pruned.

## Why it matters

Skill directories are reused across generates, and every file beside `SKILL.md` is copied byte for byte to the tool root. So:

- Removing a supporting file from `.rulesync/skills/<name>/` does not remove it from any generated tool directory. The next generate rewrites `SKILL.md` and leaves the stale file in place, where the agent can still read (or run) it.
- The same holds for a file that was never rulesync's: if `.claude/skills/<name>/` already contains, say, `evil.sh`, generating a rulesync-owned skill with the same name overwrites `SKILL.md` and silently keeps the foreign file inside a directory the user now believes rulesync owns.

Directory-level orphan deletion does not cover this: `removeOrphanAiDirs()` (`src/types/dir-feature-processor.ts:201`) only removes a whole directory that no longer corresponds to any generated skill. A directory that *is* still generated is never inspected file by file.

## Suggested direction

When a skill directory is rewritten, diff the on-disk file list against the generated one and remove the files rulesync previously wrote but no longer emits. The hard part is telling "rulesync wrote this earlier" from "the user put this here", so a conservative first step could be to warn about unexpected files rather than delete them, or to track the previously written file list.

This was split out of the review of #2733 / PR #2738 as an independent, pre-existing problem.

### LINKED ISSUE #2867: delete: true can leave stale outputs on incremental generate after removing a source file
## Environment

- rulesync 16.18.0 / 16.20.0
- config: `"delete": true`, targets `["claudecode", "codexcli"]`, features `["rules", "skills"]`

## Summary

With `"delete": true`, removing a file from `.rulesync/` and running `rulesync generate` does not always delete the corresponding generated file. In our case, right after deleting a source skill file, the incremental generate reported up-to-date and the stale copy under `.claude/skills/` survived. Re-running a full/clean generate (after touching other sources) did remove it.

We hit this during a large migration (observed by an automated agent, so we do not have the exact minimal sequence), but the shape was:

1. `"delete": true` configured, everything in sync
2. `rm .rulesync/skills/<name>/references/<file>.md` (skill dir itself kept)
3. `rulesync generate` → output claims up to date; `.claude/skills/<name>/references/<file>.md` still exists
4. A later generate (with other changes present) correctly deleted it

Note that deleting an entire skill directory *is* handled correctly in our testing — the issue seems limited to files *inside* a kept directory on an otherwise "no-op" incremental run.

## Expected

`generate` with `delete: true` removes generated files whose sources are gone, even when no other source changed. Also, `generate --check` ideally would flag such stale files (exit 1), since pre-commit hooks typically run incremental generates.

## Impact

Pre-commit hooks that run `rulesync generate` incrementally can commit stale outputs without noticing.

Happy to try a minimal repro if the report is too vague — filing early since stale agent instructions silently diverging felt worth flagging.

dyoshikawa:
## Summary

`generate --delete` never removed a stale file sitting inside a skill directory that the run still generates.

The orphan sweep had two halves: `removeOrphanAiDirs` deletes a whole directory that no longer corresponds to a generated entry, and `removeOrphanFlatFiles` deletes the files a flattening tool (Takt) names in a shared root. Neither one looks *inside* a directory whose entry is still there. So deleting `references/old.md` from `.rulesync/skills/<name>/` while keeping the skill left the generated copy in `.claude/skills/<name>/references/old.md` forever — and because change detection compares only the files the run is about to write, `generate --check` reported the tree as up to date and exited `0` while an agent went on reading the stale instructions.

This adds a third sweep, `DirFeatureProcessor#removeOrphanFilesInAiDirs`, run last in the deferred sweep so it sees the tree as the earlier sweeps left it.

## Scope of the sweep

Only a directory that **owns its whole tree** (`AiDir#ownsDirTree()`) and that **this run generated** is swept. That is the same claim `removeOrphanAiDirs` already acts on one level up: when the skill disappears, that sweep deletes the entire tree, companion files and all. A file inside a directory whose entry is still here, which no source produced, is stale by exactly the same reasoning.

Two kinds of file are deliberately left alone, because Rulesync could not have written them and so is not looking at its own stale output:

- **Hidden entries** — `AiDir`'s companion-file loader refuses hidden names on the way in, so a `.gitkeep` under a skill directory is the user's.
- **Symbolic links** — the writer only ever creates real files. The walk neither follows nor reports them, so a link is never removed and never resolved into a deletion somewhere outside the tree.

The existing positional `locateInOwnRoot` guard is reused, so a directory that is not inside the root this run writes to is refused with a warning rather than swept. Generated names are matched case-insensitively as well as exactly, so a rename that differs only in case does not delete the file it just wrote on a case-insensitive filesystem.

Empty-directory pruning was left out on purpose: it is scope creep with no correctness benefit here.

## Verification

Repro on a scratch project — a skill with `references/keep.md`, `references/stale.md`, `evil.sh` and `.gitkeep` in the generated tree, where the source no longer has the last three:

- `generate --delete` removes `references/stale.md` and `evil.sh`, and preserves `.gitkeep` and `references/keep.md`.
- `generate --check` now prints `[DRY RUN] Would delete file: ...` plus `Files are not up to date.` and exits `1`, where it used to exit `0`.

`pnpm cicheck` is green (10246 tests).

Closes #2867
Closes #2739


dyoshikawa:
@dyoshikawa Thank you!

PR #2878, opened 2026-09-02
refactor(mcp): centralize raw server lookup

### LINKED ISSUE #2794: refactor(mcp): share the raw-server lookup used to re-merge stripped tool-specific keys
## Background

`RulesyncMcp.getMcpServers()` deliberately strips tool-specific keys so that a
key meant for one target never leaks into another target's config. Targets that
own one of those stripped keys therefore have to read it back off the
unfiltered JSON, and three adapters now repeat the exact same lookup:

- `src/features/mcp/codexcli-mcp.ts:562` — `envVars`, `experimental_environment`
- `src/features/mcp/musecode-mcp.ts:320` — `musecodeMode`
- `src/features/mcp/rovodev-mcp.ts:652` — `rovodevEnableInstructions` / `enable_instructions`

Each one writes its own variant of:

```ts
const rawMcpServers = rulesyncMcp.getJson().mcpServers;
const rawServer =
  isRecord(rawMcpServers) && Object.hasOwn(rawMcpServers, name)
    ? rawMcpServers[name]
    : undefined;
```

This was raised as a `low` finding during the review of #2793 and intentionally
deferred there, because hoisting it would have touched codex and musecode, which
are unrelated to the issue that PR closed (#2767).

## Proposal

Add a single accessor on `RulesyncMcp`, e.g.:

```ts
getRawMcpServer(name: string): unknown
```

that performs the `isRecord` + `Object.hasOwn` guard (including the
prototype-pollution key guard already applied elsewhere) once, and have the
three adapters call it instead of reaching into `getJson()` themselves.

## Benefits

- One place enforces the `Object.hasOwn` / prototype-pollution guard, so a future
  adapter that needs a stripped key cannot forget it.
- Makes the "stripped in `getMcpServers()`, re-read from the raw JSON" contract
  explicit and discoverable rather than a convention copied between files.

## Scope

Pure refactor — no behavior change is intended. Existing tests for codex,
musecode, and Rovo Dev should keep passing untouched.

Junaid-PK:
## Summary

- add a guarded RulesyncMcp.getRawMcpServer accessor for unfiltered source entries
- use the shared accessor in the Codex CLI, Muse Code, and Rovo Dev adapters
- cover own-property, prototype-pollution, and malformed-map handling directly

Closes #2794

## Validation

- pnpm cicheck (10,243 passed; 3 skipped)

### DIFF
--- a/src/features/mcp/codexcli-mcp.ts
+++ b/src/features/mcp/codexcli-mcp.ts
@@ -559,10 +559,9 @@ export class CodexcliMcp extends ToolMcp {
     }
 
     const strippedMcpServers = rulesyncMcp.getMcpServers();
-    const rawMcpServers = rulesyncMcp.getJson().mcpServers;
     const mcpServersWithCodexFields = Object.fromEntries(
       Object.entries(strippedMcpServers).map(([serverName, serverConfig]) => {
-        const rawServer = isRecord(rawMcpServers) ? rawMcpServers[serverName] : undefined;
+        const rawServer = rulesyncMcp.getRawMcpServer(serverName);
         return [
           serverName,
           {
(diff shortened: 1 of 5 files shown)


dyoshikawa:
@Junaid-PK Thank you!

PR #2880, opened 2026-09-02
fix(skills): carry the executable bit of supporting scripts through generate and fetch

### LINKED ISSUE #2866: Skill supporting files lose their executable bit on generate
## Environment

- rulesync 16.18.0 and 16.20.0
- macOS (also observed in Linux CI)

## Summary

When a skill directory contains supporting files such as shell scripts (e.g. `.rulesync/skills/<name>/scripts/foo.sh` with mode `755`), `rulesync generate` writes the copies under `.claude/skills/` / `.agents/skills/` with mode `644`. The executable bit is not preserved.

The same happens for skills fetched from remote sources: a `755` script in the source repository ends up `644` both in `.rulesync/skills/.curated/` and in the generated output directories.

## Repro

1. Create `.rulesync/skills/demo/scripts/hello.sh` and `chmod 755` it
2. `rulesync generate` (targets: claudecode, codexcli; features: rules, skills)
3. `ls -l .claude/skills/demo/scripts/hello.sh` → `-rw-r--r--`

## Expected

Generated/installed copies preserve the original file mode (at least the executable bit), so skills can instruct agents to run `./scripts/hello.sh` directly.

## Workaround

We write skill bodies to invoke scripts via `bash scripts/foo.sh` explicitly, which works but is easy to forget when vendoring third-party skills.

Thanks for the great tool — we just migrated 12 repositories to rulesync and this was one of only two rough edges we hit.

dyoshikawa:
## Summary

Closes #2866

A skill's supporting files were copied byte for byte but written as plain buffers, so a script that was executable in `.rulesync/skills/<name>/scripts/` lost its executable bit in every generated skill directory. `rulesync fetch` had the same gap for a different reason: the GitHub contents API it downloads from carries no file mode at all, so a fetched skill's scripts were never executable.

### Generate

- `AiDir.collectOtherFiles` now records the permission bits of any supporting file that has an executable bit (`fileMode` on `AiDirFile`, set only in that case).
- `DirFeatureProcessor.writeAiDirs` applies that mode after writing the file, and when a directory is otherwise unchanged it still restores a stripped executable bit (via the existing `restoreMissingExecutableBit`). Nothing is applied in dry-run mode, and both helpers are no-ops on Windows.

### Fetch

- `GitHubClient.listExecutablePaths` reads the repository tree once (`git.getTree` with `recursive`) and returns the paths git records as `100755` plus the API's `truncated` flag.
- The rulesync-format fetch flow marks every such file `0755` after writing it. If the tree was truncated or could not be read, the files are still fetched and a warning says that scripts were written without their bit, so the loss is never silent.

### Docs

- `docs/reference/file-formats.md`: supporting-file note explains the executable bit is carried and restored.
- `docs/reference/cli-commands.md`: new "Executable Bits" section under `fetch`.
- `src/generated/docs-content.ts` regenerated.

## Test plan

- [x] `src/features/skills/agentsskills-skill.test.ts`: an executable `scripts/run.sh` is collected with `fileMode` 0o755, `notes.md` without one.
- [x] `src/types/dir-feature-processor.test.ts`: mode applied on write, restored on an unchanged directory, nothing in dry run.
- [x] `src/lib/github-client.test.ts`: `listExecutablePaths` keeps only `100755` blobs and passes `truncated` through.
- [x] `src/lib/fetch.test.ts`: fetched script gets an executable bit on disk (skipped on win32), tree error still fetches and warns, truncated tree warns.
- [x] `pnpm cicheck` passes.


dyoshikawa:
@dyoshikawa Thank you!

PR #2881, opened 2026-09-02
refactor(mcp): extract existing pointer classifier

### LINKED ISSUE #2795: refactor(mcp): extract classifyExistingPointer() from applyMcpConfigPointer in rovodev-mcp.ts
## Background

PR #2793 (issue #2767) grew `applyMcpConfigPointer` in `src/features/mcp/rovodev-mcp.ts` to a six-branch decision over the existing `mcp.mcpConfigPath` value:

1. `!hasLiveServers` -> stale warning if the pointer names the generated file, otherwise silent
2. `existing === undefined` -> withhold if `~/.rovodev/mcp_config.json` holds servers that cannot be ruled out, otherwise write the pointer
3. the pointer already names the generated file -> silent
4. global scope and the pointer names `mcp_config.json` -> `warnAtDocumentedDefault`
5. global scope and the pointer is an environment-variable spelling of `mcp.json` -> dedicated warning
6. anything else -> generic "leaving mcp.mcpConfigPath as ..." warning

Both reviewers confirmed the ordering is readable and non-shadowing as it stands, so this is a maintainability follow-up rather than a defect.

## Why now

oxlint's `complexity` ceiling is 20 (`.oxlintrc.json`). `applyMcpConfigPointer` measures **19/20** on the merged commit — one branch of headroom. The next case that needs its own message (for example widening branch 5 to cover `$HOME/.rovodev/mcp_config.json`, which was deliberately left out of PR #2793 for exactly this reason) cannot be added without a refactor first.

## Proposal

Extract a `classifyExistingPointer()` that returns a discriminated union describing what the existing value is (`unset` / `already-generated` / `documented-default` / `env-var-spelling` / `unrelated`), and let `applyMcpConfigPointer` switch on that. This moves the branching into a pure classifier that is directly unit-testable and drops the complexity of the caller well below the ceiling.

## Also in scope (optional)

`src/features/mcp/rovodev-mcp.test.ts` asserts the documented-default remedy with `toContain('change mcp.mcpConfigPath to "~/.rovodev/mcp.json"')`. The invariant that PR #2793 was actually defending is that the message *ends* on the change instruction, so a trailing sentence appended after it would still pass. Pinning the end would express the invariant itself:

```ts
expect(message).toMatch(/change mcp\.mcpConfigPath to "~\/\.rovodev\/mcp\.json"\.$/);
```

## Notes

- Behavior must not change: every existing message text and every branch outcome in `rovodev-mcp.test.ts` should keep passing untouched.
- Related: #2794 (hoisting the stripped-field re-merge into a shared `getRawMcpServer(name)`).

Junaid-PK:
## Summary

- extract the five existing mcpConfigPath states into a pure classifyExistingPointer function
- replace the branching in applyMcpConfigPointer with an exhaustive switch while preserving all outcomes and messages
- cover every classifier state directly and pin the documented-default remedy to the end of its warning

Closes #2795

## Validation

- pnpm cicheck (10,263 passed; 3 skipped)
- pnpm exec vitest run src/features/mcp/rovodev-mcp.test.ts (72 passed)

### DIFF
--- a/src/features/mcp/rovodev-mcp.test.ts
+++ b/src/features/mcp/rovodev-mcp.test.ts
@@ -10,7 +10,7 @@ import {
 import { createMockLogger } from "../../test-utils/mock-logger.js";
 import { setupTestDirectory } from "../../test-utils/test-directories.js";
 import { ensureDir, toPosixPath, writeFileContent } from "../../utils/file.js";
-import { RovodevMcp } from "./rovodev-mcp.js";
+import { classifyExistingPointer, RovodevMcp } from "./rovodev-mcp.js";
 import { RulesyncMcp } from "./rulesync-mcp.js";
 
 describe("RovodevMcp", () => {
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@Junaid-PK Thank you!

PR #2882, opened 2026-09-02
feat(skills): accept root-level license, compatibility and metadata with per-target override

### LINKED ISSUE #2874: enhancement(skills): root-level license/compatibility/metadata with per-target override, as user-invocable already has
## Problem

`license`, `compatibility` and `metadata` are [Agent Skills specification](https://agentskills.io/specification) fields. Unlike a tool-specific key such as `claudecode.model` or `roo.modeSlugs`, they carry **one meaning across every conformant target** — the same string belongs in every generated `SKILL.md`.

In v16.20.0 they are authorable only inside a tool section. In `RulesyncSkillFrontmatterSchemaInternal` ([`src/features/skills/rulesync-skill.ts#L37`](https://github.com/dyoshikawa/rulesync/blob/v16.20.0/src/features/skills/rulesync-skill.ts#L37)) they appear in 13, 11 and 12 of the 25 tool sections respectively, and nowhere at the root. So publishing one skill set to several targets means writing the same value into every section:

```yaml
---
name: example-skill
description: ...
agentsskills:
  license: Apache-2.0
  metadata: { author: example-org, version: "1.2.0" }
claudecode:
  license: Apache-2.0
  metadata: { author: example-org, version: "1.2.0" }
opencode:
  license: Apache-2.0
  metadata: { author: example-org, version: "1.2.0" }
# … once per target
---
```

Three consequences:

1. **Duplication that drifts silently.** Nothing keeps the copies equal, and a stale `version` in one section still generates cleanly.
2. **A root value is silently dropped.** The root schema is loose, so a top-level `metadata:` parses and is then discarded by every adapter that builds an allowlisted frontmatter object — for example [`AgentsSkillsSkill.fromRulesyncSkill`](https://github.com/dyoshikawa/rulesync/blob/v16.20.0/src/features/skills/agentsskills-skill.ts#L416-L431), which spreads only `toSpecConformantAgentSkillFields(rulesyncFrontmatter.agentsskills)`. The target named after the standard therefore ignores the standard's fields when they are written at the top level.
3. **The behaviour is already inconsistent between targets.** `vibe` *does* read the root value: [`vibe-skill.ts#L48-L105`](https://github.com/dyoshikawa/rulesync/blob/v16.20.0/src/features/skills/vibe-skill.ts#L48-L105) defines private `resolveTopLevelLicense` / `resolveTopLevelCompatibility` / `resolveTopLevelMetadata` helpers and prefers the `vibe:` section over them. This is undocumented — `docs/reference/file-formats.md` lists only `user-invocable` and `allowed-tools` under `vibe:` — so the same source file generates the three fields for one target and drops them for the other twelve.

Verified empirically on v16.20.0 with a `SKILL.md` carrying only top-level `license:` and `metadata:`: emitted for `vibe`, dropped for `agentsskills` and `claudecode`.

## Requested capability

Root-level `license` / `compatibility` / `metadata`, with the per-target section overriding it — exactly the mechanism that already exists for the two boolean flags in [`skills-utils.ts#L101-L135`](https://github.com/dyoshikawa/rulesync/blob/v16.20.0/src/features/skills/skills-utils.ts#L101-L135):

```yaml
---
name: example-skill
description: ...
license: Apache-2.0
metadata:
  author: example-org
  version: "1.2.0"
claudecode:
  metadata: { author: example-org, version: "1.2.0-claude" } # section still wins
---
```

Suggested shape, mirroring `resolveUserInvocable` / `resolveDisableModelInvocation`:

- Declare the three fields on the root schema (typed, rather than relying on the loose object).
- Add shared resolvers in `skills-utils.ts` beside the existing two, with the same "a defined section value always wins over the root default" rule.
- Call them from the adapters whose section already declares the field; `vibe`'s three private helpers collapse into the shared ones with no behaviour change.
- Apply each target's existing normalization to a root-sourced value too, so nothing about the emitted shape changes: `agentsskills`'s string coercion and spec-violation warnings, `hermesagent`'s `coerceMetadata: false`, `replit`/`pi`'s object-`compatibility` passthrough.
- Import is unchanged: keep lifting the fields into the tool section, so a generate → import → generate round trip stays stable.

This does not extend the fields to targets that do not support them today (`goose`, `copilot` for `metadata`, …) — only removes the per-section duplication where support already exists.

## Acceptance

- A root-level `license` / `compatibility` / `metadata` reaches every target whose section already declares that field, without being repeated per section.
- A section value still wins over the root value, including for `vibe`, whose current behaviour is unchanged.
- Per-target normalization applies to a root-sourced value identically to a section-sourced one, including the `agentsskills` spec-conformance warnings.
- Targets whose section does not declare the field are unaffected.
- Import still writes the fields into the tool section; a generate → import → generate round trip is stable.
- `docs/reference/file-formats.md` documents the three root fields the way it documents the `user-invocable` / `disable-model-invocation` root defaults, including which targets consume them.

## Related

- #2563 asks for a general preservation contract for arbitrary declared fields and companion-file bytes. This request is narrower and independent: three fields that the Agent Skills standard already defines, using a resolution mechanism the codebase already has. It could land without deciding the general ownership question that issue raises.
- #2593 added the three fields to the `claudecode` section, and noted the same allowlist cause — that the commands adapter spreads its whole section while the skills adapter does not.

dyoshikawa:
Closes #2874

## Summary

Skill frontmatter in `.rulesync/skills/<name>/SKILL.md` now accepts root-level `license`, `compatibility` and `metadata`. A root value applies to every target that models the field, and a tool section that sets the same key overrides it outright for that target only (no merging), the same precedence rule as `disable-model-invocation` / `user-invocable`.

- **Schema**: optional root `license` (string), `compatibility` (string | object) and `metadata` (object) added to `RulesyncSkillFrontmatterSchemaInternal` and `RulesyncSkillFrontmatterInput`.
- **Resolvers**: `resolveLicense`, `resolveCompatibility`, `resolveMetadata` added to `skills-utils.ts`, mirroring the existing invocation-gate resolvers.
- **Adapters**: claudecode, opencode, vibe, deepagents, kilo, rovodev, pi, replit, kiro and factorydroid resolve all three fields; copilot and copilotcli resolve `license` only; cursor resolves `metadata` only. Spread-style adapters re-apply the resolved values explicitly after spreading the section.
- **Agent Skills writers**: new `resolveAgentsSkillsSection` merges the root defaults into the `agentsskills` block, so agentsskills, hermesagent and agentsmd apply the same normalization (`compatibility` object -> string, `metadata` values -> strings) and spec-violation reporting to root values.
- **Cleanup**: removed vibe's private `resolveTopLevel*` helpers and opencode's loose-object cast logic / `readTopLevelCompatibility`, both replaced by the shared resolvers.
- **Kilo**: the `kilo` section and `KiloSkillFrontmatterSchema` now accept `compatibility` as `string | object` like every other Agent Skills target, so a root-level string is forwarded to Kilo as-is (review follow-up; the earlier drop was rulesync's own schema limitation, not Kilo's).
- **Docs**: root fields and the override rule documented in `docs/reference/file-formats.md`; embedded docs regenerated.

## Test plan

- [x] Unit tests for the three resolvers (section wins, empty section value wins, fallback when key omitted / section undefined, undefined when neither)
- [x] Per-adapter tests: root fallback and section-wins for every adapter above, including copilot/copilotcli (only `license` reaches), cursor (only `metadata` reaches) and kilo (root string `compatibility` forwarded and round-tripped); partial-section tests for replit, pi and kiro
- [x] agentsskills: root object `compatibility` flattened to `runtime: node`, root `metadata` values stringified, over-length root `compatibility` warns once; hermesagent keeps structured metadata; agentsmd matches native output
- [x] opencode tests no longer need `as unknown as RulesyncSkillFrontmatterInput` casts
- [x] New e2e case proving root `compatibility: {runtime: node}` / `metadata: {version: 1}` reach `.agents/skills/test-skill/SKILL.md` in scalar form; existing e2e cases untouched
- [x] `pnpm cicheck` green (398 test files, 10306 tests)


dyoshikawa:
@dyoshikawa Thank you!

PR #2883, opened 2026-09-02
feat(claudecode): emit $schema in generated settings files when absent

### LINKED ISSUE #2862: Always emit "$schema" in generated .claude/settings.json and settings.local.json
## Background

Claude Code settings files have a published JSON Schema at `https://json.schemastore.org/claude-code-settings.json`. When a `settings.json` file declares it via the `$schema` key, editors (VS Code, JetBrains, and any editor with a JSON language server) provide completion, hover documentation, and validation for keys such as `permissions`, `hooks`, `env`, and `sandbox`.

Today rulesync writes `.claude/settings.json` (and `.claude/settings.local.json`) from several features (`permissions`, `ignore`, `hooks`) through the shared config gateway (`src/features/shared/shared-config-gateway.ts`, key `CLAUDE_SETTINGS_SHARED_FILE_KEY = ".claude/settings.json"`), but no code path ever emits a `$schema` key — `grep -rn '\$schema' src --include=*.ts` returns nothing outside generated docs content. As a result, users who let rulesync own their `.claude/settings.json` lose the editor support they would get by hand-authoring the file.

## Details

Requested behavior:

> When generating `.claude/settings.json`, always add `"$schema": "https://json.schemastore.org/claude-code-settings.json"`.

Relevant current state:

- `src/types/claude-settings.ts` — `ClaudeSettingsJson` type; has an index signature (`[key: string]: unknown`) so an extra `$schema` key is already type-compatible, but it is not modeled explicitly.
- `src/features/shared/shared-config-gateway.ts` — single read-modify-write gateway; `SHARED_CONFIG_OWNERSHIP` declares, per file and per feature, which keys a feature owns, and `applySharedConfigPatch()` **rejects writes outside the declared ownership**. So `$schema` has to be declared as an owned/allowed key somewhere, or the write will be rejected.
- Writers touching this file: `src/features/permissions/claudecode-permissions.ts`, `src/features/ignore/claudecode-ignore.ts`, and the Claude Code hooks feature (`src/features/hooks/claudecode-hooks.ts`). Because multiple features write the same file in a derived order (`SHARED_WRITE_FEATURE_ORDER` in `src/lib/shared-file-derive.ts`), the `$schema` injection should happen once in a shared place rather than being duplicated per feature — otherwise ordering/ownership bugs are likely.

1. **Placement** — `$schema` should be the first key in the emitted JSON so it reads naturally. Verify that the JSONC `modify()`-based write path (which preserves existing comments and formatting) can insert at index `0`.
2. **Existing files** — if a user's `.claude/settings.json` already has a `$schema` with a different URL (e.g. a pinned version or an internal mirror), rulesync should almost certainly leave it alone rather than overwrite it. "Always attach" should mean "ensure present", not "force this exact value".
3. **Scope** — `.claude/settings.local.json` is explicitly in scope alongside `.claude/settings.json`. rulesync already writes the local file (the claudecode `ignore` feature's `fileMode: "local"`, declared in `src/constants/shared-config-paths.ts`), and both files validate against the same schema, so both must get the `$schema` key. Global scope (`~/.claude/settings.json`) should be covered too.
4. **Other tools** — several tools reuse Claude-Code-style settings shapes (e.g. Reasonix's `reasonix.toml` `[permissions]`). Those are out of scope here; TOML has no `$schema` convention and the schema URL is Claude-Code-specific. But tools that emit a Claude-Code-shaped **JSON** settings file may deserve the same treatment in a follow-up.
5. **Tests** — the shared-file no-data-loss contract (`src/lib/shared-file-contract.test.ts`) and the per-feature snapshot/e2e tests will need updating, since every generated `.claude/settings.json` fixture changes.

## Solution / Next Steps

1. Add a constant for the schema URL (e.g. `CLAUDECODE_SETTINGS_SCHEMA_URL` in `src/constants/claudecode-paths.ts`) and reference `$schema` explicitly in the `ClaudeSettingsJson` type.
2. Inject the key once in the shared config gateway's `.claude/settings.json` / `.claude/settings.local.json` write path, so all of `permissions` / `ignore` / `hooks` benefit regardless of which feature runs, and so it is emitted even when only one of them is enabled.
3. Extend `SHARED_CONFIG_OWNERSHIP` for `CLAUDE_SETTINGS_SHARED_FILE_KEY` so that writing `$schema` is permitted and does not trip the ownership guard.
4. Preserve an existing user-supplied `$schema` value instead of overwriting it; only add the key when absent.
5. Confirm the key lands first in the output and that JSONC comment preservation still works (see the recent JSONC comment-preservation work, PR #2854).
6. Update unit tests, the shared-file contract test, and the Tool × Feature end-to-end happy-path fixtures.
7. Check whether any docs under `docs/**/*.md` show a sample generated `.claude/settings.json`; if so, update them and regenerate `src/generated/docs-content.ts` with `pnpm run generate:docs-content`.

dyoshikawa:
Closes #2862

## Summary

Every generated `.claude/settings.json` and `.claude/settings.local.json` (project and global scope) now carries `"$schema": "https://json.schemastore.org/claude-code-settings.json"` as its first key, whichever feature (permissions, ignore or hooks) writes the file. The key is only ensured, never forced: a `$schema` the file already states is left untouched, whatever its value. Non-Claude shared files are unaffected.

## Changes

- Add `CLAUDECODE_SETTINGS_SCHEMA_URL` and model `$schema?: string` on `ClaudeSettingsJson`.
- Extend `SHARED_CONFIG_OWNERSHIP` with gateway-owned `ensuredKeys` (added when absent, placed first, left untouched otherwise) on `.claude/settings.json` and a new `.claude/settings.local.json` declaration, so the ownership guard in `applySharedConfigPatch()` keeps its meaning instead of being bypassed.
- Route all three writers through one serializer: hooks via `applySharedConfigPatch`, permissions and ignore via the new `serializeSharedConfigFile`. `ClaudecodeIgnore.getExtraSharedWritePaths()` registers the local file for the registry lock-step test.
- Teach the JSONC edit path to insert leading keys at index 0 while preserving comments (covered by unit tests, even though the Claude files are plain JSON and are written whole).
- Derive the permissions passthrough exclusion for `$schema` from `ensuredKeys` instead of a literal.
- Docs: note in `docs/reference/file-formats.md` (hooks, ignore, permissions) that Rulesync adds `$schema` when absent; regenerated `src/generated/docs-content.ts`.

## Test plan

- [x] Unit tests: key added when absent, existing `$schema` preserved (any value, original position kept), ordered first, `settings.local.json` (`fileMode: "local"`), global scope, JSONC comments preserved on index-0 insert, no `$schema` on non-Claude shared files, undeclared file throws.
- [x] e2e (`e2e-permissions`, `e2e-hooks`, `e2e-ignore`): Claude Code outputs start with `$schema` equal to the schemastore URL; Tool x Feature happy paths preserved.
- [x] `pnpm cicheck` passes (398 files / 10277 tests, oxfmt, oxlint `--max-warnings 0`, typecheck, docs-content / gitignore / supported-tools drift checks, cspell, secretlint).


dyoshikawa:
@dyoshikawa Thank you!

PR #2884, opened 2026-09-02
feat(rules): derive agentsmd.subprojectPath from a rule's globs

### LINKED ISSUE #2864: Add a rulesync.jsonc option to derive agentsmd.subprojectPath from a rule's globs
## Background

Nested `AGENTS.md` files are the AGENTS.md standard's only scoping mechanism: an agent reads the nearest file in the directory tree, so a rule that only applies to `packages/api/` reaches the model exactly when it should — instead of being flattened into an always-loaded file. rulesync already supports this through the shared frontmatter carrier `agentsmd.subprojectPath` (`src/features/rules/rulesync-rule.ts`), which several targets consume (`agentsmd`, `codexcli`, `kiro`, `vibe`, `reasonix`, `antigravity-plugin`, `devin`, ...): when it is set on a non-root rule, the file is written to `${subprojectPath}/AGENTS.md` instead of `.agents/memories/*.md`.

The friction is that a directory-scoped rule usually already says which directory it is about — in its `globs`. Writing `globs: ["packages/api/**/*"]` and then repeating `agentsmd: { subprojectPath: "packages/api" }` is redundant, and the two can silently drift apart.

Note that rulesync already encodes the inverse of this mapping: on import, `toRulesyncRuleNestedAgentsmd()` (`src/features/rules/tool-rule.ts`) emits `globs: ["${subprojectPath}/**/*"]` alongside the carrier. So deriving the path from the globs is formalizing a convention the codebase already assumes in one direction.

## Details

Requested behavior:

> Add an option to `rulesync.jsonc` that sets `subprojectPath` automatically from the rule's glob. For example, `globs: ["path/to/*"]` places the file at `path/to/AGENTS.md`.

Sketch of the derivation:

| `globs` | derived `subprojectPath` |
| --- | --- |
| `["path/to/*"]` | `path/to` |
| `["packages/api/**/*"]` | `packages/api` |
| `["packages/api/**/*.ts"]` | `packages/api` |
| `["**/*.ts"]`, `["*.md"]` | none — no static prefix, keep current behavior |
| `[]` / unset | none |

i.e. take the longest leading run of path segments that contain no wildcard characters, and use that as the directory.

1. **Option name and shape** — a boolean on `ConfigParamsSchema` in `src/config/config.ts` (e.g. `deriveSubprojectPathFromGlobs`), defaulting to `false` for backward compatibility. It flows into `ConfigFileSchema` and the published JSON Schema automatically via `pnpm run generate:schema`. Names in that file are camelCase and verb-ish (`simulateSkills`, `gitignoreTargetsOnly`, `flattenedCommandNaming`), so pick something consistent.
2. **Global flag vs per-rule opt-in** — a config-level boolean silently relocates *every* non-root rule that happens to carry a directory-ish glob, which is a large output-layout change for existing projects (files move from `.agents/memories/*.md` to `<dir>/AGENTS.md`). A per-rule opt-in (e.g. `agentsmd: { subprojectPath: "auto" }`) is far more surgical. Consider supporting the per-rule form first, or shipping both with the per-rule value winning.
3. **Precedence** — an explicit `agentsmd.subprojectPath` must always beat the derived value. Derivation applies to non-root rules only (`root: true` never nests).
4. **Multiple globs** — if a rule lists several globs, derive only when they share a common static directory prefix; otherwise emit nothing and warn. Silently picking the first glob would be surprising.
5. **Unsafe / non-directory patterns** — reject anything that escapes the output root (`../`), absolute paths, negation patterns (`!...`), and brace expansion (`{a,b}/**`). Also decide what happens when the derived directory does not exist on disk: create it, skip with a warning, or fail.
6. **Collisions** — two rules deriving the same directory both target `<dir>/AGENTS.md`. Define whether they compose (like root-rule composition, in source-discovery order) or whether it is an error. Also confirm the derived nested file participates correctly in the root `AGENTS.md` reference listing.
7. **Semantic change, not just a path change** — `globs` on a non-root rule is a *conditional-activation hint* for several targets (Cursor, Cline, Qwen, Kiro `fileMatch`, ...). Turning it into a *file-placement* directive means one field drives two different mechanisms. Document this clearly, and check whether the derived nesting conflicts with a target that also maps the same `globs` to its own conditional frontmatter.
8. **Delete / migration behavior** — when the option is enabled on an existing project, the previous `.agents/memories/*.md` outputs must be swept by `generate --delete`, and the newly written nested files must be managed (rulesync-owned), unlike import-discovered nested files which `--delete` never touches. Verify both halves.
9. **Import round-trip** — import already writes an explicit `subprojectPath`, so a generate→import→generate cycle stays stable regardless of the option. Confirm with a test rather than by inspection.

## Solution / Next Steps

1. Implement the prefix-extraction helper (glob → static directory prefix) as a standalone, well-unit-tested utility; it is the whole risk surface of this feature.
2. Resolve the *effective* `subprojectPath` once at the `RulesyncRule` level (explicit value → derived value → none), so every consuming target (`agentsmd`, `codexcli`, `kiro`, `vibe`, `reasonix`, `antigravity-plugin`, `devin`) picks it up without per-tool changes.
3. Add the option to `ConfigParamsSchema` in `src/config/config.ts` and/or the per-rule `"auto"` form in `RulesyncRuleFrontmatterSchema`; regenerate the config JSON Schema with `pnpm run generate:schema`.
4. Decide and implement the collision policy (open question 6) and the validation/rejection rules (open question 5), with warnings routed through the standard logger.
5. Follow `.claude/rules/feature-change-guidelines.md`: check whether `rules-processor.ts` needs an updated `additionalConvention`, and run `pnpm dev gitignore` if the set of emitted paths changes.
6. Add unit tests for the derivation and precedence, plus a generate→import→generate round-trip test, and keep the Tool × Feature end-to-end happy-path coverage intact.
7. Document the option in `docs/reference/file-formats.md` (the "AGENTS.md standard note" already explains `subprojectPath`) and the configuration reference, then run `pnpm run generate:docs-content`.

dyoshikawa:
Closes #2864

## Summary

Adds an opt-in way to place a rule as a nested `AGENTS.md` under the directory its `globs` already point at, so `agentsmd.subprojectPath` no longer has to repeat the directory by hand.

- New `rulesync.jsonc` option `deriveSubprojectPathFromGlobs` (default `false`), also honored from `rulesync.local.jsonc`.
- New per-rule opt-in `agentsmd: { subprojectPath: "auto" }`, which derives regardless of the config option.
- Precedence: explicit non-"auto" value > derived directory > no nesting. Root rules never derive.

## Changes

- New config option `deriveSubprojectPathFromGlobs` (default `false`) derives `agentsmd.subprojectPath` from each non-root rule's `globs` (`["packages/api/**/*"]` -> `packages/api`), so those rules are written as nested `AGENTS.md` by the targets that nest instead of `.agents/memories/<rule>.md`. Honored from `rulesync.jsonc` and `rulesync.local.jsonc`.
- Per-rule `agentsmd: { subprojectPath: "auto" }` opts a single rule in regardless of the option; per-rule `agentsmd: { subprojectPath: "" }` opts a single rule out (default placement, nothing derived) when the option is on. An explicit directory always wins; `root: true` rules never nest.
- Warning rule: only an explicit `"auto"` that cannot be honored (no shared wildcard-free prefix, disagreeing globs, negation, `..`, absolute path, or a root rule) warns, once per rule, naming the file. Config-driven derivation falls back silently, so typical activation globs such as `["src/**/*.ts", "test/**/*.ts"]` do not warn on every generate.
- `src/utils/glob-static-prefix.ts`: pure helper returning the longest leading run of wildcard-free POSIX segments shared by every glob of a rule. Rejects absolute paths, `..`, negation, backslashes and a leading brace expansion; strips `./`; `**/*.ts`, `*.md` and `[]` yield nothing. A wildcard-free final segment is read as a file name (`["packages/api"]` -> `packages`); write `packages/api/**` for a directory. Segments are joined with `"/"` rather than spread into `posix.join`, which threw a RangeError on a pattern with hundreds of thousands of segments.
- `RulesyncRule` resolves the placement once in its constructor (`getFrontmatter()`), so every consumer (agentsmd, codexcli, kiro, vibe, reasonix, ...) picks it up unchanged and `"auto"` never leaks as a path; the authored frontmatter stays behind `getAuthoredFrontmatter()` and `getFileContent()` writes `"auto"` back verbatim. An `agentsmd` block left empty after stripping an unresolvable `"auto"` is dropped rather than returned as `{}`. Collisions between a derived path and an explicit one compose via `RulesProcessor.mergeRulesByOutputPath` exactly like two explicit paths.
- MCP `rule` list/get/put return the authored frontmatter (the value in the file), not the resolved placement, so a get -> edit -> put round trip keeps `"auto"` and `put` answers with what it wrote. Documented in `docs/reference/mcp-server.md`.
- `src/features/rules/rules-processor.ts`, `src/lib/generate.ts`, `src/config/config.ts`, `src/config/config-resolver.ts`: thread the option from `Config` into `RulesyncRule.fromFile`; schema, getter, defaults, file/local-config merge.
- Docs: `docs/guide/configuration.md` (option), `docs/reference/file-formats.md` (derivation rules, precedence, opt-out, warning rule, collisions, `--delete`, and that `globs` keeps its conditional-activation meaning for Cursor/Cline/Kiro); `src/generated/docs-content.ts` regenerated.

No emitted paths change, so `.gitignore` derivation is untouched. Directories are still created lazily by the writer. The MCP `generateOptionsSchema` and the programmatic `GenerateOptions` are intentionally not extended in this PR.

## Test plan

- [x] `src/utils/glob-static-prefix.test.ts`: prefix derivation table (positive prefixes, every rejected pattern, multi-glob agreement/disagreement, wildcard-free final segment) and a 200k-segment pattern.
- [x] `src/features/rules/rulesync-rule.test.ts`: derivation via option and `"auto"`, explicit path wins, `""` opt-out with the option on (no warning), root rule never derives (`"auto"` warns), unresolvable `"auto"` warns once and drops `agentsmd` when empty / keeps other keys, config-driven underivable globs fall back silently, `"auto"` still warns with the option on, `fromFile` honors the option, `getAuthoredFrontmatter` keeps `"auto"` and schema defaults.
- [x] `src/mcp/rules.test.ts`: get returns `"auto"` verbatim, underivable `"auto"` kept as authored, put/get/list/put round trip leaves `subprojectPath: auto` in the file.
- [x] `src/features/rules/rules-processor.test.ts`: a derived path and an explicit path to the same directory compose into one `packages/app/AGENTS.md`; loaded rules derive only when the processor option is on.
- [x] `src/features/rules/agentsmd-rule.test.ts`: a rule imported from `packages/api/AGENTS.md` resolves to the same path with the option on and off.
- [x] `src/config/*.test.ts`: default, schema validation, `rulesync.jsonc` + `rulesync.local.jsonc` override.
- [x] `src/e2e/e2e-rules.spec.ts`: `generate --delete` with the option on writes `packages/api/AGENTS.md`, removes the stale `.agents/memories/api.md`, and lists the nested file from the root `AGENTS.md`.
- [x] `pnpm cicheck` green (399 test files, 10405 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #2885, opened 2026-09-02
fix(skills): close the nested skills root edges recorded in #2852

### LINKED ISSUE #2852: Follow-up: low-severity edges left from the nested skills root path handling (#2849)
Follow-up to #2849 (which closed #2804). Four low-severity observations were recorded during review and intentionally left unfixed there, because the review loop had reached its iteration cap and none of them was `mid` or above. They are collected here so they are not lost.

### 1. `relativeDirPath` is recorded in its pre-resolution spelling

`getConfiguredImportRoots` in `src/features/skills/claudecode-skill.ts` validates a nested skills root against its **resolved** relative path, but records `relative(outputRoot, scannedDirPath)` — the lexical spelling the glob reported. That spelling is also the path actually read, so recording the resolved one would change what is read, not just what is reported. The only observable effect is the scoping of the derived `paths` glob for a root reached through a symlink.

### 2. The `.claude/skills` seed is missed when the project's own skills directory is a symlink

The dedup seed is the literal `.claude/skills` posix path. If the project's own `.claude/skills` is itself a symlink, a nested root resolving to the same place is not recognised as the seed. This fails closed — the tail check rejects it — so the impact is reduced strictness only, and the precondition is narrow.

### 3. `nativePathToPosix` inside `sharedTrailingSegments` is currently a no-op

`src/utils/file.ts`: `sharedTrailingSegments` calls `nativePathToPosix` and then `splitPathSegments`, which splits on `/[/\\]/` — so the conversion never changes the result on either platform. It is harmless today, but it reads as a defence that is doing work, and it is asymmetric with the identity side (which does not convert). If `splitPathSegments` is ever narrowed to `/` only, this line quietly becomes load-bearing on Windows.

### 4. The identity fallback in `dedupeNamesByFileIdentity` is unreachable

`src/utils/file.ts`: `identities[index] ?? nativePathToPosix(join(dirPath, entry.name))` — the `??` branch cannot run, because `mapWithConcurrency` returns one element per input and `realFileIdentity` catches its own errors and never rejects. The preceding comment describes behaviour that cannot occur.

Items 3 and 4 are comment/clarity issues; items 1 and 2 are narrow behavioural edges. None is urgent.

dyoshikawa:
Closes #2852

## Summary

Follow-up to #2849. Closes the four low-severity edges recorded in #2852 around the nested `.claude/skills` root handling: one narrow behavioural gap (the dedup seed missing a symlinked project root), one behaviour pinned and documented rather than changed (the recorded root spelling), and two clarity fixes in `src/utils/file.ts`.

## Changes

- **Seed the nested-root dedup with the resolved project root** (`claudecode-skill.ts`): `getConfiguredImportRoots` now seeds `seenRealRelativeDirPaths` with both the literal `.claude/skills` and where it resolves to via `resolvedRelativePath`. A nested root that reaches the project's own (symlinked) skills directory is recognised as the tool's own root and not read a second time, leniently. Resolution failure keeps the literal seed; an unresolvable root is refused before the lookup, so the fail-closed behaviour is unchanged. One sentence added to `docs/reference/file-formats.md` (docs content regenerated).
- **Keep recording the reported spelling, and say why** (`claudecode-skill.ts`): the reported path is the one that is read and the one `deriveNestedSkillPaths` scopes the skill to; the resolved path is only used to judge and dedupe. Recording the resolved one would change what is read, and scoping to it would move the skill to a location the scan never reported. A test pins this for a root reached through a link.
- **Drop the no-op `nativePathToPosix` in `sharedTrailingSegments`** (`file.ts`): `splitPathSegments` splits on both separators, so both sides are already split symmetrically. The helper is exported for a test that a Windows-style spelling still shares its trailing segments with its posix identity, and that a backslash inside a name splits the same on both sides.
- **Remove the unreachable identity fallback in `dedupeNamesByFileIdentity`** (`file.ts`): the mapper now returns `{ entry, identity }` pairs, so entries are no longer matched back to identities by index and no `??` or non-null assertion is needed. The comment describing the impossible case is gone.

## Test plan

- [x] `pnpm cicheck` passes (398 files, 10332 tests; oxlint `--max-warnings 0`, typecheck, cspell, secretlint, docs-content drift check)
- [x] New `it.skipIf(process.platform === "win32")` test: a symlinked project `.claude/skills` is not read as a nested root while a genuine nested root still is (verified to fail before the seed fix)
- [x] New test: a nested root reached through a link is recorded under the reported spelling and derives `paths: ["a/b/**"]`
- [x] New `sharedTrailingSegments` tests covering a Windows-style path against a posix identity and a backslash inside a name on posix
- [x] Existing nested-root, backslash and symlink tests unchanged and passing


dyoshikawa:
@dyoshikawa Thank you!

PR #2886, opened 2026-09-02
fix(skills): match the label budget to the width the inquirer renderer measures

### LINKED ISSUE #2847: fix(fetch): skillLabelBudget's JSDoc is wrong about how cli-width resolves a zero width
## Summary

The `skillLabelBudget` JSDoc in `src/lib/skill-prompt.ts` explains why a reported width of zero is treated as "no width at all", and ends with a parenthetical that is wrong on both of its claims:

> (`cli-width` reads `CLI_WIDTH` from the environment before it falls back to 80, which this does not; a prompt is refused outright without a TTY, and a TTY answers before either is asked.)

## Why it is wrong

`@inquirer/core` does not hand `process.stdout` to `cli-width`. `createPrompt` pipes a `MuteStream` to the output stream and `readlineWidth()` measures *that*:

- `@inquirer/core@12.0.0/dist/lib/create-prompt.js:41` — `const output = new MuteStream();`
- `@inquirer/core@12.0.0/dist/lib/utils.js:24` — `return cliWidth({ defaultWidth: 80, output: readline().output });`

And `cli-width@4.1.0` resolves in this order:

1. `opts.output.getWindowSize()` — `MuteStream` has no `getWindowSize`, so this branch never fires.
2. `opts.tty.getWindowSize()` — `require("tty").getWindowSize` is `undefined` on modern Node, so this never fires either.
3. `opts.output.columns` — `MuteStream` exposes `get columns()` proxying the destination stream. With a TTY mid-resize this is `0`, which is falsy, so the branch is skipped.
4. `process.env.CLI_WIDTH`
5. `defaultWidth` (80)

So in exactly the case the comment is describing — a TTY reporting zero columns — `CLI_WIDTH` **is** consulted before the 80 fallback. The claim that "a TTY answers before either is asked" only holds when the TTY reports a non-zero width, which is not the case the paragraph exists to explain.

## Impact

Comment accuracy, plus one narrow real consequence: with a TTY reporting zero columns and `CLI_WIDTH` set below 77, `skillLabelBudget` budgets against 80 while the renderer breaks rows at `CLI_WIDTH`, so a label can wrap — the second line this cap exists to prevent. This is a rare combination and not a security hole (a wrapped line carries no confusable note and cannot forge one), but the comment currently asserts the opposite of the code.

## Suggested fix

Correct the parenthetical: state that the renderer measures the `MuteStream` `@inquirer/core` pipes to the output, that `MuteStream` proxies `columns` from the destination, and that a zero there falls through to `CLI_WIDTH` and only then to 80. If the wrap is worth closing rather than just documenting, read the same `CLI_WIDTH` fallback here so the budget tracks the width the renderer actually uses.

Follow-up to #2843 (found in review after that PR merged).

dyoshikawa:
Closes #2847

## Summary

`skillLabelBudget` in `src/lib/skill-prompt.ts` documented the inquirer renderer's width measurement inaccurately and, in one narrow case, budgeted against a width the renderer does not use. `@inquirer/core` pipes a `MuteStream` to the output stream and hands that stream to `cli-width`; `MuteStream` proxies `columns` from `process.stdout`, and `cli-width` skips a zero there in favor of the `CLI_WIDTH` environment variable before it falls back to 80. The budget read `process.stdout.columns || 80`, so a TTY reporting zero columns under a `CLI_WIDTH` narrower than 77 had its labels cut to 72 while the rows were broken at `CLI_WIDTH`, and the label could wrap onto the second line the cap exists to prevent.

## Changes

- Rewrite the JSDoc parenthetical to describe the real chain: `MuteStream` piped to the output, `cli-width` trying `output.getWindowSize`, `tty.getWindowSize` (undefined on current Node), the proxied `columns`, then `CLI_WIDTH`, then 80. Soften the "same stream the renderer measures" paragraph to say `process.stdout` stands in for the `MuteStream` because `checkbox` is called without an `output` option.
- Add `configuredTerminalWidth()`, which parses `CLI_WIDTH` the way `cli-width` does (`parseInt` in base ten, rejecting `NaN` and zero), and consult it when `process.stdout.columns` is falsy before falling back to 80.
- Tests: columns 0 with `CLI_WIDTH=60` budgets 55 columns; `abc`, `0` and an empty value fall back to 80 alongside the renderer; a reported terminal width leaves `CLI_WIDTH` unread. The suite now pins `CLI_WIDTH` unset in `beforeEach` and restores it with `vi.unstubAllEnvs()` so an inherited value cannot skew the existing zero-width and pipe cases.

Verified against the installed tree: `@inquirer/checkbox` 5.2.3 -> `@inquirer/core` 12.0.1 (`create-prompt.js`, `utils.js`), `mute-stream` 3.0.0 (`get columns()`), `cli-width` 4.1.0.

## Test plan

- [x] `pnpm cicheck` passes (398 test files, 10333 tests; fmt, oxlint, typecheck, cspell, secretlint, docs-content, gitignore and supported-tools checks clean)
- [x] `src/lib/skill-prompt.test.ts` covers columns 0 + `CLI_WIDTH=60` (55-column label), `CLI_WIDTH=abc` / `0` / empty (80-based budget), and `CLI_WIDTH` ignored while the terminal reports a width


dyoshikawa:
@dyoshikawa Thank you!

PR #2887, opened 2026-09-02
fix(opencode): resolve skills.paths before judging whether they escape the config directory

### LINKED ISSUE #2850: fix(skills): take opencode configured skill roots from the resolved path
`ClaudecodeSkill.getConfiguredImportRoots` now decides whether a discovered nested skills root is usable by resolving it (`resolvedPathEscapesRoot` in `src/utils/file.ts`), so a name inside the project that is a symbolic link out of it is refused.

`OpencodeSkill.getConfiguredImportRoots` (`src/features/skills/opencode-skill.ts:150-156`) still filters a configured `skills.paths` entry lexically — `!isAbsolute(candidate) && !normalize(candidate).startsWith("..")` — so a relative path naming a symlink that points outside the project is accepted and its `SKILL.md` files are read.

This is pre-existing behaviour, not a regression, and the config comes from the repository's own `opencode.json`, so the impact is limited to reading directories that contain a `SKILL.md`. It was left out of #2849 because changing another tool's import behaviour needs its own tests.

Proposed change: replace the hand-rolled lexical test with `resolvedPathEscapesRoot({ rootPath, targetPath })` so both roots enforce the same rule, and add tests covering a configured path that is a link out of the project.

Related: #2804, #2849

dyoshikawa:
Closes #2850

## Summary

`OpencodeSkill.getConfiguredImportRoots` filtered the `skills.paths` entries of `opencode.json` / `opencode.jsonc` only lexically (`!isAbsolute(candidate) && !normalize(candidate).startsWith("..")`), so a relative entry naming a symbolic link that points outside the config directory passed the test and its `SKILL.md` files were read. The nested Claude Code skills roots are already judged by where they resolve (`resolvedPathEscapesRoot` in `src/utils/file.ts`); this change makes the OpenCode configured roots follow the same rule.

## Changes

- `src/features/skills/opencode-skill.ts`: after the existing lexical prefilter, each remaining candidate is also checked with `resolvedPathEscapesRoot({ rootPath: configDir, targetPath: join(configDir, candidate) })` and dropped when it resolves outside the config directory. The helper's fail-closed semantics are kept: a path that cannot be resolved falls back to its literal spelling, so a non-existent relative entry stays accepted as before (the scan finds no directory under it either way). Dropped entries stay silent, like the lexical drops already were. JSDoc updated.
- `src/features/skills/opencode-skill.test.ts`: three new cases under `getConfiguredImportRoots`: a relative entry that is a symbolic link pointing outside the config directory is dropped, one that is a link staying inside is kept, and one that does not exist is kept. The two symlink cases are skipped on Windows.
- `docs/reference/file-formats.md`: the OpenCode skills note now says an entry is judged by where it resolves rather than by how it is spelled, so a symbolic link pointing outside the directory is ignored too. `src/generated/docs-content.ts` regenerated.

## Test plan

- [x] `pnpm cicheck` passes (398 test files / 10331 tests; fmt, oxlint, typecheck, cspell, secretlint, docs-content / gitignore / supported-tools drift checks)
- [x] New `getConfiguredImportRoots` tests: escaping symlink dropped, in-tree symlink kept, non-existent path kept
- [x] Existing `getConfiguredImportRoots` tests (absolute / `..` / empty / non-string entries dropped, global root resolved against `~/.config/opencode/`) still pass


dyoshikawa:
@dyoshikawa Thank you!

PR #2888, opened 2026-09-02
fix(fetch): drop skill names that Windows folds or aliases

### LINKED ISSUE #2848: fix(fetch): a skill name Windows folds onto a local one is written without a word
## Summary

On Windows, a fetched skill directory whose name differs from a local skill's only by a trailing dot (or by an 8.3-style `~1` short name) resolves to the local skill's directory when it is written, so the fetch replaces files of a skill the user already had. Nothing on the way in says so: the name is not refused, and it carries no confusable note either.

## Where it stands today

- The **prune** side already guards against this. `pruneStaleSkillFiles` in `src/lib/fetch.ts` refuses to walk a skill directory whose name a filesystem may fold onto another entry in the listing, and its comment states the case outright — the write and the prune agree with each other and disagree with the summary.
- The **write** side has no equivalent guard. `classifySkillPath` (`src/lib/fetch.ts`) drops names that cannot be shown honestly — control characters, invisible characters, names that draw as nothing — but a trailing dot is an ordinary visible character, so `deploy.` passes.
- The **confusable check** does not see it either. `describeConfusableNames` compares display forms and Latin skeletons; `normalizedFormOf` trims whitespace but not dots, so `describeConfusableNames({ names: ["deploy."], localNames: ["deploy"] })` returns no note. The trailing-space spelling (`deploy `) is caught, by `hasWhitespaceThatDoesNotShow`; the dot and `~1` forms are the gap.

With the default `--conflict overwrite`, the effect on Windows is that the contents of a long-trusted local `deploy` skill are replaced by a repository that never published a directory called `deploy`.

## Why the fix belongs on the write side

A note in the picker is the wrong instrument here: the collision is not one a reader is being asked to judge, it is a path that resolves somewhere other than where it reads. The name should be refused the way the other dishonest names are, or the write should be redirected to a path that cannot fold.

## Suggested fix

Bring the prune guard's notion of a Win32-foldable name to `classifySkillPath`: treat a skill directory name with a trailing dot or trailing space, or one shaped like an 8.3 short name (`NAME~<digits>`), as an `unsafe-name` and drop it with the warning the other dropped names get. Applying it on every platform rather than only on Windows matches how the existing guards are written — the tree may be created on one platform and used on another.

Found in review of #2846 (the local-skill comparison that PR adds); pre-existing and out of that PR's scope.

dyoshikawa:
Closes #2848

## Summary

`rulesync fetch` wrote skill directories whose names Windows folds or aliases: a name with a trailing dot or space (`deploy.`, `deploy `) is resolved by Win32 to the plain name, and a `NAME~<digits>` name (`DEPLOY~1`) is an 8.3 short-name alias that opens whatever long name it stands for. With the default `--conflict overwrite`, a remote `deploy.` therefore replaced the files of a local `deploy` skill that the repository never published. The prune step already refused exactly those shapes; the write side and the confusable check did not see them.

## Changes

- `src/lib/fetch.ts`
  - Lift the prune guard's two regexes into one named predicate, `hasWindowsFoldableSkillName`, shared by the write and the prune.
  - `classifySkillPath` returns `unsafe-name` for such a name, with a new `reason` field (`"hidden"` for the existing control/invisible-character drop, `"foldable"` for this one), so the skill is left out of the interactive picker, rejected by `--skills` as an unknown skill, and never written. The check is not gated on the platform, matching the existing guards.
  - Foldable names are reported through their own warning (`formatFoldableSkillsWarning`), which quotes each name as it is so a trailing space shows, and says why the shapes are dropped. The hidden-character warning keeps its wording.
  - The prune guard stays as a backstop and now uses the shared predicate.
- `docs/reference/cli-commands.md`: describe the drop under "Skill Names That Look Alike", turn the prune bullet into a cross-reference, and switch the examples that used a trailing space to a leading one (a trailing one is no longer on the list). `src/generated/docs-content.ts` regenerated.
- `src/lib/fetch.test.ts`: new cases for `deploy.`, `deploy `, `DEPLOY~1` (dropped and warned, in interactive and plain runs), `data~2parser` and `v1.0` (still fetched), `--skills deploy.` (unknown skill), and a local `deploy` left untouched under `--conflict overwrite` when the remote ships `deploy.`. The three prune tests that fed the old guard now assert the drop, and the whitespace-note test uses a leading blank.

## Test plan

- [x] `pnpm cicheck` passes (398 test files, 10335 tests; fmt, oxlint, typecheck, cspell, secretlint, docs-content, gitignore and supported-tools checks clean)
- [x] `deploy.`, `deploy ` and `DEPLOY~1` are neither offered in the picker nor written, and the warning names them quoted
- [x] `data~2parser` and `v1.0` are fetched as before
- [x] A local `skills/deploy/` keeps its content when the remote publishes `deploy.` under `--conflict overwrite`
- [x] `--skills deploy.` fails with "Unknown skill(s)" instead of writing


dyoshikawa:
@dyoshikawa Thank you!

PR #2889, opened 2026-09-02
feat(config): add a root language key that steers the answer language

### LINKED ISSUE #2863: Add a root "language" key to rulesync.jsonc to steer the AI response language
## Background

Many rulesync users are not native English speakers and want their AI coding agent to *answer* in their own language while keeping reasoning (and the code, comments, and commit messages) in English. Today the only way to express that is to hand-write a sentence such as "You must always answer in Japanese" into a rule file for every project, and repeat it for every tool whose root instruction file differs.

Because this is a cross-tool, project-wide preference rather than a rule about the codebase, it belongs in `rulesync.jsonc` as configuration — and rulesync can then translate that one setting into whatever mechanism each tool actually supports (a native settings key where one exists, an appended prompt where one does not).

## Details

Requested behavior:

1. **New root key `language` in `rulesync.jsonc` / `rulesync.local.jsonc`.**
   - Value is a union type of major language codes (`"ja"`, `"en"`, ...) rather than a free-form string, so editors can complete it and typos are rejected.

2. **Claude Code (native): write `language` into `.claude/settings.local.json`.**
   - e.g. `"language": "ja"`.

3. **AGENTS.md-style tools (simulated): append a prompt to the generated root instruction file.**
   - The generated `AGENTS.md` gets a trailing block such as:

     ```markdown
     ---

     You must always answer in Japanese. On the other hand, reasoning(thinking) should be in English to improve token efficiency.
     ```

   - The `---` separator is intentional: it makes the prompt concatenation visible, so readers can tell the block was appended by rulesync rather than authored as part of the rules.

Relevant current state:

- `src/config/config.ts` — `ConfigParamsSchema` holds the root-level config keys (`targets`, `features`, `simulateCommands`, `gitignoreDestination`, ...). `ConfigFileSchema` = `$schema` + partial params, and `scripts/generate-json-schema.ts` (`pnpm run generate:schema`) derives the published JSON Schema from it, so a new key gets editor completion for free once added here.
- `src/config/config-resolver.ts` — merges `rulesync.jsonc` and `rulesync.local.jsonc`; the new key must participate in that merge (local overrides base).
- `.claude/settings.local.json` is already a rulesync-written shared config file (claudecode `ignore` with `fileMode: "local"`, see `src/constants/shared-config-paths.ts`), written through `src/features/shared/shared-config-gateway.ts`. `SHARED_CONFIG_OWNERSHIP` rejects writes outside the declared ownership, so a new `language` key needs an ownership declaration and an owning feature.
- Root instruction files are emitted by the `rules` feature (`src/features/rules/*-rule.ts`); `AGENTS.md` comes from `AGENTSMD_RULE_FILE_NAME` in `src/constants/agentsmd-paths.ts` and is reused by several targets (agentsmd, devin, kiro, antigravity-plugin, ...), while other tools have their own root file (`CLAUDE.md`, `GEMINI.md`, ...).

1. **Language set** — which codes make up the union? Suggested starting point: `en`, `ja`, `zh-CN`, `zh-TW`, `ko`, `fr`, `de`, `es`, `pt-BR`, `ru`. Decide the code style (ISO 639-1 vs BCP 47) up front, since widening later is easy but changing the spelling is a breaking change. Also decide whether `"en"` is a no-op (emit nothing) or still emits an explicit instruction.
2. **Does the Claude Code `language` setting actually exist?** This must be verified against the official Claude Code settings documentation / `https://json.schemastore.org/claude-code-settings.json` before implementing — writing an unknown key into `settings.local.json` would be worse than falling back to the appended-prompt path. If it does not exist, Claude Code should use the `CLAUDE.md` appended-prompt path instead. (Network access was unavailable while filing this issue, so it is unverified.)
3. **Why `settings.local.json` and not `settings.json`?** Response language is arguably a personal preference rather than a team-shared one, which argues for the local (gitignored) file — but that also means the setting silently does not apply to teammates who share the repo. Confirm the intent, and consider whether the file should be selectable.
4. **Prompt template** — where does the per-language sentence live, and is it user-overridable? A `Record<Language, string>` constant is the simplest start; a future `language: { code: "ja", prompt: "..." }` object form could allow customization. Keep the schema shape in mind so the string form can be widened later without a breaking change.
5. **Which files get the appended block** — root instruction files only, or nested/scoped ones too? Root-only is the sane default (the block is a global directive, and repeating it in every nested file is noise). Confirm whether tools with their own root file (`CLAUDE.md`, `GEMINI.md`, `.github/copilot-instructions.md`, ...) all get it, or only the `AGENTS.md` family.
6. **Import round-trip** — `rulesync import` must not read the appended block back in as part of a rule body, or the text will duplicate on the next generate. The block needs to be recognizable (a stable marker/format) and stripped on import.
7. **Interaction with #2862** — that issue adds `$schema` to the same Claude Code settings files. Both touch the settings write path and `SHARED_CONFIG_OWNERSHIP`; implement them in an order that avoids conflicting edits to the same declarations.

## Solution / Next Steps

1. Add a `Language` union schema (e.g. `src/types/language.ts`) and an optional `language` key to `ConfigParamsSchema` in `src/config/config.ts`; regenerate the config JSON Schema with `pnpm run generate:schema` and confirm `rulesync.local.jsonc` overrides the base file.
2. Verify whether Claude Code supports a native `language` settings key (open question 2). If yes, write it into `.claude/settings.local.json` through the shared config gateway with a proper `SHARED_CONFIG_OWNERSHIP` declaration; if no, fall back to the appended-prompt path for Claude Code as well.
3. Add a per-language prompt template constant and append the block — preceded by a `---` separator — to the generated root instruction files (`AGENTS.md` and equivalents), emitting nothing when `language` is unset.
4. Make `rulesync import` strip the generated block so the setting round-trips without duplicating text.
5. Survey other targets for native language settings worth using instead of an appended prompt (Cline's `preferredLanguage`, Roo Code / Kilo Code language settings, etc.) and note the findings in this issue before widening scope.
6. Follow `.claude/rules/feature-change-guidelines.md`: check whether `rules-processor.ts` needs an updated `additionalConvention`, decide project vs global scope support, and run `pnpm dev gitignore` if any new output path appears.
7. Add unit tests plus coverage in the Tool × Feature end-to-end happy-path tests, and document the new key in `docs/**/*.md` (then run `pnpm run generate:docs-content`) and in `README.md` if the configuration reference lists root keys.

dyoshikawa:
Closes #2863

## Summary

Adds a root `language` key to `rulesync.jsonc` (overridable from `rulesync.local.jsonc`) that steers the language the AI answers in. Accepted values: `en`, `ja`, `zh-CN`, `zh-TW`, `ko`, `fr`, `de`, `es`, `pt-BR`, `ru`. When the key is omitted rulesync says nothing about language, so `en` is an explicit instruction rather than the default.

## Changes

- **Config**: `language` in `ConfigParamsSchema` (config-file only, no CLI flag), `Config.getLanguage()`, local-over-base merge in the resolver; JSON Schema regenerated.
- **Claude Code (native)**: for `claudecode` / `claudecode-legacy` the `rules` feature patches `"language": "<name>"` (e.g. `japanese`, `traditional chinese`) into `.claude/settings.local.json` (project) or `~/.claude/settings.json` (global; Claude Code reads no `~/.claude/settings.local.json`) through the shared config gateway. `SHARED_CONFIG_OWNERSHIP` now declares `rules` as owner of exactly the `language` key on both Claude settings files; nothing else in the file is touched, `CLAUDE.md` gets no block, and an unset key never touches the settings file.
- **All other targets (simulated)**: the rules processor appends one block to the generated root rule file only, the file built from the `root: true` rule, including targets whose adapters never mark a root (Cursor) and root mirrors (Rovo Dev). Nested rules never get it. The block is body content after a thematic break, built from a single template in `src/types/language.ts`, so no adapter carries language code:

  ```
  ---

  You must always answer in Japanese. On the other hand, reasoning (thinking) should be in English to improve token efficiency.
  ```

- **Import round trip**: `rulesync import` strips a trailing block for any supported language (regex built from the enum) from every imported rule, so import followed by generate yields exactly one block.
- **Docs**: `docs/guide/configuration.md` (example key + "Response Language" section), `docs/reference/file-formats.md` (response language note); embedded docs regenerated.

## Test plan

- [x] `src/types/language.test.ts`: schema, display data, append, strip for every code (incl. CRLF, bare instruction, non-trailing / different-language text left alone)
- [x] Config / resolver: accepts codes, rejects unknown values, `rulesync.local.jsonc` overrides base
- [x] Gateway: `rules` may write exactly `language` to both Claude settings files; unowned keys rejected; derived-writer snapshot updated
- [x] `ClaudecodeLanguageSettings`: project/global paths, patch preserves sibling keys, never deletable
- [x] RulesProcessor: block on root only for every non-Claude rules target (matrix), after reference/convention sections, Cursor root-source file, empty body, unset language, Claude Code settings ToolFile (project + global), import stripping (agentsmd and Cursor)
- [x] E2E: generate → import → generate keeps a single block; Claude Code writes `settings.local.json` and leaves `CLAUDE.md` alone; unset language creates no settings file
- [x] `pnpm cicheck` passes (10364 tests)


dyoshikawa:
@dyoshikawa Thank you!

PR #11, opened 2025-07-02
feat: improve Cursor .mdc file processing with comprehensive frontmatter handling

arrowkato:
## Overview
Significantly improved Cursor `.mdc` file processing with complete bidirectional frontmatter conversion support.

### 1. Enhanced Generate Feature (rulesync/*.md → .mdc)
- Support for 4 rule types: always, manual, auto attached, and agent_request
- Implemented accurate conversion logic based on `900_generate_siyou.mdc` specification
- Proper handling of frontmatter fields: `alwaysApply`, `description`, and `globs`

### 2. Enhanced Import Feature (.mdc → rulesync/*.md)
- Implemented detection logic for 4 patterns: always, manual, auto attached, and agent_request
- Complete conversion processing based on `901_import_siyou.mdc` specification
- Custom YAML parser to handle Cursor-specific `globs` format (comma-separated strings)
- Proper handling of edge cases

### File Changes
- `src/generators/rules/cursor.ts`: Implemented new generation logic
- `src/parsers/cursor.ts`: Implemented new parsing logic and custom YAML parser
- `src/core/parser.ts`: Allow empty string descriptions
- Test cases: Added comprehensive test suites

### New Features
- `determineCursorRuleType()`: Automatic rule type detection
- `convertCursorMdcFrontmatter()`: Conversion from .mdc to rulesync format
- `convertGlobsToArray()`: Comma-separated string to array conversion
- Custom YAML parser: Support for Cursor-specific formats

## Testing
- All 16 test cases pass successfully
- Comprehensive tests for all 4 patterns
- Edge case handling verified

## Impact
- Significantly improved processing accuracy for Cursor `.mdc` files
- No impact on existing functionality (backward compatibility maintained)
- Both `pnpm dev generate --cursor` and `pnpm dev import --cursor` commands improved

## Breaking Changes
None. This change maintains full backward compatibility while adding new functionality.

PR #2890, opened 2026-09-02
fix(prompt): count East Asian Ambiguous characters as two columns

### LINKED ISSUE #2845: fetch: an ambiguous-width name can still wrap a forged row on a terminal that draws the ambiguous class wide
`displayWidthOf` counts East Asian Ambiguous characters at one column, which is what a terminal running a Latin font draws them in. A terminal configured to draw the ambiguous class wide — a common setting in CJK locales — draws them at two, and `fast-string-width`, which is what `@inquirer/core` breaks the prompt's rows at, also counts them at one. So neither the label budget nor the renderer's own wrapping sees the extra columns; the physical terminal is what breaks the row, and it breaks it onto a continuation line that carries no pointer and no checkbox.

That is the same forged-row primitive #2813 closed for the other width mismatches, reached through a terminal setting rather than through a character the budget mismeasures.

## Reproduction

A skill directory named:

```
"─".repeat(60) + "● pdf-tools"
```

- passes `hasDeceptiveHiddenCharacters`, so it is offered as a row rather than refused;
- measures 71 columns under `displayWidthOf`, inside the 72-column cap, so it is emitted whole;
- draws at 132 columns where the ambiguous class is wide, and hard-wraps.

With the inactive prefix drawn at four columns, 38 box-drawing characters put the break exactly in front of `●`, so the continuation line reads `● pdf-tools` at column 0 — `●` being close to the shape the checkbox glyph itself draws in.

## Scope

Pre-existing rather than introduced by #2843: the ambiguous-width choice is documented in the `WIDE_CHARACTERS_PATTERN` comment in `src/utils/display-width.ts`, and #2843 only paid for the two extra columns the prompt's own prefix (`❯◯`) spends in such a terminal, by reserving five columns for it. The characters inside the label were left at one column each.

## Possible fix

Count the UAX #11 Ambiguous class at two columns when measuring an untrusted name. Overstating a width shortens a label that did not need shortening; understating one forges a row, which is the asymmetry `display-width.ts` already argues for elsewhere. The cost is that ordinary names carrying an ambiguous character — a Greek letter, a box-drawing character, an accented Latin letter from the Latin-1 supplement — would be cut a little earlier on a Latin terminal.

Worth weighing against splitting the measurement in two: one width model for bounding an untrusted name and one for laying out the tool's own text, which `displayWidthOf` currently serves both of.

Found during the security review of #2843 (https://github.com/dyoshikawa/rulesync/pull/2843).

dyoshikawa:
Closes #2845

## Summary

`displayWidthOf` counted the UAX #11 East Asian Ambiguous class at one column, which is what a Latin terminal draws, while a terminal set to draw the class wide (a common setting in CJK locales) draws it at two. The prompt's own renderer, `fast-string-width`, counts the class at one as well, so neither the label budget nor the renderer saw the extra columns: a skill directory named `"─".repeat(60) + "● pdf-tools"` passed `hasDeceptiveHiddenCharacters`, measured 71 columns, fit the 72-column budget, and was hard-wrapped by the terminal at 132 columns, leaving `● pdf-tools` at the left margin of a continuation line that carries no pointer and no checkbox. This is the forged-row primitive #2813 closed for the other width mismatches, reached through a terminal setting.

The Ambiguous class is now counted at two columns, the single-wide-model option the issue weighs as the smaller change. Overstating a width shortens a label that did not need it; understating one lets a forged row wrap.

## Changes

- `src/utils/display-width.ts`: add `AMBIGUOUS_CHARACTERS_PATTERN`, a hand-written escaped range table of the `East_Asian_Width=A` property of Unicode 17.0, derived from the table `get-east-asian-width` 1.6.0 carries (the package is not imported and no dependency is added). Adjacent ranges are merged; the combining stretches of the property (U+0300–U+036F, U+FE00–U+FE0F, U+E0100–U+E01EF) are left out because the mark rule counts them before a width is looked up, which also keeps `no-misleading-character-class` quiet. `widthInContext` returns two for the class. `ELLIPSIS_WIDTH` is derived from `displayWidthOf(SHORTENING_ELLIPSIS)` and becomes two, since the ellipsis is itself Ambiguous. The module comment that stated the one-column rule now explains the asymmetry.
- `src/lib/skill-prompt.ts`: comments only. `NOTE_SEPARATOR` (an em dash, also Ambiguous) and `NOTE_MARKER` were already measured through `displayWidthOf`, so they follow automatically; the `CHOICE_PREFIX_WIDTH` and `NOTE_SEPARATOR` comments no longer restate the one-column rule.
- `src/utils/display-width.test.ts`: U+2500, U+25CF, U+03B1, U+00E9, the em dash and the ellipsis at two columns; the Neutral U+25C9 and U+276F still at one; the 132-column measurement of the issue's name; `ELLIPSIS_WIDTH` at two and the cut it costs.
- `src/lib/skill-prompt.test.ts`: reproduction with the issue's name asserting the label is cut to 35 box-drawing characters plus the ellipsis (72 columns); the pinned label expectations move by the extra column of the separator and of the ellipsis (e.g. `pdf` + 67 `o` rather than 68 at the 80-column fallback). Every new value was checked against the budget arithmetic by hand.

Not changed: `src/features/permissions/vibe-permissions.ts` still uses `shortenToWidth` for the tool's own log text and shortens slightly earlier where an entry carries an Ambiguous character; harmless, and its tests pass unchanged. No `docs/**` page describes the width rule, so no documentation change.

The cost is that names carrying Greek, Cyrillic or accented Latin-1 letters are cut a little earlier on Latin terminals (about 36 such characters fill the 72-column budget in the worst case).

## Test plan

- [x] `pnpm cicheck` passes: 398 test files, 10369 tests, oxfmt/oxlint/typecheck clean, cspell 0 issues, secretlint clean
- [x] `src/utils/display-width.test.ts`: Ambiguous characters at two columns, Neutral at one, the issue's name at 132 columns, `ELLIPSIS_WIDTH === 2`
- [x] `src/lib/skill-prompt.test.ts`: the issue's name is offered as `"─".repeat(35) + "…"`, 72 columns wide
- [x] Range table diffed against the output of `eastAsianWidthType` from `get-east-asian-width` 1.6.0 over every code point (identical apart from the three combining stretches deliberately left out)

### DIFF
--- a/src/lib/skill-prompt.test.ts
+++ b/src/lib/skill-prompt.test.ts
@@ -1,4 +1,4 @@
-// cspell:ignore lette revievv -- a lookalike fixture and a label cut mid-word on purpose
+// cspell:ignore lett revievv -- a lookalike fixture and a label cut mid-word on purpose
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { displayWidthOf } from "../utils/display-width.js";
@@ -122,7 +122,7 @@ describe("promptSkillSelection", () => {
             // the tail of the second is cut: the marker and the first reason
             // are the ones a cut never reaches.
             name:
-              "[!] another entry differs from it only by lookalike letters; mi\u2026 " +
+              "[!] another entry differs from it only by lookalike letters; \u2026 " +
               `\u2014 ${lookalike}`,
             value: lookalike,
             checked: false,
@@ -148,7 +148,7 @@ describe("promptSkillSelection", () => {
     // names.
     expect(checkboxMock).toHaveBeenCalledWith(
       expect.objectContaining({
-        choices: [{ name: `pdf${"o".repeat(68)}\u2026`, value: long, checked: false }],
+        choices: [{ name: `pdf${"o".repeat(67)}\u2026`, value: long, checked: false }],
       }),
     );
   });
@@ -170,7 +170,7 @@ describe("promptSkillSelection", () => {
       expect.objectContaining({
         choices: [
           {
-            name: `[!] carries more whitespace than the row shows \u2014 pdf${" ".repeat(19)}\u2026`,
+            name: `[!] carries more whitespace than the row shows \u2014 pdf${" ".repeat(17)}\u2026`,
             value: padded,
             checked: false,
           },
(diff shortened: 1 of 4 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2891, opened 2026-09-02
fix(deepagents): do not create an empty global config.toml when nothing maps

### LINKED ISSUE #2859: Follow up PR #2828: all-tools category deferred items (empty deepagents config, helper extraction, review gap)
Follow-ups left over from PR #2828 (`fix(permissions): read the all-tools category in the command-only adapters`, issue #2825). That PR went through 10 automated review rounds; the round-10 `high` and both `mid` findings were fixed in the commits pushed afterwards, and the items below are the ones that were deliberately deferred. Treat this as a scrap list to re-verify before acting.

## 1. The post-cap commits merged without a review round

The two commits pushed after round 10 — the deepagents allowlist intersection fix (each allowlist entry intersected against `token` and `token *` instead of matching by leading token only), the narrowed `fromAllToolsCategory` ask warning, the deepagents intersection-budget message, and the low fixes (the astral-plane surrogate split in `warpCommandPatternToGlob`, the dead `foreignDenyCategories` field, test gaps) — went in after the loop hit its 10-iteration cap, so no review round looked at them. CI was green and each fix has a regression test; one fresh read of `deepagents-permissions.ts` and `shell-command-categories.ts` is worth doing.

## 2. An empty `~/.deepagents/config.toml` can still be created when nothing maps

If a canonical config produces no shell-command rules at all for deepagents-cli, the adapter still writes the config file with an empty `[shell]` section rather than skipping the write. Harmless, but it puts a file on disk that says nothing. Deferred as a `low` during the PR.

## 3. Extract the cross-adapter helpers

The three command-only adapters (deepagents-cli, Warp, Factory Droid) plus Cline now share `collectShellCommandRules` / `partitionCommandRules`, but each still carries its own copy of the surrounding warning assembly and pattern-to-glob conversion. A reviewer suggested pulling the common parts up; it was deferred to a separate PR to keep the diff of #2828 reviewable (the PR was already 16 files).

## 4. Warning-wording test symmetry

The import-side inert-entry warning is the only warning in these adapters whose text is not asserted in a test (its effect on the resulting config is). One assertion would make the set symmetric.

## 5. A mixed `all` + command-name import is silent

On import, a deepagents allowlist that mixes the literal `all` with individual command names returns `{permission: {}}` with no warning, while the non-string-element case does warn. Upstream treats such a list as disabling the option entirely, i.e. the user's config is already broken, so a one-line warning would help. Confirm the current behavior before changing it.

Related: #2825, PR #2828

dyoshikawa:
Closes #2859

## Summary

Issue #2859 lists five follow-ups deferred from PR #2828. Three were re-verified and need no code:

- **Item 1 (post-cap commits unreviewed):** the commits pushed after review round 10 (`a366d92a9`, `d6a5f1df7`, `551cef6f7`) were read fresh. The deepagents allowlist intersection matches each entry as both `token` and `token *` under one shared budget and fails closed on exhaustion, the Warp astral-plane surrogate split pushes a code point as one atom, and the dead `foreignDenyCategories` field is gone. No defect found.
- **Item 3 (extract cross-adapter helpers):** already done. `collectShellCommandRules`, `partitionCommandRules`, `collectUnenforcedAllToolsPatterns`, `createShadowingRestrictionsTest` and `warnAboutUnwrittenCommandRules` live in `shell-command-categories.ts` and are used by Warp, Factory Droid, deepagents and Cline. The remaining per-adapter code (Warp's regex-to-glob conversion, deepagents' executable-name partitioning) is tool-specific and not worth pulling up.
- **Item 5 (mixed `all` + names import is silent):** wrong at filing time. The mixed-list warning was added by `57ff0539b` (PR #2824, merged before #2828 was opened) and is asserted in `deepagents-permissions.test.ts`.

The two remaining items are fixed here.

## Changes

- **Item 2:** `DeepagentsPermissions` now overrides `shouldSkipCreationWhenPayloadEmpty()` to return `true`. When nothing maps, the `[shell]` table was already dropped, but `smolToml.stringify({})` yields a lone newline and the empty-payload creation skip in `FeatureProcessor.writeAiFiles` only fired for paths in `SHARED_USER_MANAGED_CONFIG_PATHS`, so a fresh `~/.deepagents/config.toml` holding one newline was created on a machine that had none. `config.toml` is dcode's own file: rulesync merges into it when it exists but has no reason to bring it into existence to hold nothing. An existing file is still rewritten as before, so no user content is dropped. The path is deliberately not added to `SHARED_USER_MANAGED_CONFIG_PATHS`, which also feeds gitignore derivation.
- **Item 4:** the inert-entry import test now asserts the warning text contains `auto-approve nothing`, so every warning in these adapters has its wording covered.
- `docs/reference/file-formats.md`: one clause in the deepagents paragraph states that the file is not created when nothing maps and it does not exist yet; `src/generated/docs-content.ts` regenerated.

## Test plan

- [x] New processor-level tests in `deepagents-permissions.test.ts` (pseudo-home via `setupTestDirectory({ home: true })` + `HOME_DIR`): `permission: {}` with no existing file leaves `~/.deepagents/config.toml` absent; a mapping allow rule still creates it; an existing file with a stale allowlist and an unrelated table is still rewritten (allowlist removed, table kept). The absent-file test fails without the override.
- [x] `pnpm cicheck:code`: 398 test files, 10362 tests passed; oxlint, typecheck clean.
- [x] `pnpm cicheck:content`: docs-content, gitignore, supported-tools, cspell, secretlint pass after the commit.


dyoshikawa:
@dyoshikawa Thank you!

PR #2893, opened 2026-09-02
refactor(logging): give the strip-then-quote idiom one home

### LINKED ISSUE #2842: logging: give the strip-then-quote idiom one home
`JSON.stringify(stripControlCharacters(x))` is now the way a remote-derived name reaches a log line, and it is spelled out at roughly 28 call sites across `src/` — 20 of them in `src/lib/fetch.ts` after #2841. `src/types/dir-feature-processor.ts` additionally hoists it into `quoted*` locals, so two spellings of the same idiom coexist.

The order matters and is not self-evident: the strip has to run first, because `JSON.stringify` escapes quotes and backslashes but not the C1 range or the bidi controls. That rationale currently lives in a comment at one call site.

Suggested shape: export a `quoteForLog(text: string): string` from `src/utils/control-characters.ts` that does both, carry the ordering rationale in its doc comment, and replace the call sites with it. Mechanical, and it gives the convention one home to point at when the next log line is written.

Found during the review of #2841.

dyoshikawa:
Closes #2842

## Summary

`JSON.stringify(stripControlCharacters(x))` is how a remote-derived name reaches a log line, and it was spelled out at 45 call sites across `src/`, with the reason the strip has to run first living in a comment at one of them. This PR exports `quoteForLog(text)` from `src/utils/control-characters.ts`, carries the ordering rationale in its doc comment, and routes every call site through it.

Pure refactor: no log message wording or behaviour changes. The one variant that quoted a prefixed name, `JSON.stringify(\`${SKILLS_DIR_PREFIX}${stripControlCharacters(variant)}\`)` in `fetch.ts`, now reads `quoteForLog(\`${SKILLS_DIR_PREFIX}${variant}\`)`; the prefix is a constant with no control characters, so the output is identical. The `quoted*` locals in `dir-feature-processor.ts` are kept and now call the helper.

## Changes

- `src/utils/control-characters.ts`: add `quoteForLog` with a doc comment explaining why the strip runs before the JSON quoting (`JSON.stringify` escapes quotes, backslashes and the C0 controls but passes the C1 range and the bidi overrides through intact).
- `src/utils/control-characters.test.ts`: unit tests for `quoteForLog` (printable text quoted, empty string becomes `""`, quotes and backslashes escaped, newline removed rather than escaped, C0/DEL/C1/bidi removed).
- Call sites replaced, per file:
  - `src/lib/fetch.ts`: 21 (20 direct plus the prefixed variant); the call-site rationale comment is dropped now that the helper carries it.
  - `src/types/dir-feature-processor.ts`: 14
  - `src/features/skills/skills-processor.ts`: 4
  - `src/features/skills/claudecode-skill.ts`: 3 (its `stripControlCharacters` import is now unused and removed)
  - `src/lib/github-client.ts`: 2
  - `src/utils/file.ts`: 2
- Files that still strip without quoting (error messages, summary lines) keep their `stripControlCharacters` import.

## Test plan

- [x] `grep -rn "JSON.stringify(stripControlCharacters" src` matches only the helper's own body.
- [x] `pnpm vitest run src/utils/control-characters.test.ts` passes (69 tests).
- [x] `pnpm cicheck` passes: format, oxlint, typecheck, 10377 tests across 398 files, docs/gitignore/tables sync, cspell, secretlint.

### DIFF
--- a/src/features/skills/claudecode-skill.ts
+++ b/src/features/skills/claudecode-skill.ts
@@ -9,7 +9,7 @@ import {
 import { SKILL_FILE_NAME } from "../../constants/general.js";
 import { RULESYNC_SKILLS_RELATIVE_DIR_PATH } from "../../constants/rulesync-paths.js";
 import { ValidationResult } from "../../types/ai-dir.js";
-import { stripControlCharacters } from "../../utils/control-characters.js";
+import { quoteForLog } from "../../utils/control-characters.js";
 import { formatError } from "../../utils/error.js";
 import {
   directoryExists,
@@ -159,9 +159,7 @@ async function checkNestedSkillsRoot({
     // Named rather than quoted when it is the project root, whose relative path is
     // the empty string and would otherwise be reported as a pair of quotes.
     const resolvedDescription =
-      realRelativeDirPath === ""
-        ? "the project root"
-        : JSON.stringify(stripControlCharacters(realRelativeDirPath));
+      realRelativeDirPath === "" ? "the project root" : quoteForLog(realRelativeDirPath);
     return {
       reason: `it resolves to ${resolvedDescription}, which is not a ${CLAUDECODE_SKILLS_DIR_POSIX_PATH} directory.`,
     };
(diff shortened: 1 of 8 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2895, opened 2026-09-02
feat(shared-config): warn once when a JSONC file falls back to whole-document rewrite

### LINKED ISSUE #2858: Follow up PR #2854: JSONC edit path deferred items (helper extraction, fallback logging, review gap)
Follow-ups left over from PR #2854 (`fix(shared): write JSONC configs back as edits so comments survive`, issue #2796). That PR went through 10 automated review rounds; every `mid`-or-above finding was fixed, and the items below are the ones that were deliberately deferred or rejected with a recorded reason, plus one review gap. None of them is a correctness bug in the merged code — treat this as a scrap list to re-verify before acting.

## 1. The final commit merged without a review round

`0fcf95a08` (the aggregate write-budget: `JSONC_EDIT_WRITTEN_BYTES`, `measureJsoncWrite`, `countJsoncEdits` returning `{ edits, written }`) was pushed after the review loop hit its 10-iteration cap, so it is the one commit in the PR that no review round looked at. It was mutation-tested (4 mutations, each caught) and CI was green, but a fresh read of `src/features/shared/shared-config-gateway.ts` around the budget gate is worth one pass.

## 2. Extract the JSONC edit helpers into their own module

`shared-config-gateway.ts` now carries ~350 lines of surgical-edit machinery (`skipJsoncTrivia`, duplicate-key detection, `countJsoncEdits`, `measureJsoncWrite`, the `modify`/`applyEdits` driver) alongside the codec/ownership/conflict-policy code that was there before. Splitting it into e.g. `shared-config-jsonc-edit.ts` was rejected *during* the PR only because moving 350 lines would have hidden the review diff — it is a reasonable follow-up now that the code has landed.

## 3. Unify the tolerant and strict readers

The gateway reads the same file twice with two different parsers (a tolerant read for the edit path, a strict one for the document). Reviewers noted the duplication; unifying them is a refactor with a behavioral edge (what a file that one parser accepts and the other rejects should do), so it was left out of scope.

## 4. A key removed from a shared line leaves a stray space

When a deleted key shared its line with another key (`{ "a": 1, "b": 2 }` → remove `"a"`), `modify`'s edit range leaves the separating space behind. The result is valid JSONC and the surviving keys and comments are intact; it is only cosmetic, and cleaning it up means post-processing `modify`'s output rather than trusting it.

## 5. No log line when the budget forces the whole-document rewrite

When the edit path bails out (unparsable file, duplicate key, prototype-pollution key, or the size budget), the file is rewritten whole and the user's comments are dropped with no warning. `serializeSharedConfig` takes no `Logger` today, so surfacing it means threading one through — worth doing, since silently losing comments is exactly what the PR set out to stop.

## 6. `whole` is computed unconditionally

`stringifySharedConfig` is called before the edit attempt so the fallback is always available. Measured at ~2.1% of the write cost, so it was accepted, but it can be made lazy if the write path ever gets hot.

## 7. `opencode-permissions.ts` alias renames

A naming cleanup a reviewer suggested in passing. No behavioral issue; it was out of scope for #2796.

Related: #2796, PR #2854

dyoshikawa:
Closes #2858

## Summary

Issue #2858 lists seven items deferred from PR #2854 (the JSONC edit path in `shared-config-gateway.ts`). Only item 5 needed code; the rest were re-verified against `main` and are resolved, cosmetic, or by design:

- **Item 1 (budget commit `0fcf95a08` unreviewed):** read in full. `countJsoncEdits` mirrors `applyJsoncObjectEdits` case for case and the gate checks both `JSONC_EDIT_WRITTEN_BYTES` and `edits * span > JSONC_EDIT_BUDGET_BYTES`; no defect found.
- **Item 2 (extract the edit helpers):** still holds, but it is a pure move of ~600 lines with no behavior change, so it stays a separate optional refactor rather than riding along here.
- **Item 3 (unify the tolerant and strict readers):** still holds; unifying them changes what happens to a file one parser accepts and the other rejects, which is out of scope for a warning-only change.
- **Item 4 (stray space after removing a shared-line key):** cosmetic only; removal now goes through `removeJsoncProperty`, the output is valid JSONC and every surviving key and comment is intact.
- **Item 5 (no log line on the whole-document fallback):** fixed in this PR.
- **Items 6 and 7 (`whole` computed unconditionally; opencode alias names):** by design and cosmetic respectively; no behavior change.

## Changes

- `serializeSharedConfig`, `serializeSharedConfigFile` and `applySharedConfigPatch` take an optional `logger` and `filePath`. When a JSONC file is handed to the whole-document writer, one warning per file per run (via `warnOnceWithFallback`) names the file (control characters stripped, JSON-quoted) and the reason: could not be parsed, root is not an object, a key stated twice, a prototype-pollution key, the write budget, the edit budget, or an editor exception. An empty file and the non-JSONC formats stay silent.
- `statesUneditableKeys` became `uneditableKeyOf`, returning which key (and why) blocked the edit so the warning can name it.
- The callers that write JSONC shared files pass their logger through: `copilot-mcp`, `copilot-permissions`, `amp-mcp`, `amp-permissions`, `opencode-mcp`, `opencode-permissions`, `kilo-mcp`. Every other caller is unchanged and reports through the fallback logger.
- `docs/reference/cli-commands.md` (the section "Comments in shared JSONC files are preserved", where the fallback is documented) gains one sentence saying the rewrite is reported as a warning naming the file; `src/generated/docs-content.ts` regenerated.
- Tests in `shared-config-gateway.test.ts`: the warning names the file and reason, each fallback reason is reported, control characters are stripped from the name, the in-place edit path does not warn, empty files and non-JSONC formats do not warn, warn-once semantics across repeated serializations, and `serializeSharedConfigFile` falls back to the file key as the name.

## Test plan

- [x] `pnpm cicheck` passes (10460 tests, format, lint, typecheck, docs-content, gitignore, supported-tools, cspell, secretlint)
- [x] New unit tests for the fallback warning and its silence on the edit path
- [ ] Manual: generate against an `opencode.jsonc` with a duplicate key and confirm one warning naming the file


dyoshikawa:
@dyoshikawa Thank you!

PR #2898, opened 2026-09-03
fix(permissions): skip creating empty global config for Warp and Goose

### LINKED ISSUE #2894: permissions: skip creating an empty global config for the other user-managed adapters
## Summary

Follow-up to #2859 / PR #2891. That PR makes `DeepagentsPermissions` override `shouldSkipCreationWhenPayloadEmpty()` so `~/.deepagents/config.toml` is not created when no permission rule maps to deepagents (an absent file and an empty one mean the same thing there: ask about everything).

Other adapters that merge into a global, user-managed config file through the same `writeAiFiles` path do not override the hook, so they may still create a fresh file holding nothing but a newline or an empty table when nothing maps. Candidates noted during review, not verified end-to-end:

- Warp (`src/features/permissions/warp-permissions.ts`): smol-toml merge into a global file, deletes the allow/deny keys when nothing maps.
- Goose and Takt permission adapters, if they follow the same shape.

## Task

For each adapter that writes a global user-managed config:

1. Check whether an empty payload has a meaning distinct from an absent file (as it does for Roo, where an empty list overrides a contributed default). If it does not, override `shouldSkipCreationWhenPayloadEmpty()` to return `true`, mirroring PR #2891.
2. Add the processor-level tests from PR #2891 (`src/features/permissions/deepagents-permissions.test.ts`, describe `shouldSkipCreationWhenPayloadEmpty`): file stays absent when nothing maps, a mapping rule still creates it, an existing file with a stale allowlist plus an unrelated table is rewritten with the table kept.
3. Add the matching clause to the adapter's paragraph in `docs/reference/file-formats.md` and regenerate `src/generated/docs-content.ts`.

## Notes

- The hook is only consulted when the file does not exist yet (`FeatureProcessor.writeAiFiles`), so an existing file with a stale allowlist is still rewritten. Keep that behaviour.
- Do not add the paths to `SHARED_USER_MANAGED_CONFIG_PATHS` as part of this.

dyoshikawa:
## Summary

Follow-up to #2859 / #2891. `WarpPermissions` and `GoosePermissions` now override `shouldSkipCreationWhenPayloadEmpty()` to return `true`, mirroring `DeepagentsPermissions`: when no permission rule maps and the global file does not exist yet, rulesync no longer creates a `settings.toml` holding only `[agents.profiles]` or a `permission.yaml` holding only three empty lists. For both tools an absent file and an empty payload mean the same thing (the tool's own defaults apply). An existing file is still rewritten, so stale lists are cleared and unrelated keys survive.

Takt was also checked and left unchanged: its patch always writes the derived `default_permission_mode`, so its payload is never empty and the hook would never fire.

## Changes

- `src/features/permissions/warp-permissions.ts`, `src/features/permissions/goose-permissions.ts`: add the override with a doc comment.
- Processor-level tests in both `*.test.ts` files, matching the ones added in #2891: file stays absent when nothing maps, a mapping rule still creates it, an existing file with a stale list plus an unrelated table/block is rewritten with the unrelated part kept.
- `docs/reference/file-formats.md`: matching clause in the Warp and Goose paragraphs; `src/generated/docs-content.ts` regenerated.

## Note for reviewers

A quick survey through `PermissionsProcessor` with an empty `permission` map shows other adapters still create an empty-payload file when it is absent (e.g. Amp, Antigravity CLI, Devin, Kilo, OpenCode, Zed in global mode, whose relocated global paths are not matched by `SHARED_USER_MANAGED_CONFIG_PATHS`; Hermes Agent, Kimi Code, Pi, Qwen Code, Junie, Reasonix). Each of those needs its own empty-vs-absent semantics check, so they are intentionally out of scope here.

Closes #2894

https://claude.ai/code/session_01G1MMs83Eh57Fqy3uppFBTp


dyoshikawa:
@dyoshikawa Thank you!

PR #2899, opened 2026-09-03
fix(fetch): skip Windows-foldable segments and reserved device names in every fetched path

### LINKED ISSUE #2892: fetch: check every path segment for Windows-foldable shapes and reserved device names
## Summary

Follow-up to #2848 / PR #2888. That PR makes `rulesync fetch` drop skill directory names that Windows folds or aliases (trailing dot or space, `NAME~<digits>` 8.3 short-name aliases), so a remote `deploy.` can no longer overwrite a local `deploy` under `--conflict overwrite`. Two adjacent gaps were noted during review and left out of that PR on purpose:

1. **Per-segment check for other fetched paths.** `hasWindowsFoldableSkillName` is applied to the skill directory name only. Files inside a skill (`references/notes. `, `scripts/run~1`) are written with their path segments unchecked, so a nested segment with a trailing dot/space or an 8.3 alias can still fold onto a sibling on Windows. Each segment of every written path should go through the same predicate.

2. **Windows reserved device names.** `CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9` (case-insensitive, with or without an extension, e.g. `nul.md`) are not valid file or directory names on Windows and either fail to create or resolve to the device. They should be treated as unsafe names in the same way, both for the skill directory and for nested segments.

## Where

- `src/lib/fetch.ts`: `hasWindowsFoldableSkillName`, `classifySkillPath`, the write step and the prune backstop.
- `docs/reference/cli-commands.md`: the "Skill Names That Look Alike" section should list the extra shapes.
- `src/lib/fetch.test.ts`: cases for nested foldable segments and for each reserved device name (bare and with an extension), plus a local sibling left untouched under `--conflict overwrite`.

## Notes

- Keep the check platform-independent, as the existing guards are, so a repository fetched on Linux does not produce a tree that cannot be checked out on Windows.
- The warning should keep quoting names as-is so a trailing space stays visible.

dyoshikawa:
## Summary

Follow-up to #2848 / #2888. `rulesync fetch` now applies the Windows-foldable-name check to **every segment of every fetched path**, and treats **Windows reserved device names** as unsafe as well.

Closes #2892

## What changed

- **Reserved device names.** `CON`, `PRN`, `AUX`, `NUL`, `COM1`–`COM9`, `LPT1`–`LPT9` — case-insensitive, with or without an extension (`nul.md`, `con.sh`) — are now matched by the same predicate as the trailing dot/space and `NAME~1` shapes. Names that merely begin like one (`console`, `com10`, `aux-tools`) stay ordinary.
- **Per-segment check.** `dropAmbiguousRemotePaths` (collection time, before anything is written) now inspects each segment of each collected path other than the skill directory itself. A folded segment — a nested `references/notes.`, a `SCRIPT~1/` directory, a `rules/con.md` — drops that file, records its remote directory as incomplete so the skill is **not pruned** (a local sibling the remote still ships cannot be told from a stale one), and warns quoting the segment as-is so a trailing space stays visible. `.`/`..` segments are still left to `validateRemoteRelativePath`, which refuses them outright.
- **Skill directory.** Still handled by `classifySkillPath` (whole skill dropped, existing "Windows resolves to a different directory" warning), now including reserved names. The prune backstop tests the bare skill name instead of the `skills/`-prefixed path so the anchored reserved-name pattern can match.
- The check remains platform-independent.

## Docs

- `docs/reference/cli-commands.md`: the backslash/colon subsection is now "Remote Paths Windows Reads Differently" and covers the per-segment shapes; the pruning bullet and the "Skill Names That Look Alike" paragraph list reserved device names. `src/generated/docs-content.ts` regenerated.

## Tests

- Skill directory named after each reserved device (bare, lowercase, with extension) is not offered/fetched; lookalike names (`console`, `null`, `com10`, …) are fetched.
- Nested file ending in a dot is skipped, the local sibling `notes` is untouched under `--conflict overwrite`, and the skill is not pruned.
- Nested segment ending in a space is quoted in the warning.
- Every file under a nested `SCRIPT~1/` directory is skipped; a local sibling directory stays.
- Nested files named after each reserved device are skipped; `console.md`, `com10.sh`, `data~2parser.md` are fetched.
- A `rules/con.md` outside the skills tree is skipped.

`pnpm cicheck` passes locally.

https://claude.ai/code/session_01G1MMs83Eh57Fqy3uppFBTp


dyoshikawa:
@dyoshikawa Thank you!

PR #2900, opened 2026-09-03
fix(fetch): put the whitespace reason first so a cut label keeps it

### LINKED ISSUE #2839: fetch: put the whitespace reason first so shortening cannot cut it away
`describeConfusableNames` pushes its reasons in a fixed order, and the whitespace reason added by #2837 is third of five. `formatSkillChoiceLabel` in `src/lib/skill-prompt.ts` shortens the reason column from the end, so a long earlier reason can cut the whitespace note away entirely:

```
[!] another entry differs from it only by lookalike letters; c\u2026 \u2014 <name>
```

`another entry differs from it only by lookalike letters` is 55 columns on its own, past the reason budget, so the fact that the row reaches further than it appears to is lost.

The four other reasons all say the row may be confused with *another* row; this one says the row itself occupies more columns than it shows, which nothing else on the label implies. It is the reason worth keeping when the column has to be cut, so it should be pushed first — or at least directly after the display-form reason.

dyoshikawa:
## Summary

- `describeConfusableNames` now pushes the "carries more whitespace than the row shows" reason first. `formatSkillChoiceLabel` cuts an overlong note from its tail, and the lookalike reason alone (55 columns) exceeds the note budget, so the whitespace reason that came after it was lost entirely.
- Every other reason says the row may be confused with *another* row, which a reader can still check by eye; the whitespace reason says the row itself occupies more columns than it shows, which nothing else on a cut label implies. That is the reason worth keeping.
- Adds a prompt-level test for a padded lookalike twin, reorders the "five reasons" list in `docs/reference/cli-commands.md` to match, and regenerates `src/generated/docs-content.ts`.

Closes #2839

## Test plan

- [x] `pnpm cicheck`


dyoshikawa:
@dyoshikawa Thank you!

PR #2901, opened 2026-09-03
fix(fetch): drop names that begin with a mark and note stray enclosing marks

### LINKED ISSUE #2838: fetch: enclosing marks are drawn over a name for free and are never noted
A skill name may carry an enclosing mark (`\p{Me}`) that draws over the character before it and takes no column of its own, and neither gate turns it away:

- `hasDeceptiveHiddenCharacters` refuses only invisible and control characters; an enclosing mark is neither.
- `NOTHING_DRAWN_PATTERN` in `src/lib/fetch.ts` refuses a name only when it is *entirely* marks and blanks.
- `describeConfusableNames` compares display forms, and a mark survives that comparison, so `pdf` and `pdf\u20e3` are not reported as a pair.

Measured: `pdf` and `pdf` + U+20E3 are both accepted, occupy the same three columns, and neither carries a note. A name that merely *begins* with a combining mark (`\u0301pdf`) is accepted too — the mark is drawn over whatever the terminal printed before the name, which is the checkbox glyph or a space.

There are 13 enclosing marks (U+0488, U+0489, U+1ABE, U+20DD–U+20E4, U+A670–U+A672). Restricting the fix to `\p{Me}` matters: folding `\p{Mn}` in general would break ordinary Devanagari, Arabic and Vietnamese names.

Neither case is new — #2837 only added the `FE0F`-spelled keycap to what was already accepted — but since that PR now handles U+20E3 explicitly, the rest of the family should follow.

Suggested shape: refuse a name whose first code point is a mark (a mark with nothing of its own to sit on is dishonest for the same reason an all-marks name is), and add a `describeConfusableNames` reason for an enclosing mark left over outside a legitimate keycap sequence — exporting `isKeycapSequence` from `src/utils/control-characters.ts` makes the exception reusable.

dyoshikawa:
## Summary

- `classifySkillPath` in `src/lib/fetch.ts` now drops a skill directory whose name (after stripping hidden characters) begins with a combining mark, reporting it as `hidden` and showing it without the mark. A leading mark has nothing of the name's own to sit on: it is drawn over the checkbox glyph or the quote around a name in a warning, while the name reads as the letters after it.
- New `hasEnclosingMarkOutsideKeycap` in `src/utils/control-characters.ts` reuses `isKeycapSequence`, and `describeConfusableNames` gains a sixth reason, "carries a mark drawn over the character before it", for any `\p{Me}` that is not the keycap of a UTS #51 keycap sequence. It is pushed right after the whitespace reason, since both say what the row itself is rather than what it may be confused with, and a cut label keeps the head of the note. Restricted to `\p{Me}`: folding `\p{Mn}` would mark ordinary Devanagari, Arabic and Vietnamese names.
- Tests for both gates and the reason order; `docs/reference/cli-commands.md` updated (six reasons, leading-mark drop) and `src/generated/docs-content.ts` regenerated.

Closes #2838

## Test plan

- [x] `pnpm cicheck`


dyoshikawa:
@dyoshikawa Thank you!

PR #2902, opened 2026-09-03
fix(rovodev): respect MCP config path on import

isaacgao4396:
## Summary

- Resolve Rovo Dev's configured `mcp.mcpConfigPath` during MCP import.
- Follow only paths that remain within the project or global-home import scope.
- Preserve the existing `.rovodev/mcp.json` fallback with warnings for unset or unsafe pointers.
- Add regression coverage for project-relative, global home-anchored, empty, and out-of-scope paths.

## Validation

- `pnpm run typecheck`
- `pnpm build`
- `pnpm test src/features/mcp/rovodev-mcp.test.ts` (73 passed)
- Targeted `oxfmt`, `oxlint`, and `git diff --check`
- Manual CLI import fixture confirmed that the configured active MCP file is imported instead of the inactive default file.

### DIFF
--- a/src/features/mcp/rovodev-mcp.test.ts
+++ b/src/features/mcp/rovodev-mcp.test.ts
@@ -1,4 +1,5 @@
-import { join } from "node:path";
+import { symlink } from "node:fs/promises";
+import { basename, join } from "node:path";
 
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@isaacgao4396 Thank you!

PR #2903, opened 2026-09-03
fix(permissions): feed Hermes command_allowlist from bash allows and honor the restrictions that cover them

### LINKED ISSUE #2829: Hermes command_allowlist takes allow rules from every category while deny is read from bash and webfetch only
`HermesagentPermissions.fromRulesyncPermissions` builds `command_allowlist` from the `allow` patterns of **every** canonical category, while it reads `deny` from `bash` and `webfetch` only. That asymmetry is the inverse of the one fixed for the command-only adapters in #2825 / #2828, and it fails **open** rather than closed.

## Current behavior

`src/features/permissions/hermesagent-permissions.ts` (around lines 322-336):

```ts
// `allow` patterns (all categories) feed Hermes's command allowlist, as before.
const commandAllowlist = Object.entries(permissionBlock).flatMap(([, patterns]) =>
  patternsByAction(patterns, "allow"),
);
...
const bashDeny = patternsByAction(permissionBlock.bash, "deny");
const webfetchDeny = patternsByAction(permissionBlock.webfetch, "deny");
```

Two consequences:

1. **Non-command `allow` patterns become auto-approved commands.** `{"read": {"secrets/**": "allow"}}` or `{"edit": {"src/**": "allow"}}` writes `secrets/**` / `src/**` into `command_allowlist`, where Hermes reads them as shell-command patterns. At best they are inert noise in a security-relevant list; at worst a path pattern matches a command spelling and widens what runs without a prompt. The `*` category has the same problem: `{"*": {"*": "allow"}}` auto-approves every command, which the author may only have meant as a file-tool grant.
2. **Restrictions that should withhold an allow are ignored.** An `ask` rule has no Hermes primitive and is dropped, and a `deny` outside `bash`/`webfetch` is dropped too — but the `allow` rules they cover are still written. `{"*": {"npm *": "deny"}, "bash": {"npm publish": "allow"}}` emits `npm publish` into `command_allowlist` even though the config denies it. Canonically the stricter rule wins whatever its width (rulesync collapses colliding rules as `deny > ask > allow`), so this inverts the author's intent.

## Suggested direction

Reuse the helper introduced in #2828, `src/features/permissions/shell-command-categories.ts`:

- `collectShellCommandRules` — take `allow` from `bash` only, and take the restricting (`deny` / `ask`) rules from `bash` and the all-tools `*` category.
- `createShadowedAllowTest` — withhold any `allow` that a restriction Hermes cannot express (an `ask`, or a `deny` outside `bash`/`webfetch`) covers, comparing patterns in both directions so a catch-all restriction is not silently lost.
- Warn about what was withheld, matching `warnAboutUnwrittenCommandRules`.

Keep the existing `bash` deny -> `approvals.deny` and `webfetch` deny -> `security.website_blocklist` mappings as they are; only the allowlist side and the shadowing check need to change. `docs/reference/file-formats.md` needs its Hermes paragraph updated to match, and the round-trip import path should be checked for the same asymmetry.

Related: #2825, #2828.

dyoshikawa:
## Summary

`HermesagentPermissions.fromRulesyncPermissions` built `command_allowlist` from the `allow` patterns of **every** canonical category while reading `deny` from `bash` and `webfetch` only. That failed open in two ways:

1. A non-command `allow` became an auto-approved command: `{"read": {"secrets/**": "allow"}}` wrote `secrets/**` into a list Hermes reads as shell-command patterns, and `{"*": {"*": "allow"}}` auto-approved every command.
2. Restrictions that should withhold an `allow` were ignored: `{"*": {"npm *": "deny"}, "bash": {"npm publish": "allow"}}` still emitted `npm publish`, even though the canonical order is `deny > ask > allow` whatever the width.

This PR routes the adapter through the shared helpers in `src/features/permissions/shell-command-categories.ts`, the same way the four command-only adapters already work:

- Only `bash` `allow` rules feed `command_allowlist`. An all-tools `*` allow is skipped with a warning; `read`/`edit`/… allows are never written there.
- The all-tools `*` `deny` and `ask` rules are read and **withhold** the `bash` allows they cover (compared as globs in both directions, so a catch-all `{"*": {"*": "ask"}}` is not silently lost). The `*` deny is not written into `approvals.deny`, which stays `bash`-only, and the omission is reported.
- A `bash` `ask` withholds the allows it covers too, since Hermes has no ask tier.
- Restrictions Hermes cannot express (`deny`/`ask` outside `bash`/`*`/`webfetch`, and a `webfetch` `ask`) are now reported instead of dropped silently. A `webfetch` deny keeps feeding `security.website_blocklist`.
- Import: `command_allowlist` is now authoritative for the `bash` allow rules only, so a non-command allow recorded in the private `permissions.rulesync` block survives a generate → import round-trip instead of being retracted because the allowlist (correctly) omits it.
- `docs/reference/file-formats.md` Hermes paragraphs updated to match.

The existing `bash` deny → `approvals.deny` and `webfetch` deny → `security.website_blocklist` mappings are unchanged.

Closes #2829

## Test plan

- [x] New unit tests: bash-only allowlist, `*` deny withholding, catch-all `*` ask withholding, unexpressed-restriction warnings, and a non-command allow round-trip
- [x] `pnpm cicheck` passes (401 files, 10694 tests)


dyoshikawa:
@dyoshikawa Thank you!

PR #2904, opened 2026-09-03
fix(hooks): resolve native event names with own-property lookups

### LINKED ISSUE #2757: Hook event reverse lookups resolve Object.prototype members instead of falling back
## Background

Nine hooks adapters translate a tool's native event name back to the canonical
one with the same idiom:

```ts
const eventName = X_TO_CANONICAL_EVENT_NAMES[nativeEventName] ?? nativeEventName;
```

`X_TO_CANONICAL_EVENT_NAMES` is a plain object literal, so the bracket read walks
its prototype chain. When the native name happens to be an `Object.prototype`
member, the lookup **succeeds** with an inherited value instead of falling
through to the `??` fallback, and the inherited value — an object or a function —
is then used as a property key.

## Repro

```js
const M = { PreToolUse: "preToolUse" };
for (const k of ["__proto__", "constructor", "toString"]) {
  const v = M[k] ?? k;
  const out = {};
  out[v] = ["x"];
  console.log(k, "->", JSON.stringify(Object.keys(out)[0]));
}
// __proto__   -> "[object Object]"
// constructor -> "function Object() { [native code] }"
// toString    -> "function toString() { [native code] }"
```

Every one of these sites reaches the lookup from `Object.entries()` over a
user-supplied config file, so a literal `__proto__` / `constructor` / `toString`
key in `.cursor/hooks.json`, `.qwen/settings.json`, etc. is an own enumerable key
and does reach it. The imported hook is then filed under a stringified prototype
member, which lands in the tool override block and is written back out on the
next `generate`.

## Affected sites

- `src/features/hooks/kimi-code-hooks.ts:195`
- `src/features/hooks/cursor-hooks.ts:137`
- `src/features/hooks/qwencode-hooks.ts:248`
- `src/features/hooks/copilotcli-hooks.ts:341`
- `src/features/hooks/hermesagent-hooks.ts:330`
- `src/features/hooks/copilot-hooks.ts:181`
- `src/features/hooks/kiro-hooks.ts:157`
- `src/features/hooks/reasonix-hooks.ts:125`
- `src/features/hooks/vibe-hooks.ts:131`

`kiroIdeHooksToCanonical` (`src/features/hooks/kiro-ide-hooks.ts:165`) and
`antigravity-hooks.ts:78` already guard their equivalents, and #2755 added an
`isPrototypePollutionKey` guard to `vibe-hooks.ts`.

## The existing guard is only a partial fix

`isPrototypePollutionKey` covers `__proto__`, `constructor` and `prototype`
(`src/utils/prototype-pollution.ts:8`). That is the right set for *assignment*
sinks, but it is not the set that resolves on *read*: `toString`, `valueOf`,
`hasOwnProperty`, `isPrototypeOf`, `propertyIsEnumerable` and `toLocaleString`
all resolve through the prototype too, while `prototype` does not resolve on a
plain object at all. So the guarded sites still mis-handle `toString` and
friends.

## Solution / Next Steps

1. Fix the read, not the key set — make the lookup own-property-only at every
   site, e.g.

   ```ts
   const eventName = Object.hasOwn(X_TO_CANONICAL_EVENT_NAMES, nativeEventName)
     ? X_TO_CANONICAL_EVENT_NAMES[nativeEventName]
     : nativeEventName;
   ```

   which makes the `?? nativeEventName` fallback behave as it reads.
   `goose-hooks.ts:137` already uses `Object.hasOwn` this way, so the idiom
   exists in the codebase.
2. Consider a small shared helper (e.g. `lookupOwn(map, key, fallback)` in
   `src/utils/`) so the next adapter cannot reintroduce the bug, and apply it to
   the nine sites above.
3. Keep the `isPrototypePollutionKey` guards where they also protect an
   assignment sink, but stop relying on them to make the *read* safe.
4. Add a regression test per adapter with a `toString` event name, since that is
   the case the current guards miss.

## Severity

Low-to-moderate. It is not prototype pollution — nothing is assigned to
`Object.prototype` — but it silently corrupts an imported hook's event name, and
hook commands are executed by the coding agent, so an event landing in the wrong
place is worth closing off.

dyoshikawa:
## Summary

Hook adapters translated a native event name back to the canonical one with `X_TO_CANONICAL_EVENT_NAMES[name] ?? name`. The map is a plain object, so the bracket read walks its prototype chain: a config file whose event key is an `Object.prototype` member (`toString`, `valueOf`, `constructor`, `__proto__`, ...) resolved to an inherited function or object instead of falling through to the `??` fallback, and the imported hook was then filed under a stringified prototype member such as `function toString() { [native code] }`. The existing `isPrototypePollutionKey` guards only cover the assignment-sink set (`__proto__`, `constructor`, `prototype`), so `toString` and friends still slipped through.

- Add `lookupOwn({ record, key })` in `src/utils/own-lookup.ts`, an own-property-only read that keeps the `?? fallback` honest.
- Route every event-name reverse lookup through it: the shared `toolHooksToCanonical` converter (Claude Code, Goose, Devin, Augment Code, Antigravity, Codex CLI, Factory Droid, ...) plus the per-adapter sites in Copilot, Copilot CLI, Cursor, Deep Agents (v2 record and legacy `events` array), Hermes Agent, Kimi Code, Kiro, Kiro IDE, Qwen Code, Reasonix, and Vibe.
- Keep the `isPrototypePollutionKey` guards where they also protect an assignment sink, and reword the comments so they no longer claim to make the read safe.
- Add a regression test per adapter with a `toString` event name, and unit tests for the helper.

Closes #2757

### DIFF
--- a/src/features/hooks/antigravity-hooks.ts
+++ b/src/features/hooks/antigravity-hooks.ts
@@ -14,6 +14,7 @@ import {
 import { formatError } from "../../utils/error.js";
 import { readFileContentOrNull } from "../../utils/file.js";
 import type { Logger } from "../../utils/logger.js";
+import { lookupOwn } from "../../utils/own-lookup.js";
 import { isPrototypePollutionKey } from "../../utils/prototype-pollution.js";
 import type { RulesyncHooks } from "./rulesync-hooks.js";
 import type { ToolHooksConverterConfig } from "./tool-hooks-converter.js";
@@ -75,7 +76,7 @@ function flattenAntigravityHooks(parsed: unknown): Record<string, unknown> {
     if (isPrototypePollutionKey(event) || !Array.isArray(entries)) {
       return;
     }
-    const existing = Object.hasOwn(flat, event) ? flat[event] : undefined;
+    const existing = lookupOwn({ record: flat, key: event });
     flat[event] = existing ? [...existing, ...entries] : [...entries];
   };
   for (const [key, value] of Object.entries(parsed)) {
--- a/src/features/hooks/copilot-hooks.test.ts
+++ b/src/features/hooks/copilot-hooks.test.ts
@@ -918,6 +918,35 @@ describe("CopilotHooks", () => {
       expect(entry.command).toBe("run.sh");
       expect(entry.customKey).toBeUndefined();
     });
+
+    it("should carry an event named toString through as a plain string key (#2757)", () => {
+      const copilotHooks = new CopilotHooks({
+        outputRoot: testDir,
+        relativeDirPath: join(".github", "hooks"),
+        relativeFilePath: "copilot-hooks.json",
+        // JSON.parse yields an own enumerable `toString` key, unlike an object
+        // literal whose `toString` the lookup map would inherit from
+        // Object.prototype.
+        fileContent: JSON.stringify({
+          version: 1,
+          hooks: {
+            sessionStart: [{ type: "command", bash: "echo start" }],
+            toString: [{ type: "command", bash: "echo crafted" }],
+          },
+        }),
+        validate: false,
+      });
+
+      const json = copilotHooks.toRulesyncHooks().getJson();
+      expect(json.hooks.sessionStart?.[0]?.command).toBe("echo start");
+      expect(Object.keys(json.hooks)).toEqual(["sessionStart"]);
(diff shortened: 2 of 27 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2830, opened 2026-08-30
deepagents-cli writes an allow that a restriction covers, while the other command-only adapters withhold it


dyoshikawa:
Two reviews of #2828 flagged the same inconsistency: for a restriction the tool cannot express, three of the four command-only permissions adapters now **withhold** the `allow` rules it covers, while deepagents-cli **writes the allow anyway and warns**.

## Current behavior

`src/features/permissions/deepagents-permissions.ts` reduces every allow pattern to its executable name, then detects collisions in `warnAboutUnwrittenBashRules` *after* the list is built:

```jsonc
{ "permission": { "bash": { "npm *": "allow", "npm publish": "ask" } } }
```

emits `allow_list = ["npm"]` plus a warning ("Narrow or drop the allow rule that covers them"). The same shape reaches Warp, Factory Droid and Cline as an empty allowlist, because the restriction withholds the allow it covers. Since #2828, all-tools `*` deny and ask rules are read for deepagents too, so a `*` deny beside a `bash` allow lands in the same warn-and-keep path.

dcode has no denylist at all — a command it does not auto-approve is asked about, not blocked — so withholding the allow is the only way to honor a restriction there. Warning and keeping the allow auto-approves the very command the config restricts, and the warning competes with every other warning a generation emits.

## Why it was not changed in #2828

The warn-and-keep policy predates that PR, is documented in `docs/reference/file-formats.md` (the deepagents paragraph), and is fixed by several tests. Changing it alters the output of `bash`-only configs that have nothing to do with the all-tools category, which is outside the scope of #2825. Filing it separately so the decision is made on its own merits.

## Suggested direction

- Withhold an allow whose reduced executable collides with an `ask` or `deny` rule, instead of writing it, and reword the existing warnings to say the allow was withheld and why.
- Decide what `{"bash": {"*": "allow", "npm publish": "ask"}}` should emit: today the `all` sentinel yields only to `deny`, so an `ask` still turns dcode's dangerous-pattern check off. Fail-closed says `ask` should suppress `all` too.
- Update the deepagents paragraph in `docs/reference/file-formats.md` and the collision tests in `src/features/permissions/deepagents-permissions.test.ts`.

Related: #2825, #2828.

dyoshikawa:
This is already fixed on `main`, as a side effect of a different PR that never cross-referenced this issue.

[...]

PR #2905, opened 2026-09-03
fix(permissions): honor star category for command adapters and tighten Hermes allowlist

### LINKED ISSUE #2829: Hermes command_allowlist takes allow rules from every category while deny is read from bash and webfetch only

### LINKED ISSUE #2831: permissions: command-capable adapters other than the four command-only ones ignore the all-tools `*` category
PR #2828 documents the all-tools `*` category as canonical — a rule written under it is meant for every tool, shell commands included — and wires it into the four command-only adapters (deepagents-cli, Warp, Factory Droid, Cline).

Every other command-capable permissions adapter still reads only the `bash` category, so a `deny` written under `*` does not restrict the commands they auto-approve.

Reproduced with `{"*": {"rm *": "deny"}, "bash": {"rm *": "allow", "git *": "allow"}}`:

| adapter | generated | warned? |
| --- | --- | --- |
| Zed | `always_allow: [{pattern: "rm *"}, …]` | yes ("dropping the `*` category rule") |
| Kiro | `shell.allowedCommands: ["rm *", "git *"]` | yes ("do not support category: *") |
| Roo Code | `roo-cline.allowedCommands: ["rm *", "git *"]` | **no** |

So a config whose author denied `rm *` for every tool still ships an auto-approve entry for it, and for Roo Code that happens silently.

Files: `src/features/permissions/{zed,roo,kiro,cursor,codexcli,goose,devin,takt,vibe,grokcli,junie,kilo,opencode,rovodev,hermesagent,antigravity-cli,antigravity-ide}-permissions.ts`.

Suggested scope:

1. Make every adapter that skips a category emit the "skipped" warning — Roo Code's silence is the worst case.
2. Prefer routing the command-capable adapters through `collectShellCommandRules` so the `*` category is read the same way everywhere.

Found during the security review of #2828; deliberately left out of that PR as pre-existing and out of its scope.

Zsanz3:
Fixes #2831

## What this does

Command-capable adapters now honor the all-tools `*` category when building their shell command lists, instead of silently dropping it:

- New helpers alongside `collectShellCommandRules`: `honorAllToolsOnBash`, `bashRulesHonoringAllTools`, `resolveShellCommandLists`.
- Command adapters (Cursor, Codex CLI, Roo, Zed, OpenCode, Kiro, and the rest) route through those helpers, so a `*` deny/ask correctly shadows overlapping `bash` allows.
- Roo Code now emits a skip warning for categories it cannot express, rather than dropping them without a trace.

Out of scope: #2830.

## Note on #2829

This branch originally also carried a Hermes allowlist fix for #2829. While it was in review, #2903 landed upstream with a more complete implementation (including import round-trip handling for withheld `bash` allows), so I rebased onto `main` and took the upstream version for the four overlapping files (`hermesagent-permissions.ts` and its test, `docs/reference/file-formats.md`, `src/generated/docs-content.ts`). Only the #2831 work remains here, and this PR no longer claims to close #2829.

## Verification

The full local check suite passes: format check, lint, typecheck, 10728 tests across 402 files, and the docs-content / gitignore / supported-tools / cspell / secretlint content checks.

Branch is now 1 ahead / 0 behind `main`; ~+321/-18 across 22 files.

### DIFF
--- a/src/features/permissions/antigravity-cli-permissions.ts
+++ b/src/features/permissions/antigravity-cli-permissions.ts
@@ -18,6 +18,7 @@ import { formatError } from "../../utils/error.js";
 import { readFileContentOrNull } from "../../utils/file.js";
 import { fallbackLogger, type Logger } from "../../utils/logger.js";
 import { RulesyncPermissions } from "./rulesync-permissions.js";
+import { honorAllToolsOnBash } from "./shell-command-categories.js";
 import {
   ToolPermissions,
   type ToolPermissionsForDeletionParams,
@@ -393,7 +394,7 @@ function convertRulesyncToAntigravityCliPermissions(config: PermissionsConfig): 
   const ask: string[] = [];
   const deny: string[] = [];
 
-  for (const [category, rules] of Object.entries(config.permission)) {
+  for (const [category, rules] of Object.entries(honorAllToolsOnBash(config.permission))) {
     const cliToolName = toAntigravityCliToolName(category);
     for (const [pattern, action] of Object.entries(rules)) {
       const entry = buildPermissionEntry(cliToolName, pattern);
--- a/src/features/permissions/antigravity-ide-permissions.ts
+++ b/src/features/permissions/antigravity-ide-permissions.ts
@@ -11,6 +11,7 @@ import type { PermissionAction, PermissionsConfig } from "../../types/permission
 import { formatError } from "../../utils/error.js";
 import { readFileContentOrNull } from "../../utils/file.js";
 import { RulesyncPermissions } from "./rulesync-permissions.js";
+import { honorAllToolsOnBash } from "./shell-command-categories.js";
 import {
   ToolPermissions,
   type ToolPermissionsForDeletionParams,
@@ -253,7 +254,7 @@ function convertRulesyncToAntigravityIdePermissions(config: PermissionsConfig): 
   const ask: string[] = [];
   const deny: string[] = [];
 
-  for (const [category, rules] of Object.entries(config.permission)) {
+  for (const [category, rules] of Object.entries(honorAllToolsOnBash(config.permission))) {
     const action = toIdeAction(category);
     for (const [pattern, permissionAction] of Object.entries(rules)) {
       const entry = buildPermissionEntry(action, pattern);
--- a/src/features/permissions/codexcli-permissions.test.ts
+++ b/src/features/permissions/codexcli-permissions.test.ts
@@ -1467,6 +1467,25 @@ command = "node"
     expect(content).toContain('decision = "forbidden"');
   });
 
+  it("should not auto-approve a bash allow that an all-tools deny covers", () => {
+    const rulesFile = createCodexcliBashRulesFile({
+      outputRoot: testDir,
+      config: {
+        permission: {
+          "*": { "rm *": "deny" },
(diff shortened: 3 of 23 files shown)


Zsanz3:
Rebased onto `main` to clear the conflicts — thanks for your patience here.

While doing that I noticed #2903 had already landed a more thorough Hermes allowlist fix for #2829, so I took the upstream version of the four overlapping files rather than fighting it, and dropped my narrower implementation. This PR is now scoped to #2831 only (command adapters honoring the all-tools `*` category, plus Roo skip warnings), and I have updated the description accordingly. The full local check suite passes. Happy to split or reshape this further if that would be easier to review.

dyoshikawa:
@Zsanz3 Thank you!

PR #2907, opened 2026-09-03
fix(commands): resolve case-insensitive collisions across command import sources

### LINKED ISSUE #2741: Commands import de-duplicates secondary sources with an exact-match Set (same class of bug as #2733)
## Summary

`CommandsProcessor.loadToolFiles()` de-duplicates commands coming from a secondary source (OpenCode's inline `command` block, or a shared root matched by basename) with an exact-match `Set<string>`:

```ts
const seen = new Set(toolCommands.flatMap((command) => keysOf(command)));
```

`src/features/commands/commands-processor.ts:911`

## Why it matters

This is the same class of bug as #2733, which was fixed for skills and subagents in PR #2738. Two spellings that differ only in case pass the check as distinct commands, but they are written back into one `.rulesync/commands/` tree, where `commit.md` and `Commit.md` are a single file on macOS and Windows. The copy written last wins, so the precedence the sources were scanned in is inverted rather than honored — and no warning is emitted, because the exact comparison never saw a duplicate.

## Suggested direction

Reuse `ClaimedIdentities` from `src/types/feature-processor.ts` (added in PR #2738), which folds case and Unicode normalization and reports the source that claimed a name, so the warning can tell a cross-source collision from a same-source one. The `keysOf()` basename-matching logic is specific to commands and should be preserved: each candidate key would be claimed through the shared helper rather than added to a plain `Set`.

Note the trade-off documented for the skills fix: the rule applies on every platform, so on a case-sensitive filesystem the lower-precedence command is dropped with a warning rather than imported under its own name.

This was split out of the review of #2733 / PR #2738.

dyoshikawa:
## Summary

`CommandsProcessor.loadToolFiles()` de-duplicated commands from a secondary import source (OpenCode's inline `command` block, or a shared root matched by basename) with an exact-match `Set<string>`, so two spellings that differ only in case (e.g. `commit.md` / `Commit.md`) both passed as distinct commands. Written back into the single `.rulesync/commands/` tree, those spellings collide on case-insensitive filesystems (macOS, Windows), so the copy written last silently overwrote the other instead of the source precedence being honored, with no warning.

This is the same class of bug as #2733, fixed for skills and subagents in PR #2738.

## Fix

Reuse `ClaimedIdentities` from `src/types/feature-processor.ts` for the secondary-source merge, exactly as suggested in the issue:

- The primary tool root's own commands claim their keys first, so they keep precedence.
- Each secondary-source candidate key (including the basename-matched flattened twin, where applicable) is claimed through the shared case-folding helper instead of a plain `Set`.
- A collision now distinguishes a same-source duplicate ("keeping the one already loaded") from a cross-source, case-insensitive collision (names which spelling won and which source it came from).
- The existing basename-matching logic (`matchAdditionalImportsByBasename`) is preserved as-is.

## Tests

Added three regression tests to `commands-processor.test.ts`:
- A shared-root copy that differs only in case from the primary tool's own command is dropped, and the warning names which spelling won.
- A shared-root copy whose basename differs only in case from a nested command is dropped via the basename-match path.
- OpenCode's inline `command` block reports a same-source, case-only collision (`Commit` / `commit`) distinctly from a cross-source one.

## Test plan

- [x] `pnpm cicheck` (format, lint, typecheck, full test suite, docs/gitignore/supported-tools drift checks, cspell, secretlint)

Closes #2741

### DIFF
--- a/src/features/commands/commands-processor.test.ts
+++ b/src/features/commands/commands-processor.test.ts
@@ -157,6 +157,7 @@ vi.mocked(OpenCodeCommand).forDeletion = vi.fn().mockImplementation((params) => 
   isDeletable: () => true,
   getRelativeFilePath: () => params.relativeFilePath,
 }));
+vi.mocked(OpenCodeCommand).loadAdditionalImportFiles = vi.fn().mockResolvedValue([]);
 
 // Set up static methods after mocking
 vi.mocked(RooCommand).fromFile = vi.fn();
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2908, opened 2026-09-03
fix(watch): sanitize control characters in trigger-path log output

### LINKED ISSUE #2753: Sanitize control characters at a central output chokepoint instead of per call site
Raised as a `low` finding during the review of #2752 and judged out of scope
there, so it is recorded here.

## Problem

Rulesync sanitizes untrusted text with `stripControlCharacters()`
(`src/utils/control-characters.ts`) at each individual call site. Over the
course of #2752 that pattern was applied one message at a time — the curated
collision warning, the cross-root merge warning, the input-root shadowing
warning, and finally the `inspectInputRoots` diagnostics — and each round of
review surfaced the next unguarded message.

The class is not closed. Two known gaps remain:

1. **Watch change detection.** `src/cli/commands/generate.ts` logs
   `Change detected: ${formatTriggerPaths(...)}` with filenames taken straight
   from the watched tree. `src/lib/watch.ts` does not call
   `stripControlCharacters`. A repository containing a file whose name embeds an
   escape sequence re-triggers this line on every change during a watch session.

2. **CLI error sinks.** `src/cli/wrap-command.ts` and `src/cli/index.ts` print
   error messages via `logger.error` and `console.error(formatError(error))`.
   `formatError` (`src/utils/error.ts`) does not sanitize, so any error message
   built anywhere in the codebase reaches the terminal unfiltered.

Both are pre-existing on `main`, not regressions from #2724 / #2752.

## Proposal

Close the class at a chokepoint instead of continuing per-message fixes.
Candidates, in rough order of coverage:

- Sanitize in the `Logger` implementations' output stage
  (`src/utils/logger.ts`), so every diagnostic is covered regardless of where
  the string was built.
- Or sanitize at the two CLI error sinks (`wrap-command.ts`, `index.ts`) plus
  the watch line, which covers the error class but not ordinary log lines.

Points to settle before implementing:

- Whether any output intentionally carries control characters — colour codes
  emitted by the logger itself, or JSON-mode output where the payload is
  serialized rather than printed as prose. A blanket strip must not corrupt
  those.
- Whether sanitizing centrally should replace the existing per-call-site
  guards or sit underneath them as defense in depth. Keeping both is probably
  right, since some messages are also written to files rather than the terminal.
- Whether `formatError` is the better seam than the logger, given that it is
  the shared formatter for every error path.

## Threat model

Log-line forgery only: an escape sequence such as erase-line lets untrusted
text overwrite the preceding output, so a malicious repository or config file
can make a Rulesync run appear to report something it did not. It does not
affect what Rulesync generates or which files it touches.

Related: #2724, #2752

dyoshikawa:
## Summary

`formatTriggerPaths()` renders filesystem watch-event paths directly into the `Change detected: ...` log line without sanitizing them. A file whose name (or a path segment above it) contains control characters — ANSI escapes, bidirectional overrides, etc. — created in the watched tree would let those characters reach the terminal verbatim, letting a malicious repo forge or manipulate what the log line displays.

This fixes the remaining gap from #2753. The other half of that issue's claim — that `formatError()` in `src/utils/error.ts` doesn't sanitize — is already resolved on `main` by prior commits `8b76df953` and `7c39a9520`; both the `ZodError` and generic-`Error` code paths there already run through `stripControlCharacters`/`stripControlCharactersKeepingLineFeeds`. This PR narrows scope to just the still-real `formatTriggerPaths` gap.

## Changes

- `src/lib/watch.ts`: `formatTriggerPaths()` now strips control characters from every displayed path (both the relative and the outside-every-root/absolute fallback) before joining, consistent with how the codebase already sanitizes untrusted strings elsewhere (`WarningCollection`, `formatError`, `commands-processor.ts`).
- `src/lib/watch.test.ts`: two new tests covering control-character stripping for both the relative-path and absolute-path display branches.

## Verification

Full local check suite passes: format check, lint, typecheck, 10752 tests across 402 files, and the docs-content / gitignore / supported-tools / cspell / secretlint content checks.

Closes #2753

### DIFF
--- a/src/lib/watch.test.ts
+++ b/src/lib/watch.test.ts
@@ -379,6 +379,23 @@ describe("formatTriggerPaths", () => {
       }),
     ).toBe(`${join(".rulesync", "rules", "a.md")}, ${join(".rulesync", "rules", "b.md")}`);
   });
+
+  it("strips control characters from a relative trigger path", () => {
+    expect(
+      formatTriggerPaths({
+        triggers: [join(baseDir, ".rulesync", "rules", "a\u001b[31m.md")],
+        inputRoots: [baseDir],
+      }),
+    ).toBe(join(".rulesync", "rules", "a[31m.md"));
+  });
+
+  it("strips control characters from a trigger displayed absolute", () => {
+    const outside = join("/", "sibling", "b\u001b[31m.md");
+
+    expect(formatTriggerPaths({ triggers: [outside], inputRoots: [baseDir] })).toBe(
+      join("/", "sibling", "b[31m.md"),
+    );
+  });
 });
 
 describe("watchTargets", () => {
--- a/src/lib/watch.ts
+++ b/src/lib/watch.ts
@@ -2,6 +2,7 @@ import { existsSync, type FSWatcher, watch as fsWatch, statSync } from "node:fs"
 import { dirname, isAbsolute, join, relative, sep } from "node:path";
 
 import { RULESYNC_LOCAL_CONFIG_RELATIVE_FILE_PATH } from "../constants/rulesync-paths.js";
+import { stripControlCharacters } from "../utils/control-characters.js";
 
 /**
  * Trailing debounce window applied to file-system events before a regeneration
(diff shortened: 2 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2909, opened 2026-09-03
fix(rules): warn when generate deactivates Junie's import-only rule roots

### LINKED ISSUE #2744: Junie: generate silently deactivates hand-authored .junie/rules/*.md with no warning
## Summary

`rulesync generate --targets junie` writes `.junie/AGENTS.md`. From that moment Junie reads that file **exclusively** — the multi-file branch (project-root `AGENTS.md` combined with `.junie/playbook.md` and every `.junie/rules/*.md`) becomes unreachable. A repository that already maintains hand-authored `.junie/rules/*.md` therefore has those files silently deactivated by the first `generate`, with no warning and no visible change: the files stay on disk, they are not gitignored, and nothing is deleted.

The import side already handles this. #2732 / PR #2742 added `importOnlyRoots` with `onlyWhenRootAbsent`, so a later `rulesync import --targets junie` skips those paths and warns:

> Not importing .junie/rules/style.md for junie: .junie/AGENTS.md exists, and the tool reads that file exclusively. Delete them once you have checked that content is in the root file, or move it into .rulesync/rules if it is not.

That warning lives in `RulesProcessor.loadToolFiles()` (`src/features/rules/rules-processor.ts`), and the import-only branch returns early when `forDeletion` is set, so the `generate` path never reaches it. The user who adopts rulesync and runs only `generate` — never `import` — is exactly the user whose existing files are being deactivated, and they are the one who never sees the message.

## Reproduction

1. In a repo with a hand-authored `.junie/rules/style.md` and no `.junie/AGENTS.md`, add `.rulesync/rules/*.md`.
2. Run `rulesync generate --targets junie`.
3. `.junie/AGENTS.md` is created. `.junie/rules/style.md` is untouched and no warning is emitted, but Junie no longer reads it.

## Proposed follow-up

When generating a root file for a target that declares `importOnlyRoots` gated on the root being absent, check whether any of those paths currently exist *before* writing the root file, and warn once that the tool will stop reading them — pointing at `rulesync import --targets junie` (run before the first generate) as the way to carry the content over.

The same shape applies to any future target with the "root file wins exclusively" resolution order, so the check belongs next to the existing import-side gate rather than in `junie-rule.ts`.

## Notes

- Pre-existing behavior, not a regression from PR #2742; it is filed now because that PR is what first made these paths meaningful to rulesync.
- Surfaced as a `low` finding during the review of PR #2742 and deferred as out of scope for #2732 (which covers import only).

Related: #2732, #2728

dyoshikawa:
## Summary

- `rulesync generate --targets junie` writes `.junie/AGENTS.md`. Junie reads that file exclusively once it exists, so the multi-file branch (`.junie/rules/*.md`, `.junie/playbook.md`) becomes unreachable with no warning and no visible change.
- The import side already handles this: `RulesProcessor.loadToolFiles()` checks `importOnlyRoots` with `onlyWhenRootAbsent` and warns when those paths exist alongside a root file. `generate` calls `convertRulesyncFilesToToolFiles()`, which never reached that check, so a repo that only ever runs `generate` never saw the warning.
- Added `warnForDeactivatedImportOnlyRoots()`, reusing the same `importOnlyRoots` declaration used on import, and call it from `convertRulesyncFilesToToolFiles()` whenever a root rule is about to be written. It warns once per generate run, naming the paths that will stop being read and pointing at `rulesync import --targets <tool>` as the way to carry the content over first.

## Testing

- `pnpm cicheck`
- Added two tests: one confirming the warning fires and names the deactivated `.junie/rules/*.md` file, and one confirming no warning fires when no import-only paths exist on disk.

Closes #2744

### DIFF
--- a/src/features/rules/rules-processor.test.ts
+++ b/src/features/rules/rules-processor.test.ts
@@ -97,6 +97,55 @@ describe("RulesProcessor", () => {
       expect(result[1]).toBeInstanceOf(CopilotRule);
     });
 
+    it("should warn that generating Junie's root file deactivates existing .junie/rules/*.md", async () => {
+      // Junie reads `.junie/AGENTS.md` exclusively once it exists, so the
+      // multi-file branch (`.junie/rules/*.md`) becomes unreachable the
+      // moment `generate` writes the root file. Unlike `loadToolFiles`
+      // (import), `convertRulesyncFilesToToolFiles` never scanned for this
+      // before, so a repo that only ever runs `generate` saw no warning.
+      await writeFileContent(join(testDir, ".junie", "rules", "style.md"), "# Style");
+
+      const processor = new RulesProcessor({ logger, outputRoot: testDir, toolTarget: "junie" });
+      const rulesyncRules = [
+        new RulesyncRule({
+          outputRoot: testDir,
+          relativeDirPath: RULESYNC_RULES_RELATIVE_DIR_PATH,
+          relativeFilePath: "root.md",
+          frontmatter: { targets: ["*"], root: true },
+          body: "Shared team instructions",
+        }),
+      ];
+
+      await processor.convertRulesyncFilesToToolFiles(rulesyncRules);
+
+      const warning = logger.warn.mock.calls
+        .map(([message]) => String(message))
+        .find((message) => message.includes("will no longer be read"));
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2910, opened 2026-09-03
fix(rules): warn on import when a fold target still has folded non-root rules

### LINKED ISSUE #2743: Fold-collision targets duplicate their folded rules on every generate/import cycle
## Summary

For a target whose non-root rules are folded into its root file (`collisionPolicy: "fold"` — `codexcli`, `junie`, and the other fold targets), the cycle `generate` → `import` → `generate` duplicates the folded content once per cycle. Nothing warns.

## Reproduction (on `main`, `codexcli`)

`.rulesync/rules/overview.md` (`root: true`, body `root body`) and `.rulesync/rules/style.md` (`root: false`, body `style body`), then:

```bash
for i in 1 2 3; do
  rulesync generate --targets codexcli --features rules
  rulesync import  --targets codexcli --features rules
  grep -c 'style body' AGENTS.md
done
```

```
cycle 1: 1
cycle 2: 2
cycle 3: 3
```

`.rulesync/rules/` keeps `overview.md` and `style.md` throughout.

## Cause

`generate` folds `style.md` into the root file, so `AGENTS.md` contains both bodies. `import` then reads that root file back into `.rulesync/rules/overview.md` — now carrying the folded copy of `style body` — while `.rulesync/rules/style.md` is untouched, because import never deletes rulesync sources. The next `generate` folds `style.md` in again, on top of the copy that is already inside the imported root rule.

The same shape applies to any target that concatenates several rulesync rules into one output file: importing that file back cannot tell rulesync's own fold from hand-authored content.

## Why this is filed separately

It surfaced while reviewing #2742 (Junie's `.junie/rules` / `.junie/playbook.md` import-only roots), but it is not specific to that change — it reproduces on `main` with `codexcli`, which #2742 does not touch. #2742 gates Junie's read-only roots on `.junie/AGENTS.md` being absent, which removes Junie's own contribution to the loop; the general case is untouched.

## Possible directions

- On import, when the tool's root file was produced by a fold, drop the rulesync non-root rules whose bodies are already contained in it — matching content rather than re-adding it.
- Or make the fold emit a machine-readable marker per folded section, so import can split the root file back into the rules it came from (round-trip rather than re-read).
- Or, at minimum, warn on import of a fold target that still has non-root rules in `.rulesync/rules/`, so the duplication is visible instead of silent.

dyoshikawa:
## Summary

- `collisionPolicy: "fold"` targets (`codexcli`, `junie`, and others) concatenate every targeted non-root rule into one root output file on `generate`. Importing that root file back re-reads the already-folded content as a single new rulesync rule, while the original non-root rules stay in `.rulesync/rules/` untouched — the next `generate` folds both together, duplicating the content once per generate/import cycle with nothing to indicate why.
- This is the minimal "at minimum, warn on import of a fold target that still has non-root rules in `.rulesync/rules/`" option the issue records as the safe minimum, rather than attempting to detect or split out the specific duplicated content (which the root file has no marker to recover).
- Added `RulesProcessor.warnForFoldImportDuplicationRisk()`, called from `loadToolFiles()` only on the actual import path (not the `forDeletion`/orphan-sweep path). It checks whether any non-root rulesync rules already target the fold tool before the import runs, and if so warns that generating again will duplicate them.

## Testing

- `pnpm cicheck`
- Added three tests: the warning fires and names the duplicated rule for a `codexcli` fold-target import, it stays silent when no non-root rules exist yet, and it stays silent for a non-fold target.

Closes #2743

### DIFF
--- a/src/cli/commands/import.test.ts
+++ b/src/cli/commands/import.test.ts
@@ -131,6 +131,7 @@ describe("importCommand", () => {
         loadToolFiles: vi.fn().mockResolvedValue([{ file: "rule1" }]),
         convertToolFilesToRulesyncFiles: vi.fn().mockResolvedValue([{ rule: "converted" }]),
         writeAiFiles: vi.fn().mockResolvedValue({ count: 1, paths: [] }),
+        warnForFoldImportDuplicationRisk: vi.fn().mockResolvedValue(undefined),
       };
       vi.mocked(RulesProcessor).mockImplementation(function () {
         return mockRulesProcessor as any;
@@ -326,6 +327,7 @@ describe("importCommand", () => {
         loadToolFiles: vi.fn().mockResolvedValue([{ file: "rule1" }]),
         convertToolFilesToRulesyncFiles: vi.fn().mockResolvedValue([{ rule: "converted" }]),
         writeAiFiles: vi.fn().mockResolvedValue({ count: 2, paths: [] }),
+        warnForFoldImportDuplicationRisk: vi.fn().mockResolvedValue(undefined),
       };
       vi.mocked(RulesProcessor).mockImplementation(function () {
         return mockRulesProcessor as any;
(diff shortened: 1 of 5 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2749, opened 2026-08-25
Factory Droid: the DESIGN.md design-guidelines channel is never emitted


dyoshikawa:
`#2725` covered two gaps; only gap 1 (the four packaging frontmatter fields) was implemented in #2748, and the `Closes` line auto-closed the issue. Recording gap 2 separately so it is not lost.

[Factory Droid's AGENTS.md docs](https://docs.factory.ai/cli/configuration/agents-md) list `DESIGN.md` / `Design.md` / `design.md` as an instruction surface distinct from `AGENTS.md`: *"Always-on design-system, UX, visual, and interaction guidance... Loaded separately as design guidelines"*, versus `AGENTS.md` which is *"Loaded as coding guidelines"*. It shipped in v0.96.0 (2026-04-07).

rulesync never emits it: `FactorydroidRule.getSettablePaths` (`src/features/rules/factorydroid-rule.ts`) writes only root `AGENTS.md` (global `.factory/AGENTS.md`) plus `.factory/rules/*.md`, and `grep -rn "DESIGN.md" src/ --include=*.ts` finds nothing.

The content still reaches the model through `AGENTS.md`, so the loss is channel classification rather than content. Supporting it needs a way to mark a rule as design guidance plus an `extraFixedFile`-style second root output, which is a design decision rather than a mechanical fix - hence a separate issue.

PR #2748, opened 2026-08-24
feat(skills): keep Factory Droid skill packaging frontmatter

dyoshikawa:
## Summary

Follows up the skills half of #2725: the four Factory Droid packaging frontmatter fields (`license`, `compatibility`, `metadata`, `version`) documented at <https://docs.factory.ai/cli/configuration/skills> were unmodeled and silently dropped in both conversion directions, because `toRulesyncSkill` and `fromRulesyncSkill` built explicit object literals that the `looseObject` passthrough could not rescue.

## Changes

- `FactorydroidSkillFrontmatterSchema` and the `factorydroid` block of the rulesync skill frontmatter now declare `license`, `compatibility`, `metadata` and `version`.
- Both conversion directions adopt the rest-spread pattern already used by `KiroSkill`, so every key other than `name`/`description` survives the round trip - including keys rulesync does not model yet. On the way back the section is spread first and then overridden by the resolved `disable-model-invocation` / `user-invocable` values, so root-level defaults keep winning; both resolvers already prefer a defined section value, so the override never discards one.
- `version` accepts a number as well as a string: an unquoted YAML `version: 1.0` parses as a number, and rejecting it would fail the import of a skill file Droid itself accepts.
- `docs/reference/file-formats.md` documents the four fields (and the embedded docs are regenerated).

## Not included

Gap 2 of the issue - emitting the `DESIGN.md` design-guidelines channel - is deliberately left out. The issue itself marks it optional and lower priority: the content still reaches the model through `AGENTS.md`, so the loss is channel classification rather than content, and supporting it needs a way to mark a rule as design guidance plus a second fixed root output. That deserves its own issue and design discussion rather than riding along here.

## Testing

- `pnpm cicheck` (green)
- New unit tests cover both directions for the four fields, the beyond-schema passthrough key, the root-flag precedence over the spread section, and the numeric `version`. Each was verified to fail when the corresponding fix is reverted.

Closes #2725

PR #2911, opened 2026-09-03
fix(skills): stop rejecting packaging metadata shapes tools tolerate

### LINKED ISSUE #2750: skills: a strictly typed optional frontmatter field turns a tolerated value into an aborted run
While reviewing #2748 it turned out that declaring an optional frontmatter field with a concrete type is not free: it converts a value that used to ride the `looseObject` passthrough into a hard parse failure, and a single such file aborts the whole `import` / `generate` run.

Measured on `main` before #2748, with `.factory/skills/bad/SKILL.md`:

| value | before the field was declared | after |
| --- | --- | --- |
| `license: 123` | imports | `Invalid input: expected string, received number`, run aborts |
| `compatibility: [droid, other]` | imports | `invalid_union`, run aborts |
| `metadata: just-a-string` | imports | `expected object, received string`, run aborts |
| `version: 2026-01-01` (js-yaml returns a `Date`) | imports | `invalid_union`, run aborts |

#2748 resolved this for `factorydroid` by declaring the four packaging fields as `z.unknown()`: the field stays documented and is carried through both directions, but its value is not constrained. The reasoning is that the tool documents these fields without a type and never validates them, so rulesync must not be stricter than the tool it writes for.

The same fields are still declared strictly elsewhere, and the specs behind them do not define a type either:

- `KiroSkillFrontmatterSchema` (`src/features/skills/kiro-skill.ts`) and the `kiro` block in `src/features/skills/rulesync-skill.ts`
- the `rovodev`, `kilo`, `opencode`, `deepagents`, `agentsskills` blocks in `src/features/skills/rulesync-skill.ts`

So a Kiro `SKILL.md` carrying `license: 123` still fails the whole import today.

Two things to decide:

1. Whether the `z.unknown()` treatment should be extended to the other targets' packaging fields, or whether a spec that does pin a type (the Agent Skills spec types `compatibility` as a 1-500 character string) justifies keeping the strict declaration there.
2. Whether a value that fails a declared field should degrade to a warning plus passthrough instead of aborting the run, which would make the question moot for every field at once.

Also noted in the same review: `KiroSkill.fromRulesyncSkill` spreads the section before `name`/`description`, so those two end up last in the emitted frontmatter; #2748 moved `factorydroid` to emitting them first. Worth aligning.

dyoshikawa:
## Summary

PR #2748 fixed Factory Droid's packaging frontmatter fields (`license`, `compatibility`, `metadata`, `version`) by declaring them `z.optional(z.unknown())`, because Factory Droid documents these fields without pinning a type and never validates them: a strictly-typed optional field turns an otherwise-tolerated value into a hard Zod parse failure that aborts the entire `import`/`generate` run for that file.

The same three fields (`license`, `compatibility`, `metadata`) were still declared strictly for six other skill targets. This PR researches each target's own documentation and, where accessible, its actual implementation, and applies the same fix only where the evidence supports it.

## Per-target research and decisions

**Loosened to `z.optional(z.unknown())`** (in both `{tool}-skill.ts` and the matching `rulesync-skill.ts` section):

- **Kilo Code** — docs list `license`/`compatibility`/`metadata` (https://kilo.ai/docs/customize/skills), but Kilo's SKILL.md parser is forked from OpenCode's engine, and its actual `Frontmatter` schema (`packages/core/src/skill.ts` in `Kilo-Org/kilocode`, confirmed via source inspection) does not model these fields at all, so it cannot reject any shape rulesync writes.
- **OpenCode** — docs list the same three fields (https://opencode.ai/docs/skills/), but the actual `Frontmatter` schema it parses SKILL.md with (`packages/core/src/skill.ts` in `sst/opencode`, confirmed via source inspection) omits them entirely.
- **DeepAgents** — dcode's `_parse_skill_metadata` (`libs/deepagents/deepagents/middleware/skills.py` in `langchain-ai/deepagents`, read directly) never hard-fails on an unexpected shape: values are coerced with `str()`, an overlong `compatibility` is truncated with a logged warning, and non-dict `metadata` is replaced with `{}` and a logged warning.

**Left strictly typed (no change)**:

- **Kiro** (https://kiro.dev/docs/skills/) and **Rovo Dev** (https://support.atlassian.com/rovo/docs/extend-rovo-dev-cli-with-agent-skills/) are both closed-source, and both docs explicitly defer to the Agent Skills spec for these fields. The spec (https://agentskills.io/specification) itself pins `compatibility` to a free-form string of 1-500 characters and `metadata` to a string-to-string map — contrary to the issue's premise that "the specs behind them do not define a type either," the spec does define types for these two fields. With no implementation access and no evidence of runtime leniency, these stay strict.
- **agentsskills** is the reference implementation of the Agent Skills spec itself. It already handles nonconforming input via its own warn-and-continue `collectAgentSkillViolations` mechanism rather than via Zod strictness, so it needs no change here.

## Deferred: blanket warn-and-passthrough policy

The issue's second question — whether *any* declared-field validation failure should degrade to a warning plus passthrough as a general policy, rather than deciding field-by-field — is intentionally **not** addressed in this PR. That's a separate, larger design decision for the maintainer, mirroring how PR #2748 deferred its own second gap to issue #2749.

## Also fixed: Kiro frontmatter key order

`KiroSkill.fromRulesyncSkill` spread the `kiro` section before `name`/`description`, so a `kiro` section with any of its own keys pushed `name`/`description` to the end of the emitted frontmatter — unlike every sibling adapter (e.g. `FactorydroidSkill.fromRulesyncSkill`), which always emits `name`/`description` first. Fixed by destructuring `name`/`description` out of the section before building the frontmatter object, matching the established pattern.

## Test plan

- [x] Added a schema round-trip test to `kilo-skill.test.ts` and `opencode-skill.test.ts` proving a previously-rejected shape (numeric `license`, array `compatibility`, scalar `metadata`) now parses.
- [x] Added a `fromDir` test to `deepagents-skill.test.ts` proving the same previously-rejected shapes now import successfully.
- [x] Added a new test to `kiro-skill.test.ts` asserting `name`/`description` are always the first two emitted frontmatter keys, even when the `kiro` section defines other keys.
- [x] `pnpm cicheck` passes (402 test files, 10754 tests; format, lint, typecheck, docs sync, gitignore, supported-tools tables, cspell, secretlint all clean).

Closes #2750

### DIFF
--- a/src/features/skills/deepagents-skill.test.ts
+++ b/src/features/skills/deepagents-skill.test.ts
@@ -99,6 +99,39 @@ Do the thing.`;
       expect(skill.getFrontmatter().name).toBe("My Skill");
       expect(skill.getFrontmatter().description).toBe("A useful skill.");
     });
+
+    it("should import the packaging metadata in the shapes YAML actually produces", async () => {
+      const skillDir = join(testDir, ".deepagents", "skills", "packaged-skill");
+      // dcode never hard-fails on these fields: it coerces with `str()`,
+      // truncates an overlong `compatibility`, and drops non-dict `metadata`
+      // with a warning, so none of them may fail the import of a SKILL.md it
+      // happily loads.
+      const skillContent = `---
+name: packaged-skill
+description: A skill shared through a catalog
+license: MIT
+compatibility:
+  - deepagents
+  - claude-code
+metadata: platform-team
+---
+
+Packaged body`;
+      await writeFileContent(join(skillDir, SKILL_FILE_NAME), skillContent);
+
+      const skill = await DeepagentsSkill.fromDir({
+        outputRoot: testDir,
+        dirName: "packaged-skill",
+        global: false,
+      });
+
+      const section = skill.toRulesyncSkill().getFrontmatter().deepagents;
+      expect(section).toEqual({
+        license: "MIT",
+        compatibility: ["deepagents", "claude-code"],
+        metadata: "platform-team",
+      });
+    });
   });
 
   describe("fromRulesyncSkill", () => {
--- a/src/features/skills/deepagents-skill.ts
+++ b/src/features/skills/deepagents-skill.ts
@@ -28,14 +28,17 @@ const DeepagentsSkillFrontmatterSchema = z.looseObject({
   // still accepted here for tolerant parsing of hand-written files, but on emit
   // rulesync always writes the space-delimited string form.
   "allowed-tools": z.optional(z.union([z.string(), z.array(z.string())])),
-  // Agent Skills spec fields read by dcode's `_parse_skill_metadata`.
-  // https://agentskills.io/specification
-  license: z.optional(z.string()),
-  // The Agent Skills spec defines `compatibility` as a free-form string
-  // (1–500 chars), which is what dcode actually reads; an object form is also
-  // tolerated for backward compatibility (matches the `agentsskills` adapter).
-  compatibility: z.optional(z.union([z.string(), z.looseObject({})])),
-  metadata: z.optional(z.looseObject({})),
+  // dcode's `_parse_skill_metadata` reads `license`, `compatibility`, and
+  // `metadata` (see langchain-ai/deepagents
+  // libs/deepagents/deepagents/middleware/skills.py), but it never hard-fails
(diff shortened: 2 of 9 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2912, opened 2026-09-03
fix(orphan-sweep): case-fold path identity before classifying orphans

### LINKED ISSUE #2740: Orphan deletion compares paths exactly, so a case-only rename deletes the just-generated output on macOS/Windows

dyoshikawa:
## Summary

`FeatureProcessor.removeOrphanAiFiles()` and `DirFeatureProcessor.removeOrphanAiDirs()` compared existing vs. generated paths with an exact-string `Set` lookup. On a case-insensitive filesystem (macOS, Windows), renaming a rule/skill by case only (e.g. `my-skill` -> `My-Skill`) writes the generated output over the existing on-disk entry, but the exact-match comparison did not recognize the two spellings as the same identity — so the sweep classified the pre-existing spelling as an orphan and deleted the file/directory that `generate` had just (re)written.

## Fix

Both sweeps now compare `caseFoldIdentity()` of the paths instead of the raw strings, matching the case-folding + Unicode normalization already used for cross-root de-duplication in `feature-processor.ts` (added in #2738). This is a comparison-only change — the paths that are actually deleted are still the original, un-folded spellings.

## Testing

- Added a case-only-rename regression test to both `removeOrphanAiFiles` (`feature-processor.test.ts`) and `removeOrphanAiDirs` (`dir-feature-processor.test.ts`).
- `pnpm cicheck` passes in full.

Closes #2740

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -332,6 +332,7 @@
     "pkgjs",
     "POWERLEVEL",
     "powerline",
+    "precomposed",
     "procps",
     "projectbrief",
     "pycache",
--- a/src/types/dir-feature-processor.test.ts
+++ b/src/types/dir-feature-processor.test.ts
@@ -210,6 +210,24 @@ describe("DirFeatureProcessor", () => {
       expect(logger.warn).not.toHaveBeenCalled();
     });
 
+    it("should not treat a case-only rename as an orphan", async () => {
+      const processor = new TestDirProcessor({
+        logger: createMockLogger(),
+        outputRoot: "/path/to",
+      });
+
+      // On a case-insensitive filesystem, renaming `my-skill` to `My-Skill`
+      // writes the generated directory over the existing one, so the existing
+      // path here is not actually a leftover from the old spelling.
+      const existingDirs = [createMockDir({ dirPath: "/path/to/my-skill" })];
+      const generatedDirs = [createMockDir({ dirPath: "/path/to/My-Skill" })];
+
+      const count = await processor.removeOrphanAiDirs(existingDirs, generatedDirs);
+
+      expect(count).toBe(0);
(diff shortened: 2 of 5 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2914, opened 2026-09-03
fix(rules): stop fold-import duplication check from tripping generate-time validation

dyoshikawa:
## Summary

- `warnForFoldImportDuplicationRisk()` (added in #2910) called the full `loadRulesyncFiles()` to inspect the merged non-root rules for a fold-policy tool. That method also emits a "no root rule found" warning whenever `.rulesync/rules/` is non-empty but has no root rule targeting the tool yet, and throws when a `localRoot: true` rule exists for the tool without an accompanying `root: true` rule.
- Since the duplication check runs during `rulesync import`, *before* the imported root file is written, no root rule targets the tool yet at that point by construction — so the "no root rule found" warning fired spuriously on every fold-tool import that had any pre-existing rules for other tools. Worse, a pre-existing `localRoot` misconfiguration for that tool turned an import that previously completed successfully into a thrown error, since import never called `loadRulesyncFiles()` before this change.
- This was caught in a post-merge re-review of #2910, credited to `code-review-2910-r1-2`.

## Fix

- Extracted the side-effect-free "load and merge rules from every input root" step out of `loadRulesyncFiles()` into a new private `loadMergedRulesyncRules()`.
- `loadRulesyncFiles()` now calls that helper and keeps all of its existing generate-time validation (root-rule warning, `localRoot` checks) on top.
- `warnForFoldImportDuplicationRisk()` now calls `loadMergedRulesyncRules()` directly and does its own lightweight root/target filtering, so it never triggers `loadRulesyncFiles()`'s generate-time checks.

## Test plan

- [x] Added a regression test confirming the "no root rule found" warning is not emitted by `warnForFoldImportDuplicationRisk()` even when other-tool rules exist and none yet target the tool being imported.
- [x] Added a regression test confirming `warnForFoldImportDuplicationRisk()` does not throw when a pre-existing `localRoot` misconfiguration exists for the target tool.
- [x] Verified both new tests fail against the pre-fix code and pass after the fix.
- [x] `pnpm cicheck` passes (402 files, 10767 tests, all content checks clean).

### DIFF
--- a/src/features/rules/rules-processor.test.ts
+++ b/src/features/rules/rules-processor.test.ts
@@ -1414,6 +1414,43 @@ describe("RulesProcessor", () => {
         expect.stringContaining("will re-add content already folded from"),
       );
     });
+
+    it("should not emit loadRulesyncFiles's 'no root rule found' warning, since this runs before the imported root file is written", async () => {
+      // `rulesync import` calls this before writing the imported root file to
+      // `.rulesync/rules/`, so at this point no rule targets codexcli yet —
+      // that's expected and not something to warn about here. A rule for an
+      // unrelated tool is present so `loadRulesyncFiles`'s "no root rule
+      // found" check (which fires whenever `.rulesync/rules/` is non-empty
+      // but has no root targeting this tool) would otherwise trip.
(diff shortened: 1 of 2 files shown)


dyoshikawa:
@dyoshikawa Thank you!

PR #2919, opened 2026-09-03
fix(permissions): honor all-tools rules in AugmentCode shell

### LINKED ISSUE #2917: permissions: AugmentCode adapter ignores the all-tools `*` category for shell commands
`convertRulesyncToAugmentEntries` in `src/features/permissions/augmentcode-permissions.ts` iterates `config.permission` by category and maps each to an AugmentCode `toolName` via `CANONICAL_TO_AUGMENT_TOOL_NAMES`. That map has no entry for the all-tools `*` category, so `toAugmentToolName("*")` falls through to `?? canonical` and returns the literal string `"*"` as the `toolName` — an unmanaged, unrecognized tool name from AugmentCode's point of view. The `*` category's rules are never routed to `launch-process` (the `bash`-mapped tool name), so a `deny` written under `*` never restricts commands that `bash`/`launch-process` auto-approves.

Reproduced with `{"*": {"rm *": "deny"}, "bash": {"rm *": "allow", "git *": "allow"}}`:

- Generated `toolPermissions` include `{toolName: "launch-process", shellInputRegex: "...rm.*...", permission: {type: "allow"}}` for `rm *` (from `bash`), plus a separate `{toolName: "*", permission: {type: "deny"}}` entry (from the all-tools category, logged as "passing through unknown tool category '*' as toolName").
- AugmentCode matches `toolPermissions` by exact `toolName` under first-match-wins, so the `toolName: "*"` entry never matches a `launch-process` invocation. The `rm *` allow stands unshadowed.

So a config whose author denied `rm *` for every tool still ships an auto-approve entry for it via AugmentCode, with no warning that the `*` category rule was effectively dropped for shell commands.

`src/features/permissions/augmentcode-permissions.ts` (`convertRulesyncToAugmentEntries`, `CANONICAL_TO_AUGMENT_TOOL_NAMES`, `toAugmentToolName`).

Suggested scope: route AugmentCode's `launch-process` (bash) entries through the same all-tools-aware resolution used elsewhere (e.g. `bashRulesHonoringAllTools` / `collectShellCommandRules` in `shell-command-categories.ts`) so a `*` deny/ask is folded into the generated `launch-process` entries instead of becoming an inert `toolName: "*"` entry; and/or warn when the `*` category is passed through as an unmanaged toolName, the way other "unknown tool category" pass-throughs already do.

Found during the review and fix of #2831 / PR #2905 (which fixed the same class of bug for Rovo Dev, Zed, and Roo Code); deliberately left out of that PR as a distinct adapter and out of its scope.

Junaid-PK:
Fixes #2917

## What changed

- Resolve AugmentCode shell permissions through the shared all-tools-aware helper, including configs with no separate bash category.
- Prevent an all-tools deny or ask from being bypassed by a conflicting launch-process allow.
- Skip the unsupported wildcard toolName entry, remove legacy copies during regeneration, and warn when non-shell all-tools effects cannot be represented.
- Add regressions for the reported rm conflict and an all-tools-only deny/ask configuration seeded with legacy output.

## Verification

- pnpm cicheck
- 402 test files passed; 10,811 tests passed and 3 skipped
- Formatting, lint, typecheck, generated-content checks, cspell, and secretlint passed


dyoshikawa:
@Junaid-PK Thank you!

PR #2925, opened 2026-09-04
fix(shared-config): bound the sanitizer walk against YAML alias bombs and cycles

### LINKED ISSUE #2915: Shared config YAML parsing has no protection against alias-bomb / circular-anchor DoS
## Summary

`shared-config-gateway.ts`'s YAML/JSON parsing path for shared config files (used by Rovodev, and other tools' shared-config readers) has no protection against YAML alias/anchor expansion bombs or self-referencing anchors, unlike the hardening already applied to frontmatter parsing in `src/utils/frontmatter.ts` (commits 6d479eb45, bc10edc9b, 338537031, closing #2751).

## Details

- `loadYaml` (`src/utils/yaml.ts`) is a thin wrapper around `js-yaml`'s `load()` with no `maxAliasCount` or depth limit configured.
- `sanitizeSharedConfigValue` (`src/features/shared/shared-config-gateway.ts:100-113`) recursively walks the parsed result to strip prototype-pollution keys, but does so with a **plain recursive walk with no memoization and no cycle guard**.
- js-yaml resolves anchors/aliases into real shared JS object references. A "billion laughs"-style document (nested anchors, each aliased many times) makes the *sanitizer's* naive walk redundantly re-visit the same shared subtree exponentially, even though the parsed object itself is cheap — CPU-exhaustion DoS.
- A self-referencing anchor (e.g. `a: &a {b: *a}`) produces a truly circular JS object. The unguarded recursive sanitizer recurses infinitely and crashes the process with a stack overflow (`RangeError: Maximum call stack size exceeded`).

Both are triggerable simply by committing a crafted config file (e.g. `.rovodev/config.yml`) into a repository and having a victim run `rulesync import` on it — a "clone and run" scenario. `sanitizeSharedConfigValue` has several call sites in `shared-config-gateway.ts` (e.g. lines 170, 900, 976, 1009, 1818), so this affects every tool that reads its native config through this shared gateway, not just Rovodev.

This is the same bug class the recent frontmatter-cleaning commits hardened:
- 6d479eb45 — memoize shared references (fixes alias-bomb blowup) and drop cycles via an ancestor set (fixes self-reference stack overflow)
- bc10edc9b — bound alias expansion via a `MAX_FRONTMATTER_VALUES` budget so an alias bomb is refused with a clear error instead of exhausting the heap
- 338537031 — charge nullish leaves against the same budget

`shared-config-gateway.ts`'s `sanitizeSharedConfigValue`/`loadYaml` path shares none of that protection.

## Suggested fix

Route `sanitizeSharedConfigValue` (and/or `loadYaml`'s call sites in the shared-config gateway) through the same alias-budget/memoization/cycle-guard approach already built for `src/utils/frontmatter.ts`, ideally by extracting a shared utility both modules can use.

## Provenance

Found during the security review of #2902 (`security-review-2902-r1`), which fixes #2736. Rated MID by that review: local CLI, user-initiated action, no data compromise or remote code execution — but a real crash/hang DoS reachable from an untrusted cloned repo. Out of scope for #2902 itself since the vulnerable code lives entirely in `shared-config-gateway.ts`/`yaml.ts`, files that PR does not touch.

dyoshikawa:
## Summary

`sanitizeSharedConfigValue` in `src/features/shared/shared-config-gateway.ts` rebuilt every parsed value with a plain recursive walk that had no budget and no cycle guard. js-yaml resolves anchors and aliases into shared JS object references, so:

- a small "billion laughs" document of nested anchors made the walk exponential (CPU-exhaustion DoS), and
- a self-referencing anchor (`a: &a {self: *a}`) recursed until the process crashed with a stack overflow.

Both are reachable from a config file committed to a cloned repository (for example `.rovodev/config.yml`) when the victim runs `rulesync import`, and the sanitizer is shared by every tool whose native config is read through the gateway.

## Changes

- The sanitizer now walks with the same bounds the frontmatter cleaner gained in the fix for #2751: every value is charged against `MAX_SHARED_CONFIG_VALUES` (100,000), every string leaf and mapping key against `MAX_SHARED_CONFIG_STRING_CHARS` (4,000,000), nesting is capped at `MAX_SHARED_CONFIG_DEPTH` (64), and a reference back to an ancestor is refused outright with a clear error instead of walked until the stack overflows.
- The value under a dropped prototype-pollution key is still walked and charged, so `__proto__:` cannot hide an unbudgeted alias chain.
- `parseSharedConfig` prefixes these refusals with the file path (`Failed to parse shared config at <path>: ...`) the same way it does for syntax errors.
- Ordinary shared references that are not cycles are still expanded into independent copies, matching the existing writer behaviour (`dump` with `noRefs: true`).

Memoizing shared references was deliberately not used: the writers dump with `noRefs: true`, so it would only move the blowup from the sanitizer into serialization.

## Tests

New cases in `shared-config-gateway.test.ts` cover the alias bomb (with the path prefix), a self-referencing mapping and array, a long string aliased past the character budget, a document nested past the depth cap, a non-cyclic shared reference that must keep working, and an alias bomb hidden under `__proto__`.

`pnpm cicheck` passes.

Closes #2915

### DIFF
--- a/src/features/shared/shared-config-gateway.test.ts
+++ b/src/features/shared/shared-config-gateway.test.ts
@@ -14,6 +14,9 @@ import {
   CLAUDE_SETTINGS_SHARED_FILE_KEY,
   HERMES_CONFIG_SHARED_FILE_KEY,
   isReadDenyEntry,
+  MAX_SHARED_CONFIG_DEPTH,
+  MAX_SHARED_CONFIG_STRING_CHARS,
+  MAX_SHARED_CONFIG_VALUES,
   mergeSharedConfigDeep,
   mergeSharedConfigShallow,
   parseSharedConfig,
@@ -23,6 +26,22 @@ import {
   stringifySharedConfig,
   TAKT_CONFIG_SHARED_FILE_KEY,
 } from "./shared-config-gateway.js";
+
+/**
+ * A YAML document whose `metadata` key holds `levels` nested anchors, each an
+ * array of `width` aliases to the previous one: `width ** levels` values once
+ * every alias is expanded, from a file of a few hundred bytes.
+ */
+function aliasChain(levels: number, width: number): string {
+  const names = "abcdefghijklmnopqrstuvwxyz";
+  const lines = ["metadata:"];
+  for (let i = 0; i < levels; i += 1) {
+    const items = Array.from({ length: width }, () => (i === 0 ? "x" : `*${names[i - 1]}`));
+    lines.push(`  ${names[i]}: &${names[i]} [${items.join(", ")}]`);
+  }
+  lines.push("");
+  return lines.join("\n");
+}
 
 /** The whole-document fallback warnings `logger` received, in order. */
 const wholeRewriteWarnings = (logger: ReturnType<typeof createMockLogger>): string[] =>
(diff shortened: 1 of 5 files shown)


dyoshikawa:
Merged. Thank you for the review rounds; the bounded walk now lives in src/utils/bounded-walk.ts and guards both frontmatter and shared config parsing.

Issue #2668, opened 2026-08-13
Follow up Qwen Code upstream updates: .qwen/review-rules.md checks surface, fork profiles, server-level mcp.allowed/excluded


dyoshikawa:
## Summary

Three Qwen Code surfaces are unfollowed: the `/review` project rules file `.qwen/review-rules.md` (the `checks` dimension is blank for this target), named fork profiles `.qwen/fork-profiles/*.md`, and the server-level `mcp.allowed` / `mcp.excluded` allow/deny lists. Every other dimension was verified current — in several places rulesync is already ahead of the upstream docs table.

Sources: QwenLM/qwen-code `main` docs and CHANGELOG; latest stable **v0.21.11** (2026-08-13). Both dry-runs succeed. (No commands appear in either run simply because `/workspace/.rulesync/commands` is empty, not because the adapter is gated; and the absent global ignore file is correct, since upstream has no global `.qwenignore`.)

## Recent Releases

- **v0.21.11** (2026-08-13) — current stable baseline.
- **v0.21.6** (2026-08-05) — `.qwen/review-context.json`, a strict JSON manifest consumed by `/review` ([PR #8401](https://github.com/QwenLM/qwen-code/pull/8401), [PR #8654](https://github.com/QwenLM/qwen-code/pull/8654)).
- **v0.21.3** (2026-08-01) — named fork profiles ([PR #8148](https://github.com/QwenLM/qwen-code/pull/8148)).
- **PR #2932** (merged 2026-04-09) — the deterministic `/review` rework that introduced `.qwen/review-rules.md`.

### 1. checks — the project review rules file `.qwen/review-rules.md` is not emitted

- **Upstream**: `/review` reads review criteria from `.qwen/review-rules.md` first, then falls back to `.github/copilot-instructions.md` and to a `## Code Review` section in `AGENTS.md` / `QWEN.md`. Documented at https://qwenlm.github.io/qwen-code-docs/en/users/features/code-review/ (§ Project Review Rules); introduced by [PR #2932](https://github.com/QwenLM/qwen-code/pull/2932) (2026-04-09).
- **rulesync**: `unsupported` — there is no `qwencode-check.ts` and the checks matrix row is blank for this target.
[...]

PR #2935, opened 2026-09-04
feat(permissions): author Qwen Code's tools.listDirectory and tools.workflowsEnabled

dyoshikawa:
## Summary

Closes the `tools.workflowsEnabled` and `tools.listDirectory` gaps recorded on #2668: neither key could be authored through the `qwencode` permissions override, and `workflowsEnabled` would additionally have been written into a scope where Qwen Code ignores it.

Both keys are now documented in a **stable** release — Qwen Code [v0.23.0](https://github.com/QwenLM/qwen-code/releases/tag/v0.23.0) (2026-09-03), `docs/users/configuration/settings.md`:

- `tools.listDirectory.enabled` — "Enable the built-in `list_directory` tool", default `false`.
- `tools.workflowsEnabled` — "Enable the Workflow tool… Off by default", **"User, System, and SystemDefaults scopes only; workspace values are ignored."** ([PR #9098](https://github.com/QwenLM/qwen-code/pull/9098))

The scope note is what made this more than a two-line list addition: `patch.tools` previously spread the override verbatim with no scope filter, so `qwencode: { tools: { workflowsEnabled: true } }` would land in the project `.qwen/settings.json` as dead configuration that reads as if it took effect.

## What changed

- **`QWEN_OVERRIDE_TOOLS_KEYS`** gains `listDirectory` and `workflowsEnabled`, and `QWEN_OVERRIDE_SECURITY_KEYS` gains `allowedInsecureVoiceBaseUrls` — the base URLs a voice provider may be reached at over cleartext HTTP or a private address, which also carries the provider API key in its `Authorization` header, and which Qwen Code strips out of workspace settings for that reason. `rulesync import --targets qwencode` now lifts all of them back into the override instead of dropping them.
- **A per-key rule table (`QWEN_SCOPE_RULES`)** replaces the old `QWEN_GLOBAL_ONLY_*_KEYS` lists, because Qwen Code applies three different scope rules and the old boolean could only express one:
  - `workspace-stripped` — `tools.workflowsEnabled`, `security.allowPrivateNetworkHooks`, `security.allowedInsecureVoiceBaseUrls` are on upstream's `WORKSPACE_RESTRICTED_SETTINGS`, so they are dropped from project scope with a warning and written (and announced) only in global scope.
  - `workspace-non-overriding` — `security.allowedHttpHookUrls` is on upstream's `WORKSPACE_NON_OVERRIDING_SETTINGS`: a workspace value is honored only while no higher scope sets the key. It is written in both scopes, with a note in project scope; because an empty list is Qwen Code's allow-all, *every* global change is announced, not just a granting one.
  - `user-scope-trust-check` — `security.folderTrust` is read on both sides of the merge. The *initial* trust decision comes from system + user settings alone (`loadSettings`' `initialTrustCheckSettings`, defaulting to trusted), so a project value cannot vouch for its own repository; once the workspace is trusted the merged value still drives the folder-trust gate, so it is emitted with a note saying which half it cannot decide.
- **Every global override write is announced**, not only the modeled keys. `qwencode.tools` / `qwencode.security` are `looseObject`s, so an unmodeled key previously reached `~/.qwen/settings.json` with no warning at all — including potent ones such as `tools.discoveryCommand` and `tools.callCommand`, which Qwen Code spawns at startup. Unlisted keys now fall back to an `unmodeled` rule: still written, but named. `tools.approvalMode`, `autoAccept`, `sandbox`, `sandboxImage`, `disabled` and `visible` get a `global-machine-wide` rule with wording of their own.
- **Announced values are quoted through `quoteValueForWarning`**, not `JSON.stringify`, which escapes C0 only and passes the C1 range, the bidi overrides and unbounded length straight through — and these values come out of files rulesync did not write.
- **Import warns on promotion** for the scoped keys, and only when the file came from the project scope: a value read out of the global file is already where the warning would send it.
- **Refactor:** `scopeOverrideSecurity` is generalized into `scopeOverrideGroup`, and the `tools`/`security` overlay moves into `buildOverrideGroupsPatch`, so both groups share one code path. This also keeps `fromRulesyncPermissions` under the project's complexity ceiling.
- `docs/reference/file-formats.md` updated (key list plus the full scope-caveat passage), with `src/generated/docs-content.ts` regenerated. `QwencodePermissionsOverrideSchema`'s doc comment updated too — the schema itself is `looseObject`, so no schema change was needed.

## Not in scope

#2668 stays open. It is a standing upstream-tracking issue and its title items are untouched here: the `.qwen/review-rules.md` checks surface, `.qwen/fork-profiles/*.md`, and the server-level `mcp.allowed` / `mcp.excluded` keys. Those are scope decisions (a new `checks` adapter, a new file surface) rather than facts to verify, so they are left for a maintainer.

## Verification

`pnpm cicheck` passes: oxfmt, oxlint (`--max-warnings 0`), tsgo typecheck, the full vitest suite (11036 tests), and the content checks (docs-content drift, gitignore, supported-tools, cspell, secretlint). The Qwen permissions suite is now 60 tests, covering the `listDirectory` round trip, each scope rule in both directions, the global announcements (including one for a key rulesync does not model and one for a shrinking `tools.disabled` list), the control-character stripping, the equality guards that keep a re-run silent, and the empty-group case. Every claim about upstream behavior was checked against Qwen Code v0.23.0's own `packages/cli/src/config/settingsUtils.ts` and `settings.ts`.

Refs #2668

https://claude.ai/code/session_01QrwSR9PWbPsPdZm5zibyDQ


dyoshikawa:
@dyoshikawa Thank you!

PR #2942, opened 2026-09-05
fix(hooks): preserve unowned dest handlers when opted in

### LINKED ISSUE #2940: generate: preserve existing third-party hook command entries
## Problem

`rulesync generate` replaces the `hooks` object for Claude Code, Codex CLI, and Cursor. Third-party command hooks installed after generate are wiped on the next generate unless they are duplicated into `.rulesync/hooks`.

This is generate-time data loss, not the import-merge case in #2797. #2913 preserved undocumented event keys for AugmentCode; it does not keep unmatched command entries on native events.

Moshi install is documented as non-destructive of user-owned hooks. RuleSync is the destructive side.

## Repro

1. `rulesync generate` so `.claude/settings.json` (or `.codex/hooks.json` / `.cursor/hooks.json`) contains only RuleSync-owned commands.
2. Install a third-party command hook on a native event (e.g. `moshi-hook install`).
3. `rulesync generate` again.

Expected: RuleSync-owned commands stay current; the third-party command is still present.
Actual: the third-party command is gone.

## Proposal

When writing those JSON hook files, keep existing hook entries whose command is not in the generated set. Deduplicate by command string (and matcher when the tool uses matcher groups). Do not change import semantics.

Generic, not Moshi-named.

## Out of scope

- Import merge (#2797)
- Plugin-style dests (OpenCode/Pi generated TS/Python)
- Grok CLI `rulesync.json` (dedicated owned file; other `*.json` in `.grok/hooks/` already coexist)

RaviTharuma:
## Summary

`rulesync generate` still replaces the `hooks` object by default for Claude Code, Codex CLI, and Cursor.

Set `preserveUnowned: true` in `.rulesync/hooks.jsonc` to keep dest handlers that rulesync does not own:

- Identity is type-aware (command / url / server+tool / prompt), not command-string-only.
- Callers pass the dest shape (`matcher-groups` vs `flat`).
- Commands that refer to `.rulesync/hooks` are treated as stale rulesync writes and are dropped when generate no longer emits them.
- Plugin dests always replace.
- Result maps are `Object.create(null)`.

This addresses the review on the previous identity model: generate remains a reset unless the project opts in.

Fixes #2940.

## Test plan

- [x] Helper tests for matcher-group and flat merge, stale `.rulesync/hooks` drop, matcher edit, http preserve, `toString` event, mixed shape skip
- [x] Claude/Cursor/Codex regenerate tests with `preserveUnowned: true`
- [x] Default generate still wipes third-party commands
- [x] Plugin dest still replaces when `preserveUnowned` is set
- [x] `pnpm cicheck` (CI)


dyoshikawa (changes requested):
Thank you for taking this on, and sorry for the long review — the problem in #2940 is real and worth fixing, but the identity model this PR uses ("an entry is third-party iff its exact command string is absent from the generated set") turns out to change more than the issue asked for. I ran the helper in isolation against the cases below, so these are observed outputs rather than code reading.

## Blocking

### 1. `generate` can no longer remove a hook it previously wrote

`preserveUnownedHookCommands` has no ownership marker, so a command rulesync itself wrote on an earlier run is indistinguishable from a third-party one:

```
existingHooks:  { SessionStart: [{ hooks: [{ type: "command", command: "echo A" }] }] }
generatedHooks: {}
=> { "SessionStart": [ { "hooks": [ { "type": "command", "command": "echo A" } ] } ] }
```

Deleting a hook from `.rulesync/hooks` therefore never removes it from `.claude/settings.json`, `.codex/hooks.json` or `.cursor/hooks.json`. That contradicts the ownership model this repo documents elsewhere — `src/features/shared/shared-config-gateway.ts` states for Takt that "a gate deleted there must not survive in config.yaml" and that authoritative snapshots must not "keep a default-deny capability switched on after the user revoked it" — and `docs/reference/file-formats.md` carries the same prose for other targets.

It also has a security consequence that I do not think we can accept as the default: hook entries are arbitrary shell commands. Once a command reaches one of these files by any route (a malicious PR, a compromised postinstall, a `settings.json` committed in a cloned repo), `rulesync generate` will rewrite it back on every subsequent run. Today `generate` is the way to get such a file back to a known state; after this change it becomes the mechanism that keeps the command alive.

`SHARED_CONFIG_OWNERSHIP` still declares `hooks: { kind: "replace-owned-keys", ownedKeys: ["hooks"] }` for `.claude/settings.json`, so the declaration and the actual semantics now disagree.

Suggested direction: make rulesync-written entries identifiable — an ownership manifest, or a marker derived from the existing `${CLAUDE_PROJECT_DIR}` / `.rulesync/hooks/` command conventions — and preserve only entries that are *not* marked. If that is too large a change for this PR, please make preservation opt-in (a `.rulesync/hooks` setting or a CLI flag) and keep replacement as the default.

### 2. Editing a command or a matcher duplicates the hook

Identity is the exact `(matcher, command)` pair, so any edit in `.rulesync/hooks` leaves the previous entry behind as "third-party":

```
# matcher changed Bash -> Edit in .rulesync/hooks, then regenerate
{"PreToolUse":[{"matcher":"Edit","hooks":[{"command":"a"}]},
               {"matcher":"Bash","hooks":[{"command":"a"}]}]}
```

Both now run, and the stale entry survives every later regeneration. Renaming a script or changing its arguments has the same effect, and the duplicates accumulate.

### 3. Non-command handlers are silently dropped

`commandOf()` returns `undefined` for any handler without a string `command`, and those handlers are then filtered out of `leftovers`. Claude Code settings can contain `http`, `mcp_tool`, `agent` and `prompt` handlers — `HOOK_TYPES` in `src/types/hooks.ts` lists all six, and `claudecode-hooks.test.ts` already covers importing them. Observed: an existing `{ hooks: [{ type: "http", url: "https://…" }] }` group is not merged and the whole group disappears. For those types the PR makes the reported data loss worse rather than fixing it.

Suggested fix: derive an identity per handler type (command / url / server+tool / prompt), or compare normalized JSON, and preserve every handler that is not in the generated set.

### 4. `generate` crashes on an event key inherited from `Object.prototype`

`result` is `{ ...generatedHooks }`, so `result[event]` resolves inherited members. `isPrototypePollutionKey` does not cover `toString` / `valueOf` / `hasOwnProperty`:

```
existingHooks: JSON.parse('{"toString":[{"hooks":[{"command":"x"}]}]}')
=> TypeError: values.some is not a function
```

A hand-written or hostile `.claude/settings.json` can abort the whole `generate` run. Please build `result` with `Object.create(null)` (or a `Map`) and guard the lookup, e.g. `Object.hasOwn(result, event) && Array.isArray(result[event]) ? result[event] : []`.

### 5. Shape sniffing breaks on mixed input

`isMatcherGroupList(generated) || isMatcherGroupList(existing)` decides the merge mode from whichever side happens to look like a matcher group. Both directions misbehave:

- existing flat + generated matcher groups: `mergeMatcherGroups` skips every `!isMatcherGroup` entry, so third-party entries vanish with no warning.
- existing matcher-shaped + generated flat (Cursor): the output mixes both shapes in one array, e.g. `{"sessionStart":[{"type":"command","command":"shared.sh"},{"hooks":[{"command":"tp.sh"}]}]}`. If Cursor rejects the file on schema validation, the rulesync-generated hooks stop working too.

Please pass the expected shape in from the caller (`shape: "matcher-groups" | "flat"`) rather than inferring it, and warn instead of discarding when an existing entry does not match.

### 6. `ClaudecodePluginHooks` inherits the behaviour

`ClaudecodePluginHooks` extends `ClaudecodeHooks` and does not override `fromRulesyncHooks`, so plugin builds now read back their own previous output (`hooks/hooks.json`, a fully generated artifact with `isDeletable() === true`). Removed hooks would persist into the distributed plugin. This path should keep replacing.

## Non-blocking

7. **Only three of the affected targets are changed.** The same "replace the whole hooks block in a user-owned file" pattern exists for qwencode, kiro, junie, reasonix, devin, kimi-code and antigravity. Either apply the helper there too, or state the scope in the docs and open a follow-up issue, so the behaviour is not inconsistent across targets.

8. **Duplicated parse block / bypassed gateway.** The "parse existing file, pull out `hooks`" block is copied into `claudecode-hooks.ts`, `codexcli-hooks.ts` and `cursor-hooks.ts`, and `augmentcode-hooks.ts` already has a fourth copy; the claudecode copy is missing the `existingContent.trim() === ""` guard the other two have. Worse, claudecode now parses the same content twice — once with a bare `JSON.parse`, then again inside `applySharedConfigPatch` via `parseSharedConfig`, which applies prototype-pollution sanitising and the empty-file / `invalidRootPolicy` rules. The bare parse fails silently into `catch {}`, so any divergence turns preservation off with no signal. Please export a shared `extractExistingHooks({ fileContent })`, or route this through the gateway (`kind: "custom", policyFunction: …`).

9. **Test coverage.** `codexcli-hooks.ts` is modified but `codexcli-hooks.test.ts` gains no test. Missing across all three: regeneration idempotency, deletion from `.rulesync/hooks`, matcher/command edits, global scope, and a malformed existing file. Findings 1–3 are exactly what those tests would catch, which is why CI is green.

10. **No logging.** The helper takes no logger, so an entry appears in the output with no trace of why. `hooks-processor.ts` warns for skipped matchers; preserved third-party commands deserve at least the same, and given finding 1 I would make it a `warn` listing each preserved command.

11. **Docs.** Visible generate behaviour changes for three targets with no `docs/reference/file-formats.md` update; that file documents ownership prose for other targets. Remember `pnpm run generate:docs-content` after editing docs, since CI fails on drift.

12. **Minor.** `matcherKey` collapses every non-string matcher (including `undefined`) to `""`, so `matcher: ""` and an absent matcher share a key and a narrowly scoped third-party hook can be merged into an unmatched group; group-level keys other than `matcher`/`hooks` are lost when leftovers are pushed into a generated group; `mergeFlatHandlers` and `cloneMatcherGroup` sometimes return the input by reference, so the function is not consistently pure; and `.codex/hooks.json` is still rewritten with `JSON.stringify({ hooks })`, dropping any other top-level key.

## Summary

Findings 1–6 all trace back to the same root: identity is "the exact command string", and there is no record of what rulesync itself wrote. I think the fix is to make ownership explicit (marker or manifest) and to make handler identity type-aware; with that in place the matcher-change duplication and the non-command loss disappear as well. Happy to discuss the ownership-marker shape before you invest more work in it.

RaviTharuma:
Pushed a replacement commit that takes the opt-in path you suggested.

**Default is still replace.** `generate` only keeps dest handlers when `.rulesync/hooks.jsonc` sets `preserveUnowned: true`. That keeps generate a reset, including for `echo A` and for hostile entries.

When opted in:

1. Stale rulesync writes whose command refers to `.rulesync/hooks` are dropped even if generate no longer emits them, so deleting/editing those script hooks does not leave duplicates.
2. Identity is type-aware (`command` / `url` / `server`+`tool` / `prompt`), so http/mcp_tool/prompt/agent handlers are preserved.
3. The merge result is `Object.create(null)` and event lookup uses `Object.hasOwn`.
4. Callers pass `shape: "matcher-groups" | "flat"` instead of sniffing. Mismatched existing entries are skipped with a warning.
5. `ClaudecodePluginHooks` never preserves: only `.claude/settings.json` is eligible, and only when the flag is set.

Codex CLI gained a regenerate test. Other JSON targets (qwencode, kiro, ...) are unchanged in this PR; happy to follow up.

I could not request you as a reviewer from this fork (the review-request API 404s). Please take another look when you have time.

dyoshikawa (changes requested):
Thanks for the thorough round 2 — the opt-in gate, `Object.create(null)` + `Object.hasOwn`, the caller-supplied `shape`, the type-aware identity, and the plugin-destination guard all address round-1 findings, and I verified each of those against the head commit. Four of the six round-1 blockers are genuinely fixed.

Unfortunately the core one is not, and the new identity code introduces a growth bug, so I still have to request changes.

## Round-1 findings: status

| # | Finding | Status |
|---|---------|--------|
| 1 | Deletion semantics broken (no ownership marker) | **NOT FIXED** |
| 2 | Duplicates when a generated command/matcher changes | **PARTIALLY FIXED** (matcher-only edits fixed; command edits still duplicate) |
| 3 | Non-command handlers dropped | **FIXED** (except `function`, see #4 below) |
| 4 | Crash on inherited-prototype event keys | **FIXED** |
| 5 | Shape sniffing breaks on mixed input | **FIXED** |
| 6 | `ClaudecodePluginHooks` inherits the behavior | **FIXED** |

I also confirmed the claim that the default path is unchanged: with `preserveUnowned` unset or `false`, `mergeGeneratedHookLists` early-returns and all three targets plus the plugin destination still fully replace. `tsc --noEmit` passes and the hook test suites are green.

## Blocking

### 1. [critical] Ownership is still a heuristic, so deletions are still lost

`src/features/hooks/preserve-unowned-hook-commands.ts:237-243`

```ts
function isStaleRulesyncCommand(handler: unknown): boolean {
  return (
    isPlainObject(handler) &&
    typeof handler.command === "string" &&
    handler.command.includes(".rulesync/hooks")
  );
}
```

`.rulesync/hooks` is a convention used by the scaffold example and the README, not an ownership marker. A user who writes `"command": "npm run lint"`, `"./scripts/fmt.sh"`, or `"$CLAUDE_PROJECT_DIR/scripts/x.sh"` in `.rulesync/hooks.jsonc` produces a generated command that this predicate never matches. Once such a hook is removed from the source and `generate` is re-run, it is classified as *unowned* and written back into `.claude/settings.json` forever.

Reproduced by running the helper in isolation:

```
run 1: source has old.sh  -> dest [old.sh]
run 2: source has new.sh  -> dest [new.sh, old.sh]   (old.sh is now permanent)
```

The warning text makes it worse: it says `Preserving unowned hook on <event>` for a command rulesync itself wrote on the previous run.

Hooks are arbitrary shell, so this turns `generate` from a reset into an append-only persistence mechanism, which is the same consequence as round 1 — just now gated behind an opt-in. The ownership stance the codebase already documents for a comparable case is explicit (`src/features/shared/shared-config-gateway.ts:1318-1320`): a gate deleted in the rulesync source "must not survive in config.yaml".

Suggested direction: replace the heuristic with real ownership. Either (a) stamp each generated handler with a marker key that the tool ignores and treat "has the marker but is not in this run's generated set" as *delete*, or (b) persist the previous run's generated identities in a rulesync-side lock file (e.g. `.rulesync/.generated-hooks.json`) and retract against that. If neither is workable for a given target, I would rather not ship preservation for that target than ship a mechanism that cannot retract.

### 2. [high] Handlers with no derivable identity grow the destination file on every run

`preserve-unowned-hook-commands.ts:226-235`

```ts
const identity = handlerIdentity(handler);
if (identity === undefined) {
  return isPlainObject(handler);   // always preserved, and never remembered as owned
}
```

`handlerIdentity` returns `undefined` unless `command` / `url` / `prompt` / `server`+`tool` is a non-empty string. When the generated set contains such a handler, the matching existing entry is classified unowned, appended, and — because `rememberOwned` cannot register it either — appended again on the next run:

```
run 1 handlers: 2
run 2 handlers: 3
run 3 handlers: 4
run 4 handlers: 5
```

This is reachable today: `HookDefinitionSchema` makes `command` optional and `buildToolHooks` emits `{type: "prompt"}` / `{type: "agent", model: "..."}` without a body (the warning text at `src/features/hooks/qwencode-hooks.ts:103`, "prompt hook without a prompt body", exists precisely because that state occurs). `command: ""` hits the same path.

Suggested fix: fall back to a stable canonical serialization (sorted-key `JSON.stringify`) as the identity, and route `rememberOwned` through the same fallback so the generated and existing sides can always be matched.

### 3. [mid, blocking in combination] `preserveUnowned` lives in a file that `rulesync fetch` copies

`src/types/hooks.ts:1051` puts the key in `.rulesync/hooks.jsonc`. Every other generation-behavior toggle (`simulateCommands`, `deriveSubprojectPathFromGlobs`, `gitignoreTargetsOnly`, `delete`) lives in `ConfigParamsSchema` (`src/config/config.ts:111-167`), and for good reason here: `.rulesync/hooks.jsonc` is a shareable file that `rulesync fetch` copies into other projects. A fetched `hooks.jsonc` carrying `preserveUnowned: true` silently switches the receiving project's `generate` from reset to append, after which previously planted hooks can no longer be retracted. The same file is committable, so the flag can also be flipped in a pull request.

Suggested fix: move it to `ConfigParamsSchema` as e.g. `preserveUnownedHooks`, passed down through `HooksProcessor`. That also resolves finding #9 below.

### 4. [mid, blocking in combination] The new key is undocumented

`docs/reference/file-formats.md` still lists only `version`, `hooks`, and `{tool}.hooks` as the top-level keys of `.rulesync/hooks.jsonc` (around lines 171-230), and `src/generated/docs-content.ts` was not regenerated. Since the distributed skill delegates lookup to `rulesync docs`, an undocumented key is invisible to both users and agents. Please document what changes about deletion semantics, which targets are supported, and that plugin output always replaces — then run `pnpm run generate:docs-content`.

(`pnpm run generate:tables` is not needed: `getToolTargets` is unchanged. The Tool x Feature e2e happy paths in `src/e2e/e2e-hooks.spec.ts` are correctly left intact.)

### 5. [mid] Project-scope opt-in leaks into global scope

`src/features/hooks/claudecode-hooks.ts:153-154` guards on `paths.relativeDirPath === CLAUDECODE_DIR`, but `getSettablePaths` returns `CLAUDECODE_DIR` for *both* scopes (`claudecode-hooks.ts:109-113`, with the comment "both global and project mode use the same paths"). So a `preserveUnowned: true` committed in a project's `.rulesync/hooks.jsonc` also changes how `rulesync generate --global` treats `~/.claude/settings.json`: a hook there that the user removed from the source is retained. That crosses a trust boundary the repository otherwise keeps.

## Non-blocking

6. [mid] `function` is missing from the type-aware identity. `HOOK_TYPES` (`src/types/hooks.ts:32`) includes `function`, but the `switch` at `:245-262` has no case for it, so a Qwen-shaped `{type: "function", name: "..."}` falls through to `default:` and yields `undefined` — feeding straight into finding #2. Adding `case "function": return taggedIdentity("function", handler.name);` fixes it. Also consider making `default:` include the type in the key, so `{type: "command", command: "x"}` and `{type: "function", command: "x"}` cannot collide.

7. [mid] Third-party commands that reference `.rulesync/` are silently deleted. The inverse of finding #1: a hand-written `bash .rulesync/hooks/mine.sh` in `.claude/settings.json` is removed even with `preserveUnowned: true`, with no log line. Calling a script under `.rulesync/` directly is an ordinary usage. The ownership marker in #1 fixes this too; at minimum, log a warning on deletion.

8. [mid] `preserveUnowned` does not round-trip through `import`. `buildImportedHooksConfig` (`src/features/hooks/tool-hooks-converter.ts`) rebuilds `{ version, hooks, [overrideKey] }` and drops the flag; since `rulesync import` rewrites `.rulesync/hooks.jsonc` wholesale, one import silently disables the opt-in and the next `generate` wipes third-party hooks. Moving the key per #3 removes this; if it stays, carry it through and add a round-trip test.

9. [mid] Unsupported targets are silently ignored. `preserveUnowned: true` has no effect on qwencode, kiro-cli, kiro-ide, junie, reasonix, devin, kimi-code, antigravity, etc. Please warn once from `HooksProcessor` when the flag is set but the target does not support it.

10. [mid] A malformed existing file silently discards everything. `preserve-unowned-hook-commands.ts:11-19` swallows parse errors with `catch { return undefined; }`. Codex CLI and Cursor do not go through the shared-config gateway, so a corrupt `.codex/hooks.json` / `.cursor/hooks.json` drops third-party hooks without a word — the opposite of the gateway's `jsoncParseErrors: "error"` fail-closed policy. Warn, or throw while `preserveUnowned` is on.

11. [mid] Prototype-pollution keys can reach Cursor/Codex output. `cloneMatcherGroup` / `cloneHandler` (`:196-203`) use `{...handler}`, and those two writers serialize with raw `JSON.stringify` rather than the sanitizing `parseSharedConfig` path, so a `__proto__`/`constructor` key inside a preserved handler object survives into the emitted file. Please route preserved values through `omitPrototypePollutionKeys` (`src/utils/prototype-pollution.ts`).

12. [low] Merging preserved handlers loses group-level keys. At `:139-144` leftovers are pushed into the generated group with the same matcher, so the existing group's `sequential` (and any other group attribute) is dropped. Either match on the full group shape or push a separate `{ ...group, hooks: clonedLeftovers }`.

13. [low] Non-object existing entries are dropped without a warning (`:226-235`), e.g. a hand-written `hooks: ["echo raw", {...}]`. Route them through `warnSkip`.

14. [low] `matcherKey` (`:277-282`) collapses every non-string matcher to `"other"`, so `matcher: ["a"]` and `matcher: {x: 1}` merge into one group. A stable `JSON.stringify` key would be safer.

15. [low] `mcpToolIdentity` (`:271-275`) ignores `input`, so a third-party `mcp_tool` hook with the same `server`+`tool` but a different `input` is treated as owned and deleted.

16. [low] The plugin exclusion is expressed as a directory-name comparison. It works because `ClaudecodePluginHooks` overrides `getSettablePaths`, but the intent ("fully generated output always replaces") is buried in a path check. Consider `static supportsPreserveUnowned()` overridden to `false` in the plugin subclass, mirroring the existing `isDeletable()` pattern.

17. [low] `claudecode-hooks.ts:150-168` still parses `existingContent` twice — once via `parseExistingHooksValue` (raw `JSON.parse`) and once inside `applySharedConfigPatch` via `parseSharedConfig`. Behavior matches today, but it bypasses the `SHARED_CONFIG_OWNERSHIP` declaration; reading the existing value through the gateway would keep a single source of truth.

18. [low] `warnPreserved` (`:284-298`) logs raw command strings. Please pass them through `quoteForLog` (`src/utils/control-characters.ts`) so control characters in a destination file cannot corrupt terminal output.

19. [low] Test coverage does not pin the claimed behavior: there is no test that removes a source hook whose command lacks `.rulesync/hooks` (would catch #1), no idempotency test running `generate` two or three times with identical input (would catch #2), no `mcp_tool` / `agent` / `function` identity tests, no global-scope test, no malformed-existing-file test, and `src/types/hooks.test.ts` only asserts that `preserveUnowned: true` parses, not that a wrong type is rejected.

## Summary

Findings 1-5 block the merge. The design question in #1 is the one that matters most: preservation needs a real ownership record, because without it opting in permanently disables rulesync's ability to retract a hook, and hooks execute arbitrary shell. #2 is a straightforward bug with a clear fix, and #3-#5 are placement/scope/documentation issues that follow from the same design choice.

Verified clean: no malicious code, no changes to CI workflows, `package.json`, or release configuration, opt-in is a strict `=== true` with no path that writes the flag automatically, and the default (non-opted-in) behavior is unchanged across all three targets and the plugin destination.

Happy to keep reviewing as this evolves — thank you for sticking with it.

dyoshikawa:
Thanks for this — the behavior it fixes is real, and I pushed a follow-up commit on top of your work rather than sending it back, so your commit stays as it is.

What I changed, and why:

**Ownership needs a record, not a heuristic.** As submitted, "owned" meant *the generated set has the same identity, or the command mentions `.rulesync/hooks`*. That makes preservation one-way: once a project opts in, a hook rulesync wrote and the sources later remove can never be retracted, because on the next run it is indistinguishable from a third-party handler — and hooks execute arbitrary shell. Each run now writes what it generated to a `.rulesync-hooks-lock.json` next to the hooks file (`.claude/.rulesync-hooks-lock.json`, etc.), emitted through `getAuxiliaryFiles` so it respects dry-run. An existing handler is replaced when it is generated again, **retracted** when the lock says rulesync wrote it and the sources no longer define it, and preserved otherwise. A missing, malformed or foreign-version lock degrades to preserving everything, so the failure direction is never destructive. The `.rulesync/hooks` substring rule is gone.

**The opt-in moved to `rulesync.jsonc`.** `.rulesync/hooks.jsonc` is a fetchable, committable source file, so a flag there travels with shared hook definitions and flips a destructive-behavior switch in whatever project fetches it. It is now `"preserveUnownedHooks": true` in `rulesync.jsonc` (config-file only, no CLI flag), plumbed through `Config` → `HooksProcessor` → the adapters, and gated on a new `ToolHooks.supportsPreserveUnowned()` so the Claude Code plugin bundle — which rulesync owns outright — always replaces, instead of relying on a `relativeDirPath` check.

Smaller fixes folded in:

- A handler with no recognizable identity used to hash to `undefined` and get appended again on every run; identity is now always a string, falling back to a stable structural hash.
- `function` hooks take part in identity (they were in `HOOK_TYPES` but not in the identity switch), and an `mcp_tool`'s `input` is part of it too, so two calls to the same server/tool with different arguments are no longer collapsed.
- Non-string matchers get a stable key instead of being compared by `String(matcher)`.
- Non-array event values and non-object handlers now warn instead of being dropped silently.
- A wholly preserved matcher group keeps its group-level keys.

Documented in `docs/guide/configuration.md` and `docs/reference/file-formats.md`, including the one caveat: the first run after opting in has no lock yet, so a hook rulesync wrote before is kept once and retracted from the run after that.

Merging once CI is green. Thanks again for finding and fixing this.

dyoshikawa:
@RaviTharuma Thank you!

PR #2944, opened 2026-09-05
feat(hooks): support ZCode hooks for global scope

gdm257:
Generate configuration-file hooks into `~/.zcode/cli/config.json` (global) from `.rulesync/hooks.jsonc`, and import them back.

A pre-existing `hooks.enabled: false` is preserved as the user's kill switch; otherwise `enabled: true` is stated. ZCode's native process hook type has no canonical equivalent, so those entries are skipped with a warning on import.


dyoshikawa (changes requested):
Thanks for this — the implementation is careful and idiomatic, and it lines up with how the sibling tool adapters in this repo are written. Before the findings, what I verified as clean:

- All six CI checks pass. No changes to CI workflows, `package.json`, lockfiles, or release configuration.
- The generated artifacts are honest: re-running `pnpm run generate:tables`, `generate:docs-content`, `generate:schema`, and `pnpm dev gitignore` against the PR head produces byte-identical `README.md`, `docs/reference/supported-tools.md`, `docs/reference/file-formats.md`, `src/generated/docs-content.ts`, and `.gitignore`. No drift, and nothing smuggled into the large generated file.
- The ownership declaration `hooks: { kind: "replace-owned-keys", ownedKeys: ["hooks"] }` matches how the sibling tools declare theirs, and deletion semantics are actually correct: an entry removed from `.rulesync/hooks.jsonc` disappears from `config.json` on regenerate, while `timeoutMs` / `maxOutputBytes` and unrelated top-level keys survive. No append-only hole.
- Reads and writes go through `parseSharedConfig` / `applySharedConfigPatch`, so prototype-pollution keys are stripped in both directions, and `invalidRootPolicy: "error"` fails closed on an unparseable `config.json` instead of overwriting the user's primary config.
- `getSettablePaths` returns constant relative paths only, so nothing in `.rulesync/hooks.jsonc` can influence where the file is written — a project-scope run cannot reach `~/.zcode/cli/config.json`.
- The Tool x Feature end-to-end happy paths are extended in all three places (project generate, global generate, import).
- `rules-processor.ts` `additionalConventions` covers commands/subagents/skills only, so no change is needed there, and the gitignore derivation is unchanged because ZCode MCP already emits the same `config.json`.

Unfortunately two findings block the merge, both because ZCode's own hooks documentation contradicts what the adapter assumes. All quotes below are verbatim from https://zcode.z.ai/en/docs/hooks, which I fetched independently to confirm.

## Blocking

### 1. [critical] ZCode ignores project-scope hooks entirely, so `supportsProject: true` emits dead configuration

`src/features/hooks/hooks-processor.ts:749` declares `meta: { supportsProject: true, supportsGlobal: true, supportsImport: true }`.

The docs say:

> Project-level hooks are not executed in the current version. For security reasons, any `hooks` configuration in `<workspace>/.zcode/config.json` or `<workspace>/zcode.json` is ignored as a whole — regardless of `hooks.enabled` — and logged as `config_project_hooks_ignored`. The workspace scope has also been hidden from the settings page. To share hooks across a team, distribute them via a plugin (installed with the plugin, version-controllable in your repo), or have each member configure user-level hooks.

So `rulesync generate --features hooks --targets zcode` in project mode writes a `hooks` block that ZCode reads and discards — the only lasting effect is the `enabled: true` it leaves behind in the user's primary config file. `.claude/rules/feature-change-guidelines.md` asks us to "implement both project and global scope **when the target tool supports them**. Consult the tool's official documentation to decide scope support."

This repository already has the exact precedent, at `src/features/hooks/hooks-processor.ts:640-643`:

```ts
// Junie CLI only runs user-scope hooks (~/.junie/config.json); project
// hooks in .junie/config.json are ignored by default for safety, so
// rulesync treats Junie hooks as global-only.
supportsProject: false,
```

Suggested fix: mirror Junie. Set `supportsProject: false`, pin `ZcodeHooks.getSettablePaths` to the user-scope path the way `JunieHooks.getSettablePaths` (`src/features/hooks/junie-hooks.ts:68-73`) does, re-run `pnpm run generate:tables` so the Hooks column becomes global-only, update the ZCode paragraph in `docs/reference/file-formats.md` (and `pnpm run generate:docs-content`), and move `zcode` out of `hooksGenerateTargets` / the project import case in `src/e2e/e2e-hooks.spec.ts` into the global list.

Note that MCP support for the project scope is unaffected — this is specific to hooks.

### 2. [high] `$ZCODE_PROJECT_DIR` does not exist, so every dot-relative hook command breaks

`src/features/hooks/zcode-hooks.ts:102` sets `projectDirVar: "$ZCODE_PROJECT_DIR"`, which produces:

```json
"command": "\"$ZCODE_PROJECT_DIR\"/.rulesync/hooks/session-start.sh"
```

The hooks page documents exactly six environment variables — `ZCODE_PLUGIN_ROOT`, `ZCODE_PLUGIN_ID`, `ZCODE_PLUGIN_NAME`, `ZCODE_PLUGIN_DATA`, plus the compatibility aliases `CLAUDE_PLUGIN_ROOT` and `CLAUDE_PLUGIN_DATA` — all of them plugin-scoped. There is no `*_PROJECT_DIR` variable anywhere on the page. And `command` hooks run through a shell:

> `command` — Hands the full string to the system shell; supports `shell`, `async`, `timeout`.

An undefined variable expands to the empty string, so the command actually executed is `/.rulesync/hooks/session-start.sh`, which always fails. Combined with finding 1, there is no project root in the picture at all once generation is global-only.

Suggested fix: drop the `projectDirVar` prefixing for ZCode and emit the path in a form that works from the user scope, then update the assertions in `src/features/hooks/zcode-hooks.test.ts:148-178` and `src/e2e/e2e-hooks.spec.ts` to match. Whatever the final shape, emitting a variable the tool never defines should not ship.

## Non-blocking

3. [mid] `UserPromptSubmit` and `Stop` do not use matcher filtering, but the adapter emits one. `ZCODE_CONVERTER_CONFIG` (`zcode-hooks.ts:98-117`) has no `noMatcherEvents`, and the comments at `src/types/hooks.ts:909-911`, `hooks-processor.ts:744-745`, and `docs/reference/file-formats.md:179` all claim every event is matcher-capable. The docs say:

   > `SessionStart` matches `source`; `UserPromptSubmit` and `Stop` do not use matcher filtering.

   The result is `"Stop": [{ "matcher": "Bash", ... }]`, which ZCode silently ignores — the user believes they narrowed the hook, and it fires on every stop. The shared converter already has the hook for this: `noMatcherEvents: new Set(["UserPromptSubmit", "Stop"])`, as `KimiCodeHooks` does at `src/features/hooks/kimi-code-hooks.ts:101`. Please also correct the three comments and the docs sentence, and add a test.

4. [mid] The unit tests do not pin the deletion semantics. None of the cases in `src/features/hooks/zcode-hooks.test.ts` start from an existing `hooks.events` entry and assert it disappears (`:86-116` uses an existing block with no `events`). The behavior is correct today, but nothing would catch a future switch to `deep-merge` or a mistake in the ownership declaration. Please add a case that seeds `events.PostToolUse`, generates only a different event, and asserts `PostToolUse` is gone.

5. [mid] Writing `hooks.enabled: true` when there are no events at all lifts ZCode's default-off safety posture for nothing. With an existing `{"model": "..."}` and a source that yields no ZCode-supported events, the output is `{"model": "...", "hooks": {"enabled": true, "events": {}}}`. Nothing runs yet, but the switch stays flipped, so a hook the user later adds by hand runs without them having deliberately enabled config-file hooks. Consider only touching `enabled` when `events` is non-empty:

   ```ts
   const hasEvents = Object.keys(events).length > 0;
   // ...
   ...(hasEvents ? { enabled: existingHooks.enabled !== false } : {}),
   ```

6. [low] `enabled: existingHooks.enabled !== false` (`zcode-hooks.ts:218`) preserves only a boolean `false`. An existing `"false"`, `0`, or `null` is rewritten to `true`, which contradicts the PR's own statement that an explicit off switch survives regeneration. `enabled: "enabled" in existingHooks ? existingHooks.enabled : true` keeps whatever the user wrote.

7. [low] The `wildcardMatcherMeansAll` rationale is factually wrong. `zcode-hooks.ts:114-116` says a bare `*` "is a syntax error that silently never matches", but the docs say "Missing, empty string, or `*`: matches everything" — `*` is an explicit match-all, not a regex. The emitted output happens to be equivalent, so this is a comment fix only; the same wording also appears at `src/types/hooks.ts:909` and `docs/reference/file-formats.md:179`.

8. [low] `async` and per-hook `enabled` are silently dropped. The docs list `shell`, `async`, and `timeout` for `command` hooks, and "A single hook may set `enabled: false`; the runtime genuinely skips it". The PR passes `shell` and `statusMessage` through and `timeout` arrives via the shared converter, but a canonical `async: true` disappears with no warning. Either add `booleanPassthroughFields: [{ canonical: "async", tool: "async", commandOnly: true }]` or state the omission in the adapter comment.

9. [low] `{ canonical: "shell", tool: "shell" }` (`zcode-hooks.ts:110-113`) lacks the `commandOnly: true` that the equivalent Claude Code declaration carries (`src/features/hooks/claudecode-hooks.ts:70`). Behavior is identical today because only `command` hooks are emitted, but the divergence is worth closing.

10. [low] The process-hook skip warning is not covered. `src/features/hooks/zcode-hooks.test.ts:344-371` calls `toRulesyncHooks()` without a logger, so the `logger?.warn(...)` at `zcode-hooks.ts:83-87` never executes — the test pins the skip but not the warning the PR description promises. Passing a mock logger and asserting the message would close that.

11. [low] `stripProcessHooks` (`zcode-hooks.ts:69-93`) assigns into a plain object with `result[eventName] = ...`. It is unreachable today because `this.json` has already been through `sanitizeSharedConfigValue`, but `Object.create(null)` or an `isPrototypePollutionKey` skip would make the defense local rather than inherited.

## Summary

Findings 1 and 2 block. Both are external-fact issues rather than code-quality ones, and both are fixable without redesigning the adapter: restrict hooks to the global scope the way Junie already is, and stop emitting a project-directory variable ZCode does not define. Everything else on the list is a follow-up.

The rest of the work here is genuinely good — the ownership declaration, the fail-closed parse policy, the `process`-hook import guard with its reasoning, the constants file with per-path citations, and the three-way e2e coverage are all exactly what this codebase wants. Thanks for the care, and I'm happy to re-review once the scope question is settled.

gdm257:
Thanks for reviewing. Pushed fix commits covering the findings:

1. ZCode hooks are now global-only, writing `~/.zcode/cli/config.json`; the project `.zcode/config.json` target and its e2e coverage were removed.
2. Dropped the `$ZCODE_PROJECT_DIR` prefix which comes from legacy zcode docs.
3. Matchers on `UserPromptSubmit`/`Stop` are now dropped with a warning.
4. Added a deletion test asserting an event removed from the source disappears from the output.
5. `hooks.enabled` is stated only when events are written.
6. Any existing `enabled` value, including `"false"`, `0`, or `null`, is preserved as written.
7. The `wildcardMatcherMeansAll` comment now states `*` is an explicit match-all.
8. per-hook async round-trips.
9. Shell is scoped to command hooks.
10. Added a test asserting the process-hook skip warning via a mock logger.
11. The strip accumulator uses `Object.create(null)`.

src/features/hooks/zcode-hooks.ts
```diff
@@ -0,0 +1,293 @@
...
+  // Only canonical `command` hooks are emitted. ZCode's native `process` type
+  // (an argv run without a shell) has no canonical equivalent: its `args` *is*
+  // the command line, whereas the canonical `args` field means extra argv
+  // appended by the runner, so mapping between them would change what runs.
+  // Process hooks are skipped with a warning on import instead.
+  supportedHookTypes: new Set(["command"]),
+  // ZCode hook objects support `async` (run in the background), which maps onto
+  // the canonical `async` field. A per-hook `enabled` has no canonical
+  // equivalent and is dropped on round-trip.
+  booleanPassthroughFields: [{ canonical: "async", tool: "async" }],
```
dyoshikawa: This is the one blocker. "A per-hook `enabled` has no canonical equivalent" isn't accurate — it does, and it was added for exactly this failure mode. From `src/types/hooks.ts`:

```ts
// Kiro IDE: whether the hook is active. Kiro writes `enabled` on every hook
// entry and defaults it to `true`; without a canonical home, importing a
// deliberately disabled hook and regenerating would silently switch it back
// on. Only the Kiro IDE emitter reads and writes it.
enabled: z.optional(z.boolean()),
```

ZCode's semantics are identical — a hook may set `enabled: false` and the runtime genuinely skips it, with omission meaning active — so the same reasoning applies.

Concretely: say `~/.zcode/cli/config.json` holds `PreToolUse: [{ matcher: "Bash", hooks: [{ type: "command", command: "./scripts/auto-format.sh", enabled: false }] }]`, disabled on purpose. `rulesync import --global --targets zcode --features hooks` writes it into `.rulesync/hooks.jsonc` with no `enabled`, because the import path only carries fields declared in the converter config. The next `rulesync generate --global --targets zcode` re-emits it without `enabled`, i.e. active, and ZCode starts running it before every Bash call. Nothing warns.

This is the same loss you correctly refused to accept for `process` hooks ("importing its executable alone would change what it runs"), except here it runs more rather than less.

It isn't a one-liner: `booleanPassthroughFields.canonical` is a closed union (`"failClosed" | "async" | "once" | "asyncRewake" | "continueOnBlock"` in `tool-hooks-converter.ts`), so either widen it to include `enabled`, or handle it bespoke the way Kiro IDE does — `enabled: def.enabled ?? true` on emit in `kiro-ide-hooks.ts`, and `if (entry.enabled === false) def.enabled = false` on import, which deliberately carries only the explicit `false` so the default doesn't add noise to every imported definition. Either way please add a test asserting `enabled: false` survives import → generate.

dyoshikawa (commented):
Thanks for the quick turnaround — I re-checked both blockers from the last round and they're properly settled. `supportsProject: false` with `getSettablePaths` pinned to the user path mirrors Junie exactly, the regenerated tables show Hooks as global-only, the e2e case moved to the global list, and `projectDirVar` is now empty so commands are emitted verbatim. I also verified the generated artifacts are honest again at this head: `generate:tables`, `generate:docs-content`, `generate:schema` and `dev gitignore` all produce zero drift, and all six CI checks pass. Nice catch on `ZCODE_NO_MATCHER_EVENTS` too — you used the canonical names (`beforeSubmitPrompt`/`stop`), which is right; my suggested `["UserPromptSubmit", "Stop"]` would have been a no-op, since the converter tests `noMatcherEvents` against the canonical event name before mapping. Items 4 through 11 all check out as described.

One new finding blocks, and it comes out of item 8 rather than the items you fixed. When I wrote "state the omission in the adapter comment" I was talking about `async`; I should have been clearer that per-hook `enabled` is a different matter. It has a canonical home already, added for precisely this hazard, so dropping it silently reactivates a hook the user deliberately turned off. Details inline.

The rest is follow-up material and I'm happy to take it separately rather than hold the PR: the matcher-semantics comment, the override-block warning, the two untested passthrough branches, and a handful of low security notes (rulesync flipping ZCode's default-off kill switch, `events` being replaced wholesale in a global config, and `stripProcessHooks` keying only on an explicit `type: "process"`). A security pass found no malicious code, no new dependencies, no path traversal in the home-directory write, and no command injection — `projectDirVar: ""` means nothing is interpolated into a shell string. I also confirmed every entry in the regenerated `src/generated/docs-content.ts` byte-matches the corresponding `docs/**/*.md`, so nothing is smuggled into that blob.

Fix the `enabled` round-trip with a test and I'll merge.

src/features/hooks/zcode-hooks.ts
```diff
@@ -0,0 +1,293 @@
...
+  stringPassthroughFields: [
+    { canonical: "statusMessage", tool: "statusMessage" },
+    { canonical: "shell", tool: "shell", commandOnly: true },
```
dyoshikawa, line 127: Minor, and fine as a follow-up: `async`, `shell` and `statusMessage` are all newly declared capabilities here and I checked they're real ZCode hook-object fields, but none of them is exercised in either direction in `zcode-hooks.test.ts`. A regression in the passthrough wiring would go unnoticed. Everything genuinely risky is covered well — `enabled: false`, a non-boolean `enabled`, the no-events case, the stale-event-key replacement and the unparseable root all have tests.

src/types/hooks.ts
```diff
@@ -900,6 +900,48 @@ export const KIMI_CODE_TO_CANONICAL_EVENT_NAMES: Record<string, string> = Object
...
 );
 
+/**
+ * Hook events supported by ZCode.
+ *
+ * ZCode's configuration-file hooks expose exactly seven PascalCase events, all
+ * of which have a clean canonical equivalent: `SessionStart`, `PreToolUse`,
+ * `PostToolUse`, `PostToolUseFailure`, `PermissionRequest`, `Stop`, and
+ * `UserPromptSubmit` ← `beforeSubmitPrompt`. An optional matcher (a
+ * case-sensitive regular expression) is honored on all of them except
```
dyoshikawa, line 911: Comment only, no behavior change: "a case-sensitive regular expression" is narrower than what ZCode implements. A matcher made up solely of letters, digits, underscores and `|` (e.g. `Write|Edit`) is an exact name-list match, and only other characters make it a JS regex. Rulesync passes matchers through verbatim so nothing is wrong today, but the wording will mislead whoever next has to decide whether a canonical matcher needs escaping.

src/features/hooks/hooks-processor.ts
```diff
@@ -733,6 +735,21 @@ export const toolHooksFactories = new Map<HooksProcessorToolTarget, ToolHooksFac
...
+      // ZCode hooks live under the `hooks` key of its user config file,
+      // `~/.zcode/cli/config.json`, with the event map nested under
+      // `hooks.events`. ZCode never executes workspace config hooks — the
+      // workspace file is ignored regardless of `hooks.enabled` — so rulesync
+      // treats ZCode hooks as global-only.
+      class: ZcodeHooks,
+      meta: { supportsProject: false, supportsGlobal: true, supportsImport: true },
+      supportedEvents: ZCODE_HOOK_EVENTS,
+      supportedHookTypes: ["command"],
+      supportsMatcher: true,
```
dyoshikawa, line 754: Worth a deliberate decision rather than a change request: `passthroughOverrideEvents` isn't set here, so `unsupportedEventNames` checks `effectiveHooks`, which includes override-block keys. Someone who writes `zcode.hooks.SessionStart` — ZCode's own spelling, and the spelling `buildImportedHooksConfig` falls back to for any event without a canonical mapping — gets `Skipped hook event(s) for zcode (not supported): SessionStart`, while `canonicalToToolHooks` emits it anyway via `lookupOwn(...) ?? eventName`. So the warning claims a drop that didn't happen. Low impact, since all seven ZCode events map cleanly and this only bites hand-authored native keys, and it matches the junie/devin/grokcli precedent rather than diverging from it.

dyoshikawa:
@gdm257 Thank you! Nice work tracking down the `PER_HOOK_ENABLED_TARGETS` warning gate as well — that counterpart wasn't in my review and it would have produced a spurious warning for zcode. Follow-ups from the review are tracked in #2946.

Issue #2960, opened 2026-09-08
Follow up ZCode upstream updates: skill `when_to_use`/`license`/`metadata` dropped, and no `zcode-plugin` packaging target


dyoshikawa:
## Summary

Re-check of ZCode (`--targets zcode`) upstream documentation on **2026-09-08**.

Two conclusions:

1. **Hooks are up to date.** Nothing in ZCode's documentation postdates the research window of #2944 (merged 2026-09-08). The hooks page still documents exactly the seven events rulesync ships, the same executors, the same matcher rules and the same "project-level hooks are not executed" restriction. The eight review follow-ups already tracked in #2946 are not repeated here.
2. **Two material capability gaps remain**, both outside #2946's scope: ZCode skills are generated lossily, and there is no `zcode-plugin` packaging target — which is upstream's *only* documented way to put ZCode hooks or ZCode subagents in a repository.

## Recent Releases

ZCode publishes no changelog or release-notes page (`https://zcode.z.ai/en/docs/changelog`, `/release-notes` and `/settings` all return 404), so there is no release feed to diff. Version evidence available in the docs themselves:

| Signal | Version | Source |
| --- | --- | --- |
| Current download build | **v3.11.2 (Latest)** | [install](https://zcode.z.ai/en/docs/install) |
| Windows close-to-tray default | since v3.4.0 | [install](https://zcode.z.ai/en/docs/install) |
| Memory feature | requires v3.6.4+ | [memory](https://zcode.z.ai/en/docs/memory) |
| Subagents inject AGENTS.md by default | since v3.7.1 | [subagents](https://zcode.z.ai/en/docs/subagents) |

Doc pages read in full on 2026-09-08: `hooks`, `plugin`, `skill`, `subagents`, `commands`, `mcp-services`, `agents`, `memory`, `configuration`, `safety-confirm`, `install`, `welcome`.

[...]

PR #2961, opened 2026-09-08
feat(skills): model ZCode's documented when_to_use/license/metadata fields

dyoshikawa:
## Summary

ZCode's [Skill `SKILL.md` Field Reference](https://zcode.z.ai/en/docs/plugin) allowlists exactly five frontmatter keys: required `name` and `description`, plus optional `when_to_use`, `license` and `metadata`. Rulesync emitted only `name`/`description`, so:

- a canonical skill carrying `license` or `metadata` had both silently dropped for `zcode`, while every other Agent Skills target received them, and
- `when_to_use` — ZCode-specific trigger-timing context — could not be authored at all.

This PR closes that gap (gap 1 of the issue).

## Changes

- `src/features/skills/zcode-skill.ts`: add `when_to_use`, `license` and `metadata` to `ZcodeSkillFrontmatterSchema`; resolve `license`/`metadata` through the shared `resolveLicense`/`resolveMetadata` helpers so they fall back to the root-level Agent Skills packaging defaults, and carry all three back into a `zcode` section on import. `compatibility` stays unwritten because ZCode documents no such field.
- `src/features/skills/rulesync-skill.ts`: add the optional `zcode` section to `RulesyncSkillFrontmatterSchemaInternal` and `RulesyncSkillFrontmatterInput`.
- `docs/reference/file-formats.md` (and the regenerated `src/generated/docs-content.ts`): document `zcode` in the shared packaging-field lists and rewrite the ZCode skills note.
- `src/features/skills/zcode-skill.test.ts`: two new tests — root-level `license`/`metadata` plus a `zcode`-section `when_to_use` are emitted while `compatibility` is dropped and the frontmatter round-trips; and the `zcode` section overrides the root values.

`pnpm cicheck` passes.

## Scope note

Refs #2960

The issue tracks two gaps. This PR resolves gap 1 only. Gap 2 — adding a `zcode-plugin` packaging target — is a product/design decision (a new output target, its directory layout, and its place in the Tool × Feature matrix) rather than a mechanical follow-up, so the issue stays open for a maintainer to decide. Hence `Refs` rather than `Closes`.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2958, opened 2026-09-08
Follow up Vibe Code upstream updates: Agent Plugins packages under .vibe/plugins/ are unreachable, and .vibe/plugin-data/ is not gitignored


dyoshikawa:
## Summary

Vibe v2.24.4/v2.24.5 shipped a **plugin package system** that did not exist at v2.24.3, the version [#2770](https://github.com/dyoshikawa/rulesync/issues/2770) surveyed. A plugin is a directory under `<project>/.vibe/plugins/<name>/` (project, committable) or `~/.vibe/plugins/<name>/` (global) that bundles skills, MCP servers, hooks, subagents and a knowledge surface behind one namespaced manifest. It is **not** behind an experimental flag: `PluginResolver.from_harness_files(...).resolve()` runs on every session open.

Rulesync writes none of it, and `.vibe/plugin-data/` — the repo-local materialization directory the loader derives for project plugins — is not gitignored.

The three gaps recorded in #2770 are all fixed and are not re-raised here: nested `AGENTS.md` is now emitted via `VibeRule.getNestedFilePatterns()` (`src/features/rules/vibe-rule.ts:64`), the managed-shell fanout is `VIBE_SHELL_ALIAS_TOOL_NAMES` (`src/features/permissions/vibe-permissions.ts:92`), and per-MCP-tool tables are translated to `<server>_<tool>` (`src/features/permissions/vibe-permissions.ts:1072`).

Version surveyed: **v2.25.0** (released 2026-09-04, changelog dated 2026-09-03), repo HEAD `6c79ef0e1ee484d7069bc38590d5917d3914cd48`.

## Recent Releases

Everything since #2770's survey point (v2.24.3, 2026-08-20):

- **v2.24.4 (2026-08-26)** — LLM-generated session titles, queue edit mode, git info in the session header; `/mcp login <connector>` routing fix. The `vibe/core/plugins/` tree first appears in this window: `GET repos/mistralai/mistral-vibe/contents/vibe/core/plugins?ref=v2.24.3` is 404, `?ref=v2.24.5` returns 20 files.
- **v2.24.5 (2026-08-27)** — fuzzy search in the MCP/connector browser, lazy-import startup work, `--worktree`/`--trust` trust-prompt fix.
[...]

Issue #2729, opened 2026-08-24
Follow up Devin Desktop upstream updates: no devin-plugin target for Agent Plugins 1.0.0 plugin bundles


dyoshikawa:
## Summary

Devin CLI **v3000.5.20 (2026-08-21)** added "Plugin compatibility with Agent Plugins 1.0.0 format", and the plugin docs page has grown from a one-line mention into a full specification: manifest precedence, a documented field set including dependency keys, and a six-directory bundle layout. rulesync has no `devin-plugin` target, so this authorable in-repo surface is `unsupported`. Everything else in the window is in parity, and #2688 is fully shipped. No Devin issue is currently open, so this is filed as a new one.

## Recent Releases

- **CLI v3000.5.20 — 2026-08-21** — https://docs.devin.ai/cli/changelog/stable. The only CLI release after #2688's coverage window (which ended at v3000.4.25, 2026-08-13). Notable: *"Plugin compatibility with Agent Plugins 1.0.0 format"*, `devin rm`, `devin desktop`, `devin doctor` subagent-profile validation, `/recap`, `/rename`, `/handoff` OS selection, Plan-mode plan files, plus fixes for `SessionEnd` hook execution, `Stop` hooks receiving `last_assistant_message`, permission-rule glob metacharacters, and hooks running in the correct directory for secondary workspaces.
- CLI v3000.4.25 (2026-08-13), v3000.4.16 (2026-08-10), v3000.3.27 (2026-08-01) — already covered by #2688.
- **Desktop v3.8.20 — 2026-08-21** — https://docs.devin.ai/desktop/changelog. Agent Command Center space-following, a separate-window location setting, Windows CLI / Linux packaging. Nothing config-file bearing. Earlier: v3.7.25 (2026-08-13), v3.7.16 (2026-08-10), v3.6.27 (2026-08-01).

### `plugins` (new target) — there is no `devin-plugin` target for Devin plugin bundles

**Upstream.** https://docs.devin.ai/cli/extensibility/plugins documents an authorable, in-repo plugin bundle. Manifest resolution order is `.devin-plugin/plugin.json` → `.claude-plugin/plugin.json` → root `plugin.json` (Agent Plugins 1.0.0). Only `name` is required; the documented fields are `version`, `description`, `author{name,email}`, `homepage`, `repository`, `license`, `keywords`, plus `requiredPlugins` / `optionalPlugins` / `forbiddenPlugins` and a `skills` path override. The documented layout:

```
[...]

Issue #2666, opened 2026-08-13
Follow up Google Antigravity upstream updates: sidecars (sidecar.json) unsupported, CLI skill disable-slash-command lossy, plugin.json manifest never emitted


dyoshikawa:
## Summary

Three Antigravity surfaces are unfollowed. Antigravity 2.0 introduced **sidecars** (`sidecar.json`), a persistent-process surface with no rulesync dimension; the CLI added a `disable-slash-command` skill frontmatter flag that is lossy in both directions; and the `antigravity-plugin` target never emits the `plugin.json` manifest that upstream *requires* for a directory to be recognized as a plugin at all.

Existing coverage note: #2509 tracks a separate `antigravity-cli` permissions item. This issue is scoped to `antigravity-ide` and `antigravity-plugin`, plus the CLI skills flag; see the correction to #2509's premise at the end.

## Recent Releases

- **Antigravity 2.0, current version v2.8.0** (2026-08-12) — the Customization nav now carries a Sidecars page. — https://antigravity.google/docs/sidecars
- **Antigravity CLI v1.1.12** (2026-08-11) — [changelog](https://antigravity.google/changelog), verbatim: "Added a `disable-slash-command: true` flag for a skill's `SKILL.md` frontmatter, which hides that skill from the `/` menu".
- **Antigravity CLI v1.0.11** (2026-06-24) and **v1.0.5** (2026-06-03) — [changelog](https://antigravity.google/changelog) — CLI permission merging and the project-level permission location; relevant to the #2509 correction below.

### 1. New dimension — Sidecars (`antigravity-ide`, and the plugin variant)

- **Upstream**: a `sidecar.json` per directory, discovered at `~/.gemini/config/sidecars/<id>/` and `~/.gemini/config/plugins/<pluginName>/sidecars/<id>/`. Fields: `command` or `builtin` (mutually exclusive; `builtin` currently supports `schedule`), `args`, `restart_policy` (`always` / `on-failure` / `never`), `description`, `env`, `display_name`. A sidecar is disabled unless enabled per ID in `~/.gemini/config/config.json` — `{"sidecars": {"<id>": {"enabled": true, "projectId": "..."}}}` — and the plugin form is keyed `"<plugin>/<sidecar>"`. — https://antigravity.google/docs/sidecars
- **rulesync**: `unsupported` — no sidecar feature exists.
[...]

Issue #2502, opened 2026-07-30
Follow up Kimi Code upstream updates: kimi.plugin.json plugin target (0.31.0 adds agents + dual-engine systemPrompt; only command surface)


dyoshikawa:
## Summary

Kimi Code 0.31.0 (2026-07-30) widened the `kimi.plugin.json` plugin manifest to contribute `agents` and `systemPrompt`/`systemPromptPath` on **both** agent engines, alongside the existing `skills`/`commands`/`mcpServers`/`hooks`. A plugin target would now cover rules, commands, subagents, skills, mcp and hooks in one artifact — and the manifest remains the *only* command surface Kimi Code has, so `commands` stays unreachable for the plain `kimi-code` target.

Baseline: #2425 (closed) was verified at 0.29.1; its gaps were implemented in #2438, and it deliberately deferred the plugin manifest as "a new target rather than a gap in the existing one". 0.31.0 changes that calculus.

## Recent Releases

- **0.31.0** (2026-07-30) — plugin-contributed **agents** (an `agents/` directory or explicit paths, auto-discovered and delegable as sub-agents, [PR #2365](https://github.com/MoonshotAI/kimi-code/pull/2365)); plugin **`systemPrompt` / `systemPromptPath`** (inline or file-referenced instructions, max 32 KB, "effective on both agent engines", [PR #2314](https://github.com/MoonshotAI/kimi-code/pull/2314)); Markdown custom agents on agent-core; `/secondary_model`. — [releases](https://github.com/MoonshotAI/kimi-code/releases)
- **0.30.0** (2026-07-29) — `[status_line]` in `tui.toml` (TUI presentation; no rulesync feature dimension).
- **0.29.2** (2026-07-27) — no config-surface change.

### commands / rules / subagents / skills / mcp / hooks — `kimi.plugin.json` manifest surface unmapped

- **Upstream**: as of 0.31.0, a plugin manifest contributes `skills`, `commands`, `mcpServers`, `hooks`, `agents`, and `systemPrompt`/`systemPromptPath`. — [plugins doc](https://moonshotai.github.io/kimi-code/en/customization/plugins.html)
- **rulesync**: the `kimi-code` target writes only the plain `.kimi-code/` and `~/.kimi-code/` trees; `commands` is reported unsupported and there is no way to produce one (the plugin manifest is Kimi Code's only command surface). Nothing under `src/` emits `kimi.plugin.json`.
[...]

PR #2963, opened 2026-09-08
chore(gitignore): ignore Vibe Code plugin-data

dyoshikawa:
## Summary

Resolves gap 2 of #2958 — `.vibe/plugin-data/` is a repo-local by-product with no gitignore entry.

## Changes

A Vibe plugin installed under the project's `.vibe/plugins/<name>/` derives its data root as `root.parent.parent / "plugin-data"` whenever no `data_root_base` override is supplied ([`_native.py:1009`](https://github.com/mistralai/mistral-vibe/blob/v2.25.0/vibe/core/plugins/_native.py)), and `PluginResolver.from_harness_files` — the path Vibe runs on every session open ([`_plugins.py:176`](https://github.com/mistralai/mistral-vibe/blob/v2.25.0/vibe/app_server/_plugins.py)) — supplies none. So a project plugin materializes into `<project>/.vibe/plugin-data/<plugin>/`, holding generated knowledge documents (`knowledge/<name>/KNOWLEDGE.md`) and vendored `libraries/` dependencies (`node_modules`, Python packages).

Rulesync never writes that tree, so per `.claude/rules/feature-change-guidelines.md` it belongs in `HAND_MAINTAINED_GITIGNORE_ENTRIES` rather than being derived from `getSettablePaths` — the same treatment `**/.muse/worktrees/` got.

- `src/cli/commands/gitignore-entries.ts`: new `{ target: "vibe", feature: "general", entry: "**/.vibe/plugin-data/" }` entry, with a comment recording the derivation and the primary source.
- `src/cli/commands/gitignore-entries.test.ts`: the entry added to the `justified` allowlist that the "every hand-maintained tool entry is an explicitly justified exception" test enforces.
- `.gitignore`: regenerated with `pnpm dev gitignore`.

Only the `plugin-data` directory is ignored, not `.vibe/` as a whole — `.vibe/hooks.toml`, `.vibe/agents/` and the rest stay committed project config.

`pnpm cicheck` passes.

## Scope note

Refs #2958

The issue tracks two gaps. This PR resolves gap 2 only, which the issue itself lists as landable "independent of the plugin decision".

**Gap 1** — a `.vibe/plugins/<name>/` Agent Plugins package output root — is left open because it is a product decision: it would add a new packaging target with its own manifest writer and matrix row, and `KNOWLEDGE.md` has no canonical home at all today. As the issue notes, it is the same open question as #2666 (Antigravity `plugin.json`), #2502 (Kimi), #2729 (Devin) and #2960 (ZCode), so it is likely worth settling once across targets rather than adding another one-off. That call is for a maintainer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -245,6 +245,7 @@ rulesync.local.jsonc
 **/.takt/.cache/
 **/.takt/config.yaml
 **/.muse/worktrees/
+**/.vibe/plugin-data/
 **/.augment-guidelines
 **/.devin/workflows/
 **/.junie/memories/
--- a/src/cli/commands/gitignore-entries.test.ts
+++ b/src/cli/commands/gitignore-entries.test.ts
@@ -171,6 +171,10 @@ describe("registry derivation", () => {
       // Muse Code's `--subagent-worktree-isolation` worktrees, which the
       // runtime removes only when they are clean (issue #2727).
       "musecode::general::**/.muse/worktrees/",
+      // Vibe Code's plugin data root: knowledge documents and vendored
+      // `libraries/` dependencies that a project plugin under `.vibe/plugins/`
+      // materializes at session open (issue #2958).
+      "vibe::general::**/.vibe/plugin-data/",
       // Legacy/aggregate/ghost outputs not produced via getSettablePaths.
       "augmentcode::rules::**/.augment-guidelines",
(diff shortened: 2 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2957, opened 2026-09-08
Follow up Factory Droid upstream updates: the .factory/output-styles channel and five missing settings.json keys


dyoshikawa:
## Summary

Continuation of #2765, which surveyed the droid CLI through v0.203.0 (2026-08-25) and whose four gaps are all implemented (`.factory/skills/review-guidelines/SKILL.md`, the legacy hooks fallback, the `settings.local.json` overlay, and `modelPolicy` / `missionPolicy`). Five releases have landed since, up to **v0.209.0 (2026-09-01)**. None of them introduces a config key on its own, but two documented surfaces rulesync does not reach came into view:

- **Output styles** — `.factory/output-styles/<name>.md`, a committed, frontmatter-carrying, system-prompt-shaping instruction surface with four discovery scopes. rulesync emits nothing there; `grep -rn "output-styles\|outputStyle" src` matches one unrelated Claude Code line.
- **Five `settings.json` keys** documented as direct siblings of keys rulesync already carries, missing from `FACTORYDROID_OVERRIDE_KEYS`, so they are silently dropped on import and unauthorable on generate.

Hooks, skills, commands, MCP and checks were all re-verified as fully modeled. Re-check on 2026-09-08.

## Recent Releases

| Version  | Date       | Content |
| -------- | ---------- | ------- |
| v0.209.0 | 2026-09-01 | `/archive` and Ctrl+X session archiving; GPT-5.6 Sol as the default model; faster startup with organization settings; Droid Core model suggestion |
| v0.208.1 | 2026-08-29 | Temporary GPT/Grok pricing discount |
| v0.208.0 | 2026-08-29 | Sharper image reading; verified domain invitations; MCP registry additions; built-in slash commands run immediately; hooks / MCP / session bug fixes; Windows plugin hooks preserve command quoting |
| v0.206.0 | 2026-08-28 | Cross-repository project guidance; faster CLI startup |
| v0.204.0 | 2026-08-25 | `droid doctor` diagnostics; MCP connection reliability |

### 1. rules: the `.factory/output-styles/` channel is unsupported

Per the output-styles documentation:

- Path `<scope>/output-styles/<name>.md`, discovered at four scopes — User `~/.factory/output-styles/`, **Project** `<repo>/.factory/output-styles/`, Folder `<repo>/<project-area>/.factory/output-styles/`, and Plugin `<plugin>/output-styles/`.
- "Files must use the `.md` extension"; nested files and other file types are ignored.
[...]

Issue #2765, opened 2026-08-25
Follow up Factory Droid upstream updates: review-guidelines checks surface, legacy hooks layout, settings.local.json overrides


dyoshikawa:
## Summary

Factory Droid ships three repository-authored surfaces that rulesync does not follow: the `.factory/skills/review-guidelines/SKILL.md` review policy (rulesync's `checks` dimension is `unsupported` for `factorydroid`), the legacy `.factory/hooks/hooks.json` layout that upstream still loads but rulesync never reads on import, and `.factory/settings.local.json` permission overrides that never round-trip. The first is a missing feature; the other two cause rulesync to silently drop configuration a user already has.

Research window: Factory's docs plus the full changelog through **v0.203.0 (2026-08-25)**. Both dry runs (`--targets factorydroid --features "*" --dry-run` and the `--global` variant) complete cleanly today — project writes `AGENTS.md`, `.factory/rules/*.md`, `.factory/mcp.json`, `.factory/droids/*`, `.factory/skills/**` and `.factory/hooks.json`; global writes `.factory/AGENTS.md` plus the same tool directories under the home root.

## Recent Releases

- **v0.203.0 (2026-08-25)** — current release at the time of this survey ([changelog](https://docs.factory.ai/changelog/release-notes)).
- **v0.185.0 (2026-07-30)** — MCP `connectTimeout` ([MCP configuration](https://docs.factory.ai/cli/configuration/mcp)).
- **v0.181.0 (2026-07-27)** — MCP `oauth` and `resource` keys.
- **v0.147.0 (2026-06-12)** — `commandBlocklist` in settings ([settings](https://docs.factory.ai/droid-cli/settings)).
- **v0.143.0 (2026-06-08)** — hook `suppressOutput`, which is a hook **stdout response** field rather than a config field, so its absence from rulesync's config schema is correct.
- **v0.142.0 (2026-06-05)** — GitLab CI support for the review pipeline; **v0.136.0 (2026-05-28)** `/deep-review`; **v0.126.0 (2026-05-14)** `/security-review` bundled by default. These establish that the review-droid product line behind gap 1 is current, not legacy.

### 1. checks — the `review-guidelines` repository review policy is unsupported

[...]

Issue #2209, opened 2026-07-10
Follow up Factory Droid upstream updates: autonomy keys mcpAutonomyOverrides + subagentAutonomyLevel in permissions override


dyoshikawa:
## Summary

Factory Droid added two newer autonomy keys to `settings.json` — `mcpAutonomyOverrides` (per-individual-MCP-tool autonomy) and `subagentAutonomyLevel` — that rulesync's `factorydroid` permissions override does not lift, so they are silently dropped on the canonical round-trip. This extends the tool-scoped override namespace established in #2136 (which covered `commandBlocklist`/`networkPolicy`/`sandbox`, now implemented).

## Recent Releases

- **`mcpAutonomyOverrides`** — "Per-tool MCP autonomy overrides", introduced **v0.109.3 (2026-04-26)**: [Factory changelog](https://docs.factory.ai/changelog/release-notes).
- **`subagentAutonomyLevel`** — subagent authorization level (`inherit`/`off`/`low`/`medium`/`high`), documented in [settings configuration](https://docs.factory.ai/cli/configuration/settings).

## Gaps

| Feature | Factory Droid upstream (current) | rulesync side |
| --- | --- | --- |
| permissions | `settings.json` autonomy keys `mcpAutonomyOverrides` (v0.109.3) and `subagentAutonomyLevel` | `project`/`global` override namespace, but these two keys are **not** in `FACTORYDROID_OVERRIDE_KEYS` (`src/features/permissions/factorydroid-permissions.ts:41-50`, which lists `commandBlocklist`, `networkPolicy`, `sandbox`, `mcpPolicy`, `enableDroidShield`, `sessionDefaultSettings`, `maxAutonomyLevel`, `interactionMode`). On export they survive only if already in the user's `settings.json` (via `...settings` spread); on import (`toRulesyncPermissions`, `factorydroid-permissions.ts:199-207`) they are not lifted → silently dropped from the canonical round-trip. |
| hooks (minor) | Droid migrated hooks from `.factory/hooks/hooks.json` to `.factory/hooks.json`, archiving the old file ([hooks reference](https://docs.factory.ai/reference/hooks-reference)) | Import fallback chain is `.factory/hooks.json` → `.factory/settings.json` only (`factorydroid-hooks.ts:63-71`); a not-yet-migrated project's legacy `.factory/hooks/hooks.json` is not imported. Low priority — Factory auto-migrates on next save. |

## Proposed Follow-up

[...]

PR #2964, opened 2026-09-08
feat(factorydroid): carry five documented settings keys, refresh the research map

dyoshikawa:
## Summary

Resolves gaps 2 and 3 of #2957 — the five missing `settings.json` keys, and the stale Factory Droid research reference map.

### `src/constants/factorydroid-settings-keys.ts` — five documented keys added

`FACTORYDROID_OVERRIDE_KEYS` carried one half of three documented pairs, so the other half was silently dropped on import and unauthorable on generate:

| Added key | Documented as | Sibling already carried |
| --- | --- | --- |
| `mcpAutonomyUrlOverrides` | "Autonomy overrides matched by URL. Each entry has a `urlPattern` and a `defaultLevel`" | `mcpAutonomyOverrides` |
| `builtInToolAutonomyOverrides` | "Per-tool autonomy level required to run a built-in network tool without a confirmation prompt" | `subagentAutonomyLevel` |
| `allowManagedHooksOnly` | "When true, only org-managed hooks and hooks from org-enabled plugins are loaded" | `hooksDisabled` |
| `strictEnabledPlugins` | "When true, the `enabledPlugins` maps at this level and higher levels become the exhaustive plugin allowlist" | `enabledPlugins` |
| `strictKnownMarketplaces` | "Marketplace sources that are strictly enforced. A higher-level list is the ceiling" | `extraKnownMarketplaces` |

Every spelling was verified verbatim against the raw text of [`/enterprise/hierarchical-settings-and-org-control`](https://docs.factory.ai/enterprise/hierarchical-settings-and-org-control); `mcpAutonomyUrlOverrides` and `allowManagedHooksOnly` are also on [`/droid-cli/settings`](https://docs.factory.ai/droid-cli/settings).

These are primarily org-delivered, but Factory's hierarchy uses the same schema at all four authored levels, so what matters is that a hand-written project or user `settings.json` round-trips instead of losing them — the same standing as #2765 gap 4 and #2209. `FactorydroidPermissionsOverrideSchema` is a `z.looseObject` and does not enumerate the existing autonomy/policy keys either, so no schema change is needed for consistency. `showHookOutput` was left out, per the issue: it is a display preference, not a security or autonomy control.

The file's own `@see` links were updated from the retired `/cli/configuration/*` paths to the canonical ones.

### `.rulesync/skills/rulesync-feature-research/references/factorydroid.md` — rewritten

The map was stale in four ways the issue names, all fixed:

- **Every URL was a retired path.** `/cli/configuration/*` and `/reference/hooks-reference` now 308-redirect; each was replaced with the target it resolves to (`/harness/agents-md`, `/harness/mcp`, `/harness/custom-slash-commands`, `/harness/subagents`, `/harness/skills`, `/harness/hooks`, `/droid-cli/settings`, `/droid-cli/quickstart`), and the header records the move so a future run does not re-derive it. Note `/harness/output-styles` is a 404 — the live page is `/droid-cli/output-styles`.
- **No `checks` row** even though the feature ships. Added: `.factory/skills/review-guidelines/SKILL.md` via [`/software-factory/code-review-ci`](https://docs.factory.ai/software-factory/code-review-ci), matching `FACTORYDROID_REVIEW_GUIDELINES_DIR_PATH`.
- **No row for output styles or the settings hierarchy.** Both added, each followed by a paragraph saying explicitly that it is unsupported and why, so the missing row is not misread as a missing upstream surface. A `plugins` row was added on the same principle, noting that the *consumption* half (`extraKnownMarketplaces` / `enabledPlugins`) is authorable while the bundle half is not.
- **The `rules` anchor predated the `DESIGN.md` channel.** It now names the `factorydroid.channel: design` route.

The nine hook events, the `.factory/` paths at both scopes, the commands frontmatter restriction and the skills layout were all re-verified against the live pages rather than carried over.

`pnpm cicheck` passes (`check:sync-skill-docs` confirms `.claude/skills/` is regenerated in step).

## Scope note

Refs #2957

**Gap 1 — the `.factory/output-styles/` rules channel — is left open.** It is a design decision rather than a mechanical follow-up: unlike `design`, which routes to one fixed file, an output-style channel is directory-shaped, so the `factorydroid.channel` frontmatter value would have to carry a style name, and the surface has a home-directory equivalent so it needs both scopes. That is a change to the shape of a canonical frontmatter key, and it is a maintainer's call. The map now carries a row for it so the surface stays visible.

**Gap 3's context note — the folder-level `.factory/` scope — is also left open**, as the issue itself frames it: rulesync has no per-subdirectory output root at all, which is a rulesync-wide architectural question rather than a factorydroid adapter bug. The map header now records it so a future run does not re-raise it as a gap.

### Side finding, not addressed here

Fourteen other files under `src/` and `docs/` still cite the retired `docs.factory.ai/cli/configuration/*` paths (they redirect, so nothing is broken). Sweeping them is a mechanical chore outside this issue's scope — it would also pull in `docs/reference/file-formats.md` and the regenerated `src/generated/docs-content.ts`.

### DIFF
--- a/.rulesync/skills/rulesync-feature-research/references/factorydroid.md
+++ b/.rulesync/skills/rulesync-feature-research/references/factorydroid.md
@@ -1,28 +1,63 @@
 # Factory Droid Map
 
+The `droid` CLI keeps everything under one `.factory/` tree at two authored
+scopes — the project `<repo>/.factory/` and the user `~/.factory/`. Factory's
+documentation moved in 2026: the old `/cli/configuration/*` and
+`/reference/hooks-reference` paths now 308-redirect to `/harness/*` and
+`/droid-cli/*`, so cite the canonical URLs below rather than the redirecting
+ones. Two upstream surfaces are worth keeping in view when reading the table:
+Factory documents **seven** settings levels (Org, Runtime, Folder, Project,
+User, Dynamic, BuiltIn) of which four are authored, and the Folder level is a
+per-subtree `.factory/` inside a monorepo carrying the full set of files —
+Rulesync writes one project root and one home root and has no per-subdirectory
+output root, so that level is out of reach by design rather than by oversight.
+
 ## Official Docs
 
-| Feature       | Official docs                                                     | Upstream surface                                                                                                |
-| ------------- | ----------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
-| index         | `https://docs.factory.ai/cli/getting-started/quickstart`          | Factory Droid CLI documentation                                                                                 |
-| `rules`       | `https://docs.factory.ai/cli/configuration/agents-md`             | `AGENTS.md`, nested AGENTS.md, personal `~/.factory/AGENTS.md`                                                  |
-| `ignore`      | No dedicated upstream ignore surface in map                       | No Rulesync-supported Factory Droid ignore target in map                                                        |
-| `mcp`         | `https://docs.factory.ai/cli/configuration/mcp`                   | `.factory/mcp.json`, `~/.factory/mcp.json`, stdio and HTTP servers                                              |
-| `commands`    | `https://docs.factory.ai/cli/configuration/custom-slash-commands` | `.factory/commands` and `~/.factory/commands`, `description`/`argument-hint` frontmatter only (no tool scoping) |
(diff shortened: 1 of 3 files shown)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2956, opened 2026-09-08
Follow up deepagents-cli upstream updates: [extensions] trust keys are unauthorable and DEEPAGENTS_HOME relocates every global path


dyoshikawa:
## Summary

Continuation of #2764, which surveyed `deepagents-code` through 0.1.61 (2026-08-24) and whose four gaps are all implemented. Five releases have landed since, up to **0.1.66 (2026-09-03)**, and two of them change surfaces rulesync writes.

- **0.1.65 (2026-08-28)** added a Python **extension** system with a three-key `[extensions]` config block, including `trust`, the gate on executing repository-authored Python. rulesync's `deepagents.*` permissions override namespace accepts only `startup`, so the block is unauthorable.
- **0.1.62 (2026-08-26)** added **`DEEPAGENTS_HOME`**, which relocates the entire user profile root. Every global path rulesync emits is hardcoded under the OS home directory, so on a machine that sets the variable a `--global` generate writes a complete tree that dcode never reads — the same failure mode as gap 1 of #2764, reintroduced through a different mechanism.

Everything else was re-verified as already correct: the 12 hook events, subagent and skill frontmatter, MCP transports and tool filters, and the absence of commands / ignore / checks surfaces. Re-check on 2026-09-08.

## Recent Releases

| Version | Date       | Relevant content |
| ------- | ---------- | ---------------- |
| 0.1.66  | 2026-09-03 | General-purpose subagents default to fork mode (#6024); conversations bound to a recorded workspace (#5946) |
| 0.1.65  | 2026-08-28 | **Python extension support** — discovery, configuration and loading, runtime hosting, trust and inspection UX (#5631, #5632, #5633, #5634); `/summarization-model` |
| 0.1.64  | 2026-08-28 | Trace metadata for skill invocations; MCP servers-requiring-login list |
| 0.1.63  | 2026-08-26 | Managed config loadable from a remote source (#5776); retry middleware |
| 0.1.62  | 2026-08-26 | **Added support for `DEEPAGENTS_HOME`** (#5773); `/context-doctor`; CLI config provider support (#5774) |

[...]

PR #2965, opened 2026-09-08
feat(deepagents): author the [extensions] trust gate through the permissions override

dyoshikawa:
## Summary

Resolves gap 1 of #2956 — dcode's `[extensions]` trust gate, which no `.rulesync/permissions.jsonc` could reach — plus the reference-map refresh from its follow-up 3.

### `deepagents.extensions` override block

`DeepagentsPermissionsOverrideSchema` accepted exactly `permission` and `startup`, so the three-key `[extensions]` table added in deepagents-code **0.1.65 (2026-08-28)** was unauthorable. Two of those keys are policy-load-bearing and now round-trip:

| Key | Values | Upstream default | What it decides |
| --- | --- | --- | --- |
| `enabled` | boolean | `true` | Whether extension discovery runs at all |
| `trust` | `ask` / `always` / `never` | `ask` | The fallback applied to **project** extension code when no per-project grant exists |

Verified against primary source, not the issue text: [`config_manifest.py`](https://github.com/langchain-ai/deepagents/blob/main/libs/code/deepagents_code/config_manifest.py) carries `toml_keys=("extensions", "enabled")` / `("extensions", "trust")` with defaults `True` and `"ask"`, and [`extensions/settings.py`](https://github.com/langchain-ai/deepagents/blob/main/libs/code/deepagents_code/extensions/settings.py) defines `TrustPolicy` as exactly `ask` / `always` / `never` and reads the section from `resolver.toml_snapshot(USER_RANK)` — global only, matching the rest of the adapter. [`extensions/discovery.py`](https://github.com/langchain-ai/deepagents/blob/main/libs/code/deepagents_code/extensions/discovery.py) auto-loads `*.py` from `project_root / ".deepagents" / "extensions"` as well as the user directory, which is what makes `trust` a security control rather than a preference: it is the same class of knob as `[startup].mode`, which this override already carries, and it ships defaulting to a prompt a user can accept.

`extra_paths` is **not** modeled, as the issue proposed: it names machine-local files and directories, which do not belong in a committed permissions file, and it only ever widens what loads. It is not lifted on import; a hand-written one still passes through on generate like any other unknown key, named in a warning.

Two behaviors worth calling out:

- **`trust` is read the way upstream reads it.** `parse_trust_policy` trims and lowercases, so `" Always "` is a policy dcode *is applying*; importing it verbatim would write a permissions file the next generate could not parse (the canonical enum holds only lowercase spellings), and dropping it would lose a live setting. It is normalized instead — the same courtesy the `all` / `recommended` allowlist sentinels already get.
- **Relaxations are named.** This block lands in the machine's **global** config from a `.rulesync/permissions.jsonc` a repository can carry, so `trust = "always"` (project Python imported with no prompt at all) and an `enabled` that was explicitly `false` being turned back on are warned about by name, alongside the value they replace. `never` and `false` restrict, so neither says anything; `enabled = true` over an unset key is a no-op upstream and says nothing either.

### Shared merge/lift

`mergeStartupOverride` / `liftStartupOverride` were table-specific. They now delegate to a `mergeOverrideTable` / `liftOverrideTable` pair, with the non-table guard and the unchecked-key warning generalized alongside. Every existing warning string is unchanged for `[startup]`, so the existing tests assert the same text they did before.

### `.rulesync/skills/rulesync-feature-research/references/deepagents.md` — refreshed

Follow-up 3 of the issue, minus the part already done (`Client Anchors` gained its `permissions` row in an earlier release):

- The `hooks` row claimed the pre-v2 flat list "still loads until 2026-09-01". That date has passed; `hooks/legacy.py`, `hooks/migration.py` and the compatibility `argv` field in `hooks/models/config.py` are all still on `main` (checked today), so the row now says the importer is still needed and to re-check before dropping it.
- The `permissions` row named only `[startup].mode`; it now names all three startup keys and the new `[extensions]` gate.
- A header paragraph records `DEEPAGENTS_HOME` and an `extensions` row records the discovery surface, each followed by a note saying what rulesync does and does not cover — so a missing row is never misread as a missing upstream surface.
- The `paths` anchor row is new and says explicitly that `deepagents-paths.ts` holds plain literals with no `DEEPAGENTS_HOME` awareness.

`pnpm cicheck` passes (411 files, 11115 tests).

## Scope note

Refs #2956

**Gap 2 — `DEEPAGENTS_HOME` relocating every global path — is left open.** [`_paths.py`](https://github.com/langchain-ai/deepagents/blob/main/libs/code/deepagents_code/_paths.py) confirms the issue's description exactly: `DEEPAGENTS_HOME_ENV = "DEEPAGENTS_HOME"` is "the variable that selects the user profile and trust root", it is captured once at import time before any dotenv loader runs, re-exported to every child, and `DEFAULT_PROFILE_DIR_NAME = ".deepagents"` applies only "when no profile is configured". So on a machine that sets it, a `--global` generate does write a tree dcode never reads.

The issue's own proposal is a warning rather than a new output-root mechanism, and that is the right shape — but there is no single place to put it. The global paths are spread across six adapters (`deepagents-rule.ts`, `-mcp.ts`, `-skill.ts`, `-subagent.ts`, `-hooks.ts`, `-permissions.ts`), rulesync has no per-target global-root hook to hang a detector on, and the issue names `[agents].default` as the same problem arriving through a config key rather than an env var — which a warning keyed on `process.env` would not cover. Where that check lives, and whether it generalizes past deepagents, is a maintainer's call rather than a mechanical follow-up. The map now records the behavior so a future research run does not re-derive it.


dyoshikawa:
@dyoshikawa Thank you!

PR #2967, opened 2026-09-08
fix(rovodev): carry `enable_instructions: false` in both directions

### LINKED ISSUE #2955: Follow up Rovo Dev upstream updates: enable_instructions default inverted upstream, so MCP instruction suppression is unauthorable and dropped on import
## Summary

Atlassian republished [Connect to an MCP server in Rovo Dev CLI](https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/) on **2026-09-02** and **inverted the default of `enable_instructions`**. The key that decides whether a third-party MCP server's own initialization instructions are pasted into the agent's system prompt changed from opt-**in** (absent/`false` = ignored as untrusted) to opt-**out** (absent/`true` = surfaced; `false` = suppressed).

Rulesync models the old, opt-in semantics end to end. `src/features/mcp/rovodev-mcp.ts` writes the key only when the value is literally `true`, and lifts it back only when it reads literally `true`. Under the new default that makes rulesync fail-open on the one MCP key whose entire purpose is a trust decision:

- an authored `rovodevEnableInstructions: false` writes nothing, so Rovo Dev surfaces the instructions anyway — silently, with no warning;
- a hand-authored `"enable_instructions": false` in an existing `mcp.json` is **dropped on import** and never re-emitted, so an import → generate round-trip re-enables instruction injection the user had explicitly switched off;
- the one warning the adapter does emit (`writing enable_instructions: true …`) now flags the default state, while the state that actually changes behavior is written silently — or rather, not at all.

This is the same key already handled once in #2767 (closed 2026-08-26); that work was correct against the wording live at the time and is what the doc change invalidates. No other Rovo Dev docs page changed since that pass, so this is the only material gap in this window.

## Recent Releases

Atlassian publishes no changelog for the Rovo Dev CLI. As in #2767, "recent" is grounded in each page's `lastPublished` timestamp from the embedded Contentful metadata in the page HTML. Enumerating the whole CLI docs collection at https://support.atlassian.com/rovo/docs/use-rovo-dev-cli/ on 2026-09-08:

| Page | `lastPublished` | Changed since #2767 (2026-08-25)? |
| --- | --- | --- |
| [Connect to an MCP server in Rovo Dev CLI](https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/) | **2026-09-02** | **Yes** — the `enable_instructions` inversion below |
| [Set custom instructions for code reviews](https://support.atlassian.com/rovo/docs/set-custom-instructions-for-code-reviews/) | 2026-05-29 | No |
| [Extend Rovo Dev CLI with Agent Skills](https://support.atlassian.com/rovo/docs/extend-rovo-dev-cli-with-agent-skills/) | 2026-03-11 | No |
| Use Rovo Dev CLI / Install and run / CLI commands / Manage settings / Manage sessions / Use memory / Save and reuse a prompt / Use tools / Use subagents / Use worktree mode | 2026-04-10 | No |
| Switch between LLMs | 2026-01-07 | No |
| Use server mode / Get help | 2025-10-07 | No |

https://developer.atlassian.com/cloud/acli/changelog/ still has no entry after **2026-03-25** (v1.3.15-stable, an OAuth re-authorization announcement), and nothing Rovo-Dev-config-related.

Dry runs on 2026-09-08 (`pnpm run dev generate --targets rovodev --features "*" [--global] --dry-run`): project → 55 files (6 rules + 2 MCP + 5 subagents + 42 skills); global → 50 files (1 rule + 2 MCP + 5 subagents + 42 skills). `ignore` is the only feature reported as unsupported. Both scopes now write the `mcp.mcpConfigPath` pointer into `config.yml`, which closes gap 2 of #2767 — and the 2026-09-02 page corroborates the value rulesync writes, showing `mcpConfigPath: "~/.rovodev/mcp.json"` in its own `config.yml` example.

### mcp — `enable_instructions` is now default-on upstream; rulesync cannot express the suppression and destroys an existing one on import (`project` + `global`)

**Upstream, 2026-09-02** ([MCP page](https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/), section "MCP server instructions"):

> By default, Rovo Dev CLI surfaces a user-configured server's instructions into the agent's system prompt.
>
> To prevent a specific server's instructions from being passed to the agent, set `"enable_instructions": false` on that server's entry in `mcp.json` […]
>
> When the field is absent or set to `true`, the server's instructions are surfaced to Rovo. You can disable this for MCP servers that are untrustworthy or misbehaving.

**Upstream, previous wording** (same URL, Wayback snapshot [2026-06-15](http://web.archive.org/web/20260615211217/https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/), the version #2767 was written against):

> By default, Rovo Dev CLI does not surface a user-configured server's instructions into the agent's system prompt. Because these instructions are arbitrary text controlled by a third-party server, they are treated as untrusted and ignored unless you explicitly opt in.
>
> To allow a specific server's instructions to be passed to the agent, set `"enable_instructions": true` […]
>
> When the field is absent or set to `false`, the server's instructions are not surfaced. […] Built-in, first-party Atlassian servers (such as Atlassian, Bitbucket, Compass, Rovo Dev, and Scout) are trusted and have their instructions enabled automatically.

A diff of the two snapshots shows the change is confined to that section: the default sentence is inverted, the example flips from `true` to `false`, the "only enable this for servers you trust" caution and the first-party auto-trust paragraph are deleted, and the note becomes "When the field is absent or set to `true`, the server's instructions are surfaced." Everything else on the page — `~/.rovodev/mcp.json`, the `transport` key with `stdio` / `http` / `sse`, `mcp.mcpConfigPath`, `mcp.disabledMcpServers` — is byte-identical.

**rulesync**: `project` + `global`, but the value is modeled as a boolean that only exists when `true`:

- `src/types/mcp.ts:103` — `rovodevEnableInstructions: z.optional(z.boolean())` (the canonical field added by #2767).
- `src/features/mcp/rovodev-mcp.ts:113-121` (`readEnableInstructions`) — returns `false` for anything that is not literally `true`, with the doc comment "matching Rovo Dev, where the key is absent by default".
- `src/features/mcp/rovodev-mcp.ts:908` — `...(readEnableInstructions(rawServer) && { rovodevEnableInstructions: true })`, so an authored `false` is collapsed away before it ever reaches the writer.
- `src/features/mcp/rovodev-mcp.ts:150-157` (`toRovodevServer`) — `if (rovodevEnableInstructions === true) { rest.enable_instructions = true; }`, with the comment "Only `true` is written: absent and `false` mean the same thing to Rovo Dev". That premise is what changed.
- `src/features/mcp/rovodev-mcp.ts:177-185` (`fromRovodevServer`) — destructures `enable_instructions` out of the entry and re-adds the canonical key only when it is `true`; a `false` is discarded, so it is neither preserved in `.rulesync/mcp.json` nor written back on the next generate.
- `src/features/mcp/rovodev-mcp.ts:130-139` (`warnEnabledInstructions`) — warns when writing `true`, described as "the only thing generate writes that widens what steers the model". Under the new default, writing `true` widens nothing (it restates the default) and the genuinely security-relevant write — `false` — cannot happen at all.

User-visible consequences, both silent:

1. **Suppression is unauthorable.** There is no value of `rovodevEnableInstructions` (nor of the raw `enable_instructions` spelling, which line 908 normalizes through the same true-only gate) that makes rulesync emit `"enable_instructions": false`. A user who wants a third-party server's instructions kept out of the prompt cannot get that from rulesync, and gets no diagnostic saying so.
2. **Import destroys an existing suppression.** `rulesync import --targets rovodev --features mcp` reads an `mcp.json` containing `"enable_instructions": false`, drops it, and the next `generate` writes the entry back without it — flipping the server from suppressed to surfaced. This is the fail-open outcome the `readEnableInstructions` doc comment explicitly set out to avoid ("fail-open, on the one key whose whole purpose is a trust decision"); the inversion moved which direction is fail-open.

Note that the cross-target leak half of #2767's gap 1 stays fixed either way: `RulesyncMcp.getMcpServers()` strips `rovodevEnableInstructions` (`src/features/mcp/rulesync-mcp.ts:637`), so nothing here re-exposes the key to other targets.

## Proposed Follow-up

1. Make `rovodevEnableInstructions` a real tri-state in the adapter: `true` → write `enable_instructions: true`, `false` → write `enable_instructions: false`, `undefined` → write nothing (matching Rovo Dev's default). `src/types/mcp.ts` already types it as an optional boolean, so only the true-only gates at `rovodev-mcp.ts:908` and `:155` need to change; `readEnableInstructions` becomes a `boolean | undefined` read, keeping the canonical-key-wins precedence it already has over the raw spelling.
2. Carry `false` through import: `fromRovodevServer` should lift any boolean, not only `true`, so `.rulesync/mcp.json` round-trips a suppression instead of erasing it.
3. Invert the warning. Warning on `enable_instructions: true` is now noise (it restates the upstream default). If a diagnostic is still wanted on the trust dimension, it belongs on the *absence* of a suppression for a non-first-party server — or the warning can simply be dropped and replaced by a docs line, since under the new semantics rulesync's write no longer widens anything.
4. Update the prose that encodes the old semantics: the doc comments at `rovodev-mcp.ts:99-110` and `:151-154`, and the `rovodevEnableInstructions` entry in the MCP reference docs, all currently tell the reader that absent and `false` are equivalent.
5. Because the page rewrote itself once already, pin the claim: cite the 2026-09-02 `lastPublished` alongside the behavior in the docs comment, so the next re-check can tell a doc correction from a runtime change. Worth confirming empirically against a live `acli rovodev` install before shipping, since only Atlassian's prose distinguishes "the default changed" from "the docs were wrong before".

- **hooks** — unchanged and still blocked, exactly as recorded in #1986 and re-confirmed in #2767. `/hooks` is still listed on the [commands page](https://support.atlassian.com/rovo/docs/rovo-dev-cli-commands/) (2026-04-10) as "Manage configurations for hook events", the [settings reference](https://support.atlassian.com/rovo/docs/manage-rovo-dev-cli-settings/) (2026-04-10) still documents no `eventHooks` section (zero occurrences of the string in the page), and eight plausible hooks slugs under `/rovo/docs/` all return 404. The only descriptions remain Atlassian's blog posts — [Streamlining your Rovo Dev CLI workflow with event hooks](https://www.atlassian.com/blog/development/streamline-rovo-dev-cli-with-event-hooks) and [Enhancing Developer Workflow with Rovo Dev and Notifications](https://www.atlassian.com/blog/development/developer-workflow-rovo-dev-notifications) (2026-04-15), neither of which publishes the config schema. Keep `hooks` blank; capture the shape empirically (`/hooks help`, `acli rovodev config`) rather than waiting on docs.
- **`mcpConfigPath` (gap 2 of #2767)** — resolved. Both scopes now write the pointer (`Rovo Dev MCP: setting mcp.mcpConfigPath to "~/.rovodev/mcp.json" in ~/.rovodev/config.yml` in the global dry run), and the 2026-09-02 page's own `config.yml` example uses that same value.
- **mcp, other keys** — the transport key is still `transport` with `stdio` / `http` / `sse` (rulesync writes `transport` and maps the canonical `type`, skipping `ws` with a warning), and `disabledMcpServers` still lives in `config.yml`. No drift.
- **ignore** — still no `.rovodevignore` or gitignore-respect surface in any page of the collection. Correctly unsupported.
- **rules / commands / subagents / skills / permissions / checks** — no page in these areas has been republished since #2767's pass (all `lastPublished` ≤ 2026-05-29), so its verification of `AGENTS.md` memory roots, `prompts.yml` + `content_file`, `.rovodev/subagents`, the six Agent Skills frontmatter fields plus the `.agents/skills` alternative roots, `toolPermissions`, and the repository-root-only `.rovodev/.review-agent.md` still holds.

## References

- https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/ — `lastPublished` 2026-09-02; the current, inverted `enable_instructions` semantics that define the gap.
- http://web.archive.org/web/20260615211217/https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/ — the 2026-06-15 snapshot of the same page, establishing that the default was inverted rather than misread.
- https://support.atlassian.com/rovo/docs/use-rovo-dev-cli/ — the CLI docs collection index, used to enumerate every page and its `lastPublished` date.
- https://support.atlassian.com/rovo/docs/manage-rovo-dev-cli-settings/ — 2026-04-10; the authoritative `config.yml` section list, still without `eventHooks`.
- https://support.atlassian.com/rovo/docs/rovo-dev-cli-commands/ — 2026-04-10; `/hooks` still listed as a command.
- https://www.atlassian.com/blog/development/streamline-rovo-dev-cli-with-event-hooks and https://www.atlassian.com/blog/development/developer-workflow-rovo-dev-notifications — the only public descriptions of event hooks; neither documents the schema.
- https://developer.atlassian.com/cloud/acli/changelog/ — checked 2026-09-08; nothing after 2026-03-25.
- #2767 — the closed issue whose gap 1 this supersedes and whose gap 2 is confirmed fixed. #1986 — the closed hooks issue whose blocker is re-confirmed.

dyoshikawa:
## Summary

Atlassian inverted the default of Rovo Dev's `enable_instructions` on 2026-09-02. [Connect to an MCP server in Rovo Dev CLI](https://support.atlassian.com/rovo/docs/connect-to-an-mcp-server-in-rovo-dev-cli/) now reads: *when the field is absent or set to `true`, the server's instructions are surfaced to Rovo* — and its only example of the key is `"enable_instructions": false`. The paragraph that used to describe first-party servers as auto-trusted is gone.

The adapter was written against the earlier opt-in wording, where absent and `false` meant the same thing, so it collapsed the flag to a plain boolean and only ever wrote it when true. Under the new default that is a fail-open bug on the one key whose purpose is a trust decision.

Closes #2955

## The gap, concretely

| Direction | Before | After |
| --- | --- | --- |
| generate | `rovodevEnableInstructions: false` wrote nothing, so Rovo Dev surfaced the instructions anyway — the suppression was impossible to author | `enable_instructions: false` is written out |
| import | `enable_instructions: false` in an existing `mcp.json` was dropped on the floor | lifted onto `rovodevEnableInstructions: false` |
| round trip | import → generate silently switched a suppression back on | the `false` survives both hops |

## Changes

- `readEnableInstructions` returns `boolean | undefined` instead of `boolean`. `undefined` means the entry says nothing, which is Rovo Dev's own default.
- `toRovodevServer` and `fromRovodevServer` both write the key whenever the value is a boolean, not only when it is `true`.
- The `fromRulesyncMcp` normalization carries the tri-state onto the canonical key.
- The canonical key still wins whenever present, matching how `codexcli-mcp.ts` resolves the same two-spelling conflict for `experimental_environment`. OR-ing them would let a stale raw `enable_instructions: true` copied out of the docs beat an explicit `rovodevEnableInstructions: false`.
- A non-boolean under the raw spelling reads as `undefined`, not as a suppression: Rovo Dev surfaces instructions for anything that is not `false`, so writing `false` there would invent a restriction the tool is not applying.
- The `enable_instructions` generate warning is **dropped** rather than inverted (the issue's own second option under proposal 3). It reported the servers whose instructions it *enabled*, which under the new default is the state every keyless server is already in — warning on `true` reports the default, not a widened surface. A comment at the old call site records why nothing replaced it.
- `src/types/mcp.ts` and `docs/reference/file-formats.md` describe the tri-state and cite the revised page.

## Tests

`src/features/mcp/rovodev-mcp.test.ts`: the two generate cases now expect `enable_instructions: false` to be emitted, the import case gains a `suppressed` server, the warning test is replaced by one asserting nothing is logged about the key, and a new round-trip test covers the exact fail-open path #2955 reports.

`pnpm cicheck` passes (411 files, 11116 tests).


dyoshikawa:
@dyoshikawa Thank you!

Issue #2953, opened 2026-09-08
Follow up CodeBuddy Code upstream updates: rules frontmatter drift (paths scalar, enabled, alwaysApply default) and the six unimplemented config surfaces


dyoshikawa:
## Summary

`codebuddy` shipped on 2026-09-03 (#2918, closing #2760) as a **rules-only** target, deliberately scoped down from the seven-feature proposal in #2760. Two things have moved since that PR was written:

1. Three concrete defects in the rules adapter, all grounded in CodeBuddy's current [Memory documentation](https://www.codebuddy.ai/docs/cli/memory): the documented scalar form of `paths` is rejected on import, the documented `enabled` field is silently dropped, and CodeBuddy's `alwaysApply` **defaults to `true`** (unlike Cursor), so rulesync's path-scoped rules are emitted as always-on rules.
2. `ignore` aside, every other rulesync feature dimension still has a documented CodeBuddy file surface that rulesync does not emit. #2760 is closed, so nothing tracks that remainder.

Dry-run baseline (2026-09-08, `pnpm run dev generate --targets codebuddy --features "*" [--global] --dry-run`): project writes `CODEBUDDY.md` + `.codebuddy/rules/*.md`; global writes `~/.codebuddy/CODEBUDDY.md` + `~/.codebuddy/rules/*.md`; `ignore`, `commands`, `skills`, `subagents`, `mcp`, `hooks`, `checks` and `permissions` all report "does not support".

## Recent Releases

- `@tencent-ai/codebuddy-code` `2.147.0`, published 2026-09-07 (npm registry metadata: https://registry.npmjs.org/@tencent-ai/codebuddy-code). #2760 cited `2.138.0` (2026-08-25) and its 2026-08-29 re-validation comment cited `2.141.0`; nine further releases have shipped since.
- The published [Release Notes index](https://www.codebuddy.ai/docs/cli/release-notes/README) lags npm and currently stops at **v2.140.0 (2026-08-26)**. The two entries published after #2760's research window were read in full:
  - [v2.139.0 (2026-08-25)](https://www.codebuddy.ai/docs/cli/release-notes/v2.139.0) — `AskUserForStructuredInput` tool, REPL authorization changes, ZIP-based plugin marketplace installs.
[...]

PR #2918, opened 2026-09-03
feat(rules): add CodeBuddy Code as a rules target

dyoshikawa:
## Summary

Adds CodeBuddy Code (Tencent Cloud's terminal coding agent, `@tencent-ai/codebuddy-code`) as a new `codebuddy` target for the **rules** feature only.

CodeBuddy Code's configuration surface mirrors Claude Code closely:
- A root `CODEBUDDY.md` memory file (project root, also read from `.codebuddy/CODEBUDDY.md`; `~/.codebuddy/CODEBUDDY.md` for global scope).
- Modular rules under `.codebuddy/rules/*.md` (project) / `~/.codebuddy/rules/*.md` (global), auto-loaded without a reference section, with optional `description` / `paths` / `alwaysApply` frontmatter (closer to the Cursor rule model than Claude Code's `paths`-only schema).
- A personal `CODEBUDDY.local.md` local-root file, mirroring Claude Code's `CLAUDE.local.md`.

This follows the repo's established precedent (e.g. Muse Code, ZCode) of rolling out a new tool feature-by-feature across separate PRs rather than implementing the full set (rules, mcp, commands, subagents, skills, hooks, permissions) in one PR. Only `rules` is implemented here, scoped down from the broader multi-feature proposal in the issue per the safety guidance to ship the minimal defensible slice.

### Changes

- `CodebuddyRule` (`src/features/rules/codebuddy-rule.ts`), modeled on `ClaudecodeRule`'s root/non-root structure with Cursor-style `description`/`paths`/`alwaysApply` non-root frontmatter.
- `codebuddy` passthrough block added to the canonical rulesync rule frontmatter schema (tool-specific overrides take precedence over the shared `description`/`globs`, per project convention).
- `codebuddy` registered in the rules tool-target tuple, tool display metadata, and the rules processor's factory map (`ruleDiscoveryMode: "auto"`, `localRootMode: "separate-local-file"`).
- Hand-maintained gitignore entries for `CODEBUDDY.local.md`, mirroring the `CLAUDE.local.md` pattern; project `.gitignore` regenerated via `pnpm dev gitignore`.
- Supported Tools tables (`README.md`, `docs/reference/supported-tools.md`) regenerated via `pnpm run generate:tables`.
- Prose note added to `docs/reference/file-formats.md`; `src/generated/docs-content.ts` re-embedded via `pnpm run generate:docs-content`.
[...]

Issue #2760, opened 2026-08-25
Propose a CodeBuddy Code target: CODEBUDDY.md plus the .codebuddy config tree


dyoshikawa:
## Summary

CodeBuddy Code is Tencent Cloud's terminal coding agent, distributed on npm as `@tencent-ai/codebuddy-code`. Its configuration surface is shaped almost exactly like Claude Code — a memory file, a `.codebuddy/` tree, settings-based permissions and hooks, subagents, skills and slash commands — so it maps onto rulesync's feature dimensions with very little new modelling. This issue proposes adding it as a target (`--targets codebuddy`).

## Recent Releases

- `@tencent-ai/codebuddy-code` latest version `2.138.0`, published 2026-08-25 (npm registry metadata: https://registry.npmjs.org/@tencent-ai/codebuddy-code).
- First published 2025-08-05; roughly two major lines shipped in a year.
- 473,426 npm downloads in the 30 days from 2026-07-26 to 2026-08-24 — an order of magnitude above the other candidates reviewed in the same discovery pass.
- Backed by Tencent Cloud, with a full documentation site at https://www.codebuddy.ai/docs/cli/reference and a source home at https://cnb.cool/codebuddy/codebuddy-code.

## Configuration Surface

| rulesync feature | CodeBuddy Code path(s) | Format |
| --- | --- | --- |
| rules | project: `CODEBUDDY.md`, `CODEBUDDY.local.md`, `.codebuddy/rules/*.md`; global: `~/.codebuddy/CODEBUDDY.md`, `~/.codebuddy/rules/` | Markdown with YAML frontmatter (`description`, `paths` glob with `matchBase` enabled, `alwaysApply`) — close to the Cursor rule model |
| ignore | not documented; no `.codebuddyignore` appears in the CLI reference | would stay `unsupported` |
| mcp | project: `<root>/.mcp.json` (recommended), `<root>/mcp.json` (deprecated); user: `~/.codebuddy/.mcp.json`, `~/.codebuddy/mcp.json` (deprecated), `~/.codebuddy.json` (legacy); local scope nested under `~/.codebuddy.json` at `projects/<workspace_path>` | JSON `mcpServers` map with `type: stdio \| sse \| http`, `command`/`args`/`env`/`url`/`headers`. Precedence local > project > user, with no intra-scope merging |
[...]

PR #2970, opened 2026-09-08
fix(codebuddy): follow the documented rules frontmatter contract

dyoshikawa:
## Summary

Fixes gaps 1–3 of #2953 — the three concrete defects in the `codebuddy` rules adapter, all verified against CodeBuddy's current [Memory documentation](https://www.codebuddy.ai/docs/cli/memory).

Gap 4 of that issue (the six documented config surfaces that are still `unsupported`) is a scoping decision for the maintainer, not a defect, so this PR deliberately does **not** close the issue. See "What this PR does not do" below.

Refs #2953

### 1. The documented scalar `paths` form is accepted

`paths` is documented as `string`/`string[]`, and every example in the Rule Control Fields section uses the bare-string form:

```markdown
---
alwaysApply: false
paths: src/api/**/*.ts
---
```

`CodebuddyRuleFrontmatterSchema` declared `z.optional(z.array(z.string()))`, so a rule file copied straight out of CodeBuddy's own docs failed to import with a Zod `expected: "array"` error. The schema now takes both shapes, and a new `normalizeCodebuddyPaths` helper collapses the scalar to a single-element list for everything downstream. The canonical `codebuddy:` block accepts both too.

### 2. `enabled` is modeled, so a disabled rule stays disabled

`enabled` (boolean, default `true`, "when set to `false`, the rule is not loaded at all") was absent from the schema, and `z.object` stripped it — an `import` → `generate` cycle silently switched a deliberately disabled rule back on. It is now part of both the tool frontmatter and the `codebuddy:` passthrough block.

### 3. `alwaysApply` follows the documented default of `true`

This is the substantive one. CodeBuddy's Rule Type Determination table reads:

| `alwaysApply` | `paths` | Rule type |
| --- | --- | --- |
| `true` (default) | any | ALWAYS — always injected into context |
| `false` | has value | MANUAL — only triggered on matching files |
| `false` | none | not supported; the rule will not be loaded |

The default is the **opposite** of Cursor's, which this adapter was modeled on. Rulesync emitted `paths` alone for every path-scoped rule, which CodeBuddy reads as ALWAYS and injects everywhere, ignoring the scoping the user asked for.

`resolveCodebuddyPaths` is replaced by `resolveCodebuddyRuleType`, which resolves the pair against that table:

- scoped `paths` now carry an explicit `alwaysApply: false`, so they actually scope;
- an `alwaysApply: false` with nothing to trigger on is left implicit rather than written as a rule CodeBuddy refuses to load;
- `alwaysApply: true` still drops a redundant universal glob and still keeps a non-universal one;
- on the import side, `isAlways` is now `alwaysApply !== false`, and the `true` default is materialized into the `codebuddy:` block when the file also has `paths` — otherwise the next generate would read those paths as a scope and downgrade an ALWAYS rule to MANUAL.

### 4. Documentation

The CodeBuddy Code note in `docs/reference/file-formats.md` now records the scalar `paths` form, `enabled`, and the default-`true` `alwaysApply` semantics with what rulesync does about them. `src/generated/docs-content.ts` regenerated.

## What this PR does not do

#2953's gap 4 lists six documented CodeBuddy surfaces (`mcp`, `commands`, `subagents`, `skills`, `hooks`, `permissions`) that rulesync does not emit — the remainder of #2760's proposal, deliberately deferred by #2918 in favour of shipping the rules-only slice. Whether to continue that rollout is a maintainer call, so #2953 stays open with its scope narrowed to that decision.

## Tests

Eleven new cases (11126 → 11137):

- constructor: the documented scalar `paths` form is accepted and emitted verbatim; `enabled` is written out; the invalid-frontmatter case now uses a genuinely invalid value.
- `fromFile`: the scalar-`paths` example from the docs parses and maps to `globs`; an `enabled: false` file round-trips.
- `fromRulesyncRule`: scoped globs produce `alwaysApply: false` alongside `paths`; an explicit `alwaysApply: false` with no paths is left implicit; a scalar `codebuddy.paths` is normalized; `codebuddy.enabled` reaches the generated file.
- `toRulesyncRule`: a rule with neither key is always-on (`globs: ["**/*"]`); the `true` default is materialized when `paths` are present; a disabled rule survives the round trip.

Verified end to end on a scratch project. Generate, from a rule with `globs: ["src/api/**/*.ts"]` and no `codebuddy:` block:

```yaml
---
description: API rules
alwaysApply: false
paths:
  - src/api/**/*.ts
---
```

Import of the documented scalar file followed by a regenerate returns the same rule, and `enabled: false` survives both directions.

`pnpm cicheck` passes (411 files, 11137 tests).

## References

- https://www.codebuddy.ai/docs/cli/memory — Rule Control Fields (`enabled`, `alwaysApply`, `paths` typed `string`/`string[]`) and the Rule Type Determination table.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2946, opened 2026-09-08
Follow-ups from the ZCode hooks review (PR #2944)


dyoshikawa:
## Background

PR #2944 added ZCode hooks support (`src/features/hooks/zcode-hooks.ts`, global scope only, writing the `hooks` block of `~/.zcode/cli/config.json`). The review of that PR ran a code-quality pass and a security pass. Two rounds of blocking findings were raised and fixed by the author — project-scope hooks being emitted although ZCode never executes them, a `$ZCODE_PROJECT_DIR` variable ZCode does not define, and the per-hook `enabled: false` flag being lost on import → generate — and the PR was cleared to merge once those were resolved.

Eight non-blocking findings were deliberately not held against the merge. This issue records them so they are not lost. None of them is a correctness bug in the shipped behavior; they are comment inaccuracies, missing test coverage, one misleading warning, and a few design notes about the blast radius of writing into a global user config.

### 1. Matcher-semantics comment is narrower than ZCode's actual behavior

`src/types/hooks.ts` (the `ZCODE_HOOK_EVENTS` doc block) describes the matcher as "a case-sensitive regular expression". Per ZCode's docs, a matcher made up solely of letters, digits, underscores and `|` (e.g. `Write|Edit`) is an exact name-list match, and only other characters make it a JS regex. Rulesync passes matchers through verbatim, so nothing is wrong at runtime — but the wording will mislead whoever next has to decide whether a canonical matcher needs escaping. The same phrasing may also appear in `docs/reference/file-formats.md`.

### 2. Native event keys in the `zcode.hooks` override block get a false "not supported" warning

`passthroughOverrideEvents` is not set on the `zcode` factory entry in `src/features/hooks/hooks-processor.ts`, so `unsupportedEventNames` checks `effectiveHooks`, which includes override-block keys. A user who writes `zcode.hooks.SessionStart` — ZCode's own spelling, and the spelling `buildImportedHooksConfig` falls back to for any event without a canonical mapping — gets `Skipped hook event(s) for zcode (not supported): SessionStart`, while `canonicalToToolHooks` emits it anyway via `lookupOwn(...) ?? eventName`. The warning claims a drop that did not happen.

Impact is low: all seven ZCode events map cleanly to canonical names, so this only bites hand-authored native keys. Note this matches the existing junie / devin / grokcli precedent rather than diverging from it, so the fix may belong at the shared level rather than on `zcode` alone.

[...]

PR #2972, opened 2026-09-08
docs(hooks): correct the ZCode matcher semantics and cover its passthrough fields

dyoshikawa:
Refs #2946

Two of the eight follow-ups recorded after the ZCode hooks review — the two that are self-contained and have a defensible answer without a maintainer scoping decision.

## Item 1 — matcher semantics

`ZCODE_HOOK_EVENTS` in `src/types/hooks.ts` described the matcher as "a case-sensitive regular expression". Per ZCode's hooks docs, a matcher made up solely of letters, digits, underscores and `|` is **exact name-list matching** ("Only letters, digits, underscores, and `|`: exact name-list matching, e.g. `Write|Edit`"), and only another character makes the whole value a case-sensitive JavaScript regex.

Nothing is wrong at runtime — Rulesync passes matchers through verbatim — but the wording would mislead whoever next has to decide whether a canonical matcher needs escaping. Corrected in the constant's doc block and in the ZCode sentence of the hooks section of `docs/reference/file-formats.md` (the issue guessed the phrasing might be duplicated there; it was not, so this adds the clarification rather than fixing a second copy).

## Item 3 — untested branches

- `async`, `shell` and `statusMessage` are declared passthrough capabilities of the ZCode converter but were exercised in neither direction. Added a round-trip case that imports all three off a ZCode config and asserts they come back out on regenerate.
- `stripProcessHooks` keeps a matcher group only when a non-`process` hook survives the filter. The surviving-group side was covered; the emptied-group side was not. Added a case with a process-only group beside a command group, asserting the process-only matcher yields no canonical definition and does not regenerate as a hook-less entry that would write a matcher running nothing.

## What this PR does not do

#2946 stays open, with its scope narrowed to the remaining six items. All of them are either shared-layer changes that should not be scoped to `zcode`, or design decisions that are a maintainer's to make:

- **Item 2** (native event keys in the `zcode.hooks` override block get a false "not supported" warning) — the issue itself notes this matches the existing junie / devin / grokcli precedent, so the fix belongs at the shared level or the warning should be documented as intentionally conservative. Either way it is a decision, not a defect.
- **Item 4** (canonical `timeout` lands in ZCode's legacy field) — shared converter behavior; the unit matches and it works, so it is only worth doing as part of a broader look at per-tool timeout field selection.
- **Item 5** (rulesync flips ZCode's default-off hooks kill switch) — whether `--global` generation should ever flip a tool's safety switch implicitly.
- **Item 6** (`hooks.events` replaced wholesale) — declared `replace-owned-keys` ownership semantics; the question is whether a global user config deserves a warning or a diff preview.
- **Item 7** (`stripProcessHooks` keys only on an explicit `type: "process"`) — reachability depends on ZCode's default hook `type`. I re-checked the upstream hooks docs for this PR: they document `process` and `command` as alternatives and state no default, so the contributor's reading that an explicit `type` is required still stands and there is nothing to harden against yet.
- **Item 8** (global-scope writers follow symlinks) — pre-existing and repo-wide; the issue itself says it should not be scoped to ZCode.

## Verification

`pnpm cicheck` green (411 files, 11146 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #2973, opened 2026-09-08
feat(fetch): note the one-column blanks that are drawn as a plain space

### LINKED ISSUE #2840: fetch: note the one-column blanks that are drawn as a plain space
`hasWhitespaceThatDoesNotShow` deliberately leaves a single blank *inside* a name alone when it is merely not the plain space: it is drawn, so it is a substitution rather than an extent, and the plain name it imitates is reported as the pair it makes under the display-form reason. The doc comment justifies this with `\u8a2d\u5b9a\u3000\u30ac\u30a4\u30c9`, where U+3000 is two columns wide and so tells itself apart.

That argument does not hold for the narrow blanks. U+00A0, U+2000–U+200A, U+202F, U+205F and U+1680 are all one column and are drawn indistinguishably from the plain space. A name like `pdf` + U+00A0 + `reader` alone on a list reads exactly as `pdf reader` and carries no note at all.

The harm is small, since with no twin on the list there is no other row to be taken for — but rulesync already maps U+2010 HYPHEN in the lookalike tables for the same reason, so treating the one-column blanks the same way would be consistent.

Suggested shape: a separate reason (not part of the whitespace-extent one) for a name containing one of those blanks, saying it carries a character drawn as a plain space. U+3000 stays out, since its width distinguishes it.

dyoshikawa:
Closes #2840

## The gap

`hasWhitespaceThatDoesNotShow` deliberately leaves a single blank *inside* a name alone when it is merely not the plain space: it is drawn, so it is a substitution rather than an extent. The doc comment justified that with `設定 ガイド` written with U+3000, which is two columns wide and so tells itself apart on the row it is printed on.

That argument does not carry to the narrow blanks. U+00A0, U+1680, U+2000–U+200A, U+202F and U+205F are all one column and are drawn indistinguishably from the plain space, so `pdf` + U+00A0 + `reader` reads on screen as exactly `pdf reader`. Beside the plain spelling the pair is already reported — names are compared with their whitespace collapsed and NFKC-folded — but **alone on a list such a name carried no note at all**, while reading as a name it is not. It is the same case the lookalike tables already cover for U+2010 HYPHEN, whose only tell is that it is not the ASCII one.

## The change

- `ONE_COLUMN_BLANK_PATTERN` and `hasBlankDrawnAsPlainSpace` in `src/utils/confusable-names.ts`, and a seventh reason, **"carries a character drawn as a plain space"**.
- A reason of its own rather than a widening of the whitespace-extent one, as the issue suggests: the two answer different questions — how far the name reaches, against which character was chosen. The doc comment on `hasWhitespaceThatDoesNotShow` now points at the new check instead of arguing the case away.
- Emitted **third**, after the whitespace and enclosing-mark reasons and ahead of the comparison ones. Like those two it says what the row itself is drawn as whatever else is on the list, and the picker cuts an overlong label from its tail, so the reasons that cannot be checked by eye go in front.
- **U+3000 IDEOGRAPHIC SPACE stays out**, per the issue: its width distinguishes it, and marking it would put a warning on an ordinary Japanese name. U+200B ZERO WIDTH SPACE is outside the U+2000–U+200A range and stays the hidden-character check's business.

## Tests

- The `it.each` case asserting `pdf` + U+00A0 + `reader` is left alone is replaced by a seven-case `it.each` covering U+00A0, U+202F, U+205F, U+1680, U+2000, U+2003 and U+200A, each asserted to carry the new note with nothing on the list to compare it to. The ideographic-space case stays in the "left alone" block, with its comment rewritten to say why the width is the reason.
- The pairing case now asserts the substituted name carries both reasons, `carries a character drawn as a plain space; another entry has the same display form`, while the plain spelling still carries only the display-form one.
- A new ordering case asserts all four reasons on one name come out as whitespace, mark, plain-space blank, then scripts.

## Docs

`docs/reference/cli-commands.md` goes from six reasons to seven, with the new bullet placed third and the first bullet's carve-out rewritten to hand the substitution question to it. `src/generated/docs-content.ts` regenerated.

## Verification

`pnpm cicheck` green (411 files, 11153 tests).


dyoshikawa:
@dyoshikawa Thank you!

Issue #2945, opened 2026-09-06
Add support for Poolside (`pool` coding agent CLI)


Fe2-O3:
## Summary

Please add support for **Poolside** as a generation target.

Poolside's coding agent is `pool` — a single binary that runs an interactive TUI in the
terminal and also acts as an Agent Client Protocol (ACP) server, so it is driven from Zed,
JetBrains, Neovim and other ACP-compatible editors as well as from the CLI.

- Repo: https://github.com/poolsideai/pool
- Docs: https://docs.poolside.ai (LLM doc index: https://docs.poolside.ai/llms.txt)
- CLI overview: https://docs.poolside.ai/cli/pool
- Install: https://docs.poolside.ai/cli/install

## Why

Poolside exposes most of the primitives rulesync already fans out — rules, skills, MCP,
subagents, hooks and permissions — and it does so on top of two conventions rulesync
already handles: it reads `AGENTS.md` for instructions, and its skills follow the
[Agent Skills specification](https://agentskills.io/specification) that rulesync's
`agentsskills` target already enforces. So a fair amount of the plumbing exists.

Being honest about demand: this is one user asking. I keep a single source of truth across
several agents and Poolside is the only one of mine that rulesync cannot target.

## Config surface → rulesync feature mapping

Read from the official docs and cross-checked against a local `pool` 1.0.14 install.

| rulesync feature | Global scope | Project scope |
| --- | --- | --- |
| **rules** | `~/.config/poolside/AGENTS.md` (honours `XDG_CONFIG_HOME`) | `AGENTS.md` at the repository root, plus nested `AGENTS.md` in subdirectories |
| **skills** | `~/.config/poolside/skills/<name>/SKILL.md`, also `~/.agents/skills/` | `.poolside/skills/<name>/SKILL.md`, also `.agents/skills/` |
| **mcp** | `mcp_servers:` in `~/.config/poolside/settings.yaml` | `mcp_servers:` in `.poolside/settings.yaml` / `.poolside/settings.local.yaml` |
| **permissions** | `tools:` and `paths:` (each `allow` / `deny` / `disabled`) in the settings file | same keys, project settings files |
| **hooks** | `hooks:` in the settings file | same key, project settings files |
| **subagents** | `subagents:` in the settings file | same key, project settings files |
| **commands** | Not user-definable — slash commands are built-in only; skills cover the custom-command use case | — |
| **ignore** | No public spec | — |

[...]

PR #2976, opened 2026-09-08
feat(rules): add Pool (Poolside) as a rules target

Fe2-O3:
Refs #2945.

Adds Pool (Poolside's coding agent CLI, `pool`) as a rulesync `rules` target. Pool reads `AGENTS.md` at personal, project, and nested per-directory scopes, so this target:

- writes the root rule to `AGENTS.md` (project) / `~/.config/poolside/AGENTS.md` (global, via `--global`, honouring `XDG_CONFIG_HOME`)
- folds plain non-root rules into that single root file (no modular rules directory)
- emits directory-scoped rules (`agentsmd.subprojectPath`) as nested `<dir>/AGENTS.md` (project scope only)
- imports nested files back through the shared AGENTS.md helper, so one rulesync rule per subproject no matter which AGENTS.md-family target discovered it

Scope note: this PR covers `rules` only. Per the mapping in #2945, `skills` (Agent Skills spec dirs), `mcp` (`mcp_servers:` in settings.yaml), `permissions` (`tools:`/`paths:`), `hooks` (`hooks:`), and `subagents` (`subagents:`) are natural follow-ups; `commands` (built-in only) and `ignore` (no public spec) are out. Happy to follow up with skills/MCP next, or to split this further if preferred.

Verified against https://docs.poolside.ai/agent-instructions and a local read of the pool 1.0.14 config surface described in #2945.

## Verification

- New `PoolRule` unit tests: 10 passed
- Full unit suite: 412 files, 11163 passed, 3 skipped
- E2E rules (incl. new pool generate/import/global entries): 144 passed
- `pnpm run check` (format, lint, typecheck): clean
- `pnpm run cicheck:content` (docs-content, gitignore, supported-tools, cspell, secretlint): clean
- Real CLI smoke: `pnpm dev generate --targets pool --features rules --dry-run` resolves to `AGENTS.md`
- Generated tables (`pnpm run generate:tables`, `generate:docs-content`, `dev gitignore`) applied; hand-written `file-formats.md` Pool note added


src/constants/pool-paths.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @see https://docs.poolside.ai/agent-instructions
+ * @see https://github.com/poolsideai/pool
+ */
+
+/** Global config directory for Pool, relative to the home directory. */
+export const POOL_GLOBAL_DIR = join(".config", "poolside");
+
+/** Instruction file name Pool reads at every scope. */
+export const POOL_RULE_FILE_NAME = "AGENTS.md";
```
dyoshikawa: This constant is not referenced anywhere on the branch - `pool-rule.ts` imports `AGENTSMD_RULE_FILE_NAME` from `agentsmd-paths.ts` instead, so `pnpm knip` reports it as an unused export. Either drop it and keep the single source of truth for `"AGENTS.md"`, or use it here the way `ZcodeRule` uses `ZCODE_RULE_FILE_NAME`. Dropping it reads better given the PR's own point that Pool's file is literally the AGENTS.md standard's.

src/constants/pool-paths.ts
```diff
@@ -0,0 +1,21 @@
+import { join } from "node:path";
+
+/**
+ * Pool (Poolside's coding agent CLI) configuration-layout conventions.
+ *
+ * Pool reads `AGENTS.md` instruction files the same way the AGENTS.md standard
+ * describes them: the personal `~/.config/poolside/AGENTS.md` (honouring
```
dyoshikawa: Same point as on the docs note below: `POOL_GLOBAL_DIR` is fixed to `.config/poolside` under the home directory, so "honouring `XDG_CONFIG_HOME`" promises more than the code does. Worth rewording to say that Pool supports the override upstream while rulesync writes the XDG default.

docs/reference/file-formats.md
```diff
@@ -168,6 +168,8 @@ Multiple files can set `root: true` for the same target in project and global mo
...
...to that single file, since it has no modular rules directory. Directory-scoped rules are the exception: Pool loads every `AGENTS.md` from the repository root down through the working directory, deeper files taking precedence — so a non-root rule carrying `agentsmd.subprojectPath` is emitted as a nested `<subprojectPath>/AGENTS.md` (project scope only) instead of being folded. On **import**, nested files are discovered by the same project scan the AGENTS.md standard uses (same dependency/build-directory exclusions; import-only, never removed by `--delete`). Like the Vibe Code note above, the imported rule is **not** suffixed or target-scoped: Pool's nested file is literally the AGENTS.md standard's own per-directory file at the same path, so it lands in `.rulesync/rules/<directory-with-hyphens>.md` with `targets: ["*"]`. Because the emitted file is the plain per-directory `AGENTS.md`, a rule scoped with `targets: ["pool"]` is still picked up by every other AGENTS.md reader working in that directory — Pool's nested surface is the shared standard's file, so target-scoping it is structurally impossible. See the [Pool AGENTS.md instructions](https://docs.poolside.ai/agent-instructions).
```
dyoshikawa: Pool does honour `XDG_CONFIG_HOME` upstream, but rulesync never reads it: `POOL_GLOBAL_DIR` is `join(".config", "poolside")` resolved against the home directory, so for anyone who sets `XDG_CONFIG_HOME` this writes somewhere Pool will not look. The Meta Muse Code note handles the same situation well - it says rulesync emits "only the XDG-default" path - and matching that phrasing keeps this accurate without any code change. Hardcoding `~/.config` is the existing convention here (amp, goose, zed, opencode, crush, kilo, devin, musecode all do it), so only the sentence needs adjusting.

dyoshikawa (commented):
I checked every path this adds against the upstream sources - docs.poolside.ai/agent-instructions and the poolsideai/pool README - and they all hold up: `AGENTS.md` as the instruction file, the repository-root file as project scope, the root-to-cwd nested scan with deeper files winning, the ignored-directory exclusions, and `~/.config/poolside/AGENTS.md` (directory `poolside`, not `pool`) for personal instructions ranking below project scope. The negatives check out too: there is no modular rules directory, no `AGENTS.local.md` variant (only `settings.local.yaml`), and Pool does not read `CLAUDE.md`, so not emitting any of those is correct. Tests, the generated tables and the docs prose line up, and the E2E rules matrix covers `pool` the same way it covers `vibe`. The target name `pool` matches the binary and follows the `pi` precedent, so I would keep it.

Two things before this can go in. The branch now conflicts with `main` in `src/generated/docs-content.ts` - #2978 landed a docs edit that touches the same generated file. Rebasing on `main` and re-running `pnpm run generate:docs-content` resolves it; the source `docs/reference/file-formats.md` merges cleanly, so the conflict is only in the generated output. The other is the `XDG_CONFIG_HOME` wording, noted inline, plus an unused export.

Nothing else is blocking. One cosmetic note: `pool` is appended at the end of `rulesProcessorToolTargetTuple` and of the `toolRuleFactories` map rather than sitting between `pi` and `qwencode`, which only shows up as the row order in the generated tables. Reorder it if you prefer, but it changes no behaviour.

Fe2-O3:
All three points addressed in ba46de1e (rebased on current main):

- Conflict: rebased; `docs/reference/file-formats.md` merged cleanly as expected, regenerated `docs-content.ts` so it now includes both #2978 and the Pool note. Branch reports MERGEABLE.
- Unused export: dropped `POOL_RULE_FILE_NAME` — `pool-rule.ts` keeps using `AGENTSMD_RULE_FILE_NAME` as the single source of truth. No `pool` hits left in knip output.
- XDG wording: both the constants comment and the Pool file-formats note now say rulesync writes only the XDG-default `~/.config/poolside/AGENTS.md` while Pool honours `XDG_CONFIG_HOME` upstream (same phrasing as the Muse note).

On the cosmetic note: left `pool` appended at the end of the rules tuple and factory map rather than between `pi` and `qwencode` — row order only, no behaviour change, and it keeps this PR's diff to the appended lines.

Verification on the rebased tree: full unit suite 412 files / 11168 passed, e2e-rules 144 passed, `pnpm run check` and `pnpm run cicheck:content` clean.

dyoshikawa:
@Fe2-O3 Thank you! Merged. The follow-up notes from the review that were not blocking are tracked in #2984.

Issue #2525, opened 2026-07-31
qwencode: import should also read the Claude-parity project .mcp.json (v0.18.1, code-only surface)


dyoshikawa:
## Background

Carved out of #2507 (its gap 3, explicitly deferred there): Qwen Code v0.18.1 ([QwenLM/qwen-code PR #4713](https://github.com/QwenLM/qwen-code/pull/4713), issue #4615) added `<projectRoot>/.mcp.json` as a Claude-parity project MCP source (`packages/cli/src/config/mcpJson.ts` normalizes Claude's `type`-based transport shape; precedence: user settings < `.mcp.json` < workspace/system settings < `--mcp-config`; servers untrusted until approved). The surface is **implemented in code only** — it is not yet in the published mcp.md docs.

rulesync's `QwencodeMcp.getSettablePaths` returns `.qwen/settings.json` for both scopes. Generated servers sit at workspace scope and outrank `.mcp.json`, so nothing rulesync writes is broken — the gap is one-directional: a repo whose servers live in a checked-in `.mcp.json` is invisible to `rulesync import --targets qwencode`.

## Proposal

Import-only: have the qwencode import path also read `<projectRoot>/.mcp.json`, merged beneath `settings.json` (settings win, matching upstream precedence). Do **not** emit to `.mcp.json` — that would subject managed servers to the approval gate for no benefit.

Deferred until upstream documents the surface in mcp.md (re-check before implementing).

dyoshikawa:
Re-checked the deferral condition (2026-07-31): still unmet, keeping this deferred.

- The code surface is still present upstream: `packages/cli/src/config/mcpJson.ts` exists on QwenLM/qwen-code main (latest release v0.21.2).
- The published MCP docs (`docs/users/features/mcp.md`) still document only `.qwen/settings.json` (project) and `~/.qwen/settings.json` (user) as MCP configuration sources — no mention of the project-root `.mcp.json` Claude-parity surface.

Per the deferral note in the issue body, implementation waits until upstream documents the surface in the MCP docs. Will re-check on a future pass.

dyoshikawa:
Re-checked the deferral condition (2026-08-17): **still unmet, keeping this deferred.**

- The code surface is still present upstream: `packages/cli/src/config/mcpJson.ts` still exists on QwenLM/qwen-code `main` (latest release v0.21.13, published 2026-08-17 — up from v0.21.2 at the last re-check).
[...]

PR #2977, opened 2026-09-09
fix(qwencode): stop dropping tools.eager on import

dyoshikawa:
## Summary

Fixes the two mechanical follow-ups recorded on #2668. The issue's remaining items are design decisions and stay open, so this PR uses `Refs` rather than `Closes`.

### 1. `tools.eager` is no longer dropped on import

`QWEN_OVERRIDE_TOOLS_KEYS` is a curated allow-list on the import path, and `eager` — added upstream in Qwen Code v0.22.3 ([QwenLM/qwen-code#10098](https://github.com/QwenLM/qwen-code/pull/10098)) — was absent, so `toRulesyncPermissions` discarded it. `generate` was unaffected because it spreads the user's existing `tools` block, so the loss was confined to `import` → `generate` round trips, where the key vanished from the rulesync source of truth and could never be re-emitted.

**One correction to the issue's own follow-up note**, which said "no `QWEN_SCOPED_TOOLS_KEYS` entry is needed". `QwenScopedKeys` is a total `Record`, so an entry is mandatory and the compiler enforces it. The rule was picked from the primary source rather than the docs page: `tools.eager` appears in neither `WORKSPACE_RESTRICTED_SETTINGS` nor `WORKSPACE_NON_OVERRIDING_SETTINGS` in `packages/cli/src/config/settingsUtils.ts`, checked at both [`v0.23.0`](https://github.com/QwenLM/qwen-code/blob/v0.23.0/packages/cli/src/config/settingsUtils.ts) and `main`, so it is honored in any scope → `global-machine-wide`.

Its notes describe what the key actually does rather than reusing the rule's default sentence: unlike `tools.disabled`, a name missing from the list is *demoted to deferred*, not taken away, and stays callable through `tool_search` — with the one carve-out that in a session where `tool_search` is not registered the demotion is final, which [the setting's own schema description](https://github.com/QwenLM/qwen-code/blob/v0.23.0/packages/cli/src/config/settingsSchema.ts) calls out and logs a warning for.

### 2. The qwencode research reference map was steering research runs away from three implemented dimensions

`.rulesync/skills/rulesync-feature-research/references/qwencode.md` recorded `commands`, `subagents` and `hooks` as *"No dedicated upstream `<feature>` surface in map"* / *"No Rulesync-supported Qwen Code hooks target in map"*. All three are documented upstream **and** fully implemented in rulesync, and the research skill's Collect step reads those exact sentinel phrases as "upstream has nothing here".

Rewritten from the primary sources, all re-verified at `v0.23.0`:

- `commands` → [`/users/features/commands/`](https://qwenlm.github.io/qwen-code-docs/en/users/features/commands/) — `.qwen/commands/*.md` at both scopes, nested directories becoming `:`-namespaced commands.
- `subagents` → [`/users/features/sub-agents/`](https://qwenlm.github.io/qwen-code-docs/en/users/features/sub-agents/) — `.qwen/agents/*.md` at both scopes, Claude Code agent files accepted, project-only `.qwen/fork-profiles/*.md` noted as separate.
- `hooks` → [`/users/features/hooks/`](https://qwenlm.github.io/qwen-code-docs/en/users/features/hooks/) — the `hooks` block in the layered `settings.json`, pointing at `QWENCODE_HOOK_EVENTS`.
- `checks` (new row) → [`/users/features/code-review/`](https://qwenlm.github.io/qwen-code-docs/en/users/features/code-review/) — `.qwen/review-rules.md`, marked `unsupported` with the reason, so a missing row is not misread as a missing upstream surface.
- `settings` (new row) → the exhaustive settings key table, which is the page a future pass should diff first.
- `mcp` now records that there is still no documented project-root `.mcp.json` at `v0.23.0` or `main` — the unmet deferral condition of #2525.

A closing section documents the allow-list trap this PR's first half fell into, so the next person adding a `tools`/`security` key knows the import path is where the data loss lives and that the scope rule comes from `settingsUtils.ts`, not the docs prose.

## What is deliberately not in this PR

The rest of #2668 stays open: `general.outputStyle` (needs a new `general` group in the override plus a change to `.qwen/settings.json` ownership in `shared-config-gateway.ts`), the `.qwen/review-rules.md` checks target, `.qwen/fork-profiles/*.md`, server-level `mcp.allowed` / `mcp.excluded`, the `skills.*` kill-switches, and the `.qwen/` by-product gitignore entries.

## Testing

- New round-trip case pinning `tools.eager` in both directions, alongside the existing `visible` / `listDirectory` / `workflowsEnabled` cases.
- `pnpm cicheck` — 411 files, 11154 tests, all green.

Refs #2668


dyoshikawa:
Merged. Thanks — `tools.eager` now round-trips through the qwencode override instead of being silently dropped on import.

Issue #2599, opened 2026-08-06
Follow up Reasonix upstream updates: MCP startup_timeout_seconds destroyed, allow_dynamic_bash unauthorable, hook cwd unmodeled


dyoshikawa:
## Summary

Three Reasonix drifts as of v1.20.0 (rulesync's implementation notes cite v1.18.0): the new per-server MCP `startup_timeout_seconds` is destroyed on regenerate, the v1.19.0 `[permissions] allow_dynamic_bash` opt-in is unauthorable (and a docs claim it obsoletes is still published), and the hook `cwd` field is unmodeled.

All claims verified by diffing the upstream docs tree between tags v1.18.0 and v1.20.0 (esengine/DeepSeek-Reasonix), so each is attributable to a released tag.

## Recent Releases

- **v1.20.0** (2026-08-05) — Extension Protocol v1 / Plugin Manifest v1 (PR #7373; `[[plugins]]` MCP schema unchanged apart from gap 1's key — checked, not a gap).
- **v1.19.2** (2026-08-02) — MCP cold-start resilience (PR #7164): per-server `startup_timeout_seconds` + global `mcp_startup_timeout_seconds` (default 30).
- **v1.19.0** (2026-08-01) — `[permissions] allow_dynamic_bash` advanced opt-in.
- v1.19.6 "global hooks migration" (PR #7445) is a one-shot legacy-home migration; the Reasonix home is still `~/.reasonix` — checked, not a gap.

### 1. mcp — per-server `startup_timeout_seconds` silently destroyed

- **Upstream**: `[[plugins]]` per-server `startup_timeout_seconds` ("initialize + tools/list cap; 0 = global/default cap") — [SPEC.md at v1.20.0](https://github.com/esengine/DeepSeek-Reasonix/blob/v1.20.0/docs/SPEC.md), [PR #7164](https://github.com/esengine/DeepSeek-Reasonix/pull/7164)
- **rulesync** (`project`, `global`): `REASONIX_PLUGIN_FIELDS` in `src/features/mcp/reasonix-mcp.ts` is a strict allowlist (`type`,`command`,`args`,`env`,`url`,`headers`,`call_timeout_seconds`,`tool_timeout_seconds`); since rulesync owns and rewrites the whole `plugins` key, a hand-authored `startup_timeout_seconds` is discarded on the next generate and cannot be authored or imported.
[...]

PR #2978, opened 2026-09-09
fix(reasonix): stop dropping the [[plugins]] concurrency and auto_start keys

dyoshikawa:
## Problem

Reasonix's `[[plugins]]` entries accept two scheduling keys that rulesync's `REASONIX_PLUGIN_FIELDS` allowlist does not carry:

- `concurrency` — `"parallel"` or `"serial"`. `serial` stops two calls to the same server from running at once across the whole session, which is what keeps sub-agents sharing one stdio process from interleaving on its state. With no explicit value Reasonix substring-matches the server *name* against a known-stateful list (`browser`, `playwright`, `puppeteer`, `chrome`, `chromium`, `selenium`) and serializes a match; an explicit value always wins.
- `auto_start` (`*bool`) — `false` switches the server off: `Config.EnabledPlugins` drops it from the enabled set, so its tools never reach the model and Reasonix reports it as `disabled` rather than `deferred`. Only a durable override in `mcp-activation.json` outranks the file value.

Because `fromRulesyncMcp` rewrites the whole `plugins` key through `applySharedConfigPatch`, this is not merely "cannot be authored from `.rulesync/mcp.json`": a value written by hand into `reasonix.toml` is **deleted on the next generate**, silently. For these two keys that deletion changes behavior: dropping the key does not fall back to a neutral default, it hands the decision back to the server's name. A `serial` dropped from a stateful server the name list does not catch puts it on the parallel path; a `parallel` dropped from a server the list does catch forces it serial; and a dropped `auto_start = false` switches a deliberately disabled server back on.

## Change

Both keys join `REASONIX_PLUGIN_FIELDS` as passthrough fields, exactly like the sibling timeout keys. The write path and the import path both iterate that list, so authoring, import and round-trip all follow from the one addition. Neither key has a canonical counterpart, so no deep mapping is introduced.

Tests cover export, import, and a round trip. Each pins a value that actually overrides what Reasonix would infer on its own — `parallel` on a server named `browser`, `serial` on one the stateful-name list does not catch, and `auto_start: false`, the value a truthiness filter would drop.

`docs/reference/file-formats.md` gains both keys in the Reasonix MCP note.

## Verification

Confirmed against the upstream primary source rather than the docs prose: `internal/config/plugin_entry.go` at `v1.38.3` declares `Concurrency string \`toml:"concurrency"\`` and `AutoStart *bool \`toml:"auto_start"\``. `docs/SPEC.md` §3.16 documents the concurrency semantics and the stateful-name default; `docs/PLUGIN_PACKAGES.md` documents `auto_start=false` for imported servers, and `internal/config/config.go` (`ShouldAutoStart` / `EnabledPlugins`) plus `internal/cli/mcp_manager.go` (the `disabled` vs `deferred` status split) are what settle that `auto_start = false` disables rather than defers.

`concurrency` has no canonical counterpart. `auto_start` overlaps canonical `disabled`, but it is left as a passthrough here rather than deep-mapped: Reasonix's activation store can flip a server independently of the file, so the two are not interchangeable, and deciding which wins when a canonical `disabled` and a hand-written `auto_start` disagree — in both the write and the import direction — is a separate call left on #2599.

`pnpm cicheck` passes (411 test files, 11158 tests).

## Scope

Refs #2599 — this closes the two MCP gaps recorded in that issue's re-check comments. The issue stays open for its gap 3 (the hook `cwd`/`env` fields), which needs a maintainer decision on whether `cwd` becomes a shared canonical hook key or a Reasonix-only passthrough.


dyoshikawa:
Merged. Thanks — this ships the two `[[plugins]]` scheduling keys as passthrough fields, so a hand-written `concurrency` or `auto_start` survives the next generate.

Two follow-ups stay on #2599: the hook `cwd` gap, and whether canonical `disabled` should deep-map onto `auto_start` (which needs a precedence rule against a hand-written value, in both the write and the import direction).

Issue #2664, opened 2026-08-13
Follow up Claude Code upstream updates: dynamic workflows (.claude/workflows/) unsupported, scheduled-task skills written to an unread project path


dyoshikawa:
## Summary

Claude Code's dynamic workflows (v2.1.154) are a slash-command surface stored in `.claude/workflows/` and `~/.claude/workflows/` that rulesync neither generates nor imports, and a plugin can distribute the same scripts through a `workflows/` component directory. Separately, rulesync routes `scheduled-task` skills to a project path Claude Code never reads.

Research window: v2.1.126 → v2.1.231 (2026-05-01 → 2026-08-13).

## Recent Releases

- **v2.1.154** (2026-05-28) — [CHANGELOG](https://github.com/anthropics/claude-code/blob/main/CHANGELOG.md) — "Introducing dynamic workflows … Run `/workflows` to view your runs". A run can be saved as a reusable slash command; plugins can ship them as a component.
- **v2.1.178** (2026-06-15) — monorepo resolution for project workflows: they load from every `.claude/workflows/` between cwd and repo root, closest name wins.
- **v2.1.187** (2026-06-23) — scheduled-task writes no longer follow a `.claude` symlink (hardening; relevant context for gap 2).
- **v2.1.219** (2026-07-24) — `DirectoryAdded` hook event — already emitted by rulesync.

### 1. commands/skills — dynamic workflows (`.claude/workflows/`, `~/.claude/workflows/`) are unsupported

- **Upstream**: a workflow run can be saved as a reusable slash command into `.claude/workflows/` (project, shared via VCS) or `~/.claude/workflows/` (user scope, honoring `CLAUDE_CONFIG_DIR`); the saved script then runs as `/<name>` in later sessions. A workflow is a single self-contained plain-JavaScript file exporting a `meta` object, with `meta.name` supplying the command name; `import()` is rejected by the runtime, so there are no module dependencies to resolve. Sources: [workflows docs](https://code.claude.com/docs/en/workflows) (sections "Save the workflow for reuse", "How a workflow runs") and CHANGELOG entries 2.1.154 / 2.1.178.
[...]

PR #3003, opened 2026-09-12
fix(skills): skip claudecode scheduled-task skills at project scope

dyoshikawa:
## Background

Gap 2 of #2664: `claudecode.scheduled-task: true` skills were routed to `.claude/scheduled-tasks/<name>/SKILL.md` regardless of scope, so a project-scope generate wrote `<project>/.claude/scheduled-tasks/<name>/SKILL.md` — a file that looks live and never fires.

Re-verified against the current docs on 2026-09-12:

- https://code.claude.com/docs/en/desktop-scheduled-tasks — "To edit a task's prompt on disk, open `~/.claude/scheduled-tasks/<task-name>/SKILL.md` (or under `CLAUDE_CONFIG_DIR` if set)". No project-level location is documented.
- https://code.claude.com/docs/en/claude-directory and https://code.claude.com/docs/en/skills#where-skills-live list no project-scope `scheduled-tasks/` (negative evidence already recorded on the issue).

The fix shape follows the precedent set by #2694 for user-scope-only sandbox keys: warn and skip at project scope, unchanged at global scope, import unchanged.

## Changes

- `src/features/skills/skills-processor.ts`: a `scheduled-task` skill targeted at `claudecode` is skipped with a warning when `global` is false. The warning names the skill, states where Claude Code reads scheduled tasks from, and points at `--global` plus any stale copy an earlier generate left behind.
- `docs/reference/file-formats.md`: the frontmatter comment and the scheduled-task paragraph now state the user-scope-only constraint; `src/generated/docs-content.ts` regenerated.
- Tests: processor unit tests for the project-scope skip (warning asserted) and the global-scope passthrough; the e2e project-scope case now asserts the skip and warning, and a new global-mode e2e case asserts the `~/.claude/scheduled-tasks/<name>/SKILL.md` output and that nothing lands in the project.

Gaps 1 and 3 on the issue (`.claude/workflows/` and the plugin `workflows/` component) are a new feature dimension and stay open, so this PR does not close the issue.

## Testing

- `pnpm cicheck` (413 test files, 11208 tests passed; content checks clean)
- `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-skills.spec.ts -t scheduled-task` (5 passed)

Part of #2664


dyoshikawa:
@dyoshikawa Thank you!

PR #3005, opened 2026-09-12
fix(kimi-code): preserve hand-written [permission] siblings on regenerate

dyoshikawa:
## Summary

Refs #2502 (the `[permission] dangerous_command_guard` gap recorded on 2026-09-08; the issue stays open for its remaining items).

Kimi Code 0.40.0 added `dangerous_command_guard` under `[permission]` in `~/.kimi-code/config.toml` (default `true`; `false` turns the built-in dangerous-command policy off — [config-files.md § permission](https://moonshotai.github.io/kimi-code/en/configuration/config-files.html#permission), verified against the current `docs/en/configuration/config-files.md` in `MoonshotAI/kimi-code`). Rulesync owns the whole `permission` key through the shared-config gateway (`replace-owned-keys`) but only ever authors `rules`, so a hand-written sibling was deleted on every `rulesync generate --features permissions --global`. In this case the loss re-enables the guard, so it fails safe rather than open, but it is still a user setting silently discarded.

## Changes

- `src/features/permissions/kimi-code-permissions.ts`: new `mergeKimiCodePermissionSection` carries the existing `[permission]` siblings of `rules` over onto the generated table (same shape as the existing `[tools]` per-key merge). `rules` always comes from the patch, so the ordered rule list is still replaced, not merged.
- `src/features/permissions/kimi-code-permissions.test.ts`: three tests — sibling preserved, `rules` still replaced, no table in the existing file.
- `docs/reference/file-formats.md` (+ regenerated `src/generated/docs-content.ts`): one sentence documenting the per-key merge and that the sibling keys are neither authorable nor imported.

## Not done here

Whether `dangerous_command_guard` should become an authorable `kimi-code` override field and round-trip on import is a schema decision left on the issue; this PR only stops the deletion.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2498, opened 2026-07-30
Follow up Grok CLI upstream updates: sandbox.toml deny-glob ignore surface (project scope now documented), per-handler hook env dropped


dyoshikawa:
## Summary

Two Grok CLI surfaces are unfollowed as of upstream v0.2.114 (2026-07-29): project-scope `.grok/sandbox.toml` custom-profile `deny` globs are now documented, giving grokcli a real path-deny surface for rulesync's `ignore` feature (currently `unsupported`), and the per-handler hook `env` map is silently dropped on generate and import.

Baseline: #2418's gaps were fixed in #2447 (`.grok/rules/`, skill invocation flags, `http` hooks) and `.grok/commands/` support has since landed too (`src/features/commands/grokcli-command.ts`). This pass verified against a clone of `xai-org/grok-build` at commit `500129c` (2026-07-29, version 0.2.114) and the changelog window v0.2.100–v0.2.114 ([x.ai/build/changelog](https://x.ai/build/changelog)).

### 1. ignore — `.grok/sandbox.toml` custom-profile `deny` globs (project + global)

- **Upstream**: [sandbox doc](https://docs.x.ai/build/features/sandbox) — "Custom profiles can be defined in `~/.grok/sandbox.toml` (user-level) or project `.grok/sandbox.toml` (project-level)", example `[profiles.my-profile]` with `extends = "workspace"`, `restrict_network = true`, `deny = ["/secrets", "**/.env", "**/*.pem"]`. Source-confirmed at 0.2.114: `crates/codegen/xai-grok-sandbox/src/profiles.rs` `load_sandbox_config(workspace)` reads `~/.grok/sandbox.toml` then merges `<workspace>/.grok/sandbox.toml`; a custom profile's `deny` list is "kernel-enforced (read + write/rename)". The glob shape is exactly what rulesync's ignore feature expresses.
- **rulesync**: `ignore` is `unsupported` for grokcli. `.grokignore` genuinely does not exist upstream (re-verified at 0.2.114), so the old verdict was right for the wrong surface — #2418 declined to file this when the docs listed `[sandbox]` as global-only, and that reasoning no longer holds now that the project-scope file is documented.
- **Design caveats** (shape, don't block): the deny list only bites while the named profile is active (`grok --sandbox <name>`, `[sandbox] profile`, or `GROK_SANDBOX`), and `merge_project_profiles` keeps a globally-defined profile of the same name, so a project file cannot silently override a user profile.

[...]

PR #3006, opened 2026-09-12
feat(grokcli): map the agent and glob permission categories onto AgentMessage and Glob

dyoshikawa:
## Summary

Refs #2498 (the 2026-09-08 `AgentMessage` item and the 2026-08-24 `Glob` item; the issue stays open for the remaining design-gated surfaces).

Grok Build's permission rule parser (`crates/codegen/xai-grok-workspace/src/permission/rules.rs`, `tool_name_to_filter`) accepts `"Grep" | "Glob" => ToolFilter::Grep` and, since 1.0.10 (`9684fa3c`), `"AgentMessage" | "SendSubagentMessage" | "SendAgentMessage" => ToolFilter::AgentMessage`. rulesync's `grokcli` permissions adapter skipped the canonical `glob` and `agent` categories on generate (a `glob` deny authored once was silently unenforced for Grok) and dropped `Glob(...)` / `AgentMessage(...)` entries on import.

- `CATEGORY_TO_GROK_TOOL`: `glob → Glob`, `agent → AgentMessage`.
- `GROK_TOOL_TO_CATEGORY`: `Glob → glob`, `AgentMessage → agent`, plus import-only `SendSubagentMessage` / `SendAgentMessage` (always re-emitted as `AgentMessage`). The verbose `[[permission.rules]]` form's `tool = "agent_message"` (upstream serde spelling) is accepted as well.
- `Grep` and `Glob` share one Grok filter but stay separate on import so each spelling round-trips to the category it was authored under.
- Comments and `docs/reference/file-formats.md` no longer claim `glob`/`agent` have no Grok tool; the docs state that an `AgentMessage` pattern is matched against the *target subagent id* (glob or prefix, `policy.rs`), not the subagent type a Claude Code `Agent(...)` rule names, so only `*` means the same thing for both — and extend the migration note (previously unmanaged `Glob`/`AgentMessage` entries become Rulesync-owned).

## Test plan

- [x] New unit tests: generate `glob`/`agent` (bare and patterned), import `Glob`/`Grep` into separate categories, import all three `AgentMessage` spellings plus the verbose `agent_message`, and legacy spellings regenerating as `AgentMessage`.
- [x] `pnpm cicheck` passes (413 files / 11216 tests).


dyoshikawa:
@dyoshikawa Thank you!

Issue #2496, opened 2026-07-30
Follow up Codex CLI upstream updates: unauthorable [agents]/[[skills.config]]/project_doc_* keys, additionalContextLimit dropped, Codex plugin bundles


dyoshikawa:
## Summary

Three Codex CLI surfaces are unfollowed as of upstream 0.146.0: feature-adjacent top-level `config.toml` keys (`[agents]`, `[[skills.config]]`, `project_doc_*`, `mcp_oauth_callback_*`, `allow_managed_hooks_only`) have no authoring path, the per-handler hook field `additionalContextLimit` (0.145.0) is dropped by the converter, and the checked-in Codex plugin layout (`.codex-plugin/plugin.json` + marketplace discovery) has no rulesync surface.

Reviewed releases 0.129.0 → 0.146.0 (2026-05-07 → 2026-07-29). Docs note: `developers.openai.com/codex/*` now 308-redirects to `learn.chatgpt.com/docs/*`; both hosts still serve content, and appending `.md` yields Markdown.

## Recent Releases

- **0.145.0** (2026-07-21) — `SessionEnd` hook event ([PR #33895](https://github.com/openai/codex/pull/33895), already followed) and configurable hook context spill limits `additionalContextLimit` ([PR #34393](https://github.com/openai/codex/pull/34393), gap 2).
- **0.143.0** — relaxed top-level `hooks.json` metadata validation ([PR #30229](https://github.com/openai/codex/pull/30229)); 0.140.0 warned on unsupported fields ([PR #26426](https://github.com/openai/codex/pull/26426)).
- **0.134.0** — plugin feature flag removed ([PR #22552](https://github.com/openai/codex/pull/22552)); `mcp_oauth_callback_url` ([PR #24120](https://github.com/openai/codex/pull/24120)).
- **0.131.0** (2026-05-18) — plugin hooks enabled by default ([PR #22549](https://github.com/openai/codex/pull/22549)), `allow_managed_hooks_only` ([PR #20319](https://github.com/openai/codex/pull/20319)), `mcp_oauth_callback_port` ([PR #20237](https://github.com/openai/codex/pull/20237)).

### 1. No authoring path for feature-adjacent top-level `config.toml` keys

- **Upstream** ([config reference](https://developers.openai.com/codex/config-reference)):
[...]

PR #3007, opened 2026-09-12
feat(codexcli): map the stopCancelled hook event onto Codex's Interrupt

dyoshikawa:
## Summary

- Add `stopCancelled` to `CODEXCLI_HOOK_EVENTS` and map it onto Codex CLI's `Interrupt` hook event (`CANONICAL_TO_CODEXCLI_EVENT_NAMES`), so a canonical `stopCancelled` hook is generated as `Interrupt` in `.codex/hooks.json` and an existing `Interrupt` block imports back as `stopCancelled`.
- Document the event's constraints in a code comment and a docs note: Codex runs it only when the user interrupts an active turn on the main thread (never for idle threads or subagents), ignores any `matcher`, honours only an optional `systemMessage` in the hook output (it cannot prevent the interruption), and caps command timeouts at 3s (1s default). Rulesync forwards the hook as authored; none of that is modelled differently, matching the precedent set for `sessionEnd`.
- The hook event × tool matrix in `docs/reference/file-formats.md` and `src/generated/docs-content.ts` are regenerated.

Refs #2496 (the `Interrupt` item; the other items there are design/docs decisions and stay open).

**Follow-up recorded, not done here:** Kimi Code's native `Interrupt` has the same shape as canonical `stopCancelled` but stays native-only (reachable via the `kimi-code` override), because mapping it would collide with existing overrides that address it by name; the Kimi comments and docs now say so instead of claiming it has no canonical counterpart.

## Verification

- `Interrupt` verified against https://learn.chatgpt.com/docs/hooks.md and `codex-rs/core/config.schema.json` at rust-v0.153.4 (`HooksToml.Interrupt`); shipped in rust-v0.150.0.
- New tests in `src/types/hooks.test.ts` (name mapping both ways, `stop` kept distinct) and `src/features/hooks/codexcli-hooks.test.ts` (generate `stopCancelled` → `Interrupt`, import `Interrupt` → `stopCancelled`).
- `pnpm cicheck` passes (413 files / 11220 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #3008, opened 2026-09-12
docs(takt): describe command gates as rule-gated since 0.62.0 and gitignore .takt/make/

dyoshikawa:
## Summary

- Reword the Takt command-gate security notes in `docs/tools/takt.md`, `docs/reference/file-formats.md`, and the `TaktCheck` doc comment: a gate still needs no policy switch to run, but since TAKT 0.62.0 it runs after rule resolution and before the transition is applied, and a workflow rule carrying `command_gates: skip` applies its transition without running it — so a gate is no longer described as an unconditional post-step check.
- Add `**/.takt/make/` to `HAND_MAINTAINED_GITIGNORE_ENTRIES` (and the project `.gitignore` via `pnpm dev gitignore`): `takt make` (0.64.0) keeps every Workflow Maker run under `.takt/make/<timestamp>/` and never removes completed or failed runs.
- Regenerate `src/generated/docs-content.ts`.

The `assistant.init_files` / `init_files` part of the issue is not addressed here; it needs a design decision on how (or whether) rulesync should emit them, so the issue stays open.

## Verification

- `command_gates` (`required` / `skip`) on workflow rules and the run order of command quality gates: TAKT workflow docs (https://github.com/nrslib/takt/blob/main/docs/workflows.md) and the 0.62.0 release notes.
- `takt make` output under `.takt/make/<timestamp>/`: TAKT CLI docs (https://github.com/nrslib/takt/blob/main/docs/cli.md), 0.64.0.
- `pnpm cicheck` passes locally (413 files / 11220 tests).

Refs #2421


dyoshikawa:
@dyoshikawa Thank you!

PR #3023, opened 2026-09-13
feat(vibe): warn about backslash hook commands and reserved skill names

dyoshikawa:
## Summary

Refs #2958 — resolves the two mechanical follow-ups from the 2026-09-12 re-check comment. The remaining item in that issue (a `.vibe/plugins/<name>/` packaging target) is a product decision and stays open.

- **Vibe hooks:** since v2.25.1 (verified at v2.25.3), `load_hooks_config` drops any `[[hooks]]` entry whose `command` contains a backslash ("Hook '<name>' skipped: backslash paths are not supported in hook commands. Use forward slashes instead."), because the exec-based executor tokenizes the command with `shlex.split`. The adapter now warns at generate time when a Vibe-targeted hook command contains a backslash; the hook is still written unchanged. The stale "Vibe hooks are always shell commands" comment is replaced by an accurate description (legacy backend: shell; unified harness: `shlex.split` + exec, per the v2.25.1 changelog entry).
- **Vibe skills:** `SkillManager` treats the built-in skill names `vibe` and `skill-creator` as reserved and skips a project/user skill by that name with only a debug log. The adapter now warns at generate time when a Vibe-targeted skill uses one of those names; the skill is still written.
- Docs: hooks and skills notes in `docs/reference/file-formats.md` mention both behaviors; `src/generated/docs-content.ts` regenerated.

Shell-metacharacter warnings were deliberately **not** added: the v2.25.3 legacy backend (`vibe/core/hooks/executor.py`) still runs hooks through `create_subprocess_shell`, and the unified harness is opt-in / rolled out via flag, so a metacharacter warning would be wrong for most current installs.

## Evidence

- https://github.com/mistralai/mistral-vibe/blob/v2.25.3/vibe/core/hooks/config.py (backslash check in `load_hooks_config`)
- https://github.com/mistralai/mistral-vibe/blob/v2.25.3/vibe/core/skills/builtins/__init__.py (`BUILTIN_SKILLS`: `vibe`, `skill-creator`)
- https://github.com/mistralai/mistral-vibe/blob/v2.25.3/vibe/core/skills/manager.py (`_reserved_builtins` skip)
- https://github.com/mistralai/mistral-vibe/blob/v2.25.3/CHANGELOG.md (v2.25.1: "Hook commands run without a shell to prevent injection via hooks.toml")

## Test plan

- [x] `src/features/hooks/vibe-hooks.test.ts`: warns (named and unnamed hook) and still writes the entry; no warning without a backslash
- [x] `src/features/skills/vibe-skill.test.ts`: warns for `vibe` / `skill-creator`, still generates; no warning for other names
- [x] `pnpm cicheck`


dyoshikawa:
@dyoshikawa Thank you!

PR #3024, opened 2026-09-13
feat(factorydroid): add threat-model channel routing to .factory/threat-model.md

dyoshikawa:
## Summary

Refs #2957 (the `threat-model` gap; the `output-styles` gap stays open as a product decision).

Factory's Security Review reads `.factory/threat-model.md` as its attack-surface map when the file exists ("if `.factory/threat-model.md` exists, Droid uses it as the attack-surface map" — https://docs.factory.ai/software-factory/security-review). Rulesync had no way to emit that file. This PR adds a second directory-shaped route on `rules`, `factorydroid.channel: threat-model`, mirroring the existing `design` route.

## Changes

- `src/features/rules/factorydroid-rule.ts`: generalize the `design` boolean into a `channel: "design" | "threat-model"` so both fixed-file routes share one code path (`getChannelPaths` / `findChannelByPath`). `getSettablePaths()` gains `threatModel = { relativeDirPath: ".factory", relativeFilePath: "threat-model.md" }`; `getExtraFixedFiles` lists both channel files for import/deletion; both are excluded from the root TOON references; both are project scope only (ignored on `root: true` and in `--global`). Import writes the threat model to `.rulesync/rules/threat-model.md`, under its own basename so it never collides with `DESIGN.md`.
- `src/constants/factorydroid-paths.ts`: `FACTORYDROID_THREAT_MODEL_FILE_NAME`.
- `src/features/rules/rulesync-rule.ts`: `factorydroid.channel` enum accepts `threat-model`.
- `.gitignore`: `**/.factory/threat-model.md` derived by `pnpm dev gitignore`.
- Docs: `docs/reference/file-formats.md` (frontmatter example + Factory Droid note), `references/factorydroid.md` (threat-model row, rules anchor), regenerated `src/generated/docs-content.ts`.
- Tests: unit coverage for routing, root/global guards, `fromFile` (including the `.factory/rules/threat-model.md` name collision), `forDeletion`, and the import round-trip; an e2e generate → import round-trip case next to the existing `design` one.

## Verification

- `pnpm cicheck` passes (415 test files, 11278 tests).
- `pnpm test:e2e` for the Factory Droid rules cases passes.


dyoshikawa:
@dyoshikawa Thank you!

PR #3025, opened 2026-09-13
fix(codebuddy): split a comma-separated scalar paths value into its patterns

dyoshikawa:
## Summary

Refs #2953 (the comma-separated `paths` import gap from the 2026-09-12 re-check; gap 4 — the six unimplemented CodeBuddy config surfaces — stays open as a product decision).

CodeBuddy's [memory docs](https://www.codebuddy.ai/docs/cli/memory) say a scalar `paths` value can "combine multiple patterns with commas", with the example `paths: {src,lib}/**/*.ts, tests/**/*.test.ts`. `normalizeCodebuddyPaths` wrapped a scalar as a single one-element list without splitting, so such a rule imported as one comma-bearing glob that every other target then received verbatim.

## Changes

- `src/features/rules/codebuddy-rule.ts`: new `splitCodebuddyPathsScalar` splits a scalar on commas at brace depth 0, so `{src,lib}` stays one pattern while `, tests/**/*.test.ts` becomes a second. `normalizeCodebuddyPaths` uses it for the string shape only — list elements are not split (the docs describe the comma form for the string shape, and a list already separates its patterns). Applies to both the tool frontmatter (import) and the canonical `codebuddy.paths` block (generate), which accept the same shapes.
- `src/features/rules/codebuddy-rule.test.ts`: import of the documented comma example → two globs (and two `codebuddy.paths` entries); a list element with a brace group is left intact; a scalar `codebuddy.paths` in the canonical rule is split the same way.
- `docs/reference/file-formats.md` CodeBuddy note + regenerated `src/generated/docs-content.ts`.

## Verification

- `pnpm cicheck` passes (415 test files, 11281 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #3031, opened 2026-09-13
fix(generate): warn when a later target overwrites a fold target's root file

### LINKED ISSUE #3022: zoocode/roo rules generation silently drops all non-root content from AGENTS.md
## Summary

Enabling the `zoocode` (or its deprecated predecessor `roo`) target's `rules` feature — **by itself, with no other target configured** — corrupts `AGENTS.md`: only the root rule's body survives, and every non-root rule's content is silently dropped.

## Background

Per #2409 (item 4), `roo` declares no `localRootMode`/`localRootFileName`, so it falls back to `append-to-root` and folds its rules into `AGENTS.md` — the same file Codex CLI (`codexcli`) and other "fold" targets also own. That shared ownership is presumably why `AGENTS.md` gets touched at all by a Roo/ZooCode-only config, but `codexcli` does **not** need to be present to trigger the bug (see reproduction below).

## Reproduction

Minimal `rulesync.jsonc` — only `zoocode`, nothing else:

```jsonc
{
  "targets": {
    "zoocode": ["rules"]
  },
  "outputRoots": ["."],
  "delete": true
}
```

`.rulesync/rules/overview.md`:

```markdown
---
root: true
targets: ["*"]
---
# root body
```

`.rulesync/rules/style.md` (one or more non-root rule files):

```markdown
---
paths:
  - '**/*'
---
# style body
```

```bash
npx rulesync@16.13.0 generate
cat AGENTS.md   # only "# root body" — "# style body" is gone
cat .roo/rules/style.md   # correct — non-root content IS written here
```

So `.roo/rules/*.md` (the intended non-root output for Roo/ZooCode) is generated correctly, but `AGENTS.md` — a file `zoocode` should arguably not even need to touch when `codexcli` isn't configured — ends up truncated to just the root body.

Additional observations from a real project config (7 non-root rule files, ~344 lines of folded content normally present in `AGENTS.md`):

- Same truncation with `roo` instead of `zoocode`.
- Same truncation whether or not `codexcli` is also configured (adding `codexcli` alongside `zoocode`/`roo` does not change the outcome).
- Same truncation with an unrelated third array-style target (`cline`) added alongside `codexcli`, suggesting the shared-root-file fold mechanism is generally fragile when more than one fold-style target is active, though the single-target (`zoocode` alone) case above is the cleanest repro.

## Versions tested

Reproduces on `16.13.0`, `16.14.0`–`16.18.0`, `16.19.0`, and the latest `16.29.0` (as of 2026-09-13).

## Impact

Any project that wants to add Roo Code / ZooCode support cannot safely run `rulesync generate` if `AGENTS.md` (Codex CLI's root file) is also present/managed — every run silently deletes the bulk of `AGENTS.md`'s content (all non-root rule bodies). This is easy to miss in a large diff and easy to accidentally commit.

## Possible related issues

- #2743 (fold-collision targets duplicate content on generate/import cycles) — same "fold into shared root file" mechanism, different symptom (duplication vs. loss).
- #2409 (Roo Code final-state gaps) — item 4 is the background cited above.

Neither issue's reproduction matches this one (both are about global scope or import/generate cycling; this is a plain project-scope `generate` with only `zoocode` configured).

## Environment

- rulesync: 16.13.0 through 16.29.0 (latest), all affected
- Node: v22
- OS: Linux (WSL2)

dyoshikawa:
## Summary

Closes #3022

`codexcli` folds every non-root rule into `AGENTS.md`, while `roo` and `zoocode` write only the root rule to `AGENTS.md` and keep the rest under `.roo/rules/`. Both layouts are correct on their own — the loss reported in #3022 happens when the two share `AGENTS.md` and the modular target is listed **after** `codexcli`: the documented last-wins rule lets `zoocode` overwrite `AGENTS.md` with the root rule alone, so Codex CLI silently loses every non-root rule.

This PR keeps last-wins (it is documented and relied on) and makes the lossy case loud instead of silent:

- `generateRulesCore` now watches the converted tool files per target and warns when a non-fold target overwrites, with different content, a root file that a fold target already claimed. The warning names the reorder that keeps the folded content (`list 'codexcli' after 'zoocode'`).
- The reverse order (`zoocode,codexcli`) produces no warning: `codexcli` wins `AGENTS.md`, `zoocode` still gets its files under `.roo/rules/`, nothing is lost.
- `docs/guide/configuration.md` "Target Order and File Conflicts" documents the fold-target ordering.

## Changes

- `src/lib/generate.ts`: `createFoldRootOverwriteWatch` + an `onToolFiles` hook on `processFeatureWithRulesyncFiles`.
- `src/e2e/e2e-rules.spec.ts`: e2e covering both orders (warning text, `AGENTS.md` and `.roo/rules/style.md` content).
- `docs/guide/configuration.md`, `src/generated/docs-content.ts`: docs note (regenerated).

## Test plan

- [x] New e2e: `pnpm vitest run --config vitest.e2e.config.ts src/e2e/e2e-rules.spec.ts -t "fold target folds"`
- [x] Manual: built CLI warns for `--targets codexcli,zoocode`, stays quiet for `zoocode,codexcli`, `zoocode` alone, and `codexcli,junie`


dyoshikawa:
@dyoshikawa Thank you!

Issue #2983, opened 2026-09-10
Guidance wanted: keeping many consumer repositories in sync with a shared source


sugurutakahashi-1234:
## Context

This is a question about recommended practice rather than a bug report.

We maintain a shared rules repository that about 15 other repositories consume through declarative `sources`. Everything inside a single repository is well covered by the docs, and we follow it as written: no tag on `source` so it tracks the default branch, `rulesync.lock` committed, `install --update` to move forward, `doctor --strict && install --frozen && generate --check` in CI.

What the docs do not cover is the layer above that: what to do after the shared repository changes and 15 lockfiles are now behind.

## What we could not find guidance on

**1. How is the update supposed to be triggered?**

`install --update` is a deliberate, per-repository action, which matches how npm and Bun lockfiles work, so we initially assumed the intent is "update when you want to". We briefly ran a scheduled GitHub Actions workflow that ran `install --update` weekly and opened a PR, then removed it again because nothing in the docs suggests rulesync is meant to be driven that way, and it felt like grafting a Renovate-shaped workflow onto a tool that deliberately does not have one.

Is manual, on-demand updating the intended model? Or do larger adopters run something scheduled?

**2. How do you see which repositories are behind?**

There is no command that answers "is my lockfile behind its source?" without mutating anything. `--dry-run` covers `generate`, not source resolution. We ended up comparing each lockfile's `resolvedRef` against `gh api repos/<owner>/<repo>/commits/HEAD` ourselves.

A read-only "sources are N commits behind" report — either a flag on `install` or a check inside `doctor` — would remove that scripting. Would that fit the tool's scope?

**3. Updating many repositories at once**

We wrote our own script that walks sibling clones, skips dirty working trees, runs `install --update && generate`, verifies, and commits. It works, but every organization consuming a shared source presumably writes the same script. The case studies list several companies doing centralized rules, so we suspect this problem is common.

We are not asking rulesync to orchestrate multiple repositories. A short FAQ entry describing the intended workflow — even if the answer is "run it per repository, on demand" — would settle it, because right now each team has to guess.

## Why we are asking rather than just doing our own thing

We already have a working setup, so this is not blocking us. The reason for the question is that we keep re-deriving these decisions from first principles and are not sure whether we are fighting the tool's design or filling a gap it intentionally leaves to users. Knowing which of the two it is would let us stop second-guessing the setup.

## Environment

- rulesync 16.26.1
[...]

PR #3033, opened 2026-09-13
docs(faq): describe keeping many repositories in sync with a shared source

dyoshikawa:
## Summary

Adds a FAQ entry, **How do I keep many repositories in sync with a shared source?**, answering the three questions raised in #2983 by a team that consumes one shared rules repository from ~15 consumer repositories:

1. **How is the update triggered?** Rulesync stops at the repository boundary: `rulesync.lock` moves only when `rulesync install --update` runs in that repository, the same model as an npm or Bun lockfile. Both on-demand runs and a scheduled CI job that opens a pull request are supported ways to drive it.
2. **How do you see which repositories are behind?** There is no read-only staleness command today (`generate --dry-run` covers generation, `install --frozen` checks coverage, not currency). The entry shows the two workable checks: compare the lockfile's `resolvedRef` with the source branch head, or run the scheduled job without committing and read its diff.
3. **Updating many repositories at once** is a loop over clones that does per repository what one repository does; rulesync does not orchestrate it.

The entry also documents one detail that makes a scheduled job noisy if missed: `install --update` stamps a fresh `resolvedAt` on every re-resolved source, so the lockfile changes even when no `resolvedRef` did. The suggested `git diff --quiet -I '"resolvedAt"' -- rulesync.lock rulesync-npm.lock.json` was verified locally to ignore a timestamp-only change and to detect a `resolvedRef` change (`git diff -I` needs Git 2.30 or newer).

`src/generated/docs-content.ts` is regenerated so `rulesync docs` embeds the new entry.

## Related issue

Refs #2983. The issue is intentionally left open: its second question doubles as a proposal for a read-only "sources are N commits behind" report (an `install` flag or a `doctor` check), which is a scope decision for the maintainer rather than something this docs change settles.

## Verification

- `pnpm cicheck` passes (content and code).
- The `git diff -I` example was exercised against a throwaway repository with a timestamp-only change (exit 0) and a `resolvedRef` change (exit 1).


dyoshikawa:
@dyoshikawa Thank you!

PR #3036, opened 2026-09-13
feat(init): seed targets and features from a user-level rulesync.jsonc

### LINKED ISSUE #3019: Feature Request: Support global user config / template for `rulesync init` (`targets` & `features`)
### Problem
Whenever running `rulesync init` in a new project, it generates a `rulesync.jsonc` with hardcoded defaults:
```jsonc
"targets": ["codexcli", "claudecode", "opencode"],
"features": ["rules", "mcp", "subagents", "skills", "hooks", "permissions"]
```

For developers who use a different set of coding agents (e.g. `antigravity-cli`, `cursor`, `pi`) or only need a subset of features (e.g. only `rules` and `skills`), this requires manually editing both `targets` and `features` on every project initialization.

### Proposed Solution: Global User Config / Template
Allow `rulesync init` to read defaults from a global user configuration file if present, such as:
- `~/.config/rulesync/rulesync.jsonc` (or `~/.rulesync/rulesync.jsonc`)

#### Desired Behavior:
1. When running `rulesync init`:
   - If a global user configuration exists:
     - Use its **`targets`** (array or object form) and **`features`** as the default template for the newly created `rulesync.jsonc`.
     - (Optionally) Inherit other common preferences like `delete`, `simulate*`, etc.
   - If no global configuration exists:
     - Fall back to the current hardcoded defaults (`codexcli, claudecode, opencode`).
2. This provides a true "set once and forget" experience across all new repositories.

### Code Reference
The template values are currently hardcoded in `src/lib/init.ts`:
```ts
// src/lib/init.ts in createConfigFile()
await writeFileContent(
  path,
  JSON.stringify(
    {
      $schema: RULESYNC_CONFIG_SCHEMA_URL,
      targets: ["codexcli", "claudecode", "opencode"],
      features: ["rules", "mcp", "subagents", "skills", "hooks", "permissions"],
      outputRoots: ["."],
      delete: true,
      // ...
    }
  )
);
```

dyoshikawa:
## Summary

`rulesync init` hardcoded the `targets` and `features` it writes into a new `rulesync.jsonc`, so developers who use a different set of tools had to edit both keys after every `init`. This PR lets `init` seed the two keys from a user-level config file.

- `init` reads `$XDG_CONFIG_HOME/rulesync/rulesync.jsonc` (`~/.config/rulesync/rulesync.jsonc` when `XDG_CONFIG_HOME` is unset or relative) and copies its `targets` and `features` into the new project config. A key the file omits keeps the built-in default; the object form of `targets` is copied as-is and the top-level `features` key is then left out.
- Only `targets` and `features` are inherited. The file is a template for `init`, not a config layer: `generate` and the other commands never read it.
- A user config that exists but fails to parse, names an unknown target, or combines object-form `targets` with `features` makes `init` fail with the file path in the message instead of silently writing the defaults.
- The CLI logs `Seeded targets and features from <path>` (and captures `seededFrom` in `--json` mode) when the file was used.
- Docs: new `Init Command` section in `docs/reference/cli-commands.md` and a pointer in `docs/guide/configuration.md`.

Closes #3019

## Test plan

- [x] `pnpm cicheck` passes
- [x] `src/lib/init.test.ts`: path resolution (default, absolute `XDG_CONFIG_HOME`, relative ignored), seeding both keys, partial keys, object-form `targets`, file without either key, existing `rulesync.jsonc` skipping the read, malformed / invalid / mutually-exclusive user configs
- [x] `src/cli/commands/init.test.ts`: the `Seeded ...` message is logged only when a user config was used


dyoshikawa:
@dyoshikawa Thank you!

Issue #3045, opened 2026-09-14
Follow-ups from the v16.31.0 target reviews (Bob, Tabnine, Cortex Code): accepted low findings


dyoshikawa:
## Background

PRs #3037 (IBM Bob follow-ups), #3040 (Tabnine CLI target) and #3041 (Snowflake Cortex Code target) were merged after their review rounds came back with only `low` findings. Those findings were accepted rather than fixed inline so the targets could ship in v16.31.0; this issue collects them in one place so they are not lost. None of them is a defect that changes generated output for a correctly authored source file — they are wording drifts, duplicated code, missing warnings for odd inputs, and small test gaps.

### Snowflake Cortex Code (#3041)

- `docs/reference/file-formats.md` — the Cortex Code MCP note says a project-scope run is "rejected with an explanatory error", but `src/lib/generate.ts` (around the `Target '<t>' does not support the feature` branch) skips the target with a warning instead. The prose should say "skipped with a warning".
- `src/features/mcp/cortexcode-mcp.ts` (`convertStdioServer`, ~L136-148) — a server with both a stdio `command` and a `url` gets a warning whose wording contradicts what is then written.
- `src/features/mcp/cortexcode-mcp.ts` `fromFile` (~L255-266) parses the file twice; `kimi-code-mcp.ts` (~L343) shows the single-parse shape.
- `src/features/mcp/rulesync-mcp.ts` — the `cortexcode` tool-scoped key is out of alphabetical order in the schema.
- `src/features/hooks/cortexcode-hooks.ts` is a near-duplicate of `bob-hooks.ts`; together with `claudecode-hooks.ts` and `tabnine-hooks.ts` (and now `continue-hooks.ts`) the Claude-Code-compatible settings.json hooks adapters are candidates for one factory.
- Unknown MCP server keys are passed through without sanitising a nested `__proto__` (shared with the tabnine and bob adapters).
- `warnAndSkipMcpServer` in `src/features/mcp/mcp-transport.ts` (~L83) interpolates the raw server name into the warning; it should go through `quoteValueForWarning`.
- `resolveLocalMcpCommand` flattens `command` and `args`, so a stdio entry with `args` but no `command` is emitted with `args[0]` as its command in the cortexcode and bob adapters (the continue adapter guards against this; see `continue-mcp.ts` `convertStdioServer`).

### Tabnine CLI (#3040)

- The override warning for `run_shell_command(git *)` prints the prefix as `git * *`.
[...]

PR #3041, opened 2026-09-14
feat: add Snowflake Cortex Code target (rules, mcp, subagents, skills, hooks)

dyoshikawa:
## Summary

Adds Snowflake Cortex Code (`cortexcode`) as a target for rules, MCP, subagents, skills and hooks.

Closes #3012

## What is generated

| Feature | Project | Global | Notes |
| --- | --- | --- | --- |
| rules | `AGENTS.md` | — | Cortex Code reads `AGENTS.md` at the project root; only the root rule is written, non-root rules are folded in the usual way. |
| mcp | — | `~/.snowflake/cortex/mcp.json` | Cortex Code documents MCP servers only in the user config; the `type` field (`stdio` / `http` / `sse`) is always written explicitly. |
| subagents | `.cortex/agents/*.md` | `~/.snowflake/cortex/agents/*.md` | Markdown with `name` / `description` frontmatter; a `cortexcode` frontmatter block passes tool-specific keys through. |
| skills | `.cortex/skills/<name>/SKILL.md` | `~/.snowflake/cortex/skills/<name>/SKILL.md` | Directory skills with supporting files copied verbatim. |
| hooks | `.cortex/settings.json` (`hooks` key) | `~/.snowflake/cortex/hooks.json` (`{ "hooks": ... }`) | Both go through the shared-config gateway so sibling keys survive; the global file is never deleted. |

## Design decisions

- **Permissions are not implemented.** The Cortex Code docs describe three mutually inconsistent permission formats, and the CLI rewrites its own permission cache at runtime, so there is no stable file rulesync could own. Left as a follow-up once the format settles.
- **Commands are not implemented.** Custom slash commands are not documented for Cortex Code.
- **No ignore file.** Cortex Code has no documented ignore file.
- **Rules are project-only `AGENTS.md`.** No global rules location is documented.
- **MCP is global-only** with an explicit `type` per server, since Cortex Code documents the transport as required.
- **Hooks:** the `*` matcher is written verbatim; a matcher on `UserPromptSubmit` / `Stop` is dropped with a warning because those events carry no tool name to match; the global `hooks.json` is written as `{ "hooks": ... }` via the gateway and is not deletable (it may carry hand-written content).

## Checks

- `pnpm cicheck` passes.
- Tool x Feature e2e happy-path specs added for rules, mcp, subagents, skills and hooks (`pnpm vitest run --config vitest.e2e.config.ts`: 5 files / 663 tests pass).
[...]

PR #3040, opened 2026-09-14
feat: add Tabnine CLI target

dyoshikawa:
## Summary

Adds **Tabnine CLI** (`--targets tabnine`) as a native target covering all eight feature dimensions, in both project and global scope.

Closes #3013

## What is generated

| Feature | Project | Global (`--global`) |
| --- | --- | --- |
| rules | `TABNINE.md` (root, lists the guideline files) + `.tabnine/guidelines/*.md` | `~/.tabnine/agent/TABNINE.md` + `~/.tabnine/guidelines/*.md` |
| ignore | `.tabnineignore` | — |
| mcp | `mcpServers` in `.tabnine/agent/settings.json` | `mcpServers` in `~/.tabnine/agent/settings.json` |
| commands | `.tabnine/agent/commands/<name>.toml` | `~/.tabnine/agent/commands/<name>.toml` |
| subagents | `.tabnine/agent/agents/<name>.md` | `~/.tabnine/agent/agents/<name>.md` |
| skills | `.tabnine/agent/skills/<name>/SKILL.md` | `~/.tabnine/agent/skills/<name>/SKILL.md` |
| hooks | `hooks` in `.tabnine/agent/settings.json` | `hooks` in `~/.tabnine/agent/settings.json` |
| permissions | `tools.allowed` / `tools.exclude` in `.tabnine/agent/settings.json` | same in `~/.tabnine/agent/settings.json` |

All `settings.json` writes go through the shared-config gateway (deep merge of the owned keys only), so `hooksConfig`, `general`, `context`, `experimental`, ... survive a regenerate and the file is never deleted.

## Design decisions

- **Rules**: Tabnine CLI auto-loads only the root `TABNINE.md`, so non-root rules are written to `.tabnine/guidelines/` (the directory the Tabnine agent documents) and listed from `TABNINE.md` (toon mode). The user-level `~/.tabnine/agent/TABNINE.md` path is inferred from the Gemini CLI convention Tabnine CLI derives from, since Tabnine documents no user memory file.
[...]

PR #3037, opened 2026-09-14
feat: add IBM Bob target (rules, ignore, mcp, commands, skills, hooks)

dyoshikawa:
## Summary

Adds the `bob` target for **IBM Bob** (IBM's agentic coding assistant, available as an IDE extension and the Bob Shell CLI), covering rules, ignore, MCP, commands, skills and hooks.

Closes #3011

## Configuration surface

| Feature | Project | Global |
| --- | --- | --- |
| rules | `AGENTS.md` (root) + `.bob/rules/*.md` | `~/.bob/AGENTS.md` + `~/.bob/rules/*.md` |
| ignore | `.bobignore` | — (Bob documents no user-level ignore file) |
| mcp | `.bob/mcp.json` | `~/.bob/mcp.json` |
| commands | `.bob/commands/**/*.md` | `~/.bob/commands/**/*.md` |
| skills | `.bob/skills/<name>/SKILL.md` | `~/.bob/skills/<name>/SKILL.md` |
| hooks | `.bob/settings.json` (`hooks` key) | `~/.bob/settings/settings.json` (`hooks` key) |

Subagents and permissions are not part of Bob's documented file surface and are left unsupported.

## Notable decisions

- **Rules** are plain Markdown with no frontmatter; rulesync frontmatter is stripped on generate. Bob's mode-specific `.bob/rules-{mode}/` directories are not emitted.
[...]

PR #3048, opened 2026-09-14
fix(mcp): quote skipped server names and reject args-only stdio entries

dyoshikawa:
## Summary

Lands the mechanical items from steps 1 and 3 of #3045 (review follow-ups accepted when PRs #3037, #3040 and #3041 shipped):

- `warnAndSkipMcpServer` (`src/features/mcp/mcp-transport.ts`) now quotes the server name through `quoteValueForWarning` instead of interpolating the raw string, so a crafted name cannot carry a line break or control characters into the warning.
- `resolveLocalMcpCommand` now returns an empty array for an entry that carries `args` but no `command`. It used to merge the two, so every adapter that writes the resolved command back out (Bob, Cortex Code, Tabnine CLI, Zed, Copilot CLI, Muse Code, ZCode as `command`/`args`; Kilo and OpenCode as a `command` array) promoted `args[0]` to the program. All of them now skip such an entry with their existing "without a command" warning, and Continue's inline guard is replaced by the shared one.
- Continue's `envFile` warning quotes the server name through `quoteValueForWarning` like the skip warnings do.
- Bob MCP JSDoc says `env`/`headers` are sanitized (which the code already did), and the Cortex Code MCP note in `docs/reference/file-formats.md` says a project-scope run skips the target with a warning (what `src/lib/generate.ts` does) instead of "rejected with an explanatory error".

Not covered here, so #3045 stays open: the hooks-adapter factory / `sanitizeEnv` dedupe, nested `__proto__` in passthrough keys, the Tabnine permissions warning sweep, and the Tabnine import alignments.

Refs #3045

## Test plan

- [x] `pnpm cicheck` passes (new unit tests for the args-only `resolveLocalMcpCommand` case and the quoted warning; an args-only stdio server added to the skip tests of all ten adapters)


dyoshikawa:
@dyoshikawa Thank you!

Issue #2959, opened 2026-09-08
Follow up AugmentCode upstream updates: the augmentcode-plugin target and the plugin-consumption settings keys are untracked after #2670 closed


dyoshikawa:
## Summary

Re-checked `augmentcode` on **2026-09-08**. There is **no new upstream release** in the window — `@augmentcode/auggie` `dist-tags.latest` is still **0.36.0 (2026-08-21)**, and the 0.37.0 prerelease line contains no config surface. Every implemented dimension (rules, ignore, mcp, commands, subagents, skills, hooks, permissions, checks) remains in parity.

What this issue is for: **#2670 closed on 2026-09-03 as `completed`, but its own design question was never decided.** The closing commit `4a88924` says so explicitly in its message — *"Refs #2670 (partial: the data-loss bug reported in a comment on that issue; the issue's own design question about a new augmentcode-plugin target is unrelated and remains open)"* — and the 2026-08-13 triage comment on #2670 had recorded *"classified as design, not implemented … Left open for that call."* The closure followed #2913 (the hook-key data-loss fix), not the decision. Two `unsupported` capability gaps are consequently untracked, plus one reference-map defect that will suppress future checks.

Filed as a new issue because no AugmentCode issue is currently open (all of #2670, #2917, #2420, #2234, #2150, #2134, #1890, #1822, #1784, #1758, #1700 are closed).

## Recent Releases

| Release | Date | Config-surface impact |
| --- | --- | --- |
| `0.36.0` — current `latest` | 2026-08-21 (npm publish); [changelog](https://www.augmentcode.com/changelog/auggie-cli-0-36-0-release-notes) dates it 2026-08-22 | **None.** A `--reasoning-effort` flag plus fixes: session-replay crashes, `secret get --reveal` removal, git version preflight, agent file-view range parsing, MCP header-secret error messages, prompt-module truncation, VFS sync, dependency CVEs. |
| `0.37.0-prerelease.202609072259` — current `prerelease` | 2026-09-07 | **None.** See below. |

`npm view @augmentcode/auggie dist-tags` → `latest: 0.36.0`, and https://www.augmentcode.com/changelog lists nothing after the 0.36.0 entry, so 0.36.0 is still the newest stable as of 2026-09-08.

[...]

PR #3056, opened 2026-09-15
fix(augmentcode): import current tool-name aliases as managed permissions

dyoshikawa:
## Summary

Refs #2959 — resolves the third ("safe now") item of the 2026-09-12 follow-up comment: importing AugmentCode's current tool names. The issue stays open for the remaining two gaps, which need a maintainer decision on when to switch generation to the current names.

AugmentCode's [permissions reference](https://docs.augmentcode.com/cli/permissions) renamed `launch-process` / `view` / `str-replace-editor` / `save-file` to `terminal` / `read` / `edit` / `write` and lists the legacy names as aliases of them. Rulesync only knew the legacy spellings, so an existing `.augment/settings.json` authored with the current names was mishandled:

- a `terminal` entry was imported as an unmanaged pass-through tool `terminal`, losing its `shellInputRegex` → `bash` pattern recovery;
- `read` / `edit` / `write` rows imported correctly only by coincidence of the canonical names, but were unmanaged on regenerate, so a stale `terminal` / `read` / `edit` / `write` `allow` row survived every regenerate next to the generated legacy row (first-match-wins made the stale row shadow it).

## Changes

- `src/features/permissions/augmentcode-permissions.ts`: add `CURRENT_TO_LEGACY_AUGMENT_TOOL_NAMES` and resolve it when reading existing entries — `toCanonicalToolName` maps both spellings to the same category, the preservation filter treats aliased rows as managed (permissive rows replaced, `deny` rows kept unless a generated row re-emits the same shape) and the `shellInputRegex` recovery accepts `terminal`. Generation still emits the legacy names, which the shipped CLI (0.36.0) matches exactly.
- Tests for importing current names (including regex recovery on `terminal`), folding an alias with its legacy twin under most-restrictive-wins, and the regenerate preservation policy for aliased rows.
- `docs/reference/file-formats.md` AugmentCode permissions note updated; `src/generated/docs-content.ts` regenerated.

## Verification

- `pnpm cicheck` passes locally.


dyoshikawa:
@dyoshikawa Thank you!

PR #3063, opened 2026-09-16
refactor(watch): replace node:fs watch with chokidar

### LINKED ISSUE #3062: Watch mode: .git churn triggers regeneration when the source tree is the repo root, and symlinked source directories are never watched
## Summary

`rulesync generate --watch` regenerates on every write Git makes inside `.git/` when the source tree is the repository root, and it never notices edits behind a symlinked source directory. Both come from the watcher being a recursive `fs.watch` with no way to prune subtrees and no symlink following.

## Motivation / Purpose

I run `rulesync generate --watch --global` as a background service against a repo where the rulesync source tree is the repo root, a layout the [separate input root guide](https://github.com/dyoshikawa/rulesync/blob/main/docs/guide/separate-input-root.md) supports:

```
repo/                                # inputRoots: ["."]
├── .git/
│   ├── fsmonitor--daemon/cookies/   # ~54% of all wakeups
│   ├── objects/maintenance.lock     # ~15%
│   └── refs/copilot/…/base.lock     # ~13%
├── rules/
│   └── overview.md                  # the only thing that should matter
├── skills/
├── README.md                        # also watched
└── rulesync.jsonc
```

After about 17 hours the log held roughly 400 regenerations. Around 90% were Git writing inside `.git/`: fsmonitor daemon cookies, scheduled-maintenance locks, and the lock files GitHub Copilot keeps under `refs/copilot/`. Nine touched a file rulesync reads. The watcher wakes every three minutes, re-reads every source file, writes nothing, and appends another "up to date" line to a log that grows forever. No config key or ignore file scopes the watch set.

The obvious workaround is worse. Moving the sources behind a symlink so that only `.rulesync/` is watched:

```
repo/
├── .rulesync/                       # inputRoots: [".rulesync"]
│   └── rules -> ../rules            # generate reads through it; the watcher never does
├── rules/
│   └── overview.md                  # edits here are never noticed
└── rulesync.jsonc
```

Edits to `rules/overview.md` stopped reaching the outputs, because recursive `fs.watch` does not descend into symlinked directories. The failure is silent: the watcher keeps running and reports nothing, while every agent on the machine reads stale instructions. That drift is exactly what `--watch` exists to prevent.

## Details

Reproduce the `.git` case:

```sh
git init
printf '{ "inputRoots": ["."], "targets": ["claudecode"], "features": ["rules"] }\n' > rulesync.jsonc
mkdir rules
printf -- '---\nroot: true\ntargets: ["*"]\n---\n\nHello.\n' > rules/overview.md
rulesync generate --watch
```

Commit in another terminal, or write a README. Expected: nothing. Actual: "Change detected" and a full regeneration.

Reproduce the symlink case:

```sh
mkdir -p shared-rules .rulesync
printf -- '---\nroot: true\ntargets: ["*"]\n---\n\nHello.\n' > shared-rules/overview.md
ln -s ../shared-rules .rulesync/rules
rulesync generate --watch
```

Edit `shared-rules/overview.md`. Expected: regeneration. Actual: nothing, even though the initial run generated from it.

Proposed fix: inside an input root, watch only the top-level entries generate reads (`rules/`, `commands/`, `subagents/`, `checks/`, `skills/`, and the MCP, hooks, permissions, and `.aiignore` files). Prune everything else at the root level before it is ever watched. Follow symlinks below those entries. Node's recursive watch can do neither, so the watcher itself has to change.

## Additional Context

In #2477, using [chokidar](https://github.com/paulmillr/chokidar) was suggested:

> Pick the watching mechanism: Node's `fs.watch` is unreliable across platforms (duplicate/missing events, recursive support), so an established watcher such as chokidar is likely worth the dependency; evaluate against the current dependency policy.
>
> — @dyoshikawa

#2488 went with `fs.watch` to avoid a new dependency. The following day brought #2516, #2520, #2531, #2532, and #2536 chasing dead watchers and test races. chokidar has been pure JavaScript with a single dependency since v4. It already appears in this repo's lockfile through another dependency, so adopting it adds no packages to the install. I have the swap implemented and will open a PR against this issue.

austintraver:
## Background

Closes #3062

I run `rulesync generate --watch --global` as a background service on a repository whose root is the source tree, via `inputRoots: ["."]`. Recursive `fs.watch` then covers everything.

```
.
├── .git/
│   ├── fsmonitor--daemon/cookies/   # ~54% of regenerations
│   ├── objects/maintenance.lock     # ~15%
│   └── refs/copilot/…/base.lock     # ~13%, Copilot's lock files
├── README.md                        # watched, never read
├── rules/                           # read by generate
└── rulesync.jsonc
```

Over about 17 hours the log recorded roughly 400 regenerations. Nine touched a file rulesync reads. Nearly all the rest were Git writing inside `.git/`. Each spurious regeneration re-reads every source file and writes nothing. The log gains one more "up to date" line each time. No config key or ignore file scopes the watch set.

I tried to escape by moving the sources behind a symlink: `inputRoots: [".rulesync"]` with `.rulesync/rules -> ../rules`.

```
.
├── .rulesync/
│   └── rules -> ../rules   # generate reads through it; fs.watch stops here
├── rules/
│   └── overview.md         # edits here are never noticed
└── rulesync.jsonc
```

Edits to `rules/overview.md` stopped reaching the outputs, because recursive `fs.watch` does not descend into symlinked directories. The failure is silent: the watcher keeps running and reports nothing, while every agent on the machine reads stale instructions. That drift is exactly what `--watch` exists to prevent.

In #2477, the design issue for watch mode, [chokidar](https://github.com/paulmillr/chokidar) was already suggested:

> Pick the watching mechanism: Node's `fs.watch` is unreliable across platforms (duplicate/missing events, recursive support), so an established watcher such as chokidar is likely worth the dependency; evaluate against the current dependency policy.
>
> — @dyoshikawa

#2488 implemented watch mode with `fs.watch` and no new dependency. The following day brought #2516, #2520, #2531, #2532, and #2536, all chasing dead watchers and test races.

Measured against the dependency policy, chokidar 5 is pure JavaScript and ESM-only. Its only dependency is `readdirp`. Both packages are already in `pnpm-lock.yaml` through another dependency. `pnpm install` adds no new packages.

## What this PR does

This PR is that implementation. `rulesync generate --watch` now runs on chokidar 5.0.0.

The larger shift is what gets watched under each input root. Instead of blacklisting known noise, the watcher whitelists the top-level entries `generate` reads: `rules/`, `commands/`, `subagents/`, `checks/`, `skills/`, the `mcp`, `hooks`, and `permissions` JSON and JSONC files, and `.aiignore`. Everything else at the root level is pruned by chokidar's `ignored` predicate during traversal. A pruned directory is never opened. Below a whitelisted entry, everything is watched. Symlinks are followed.

```
.
├── .git/                        # pruned, never opened
├── README.md                    # pruned
├── commands/                    # watched, recursively
├── rules/                       # watched, recursively
├── skills -> ../shared-skills   # watched, recursively, through the link
├── mcp.jsonc                    # watched
└── rulesync.jsonc               # watched by the existing config-file target
```

The whitelist is not new data. `rulesync fetch` already kept this per-feature table as a local `FEATURE_PATHS` in `src/lib/fetch.ts`. It now lives in `src/constants/rulesync-paths.ts` as `FEATURE_SOURCE_TREE_ENTRIES`, keyed by the `Feature` union. Fetch and watch both read it. Adding a feature fails to compile until its entries are listed.

Chokidar signals readiness asynchronously through a `ready` event. `WatchHandle` gains a `ready` promise. The `generate` command awaits it before printing the "Watching for changes in:" banner. An edit made after the banner is guaranteed to be seen.

Chokidar, like `fs.watch`, goes quiet when the watched root directory itself is deleted and recreated. The re-arm loop from #2516 and #2531 stays wrapped around the chokidar watcher.

Filtering `fs.watch` events would have fixed the log but not the link. An edit behind a symlink never produces an event, so there is nothing to filter.

## Tests

Two new cases in `src/lib/watch.test.ts`:

- Under an input root, `.git/index` and `README.md` are ignored while `rules/a.md` and `mcp.jsonc` are watched.
- An edit behind a symlinked `rules/` directory is reported. Skipped on Windows, where creating a symlink needs privileges.

The existing watch cases for the debounce, the re-arm loop, and the config-file target pass unchanged, as do the fetch, generate, and end-to-end watch tests. Full `pnpm cicheck` passes.

By hand, against the built CLI: in the repo-root layout a `git commit` and a README edit produce no regeneration and a rule edit does. In the symlink layout the initial run reads through the link and an edit behind it regenerates.


src/lib/watch.ts
```diff
@@ -271,7 +276,7 @@ function watchTargetWithRearm({
         return;
       }
     }
-    watcher.close();
+    void watcher.close();
```
dyoshikawa: If the target is deleted while chokidar is still doing its initial scan, the liveness timer lands here and `close()` runs before `ready` ever fires. chokidar's `close()` calls `removeAllListeners()`, so the `once("ready", markReady)` registered in `attach()` is dropped and `markReady` is never called on this path. The re-armed watcher registers a fresh listener, but only once the directory comes back, so in the meantime `generate.ts:346` sits on `await handle.ready` with no banner and no SIGINT/SIGTERM handlers installed. I reproduced this with a ~2,000-file tree.

Calling `markReady()` here (or in the `attach` throw path in `scheduleRearm`) fixes it; the semantics "resolves once every target that exists is being watched" still hold because a missing target counts as done. A regression test for this would be good to have.

src/constants/rulesync-paths.ts
```diff
@@ -84,6 +86,19 @@ export const RULESYNC_MCP_LEGACY_FILE_NAME = "mcp.json";
...
 
+// Top-level entries of a rulesync source tree, per feature.
+export const FEATURE_SOURCE_TREE_ENTRIES: Record<Feature, readonly string[]> = {
+  rules: [RULES_FEATURE_SUBDIR],
+  commands: [COMMANDS_FEATURE_SUBDIR],
+  subagents: [SUBAGENTS_FEATURE_SUBDIR],
+  skills: [SKILLS_FEATURE_SUBDIR],
+  checks: [CHECKS_FEATURE_SUBDIR],
+  ignore: [RULESYNC_AIIGNORE_FILE_NAME],
+  mcp: [RULESYNC_MCP_FILE_NAME, RULESYNC_MCP_LEGACY_FILE_NAME],
```
dyoshikawa: `RulesyncMcp.getSettablePaths()` still lists `.rulesync/.mcp.json` as a (deprecated) legacy path and `fromFile` reads it with a warning, so on main an edit there triggered regeneration and here it silently doesn't. Adding `".mcp.json"` to this list closes the gap. Note that `fetch` shares this table, so that also makes `rulesync fetch` pick up a remote `.mcp.json`; I think that's acceptable, but if you'd rather keep fetch unchanged, split the watch set out or derive it from each feature's `getSettablePaths()` the way `gitignore-derive.ts` does.

src/lib/watch.ts
```diff
@@ -195,34 +206,28 @@ function watchTargetWithRearm({
...
+    });
+    created.on("all", (_event, path) => {
+      onChange({ path });
+      verifyStillWatching();
+    });
     created.on("error", (error) => {
       onError({ error, directory: target.directory });
       verifyStillWatching();
     });
+    created.once("ready", markReady);
```
dyoshikawa: Small ordering thing related to this `ready` listener: in `scheduleRearm`, the `onChange` right after `attach()` fires before the new chokidar instance has finished its initial scan, so an edit made between then and its `ready` isn't reported. In practice the regeneration that `onChange` triggers reads the edit anyway unless the scan takes longer than the 300 ms debounce, so it's minor; deferring that `onChange` until the new watcher's `ready` would close it.

src/lib/watch.ts
```diff
@@ -285,17 +290,19 @@ function watchTargetWithRearm({
     try {
       attach();
     } catch (error) {
-      // If the directory disappeared between the existence check and
-      // `fs.watch`, treat it like any other temporarily absent overlay.
+      // If the directory disappeared between the existence check and the
+      // watcher's start, treat it like any other temporarily absent overlay.
       // Permission and platform errors for a still-existing directory remain
       // real attachment failures and must propagate.
```
dyoshikawa: chokidar's `watch()` doesn't throw synchronously for EACCES and similar; those arrive as an async `error` event. So this branch is effectively only reachable through the mocks now, and an unreadable root logs "Watch error on ..." and leaves a permanently inert watcher instead of failing at startup. Not a blocker since the initial `generateOnce` fails first, but the comment should describe the new behavior.

src/cli/commands/generate.ts
```diff
@@ -343,6 +343,7 @@ async function generateWatchCommand(logger: Logger, options: GenerateOptions): P
       logger.error(`Watch error on ${directory}: ${formatError(error)}`);
     },
   });
+  await handle.ready;
```
dyoshikawa, line 346: This changes the user-visible watch contract, and `docs/reference/cli-commands.md` (around line 259) still says the `.rulesync/` source tree is watched recursively. Please update it to describe the per-input-root whitelist (and that symlinked entries are followed) and run `pnpm run generate:docs-content`, since `src/generated/docs-content.ts` is checked for drift in CI.

dyoshikawa (commented):
Thanks for this. I reproduced the two symptoms from #3062 and the whitelist + symlink-following approach fixes both; the fetch refactor is behavior-preserving and I checked the chokidar/readdirp integrity against the registry. Three things need to change before I can merge, and I left inline notes on each: `handle.ready` can stall forever when a target is deleted during chokidar's initial scan (Copilot's second comment is right), the legacy `.rulesync/.mcp.json` dropped out of the watch set, and the CLI docs still describe the old recursive contract.

Two smaller items you can take or leave: the post-re-arm `onChange` fires before the new watcher is ready (only bites when the initial scan outlasts the 300 ms debounce), and the "must propagate" comment in the attach error path is now stale since chokidar reports EACCES-style failures through the async `error` event instead of throwing.

One heads-up on the dependency note in the description: chokidar and readdirp are indeed already in the lockfile, but via `repomix`, which is a devDependency. For people installing the published package this adds two runtime dependencies. That's fine with me given what it buys, just wanted the record straight.

dyoshikawa:
Thanks for the migration, @austintraver! To get this landed I pushed the review follow-ups directly onto your branch (your commit is kept as-is):

- `e95e91a7` fix(watch): never leave the `ready` promise pending — `markReady()` on every detach (liveness sweep or handle close), the post-re-arm regeneration now waits for the new watcher's initial scan, and the stale "must propagate" comment is corrected. Two regression tests cover the deleted-during-scan and closed-before-ready cases.
- `501c2868` fix(watch): keep the deprecated `.rulesync/.mcp.json` in the watch set — added as `RULESYNC_MCP_DEPRECATED_DOTFILE_NAME` to the shared `FEATURE_SOURCE_TREE_ENTRIES` table (so `fetch` picks it up too, as discussed), plus the watch-mode docs update and regenerated `docs-content.ts`.

`pnpm cicheck` passes locally; merging once CI is green.

dyoshikawa:
@austintraver Thank you!

PR #3064, opened 2026-09-16
ci(draft-release): inline the opencode wrapper to drop the last Node 20 action

saitota:
## Why

GitHub removes Node 20 from the Actions runner on 2026-09-23 ([changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)). The only Node 20 action left in this repo is `actions/cache@v4`, referenced by mutable tag inside the `anomalyco/opencode/github` composite wrapper; the last Draft Release run ([26214698542](https://github.com/dyoshikawa/rulesync/actions/runs/26214698542)) warns on exactly that. Upstream still uses `@v4` on `dev` and in v2.0.3, so bumping the pin cannot fix it. Inlining the wrapper's steps is the only way to control that dependency, and it also makes every action in the job SHA-pinned.

## Verification

- `pnpm cicheck` green, `actionlint` clean.
- Inlined steps run twice on my fork: install on cache miss ([35056570000](https://github.com/saitota/rulesync/actions/runs/35056570000)), restore on cache hit ([35056606488](https://github.com/saitota/rulesync/actions/runs/35056606488)). No Node.js 20 annotation on either.
- After merge: dispatch Draft Release once and confirm the job has no "Node.js 20 actions are deprecated" annotation. `opencode github run` receives the same env as before, so no behaviour change is expected; not verified end-to-end here (maintainer-gated).


saitota:
Closing this. On reflection the fix belongs in the `anomalyco/opencode/github` action itself (bump its `actions/cache@v4` to a Node 24 release). 
 The request is tracked at anomalyco/opencode#40101
Sorry for the noise.

PR #2083, opened 2026-06-30
chore: route npm installs through Takumi Guard

saitota:
## Background

PR #1567 added a cooldown to reduce exposure to newly published supply-chain attacks, but malicious packages can still remain after the cooldown window. Takumi Guard is an npm registry proxy that checks packages against a threat database and blocks known-malicious packages before they are installed.

## References

- Cooldown PR: https://github.com/dyoshikawa/rulesync/pull/1567
- Takumi Guard GitHub Action: https://github.com/flatt-security/setup-takumi-guard-npm#readme
- Takumi Guard npm quickstart: https://shisho.dev/docs/t/guard/quickstart/npm/

## Verification

- [x] Ran pinact run.
- [x] Ran pnpm run cicheck:code && pnpm run cicheck:content.
- [x] Ran pnpm run check.
- [x] Ran pnpm run test.
- [ ] Confirmed the updated GitHub Actions jobs on this PR.
- [ ] Confirmed bun install through Takumi Guard in CI.
- [ ] Confirmed the release publish workflow after merge.

PR #3067, opened 2026-09-16
ci: trace release and secret-bearing jobs with cicd-sensor (Takumi Runner)

saitota:
## Background

Takumi Guard (#2083) checks what gets installed; nothing observes what installed code does while a job runs. [Takumi Runner](https://flatt.tech/takumi/features/runner/) (GMO Flatt Security) offers a [cicd-sensor integration](https://shisho.dev/docs/t/runner/features/cicdsensor-integration/): an eBPF sensor on GitHub-hosted Linux runners that records process, file and network activity of a job. [Free tier](https://shisho.dev/docs/r/202609-takumi-runner-cicd-sensor-free-tier/): 3,000 traced job-minutes/org/month, no credit card; over the limit traces are dropped, jobs are unaffected.

## Scope

Job minutes over the last 31 days of this repo, rounded up per job. Only jobs holding write tokens or API keys are traced.

| Workflow / job | min/month | Traced | Why |
|---|---:|:---:|---|
| publish | 30 | ✅ | npm publish OIDC, `contents: write` |
| publish-assets | 84 | ✅ | tag push, release assets, attestation |
| security-scan | 270 | ✅ | 6 API-key secrets |
| draft-release | 0 | ✅ | API keys, `contents/pull-requests/issues: write` |
| ci (push) | 717 | ❌ | read-only, no secrets |
| ci (pull_request) | 2,073 | ❌ | same; fork PRs have no OIDC |
| e2e-binaries build (push / PR) | 692 / 2,025 | ❌ | read-only, no secrets |
| e2e-binaries test ubuntu (push / PR) | 2,144 / 6,268 | ❌ | same; macOS/Windows legs unsupported |
| docs build / deploy | 133 / 133 | ❌ | read-only build, Pages deploy only |
| pr-policy, auto-close-*, ci-failure-reminder, actionlint, trivy | 393 | ❌ | 1-minute `GITHUB_TOKEN` jobs, no installs |
| **Traced total** | **384** | | 13% of 3,000 |

## Changes

- Tracing is opt-in via the repo variable `SHISHO_BOT_ID`; unset, the step is skipped. If Takumi is down the job still completes.
- No job's runtime, matrix or release behavior changes.

Maintainer setup: create a Takumi org at https://cloud.shisho.dev/hello/takumi, add a bot with the "Takumi Runner Trace Sender" role and an OIDC trust condition for `dyoshikawa/rulesync` (audience left empty), set `SHISHO_BOT_ID` to the bot ID (public identifier, not a secret).

## Test Plan

- [x] `actionlint`, `pinact run --check`, `pnpm cicheck` (12,241 tests) green
- [x] Fork run without `SHISHO_BOT_ID`: step skipped, job continues — https://github.com/saitota/rulesync/actions/runs/35058521370
- [x] Fork run with an invalid `SHISHO_BOT_ID`: sign-in fails, job continues — https://github.com/saitota/rulesync/actions/runs/35060809480
- [ ] After setup: first release run shows a trace in Takumi Runner

Caveat: `cicd-sensor-action` is pre-release and `shisho-cloud-action` runs on Node 20 (removed from runners 2026-09-23). If either breaks, tracing stops silently; releases are unaffected.

Refs #2083


.github/actions/cicd-sensor/action.yml
```diff
@@ -0,0 +1,25 @@
...
+      if: inputs.bot-id != ''
+      uses: flatt-security/shisho-cloud-action@9b68d33503ebf75d7c285d489b288f96ef3df758 # v1.1.0
+      with:
+        bot-id: ${{ inputs.bot-id }}
+        export-token: true
+        expires-in-minutes: 360
+
+    - name: Start cicd-sensor
+      if: inputs.bot-id != ''
+      uses: cicd-sensor/cicd-sensor-action@6511eb44c91d71b2b93d71193b1bf2cb18352f66 # v0.0.38
```
dyoshikawa, line 103: This pin only covers the JS wrapper. At this commit, `src/main.js` `stageReleaseBinaries()` does `curl --fail --location` on `https://github.com/cicd-sensor/cicd-sensor/releases/download/releases/v0.0.45/cicd-sensor_0.0.45_linux_<arch>.tar.gz`, `tar xzf`, then `sudo install` and starts it as a root transient systemd unit — with no checksum or signature verification, even though the release publishes `checksums.txt.sigstore.json` and per-asset sigstore bundles. Anyone who can rewrite that release asset gets root (and every secret, plus the npm publish OIDC) inside `publish.yml` and `publish-assets.yml`.

Before this goes into the release jobs I'd want the agent binary verified: ideally upstream adds `cosign verify-blob` / `gh attestation verify` against the bundles it already ships, or we wrap the download ourselves and pass the verified path in. Our rules doc already spells out the same 'pin covers the wrapper only' caveat for opencode, and that one only runs in an actor-gated `workflow_dispatch`.

.github/actions/cicd-sensor/action.yml
```diff
@@ -0,0 +1,25 @@
...
+      with:
+        manager-url: https://manager.cicdsensor.cloud.shisho.dev
```
dyoshikawa, line 106: One thing `continue-on-error` doesn't cover: the manager pushes rules, and the baseline rule set includes `action: terminate` entries (9 in `rules/ioc.yaml`, 1 in `generic-tracking-escape.yaml`), plus `monitor_mode` and `redact_process_args` are manager-side `OutputSettings`. So a false positive or a vendor-side config change can kill `pnpm publish` mid-flight, and argv redaction is not something we control from this repo. Do you know whether the Takumi org can pin `monitor_mode` (detect-only) for now, and whether there's a published data-retention / redaction policy for traces? I couldn't find either on the integration or free-tier pages.

.github/actions/cicd-sensor/action.yml
```diff
@@ -0,0 +1,25 @@
+name: Start cicd-sensor
+description: Trace the rest of the job with cicd-sensor and send the trace to Takumi Runner
+inputs:
+  bot-id:
```
dyoshikawa, line 4: Nit: `required: false` for consistency with `minimize-comments`. Behaviour is the same either way.

.github/actions/cicd-sensor/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    description: Shisho Cloud bot ID with the "Takumi Runner Trace Sender" role. Skips tracing when empty.
+runs:
+  using: composite
+  steps:
+    - name: Sign in to Shisho Cloud
+      id: auth
+      if: inputs.bot-id != ''
+      uses: flatt-security/shisho-cloud-action@9b68d33503ebf75d7c285d489b288f96ef3df758 # v1.1.0
```
dyoshikawa, line 18: Small note: `shisho-cloud-action` is still `using: node20` on upstream `main` too, so there's no node24 build to move to. Since 2026-06 the runner forces node20 actions onto node24 (your fork run shows exactly that annotation), so it'll likely keep working past 2026-09-23, but every release job will carry the deprecation warning until upstream ships a node24 build. Worth an upstream issue.

.github/workflows/security-scan.yml
```diff
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: read
+  id-token: write
```
dyoshikawa, line 10: This job is the only one that runs `pnpm install` without `--ignore-scripts`, and it currently has `contents: read` only. Adding `id-token: write` here means a dependency lifecycle script can now mint a GitHub OIDC token for this repo (and, if the bot's trust condition is repo-wide with an empty audience, a Trace Sender token). The blast radius is small, but the value of tracing this job is also small. I'd either drop `security-scan.yml` from the scope, or switch it to `--ignore-scripts` first, and in any case scope the bot's OIDC trust condition to the specific `job_workflow_ref`s rather than the whole repo.

.github/workflows/draft-release.yml
```diff
@@ -25,12 +25,19 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
+      id-token: write
```
dyoshikawa, line 28: `.rulesync/rules/github-actions-security.md` (and its generated copies in `.claude/rules/` and `AGENTS.md`) currently states that jobs using `anomalyco/opencode/github` with `use_github_token: "true"` must not request `id-token: write`, and the residual-risk paragraph says to revisit when token scope widens. This line makes that text wrong, and opencode (installed unpinned via `curl | bash`) now runs with `ACTIONS_ID_TOKEN_REQUEST_*` in scope. Please update the rule to say why these two jobs now carry `id-token: write` (cicd-sensor sign-in, opencode itself still doesn't use OIDC), add the cicd-sensor residual risk alongside the opencode one, and regenerate with `pnpm generate`. Alternatively, leave draft-release out of the scope.

dyoshikawa (commented):
Thanks for the detailed write-up and for pinning both actions by SHA — the pins resolve to the stated tags, the `if: inputs.bot-id != ''` gating works as described, and the manager URL being hard-coded is the right call. I went through both actions at the pinned commits and the cicd-sensor log schema, and I don't see a path by which secret values (env, headers, request bodies) leave the runner in the current version.

That said, I'm not comfortable merging this as-is, mainly because of what the SHA pin does and doesn't cover. The pin fixes the JS wrapper, but the eBPF agent that actually runs as root under `systemd-run` is fetched from a mutable GitHub Release without any checksum or signature check, and this PR puts that inside the two jobs that hold npm trusted-publishing OIDC and sign our release attestations. That is a different risk class from Takumi Guard, which only redirects the npm registry. Details are in the inline comments, along with a few smaller things: the `github-actions-security.md` rule now contradicts the `id-token: write` added to draft-release, `security-scan.yml` runs `pnpm install` with lifecycle scripts and would now hand OIDC to them, Dependabot won't track pins under `.github/actions/*/action.yml`, and the manager-side `terminate` rules can stop a job in a way `continue-on-error` can't catch.

What I'd like to see before revisiting: verification of the downloaded agent (either upstream adopting the sigstore bundles it already publishes, or a verification step in our composite action), the rules doc updated to record the accepted residual risk, and a Dependabot entry for the new action directory. Whether to enable this in `publish.yml` / `publish-assets.yml` at all versus starting with a narrower scope is a call I want to make explicitly rather than by default, so please treat the scope as open. Also worth updating the PR body to make clear that `cicd-sensor` is an independent project and Flatt is the trace recipient — the two trust anchors are distinct.

.github/workflows/publish.yml
```diff
@@ -21,6 +21,12 @@ jobs:
         with:
           ref: main
 
+      - name: Start cicd-sensor
+        continue-on-error: true
+        uses: ./.github/actions/cicd-sensor
```
dyoshikawa, line 25: Dependabot's `github-actions` ecosystem with `directory: "/"` only scans `.github/workflows/*.yml` and a root-level `action.yml`; it doesn't walk `.github/actions/*/action.yml`. None of our existing composite actions pull third-party `uses:`, so this hasn't mattered until now. Please add a `directory: "/.github/actions/cicd-sensor"` entry to `.github/dependabot.yml` with the same `all-actions` group settings so these two pins get bumped — `cicd-sensor-action` cuts releases roughly every two weeks.

.github/workflows/publish-assets.yml
```diff
@@ -41,6 +41,12 @@ jobs:
           fetch-depth: 0
           fetch-tags: true
 
+      - name: Start cicd-sensor
+        continue-on-error: true
```
dyoshikawa: Both fork runs exercised the skip path and the failed-sign-in path, but the happy path (agent download, root unit, docker socket takeover, post step) has never run against one of our jobs. `post.js` throws on tamper detection, and while the runner source suggests `continue-on-error` carries over to the post step of a nested action, that hasn't been observed. Could you run it once from your fork with your own Takumi org so we can see timings and that a post-step failure really doesn't turn the job red?

dyoshikawa:
@saitota Thank you! Merged with your commits as they were.

I pushed one commit on top (dbe0f3c0) to address the review points before merging, so the composite action is a bit bigger than what you wrote:

- The same release asset is now verified before `cicd-sensor-action` runs: `cosign verify-blob` against the sigstore bundle (signing identity of the upstream release workflow) plus pinned SHA-256 digests of the tarball and the extracted agent binary, and `cicd-sensor-version` is passed explicitly. After the action installs `/usr/local/bin/cicd-sensor`, its digest is checked against the verified binary.
- Availability failures (sign-in, cosign install, Sigstore trust root, download) still fail open and skip tracing; integrity failures now fail closed, so `continue-on-error` moved from the callers into the composite action.
- `security-scan.yml` installs with `--ignore-scripts` since it now carries `id-token: write`.
- A Dependabot entry for `.github/actions/cicd-sensor`, and the GitHub Actions security rule records the trust model and residual risk.

The happy path will be exercised once `SHISHO_BOT_ID` is set on this repository; I will run `security-scan` by `workflow_dispatch` first before cutting a release with it.

PR #3068, opened 2026-09-16
fix(claudecode): sync settings scope tables with the settings reference

dyoshikawa:
## Summary

Refreshes the two per-key scope tables in `src/features/permissions/claudecode-permissions.ts` against the current **Scope** column of the [Claude Code settings reference](https://code.claude.com/docs/en/settings-reference), resolving the permissions items recorded in the 2026-09-12 follow-up on #2664.

- `enableArtifact` is documented as `Any file` ("Every file can turn the tool off, and none can turn it back on"), so it is no longer dropped at project scope.
- `autoContinueAtUsageLimit`, `bashEditDiffEnabled`, `desktopSessionCleanupPeriodDays`, `feedbackDrafts` and `modelPicker` are `User or managed`, so they are skipped at project scope and emitted only under `--global`. `autoContinueAtUsageLimit` is the one where a committed value would do the opposite of what it says (a project value turns the feature off regardless), which the table comment records.
- `disableDesktopLocalSessions`, `gatewayInternalNetworks`, `managedMcpServers`, `managedSourcesBehavior` and `modelPricing` are `Managed`, and `copyOnSelect` is `Global config`, so they are skipped in both scopes with the existing "only honored in …" warning.

No new key names were invented: every entry was checked against the `### \`<key>\`` section of the reference. `processWrapper` (also `User or managed`) stays in the command-executing refusal table, which runs first.

## Test plan

- [x] New unit tests cover `enableArtifact` at project scope, the five user-scope keys (dropped at project scope with a warning, written under `--global`) and the six unhonored keys (dropped in global mode with the source named).
- [x] `pnpm cicheck` passes.

Refs #2664 (the workflows / scheduled-task gaps A–E on that issue remain open and are product decisions).


dyoshikawa:
@dyoshikawa Thank you!

PR #3072, opened 2026-09-16
chore: release v16.34.2

### LINKED ISSUE #3062: Watch mode: .git churn triggers regeneration when the source tree is the repo root, and symlinked source directories are never watched

dyoshikawa:
Release v16.34.2.

## What's Changed

* **Watch mode:** `generate --watch` moved to chokidar, watches only the source-tree entries `generate` reads (so `.git/` churn no longer triggers a run when the source tree is the repo root), follows symlinked source directories, waits for the initial scan before printing its banner, and never stalls that wait if a watched directory disappears mid-scan. Closes #3062. by @austintraver in #3063


dyoshikawa:
@dyoshikawa Thank you!

PR #3093, opened 2026-09-16
fix(codebuddy): import a paths-only rule as path triggered

dyoshikawa:
## Summary

Refs #2953 — resolves the import-side gap from the 2026-09-16 re-check. Gap 4 (the six unsupported CodeBuddy surfaces) remains a product decision and is untouched.

The CodeBuddy docs table says `alwaysApply` defaults to `true`, but the shipped loader (`@tencent-ai/codebuddy-code` 2.151.0, `dist/codebuddy.js`, verified from the npm tarball) resolves

```
alwaysApply = fm.alwaysApply === true || (!hasPaths && fm.alwaysApply !== false)
type        = !alwaysApply && hasPaths ? MANUAL : ALWAYS
```

so a rule with `paths` and no `alwaysApply` is path triggered. `CodebuddyRule.toRulesyncRule` materialized `codebuddy.alwaysApply: true` for such a file, and the next generate wrote `alwaysApply: true` + `paths`, which CodeBuddy loads as ALWAYS — an import/generate cycle silently widened a scoped rule to every context.

## Changes

- `src/features/rules/codebuddy-rule.ts`: import follows the loader — a file is ALWAYS when `alwaysApply: true` is set or when it has no `paths` and does not say `false`; `alwaysApply` is carried over only as written (no `true` default materialized). The `resolveCodebuddyRuleType` comment records the loader's resolution; generate still writes the explicit `alwaysApply: false` next to scoped paths, which is correct under both the docs table and the loader.
- `src/features/rules/codebuddy-rule.test.ts`: the two tests that asserted the materialized `true` now assert the path-triggered round trip (including the regenerated `alwaysApply: false` + `paths`), plus a case for an explicit `alwaysApply: true` with paths staying always-on.
- `docs/reference/file-formats.md` (+ regenerated `src/generated/docs-content.ts`): the CodeBuddy note says import follows the CLI for `paths`-only files.

## Verification

- `pnpm cicheck` passes (455 files, 12274 tests).


dyoshikawa:
@dyoshikawa Thank you!

Issue #2598, opened 2026-08-06
Follow up Warp upstream updates: standalone Agent CLI config root (~/.warp_cli/) unsupported — add a warpcli target


dyoshikawa:
## Summary

Warp shipped a standalone Agent CLI with its **own config root**, separate from the Warp app: CLI settings, MCP, and permissions live under `~/.warp_cli/` (platform variants), and the docs state "the app and the CLI keep separate settings files, so changing one never affects the other." rulesync's `warp` target writes only the app's paths, so a CLI user cannot sync MCP or permissions at all. Proposed: a new `warpcli` target (copilot/copilotcli precedent). Two implementation caveats — one affecting the existing `warp` target — are recorded below.

## Recent Releases

- **2026-07-23 release (v0.2026.07.22.09.01)** — `warp-tui` renamed to `warp` (PR #13904), establishing the standalone Agent CLI.
- Execution profiles opened to all users 2026-07-31 (already followed via #2512/#2518).

### 1. mcp + permissions — the Agent CLI's config root is unreachable

- **Upstream** ([CLI configuration](https://docs.warp.dev/cli/configuration/), [CLI permissions & profiles](https://docs.warp.dev/cli/permissions-and-profiles/)):
  - CLI settings: macOS `~/.warp_cli/settings.toml`, Linux `~/.config/warp-terminal/cli/settings.toml` (respects `$XDG_CONFIG_HOME`), Windows `%LOCALAPPDATA%\warp\Warp\config\cli\settings.toml`
  - CLI MCP: `~/.warp_cli/.mcp.json` — "keeps its own MCP server configuration, separate from the Warp app's"
  - CLI permissions: `[agents.execution_profiles.default]`; the CLI always runs the `default` profile.
- **rulesync**: only app paths exist (`src/constants/warp-paths.ts`: `.warp`, `.config/warp-terminal`, `AppData/Local/warp/Warp/config`); `grep -rn "warp_cli" src/` returns nothing. Rules and skills already work for the CLI because both surfaces share `AGENTS.md` / `.agents/skills/` with the app.
- **Follow-up**: add a **`warpcli` target** (copilot/copilotcli precedent in `src/types/tool-target-tuples.ts`) scoped to mcp (global, `~/.warp_cli/.mcp.json`) and permissions (global, CLI `settings.toml`); deliberately leave rules/skills/commands/ignore to `warp`.

[...]

PR #3097, opened 2026-09-17
feat(skills): import Warp skills from the shared .agents/skills root

dyoshikawa:
## Summary

Warp discovers skills from the cross-tool `.agents/skills/` directory (project) / `~/.agents/skills/` (user) — the location its docs list first as **recommended** — in addition to `.warp/skills/`. `WarpSkill.getSettablePaths` only knew `.warp/skills`, so `rulesync import --targets warp` silently missed every skill kept under `.agents/skills/`.

- Declare the shared `AGENTSMD_SKILLS_DIR_PATH` as an `importOnlySkillRoots` entry on `WarpSkill` (same pattern as `junie-skill.ts` / `vibe-skill.ts` / `kimi-code-skill.ts`). Import-only roots are never listed for orphan deletion, so a skill another tool or the user put under `.agents/skills/` (or `~/.agents/skills/`) is never removed by a `warp`-target `--delete`. Generation is unchanged and still writes to `.warp/skills/`.
- `WarpSkill.isDirOwned` applies command-slug ownership only to the `.warp/skills` root, since `WarpCommand` writes only there; a same-named directory under `.agents/skills` stays an importable skill.
- Tests: settable-paths shape, a `fromDir` case from the `.agents/skills` root, `isDirOwned` per root, and processor-level tests for precedence (`.warp/skills` wins) and deletion exclusion.
- Docs: a Warp skills note in `docs/reference/file-formats.md`, `warp` added to the cross-root duplicate note, `src/generated/docs-content.ts` regenerated.

This is the mechanical skills sub-gap from the Warp follow-up issue; the `warpcli`-vs-extend decision, `auto_approve_bypasses_command_denylist`, and the nested-subdirectory import scan remain open there.

Refs #2598

## Evidence

- https://docs.warp.dev/agents/capabilities/skills/ — "Skill locations": `.agents/skills/` (recommended), `.warp/skills/`, … and the same set under `~/`.

## Test plan

- [x] `pnpm cicheck` green locally.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2592, opened 2026-08-06
Follow up OpenCode upstream updates: notification/permissionDenied hook mappings, references key, V2 beta tracking


dyoshikawa:
## Summary

Three OpenCode drifts as of v1.18.14: two canonical hook events (`notification`, `permissionDenied`) have faithful upstream plugin counterparts but are unmapped, and the stable `references` config key — a genuine context-provisioning surface — has no rulesync coverage. Additionally, the OpenCode V2 beta restructures nearly every key rulesync emits and deserves a tracking note.

Canonical repo note: `sst/opencode` 301-redirects to `anomalyco/opencode` (same repository id). Latest stable at research time: **v1.18.14 (2026-08-05)**.

## Recent Releases

- **v1.18.14** (2026-08-05) — latest stable — https://github.com/anomalyco/opencode/releases
- **v1.17.4** (2026-06-12) — MCP `cwd` for local servers (already followed).
- **v1.17.1** (2026-06-10) — `reference` renamed to `references` (gap 3 below).
- **V2 beta** — separate `opencode2` binary via `@opencode-ai/cli@next` — https://opencode.ai/v2/docs (gap 4).

### 1. hooks — canonical `notification` unmapped despite `tui.toast.show`

- **Upstream**: the plugin API exposes `tui.toast.show` (payload: `message`, `variant` info/success/warning/error, optional `title`, `duration`) — https://opencode.ai/docs/plugins/
- **rulesync** (`project`, `global`): `notification` is mapped for many other tools but absent from `OPENCode_HOOK_EVENTS` (`src/types/hooks.ts:323`) and `CANONICAL_TO_OPENCODE_EVENT_NAMES` (`:1115`), so OpenCode users get no plugin stub for it.
- **Follow-up**: add `notification` to the event set and map it to `tui.toast.show`.

### 2. hooks — canonical `permissionDenied` unmapped despite `permission.replied`

- **Upstream**: `permission.replied` carries `sessionID`, `requestID`, and `reply` (`"once" | "always" | "reject"`), so denial is distinguishable — https://opencode.ai/docs/plugins/
- **rulesync**: maps `permissionRequest` → `permission.asked` but nothing to `permission.replied`.
[...]

PR #3098, opened 2026-09-17
feat(mcp): pass OpenCode protocol/codemode server keys through on generate

dyoshikawa:
## Summary

OpenCode V2 documents two additional per-server MCP keys in `opencode.json`: `protocol` (`legacy` | `auto` | `2026-07-28`, PR anomalyco/opencode#48937) and `codemode` (`false` exposes tools directly instead of through Code Mode). The import side already preserved them through the `looseObject` server schema, but the generate side uses the explicit `OPENCODE_PASSTHROUGH_SERVER_FIELDS` allow-list (`timeout`, `oauth`), so a configured `protocol` / `codemode` was silently dropped on the next `rulesync generate`.

- Add `protocol` and `codemode` to `OPENCODE_PASSTHROUGH_SERVER_FIELDS` (this also makes an entry carrying only those keys count as a transport, not a toggle, via `OPENCODE_MCP_TRANSPORT_KEYS`).
- Extend the OpenCode → rulesync → OpenCode round-trip test.
- Add an **OpenCode MCP note** to `docs/reference/file-formats.md` listing the passthrough set; regenerate `src/generated/docs-content.ts`; add `codemode` to `cspell.json`.

On the issue's open question ("how V1 handles an unknown `protocol` key"): the published V1 schema at https://opencode.ai/config.json declares `additionalProperties: false` for `McpLocalConfig` / `McpRemoteConfig`, so V1 would reject the keys — but they are only ever written when the canonical server already carries them, which is documented in the note.

Refs #2592

## Evidence

- https://opencode.ai/v2/docs/mcp-servers/ — local/remote field tables listing `codemode` and `protocol`, plus the "Protocol version" section.
- https://opencode.ai/config.json — V1 schema (`McpLocalConfig` / `McpRemoteConfig` without the keys, `additionalProperties: false`).

## Test plan

- [x] `pnpm cicheck` green locally.


dyoshikawa:
@dyoshikawa Thank you!

Issue #2499, opened 2026-07-30
Follow up Kilo Code upstream updates: mode-specific rule directories (.kilo/rules-{mode}/) not authorable


dyoshikawa:
## Summary

One remaining Kilo Code gap after #2417's items were resolved: mode-specific rule directories (`.kilo/rules-{mode}/`) are read on every Kilo config load but cannot be authored from rulesync — a rulesync rule can only apply to all modes.

Baseline: #2417 surveyed through v7.4.16 (2026-07-24) and is closed; its remaining items are now implemented — the `sandbox` permissions block (with project-scope narrowing to `enabled`/`network`) and global non-root rules (`~/.kilo/rules`) are both in the codebase and visible in dry-runs. The only newer stable release, v7.4.17 (2026-07-29), adds no authorable config surface.

## Recent Releases

- **v7.4.17** (2026-07-29) — [release](https://github.com/Kilo-Org/kilocode/releases/tag/v7.4.17) — Agent Manager / VS Code UI work plus an OpenCode v1.17.5→v1.17.9 sync (PR #12460) of runtime MCP/provider bugfixes. No new config key or file surface.
- **v7.4.15** (2026-07-22, PR #12422) — "project memory" rework; agent-managed runtime state, not a user-authored config file.

### rules — mode-specific rule directories (`.kilo/rules-{mode}/`) not authorable (project scope)

- **Upstream** (confirmed against `main`, 2026-07-30): `packages/opencode/src/kilocode/rules-migrator.ts` scans `{.kilo,.kilocode}/rules-{mode}/*.md` for `KNOWN_MODES = ["code", "architect", "ask", "debug", "orchestrator"]` and injects them as instructions. `includeModeSpecific` defaults to `true` and `packages/opencode/src/kilocode/config/config.ts` calls `RulesMigrator.migrate({ projectDir })` with no options, so these directories are read on **every config load** — current behavior, not a one-shot legacy migration. Docs: [custom instructions](https://kilo.ai/docs/customize/custom-instructions).
- **rulesync**: unsupported — `KILO_RULES_DIR_NAME` is the single `"rules"` (`src/constants/kilo-paths.ts`), and `kilo-rule.ts` has no mode dimension.
- **Scope note**: project only — the migrator's mode loop runs over `projectDir` and `globalRulesDirs()` has no mode variant.

## Proposed Follow-up

[...]

PR #3100, opened 2026-09-17
feat(commands): skip the reserved goal command for the Kilo target

dyoshikawa:
## Summary

Kilo reserves the `/goal` command name for its session-goal feature. Its command loader throws on a custom command by that name found in either `.kilo/commands/` or `~/.config/kilo/commands/` — `The /goal command is reserved for session goals. Rename the custom command.` — and because the throw sits inside `Command.state`, it aborts the whole command scan rather than only the one file.

Until now `rulesync generate --targets kilo` happily wrote `.kilo/commands/goal.md` from a rulesync command with that stem, leaving the user with a Kilo that loads no custom commands at all.

## Changes

- `commands-processor.ts`: add an optional per-command write gate, `getWriteBlockReason({ rulesyncCommand, global })`, to the tool command factory interface. It is evaluated on the command as it will be written (after flattening for tools without subdirectory support), before collision bookkeeping; a non-null reason skips the command with a warning naming the path and the reason. This mirrors the skills processor's `getDirWriteBlockReason` hook.
- `kilo-command.ts`: implement the gate for the `goal` stem at both scopes, with a message that names the reservation and how to resolve it (rename, or drop `kilo` from `targets`). Import (`fromFile`) is untouched: a `goal.md` already on disk is read as it is.
- Tests: processor test that the gate runs on the flattened path, drops the command, and warns once; `KiloCommand.getWriteBlockReason` unit tests (both scopes; look-alike names such as `goals.md`/`git-goal.md` stay allowed); import of an existing `goal.md` is unaffected.
- Docs: new **Kilo Code note** under `rulesync/commands/*.md` in `docs/reference/file-formats.md`; `docs-content.ts` regenerated.

## Evidence

- Reservation: https://github.com/Kilo-Org/kilocode/blob/v7.6.2/packages/opencode/src/command/index.ts (`if (name === "goal") throw new Error(...)` while iterating `cfg.command`; still present on `main` as of v7.7.2).
- Command files feed `cfg.command`: https://github.com/Kilo-Org/kilocode/blob/v7.6.2/packages/opencode/src/config/command.ts (`Glob.scan("{command,commands}/**/*.md")`, name from the file path).

This is one mechanical sub-item of the Kilo follow-up issue; the remaining items there are design-gated and stay open.

Refs #2499


dyoshikawa:
@dyoshikawa Thank you!

PR #3101, opened 2026-09-17
feat(skills): pass the Grok CLI path-gated paths skill field through

dyoshikawa:
## Summary

Grok CLI supports path-gated (conditional) skills: a `paths` field in `SKILL.md` frontmatter — gitignore globs as a YAML list or a comma-separated string — keeps the skill out of the listing until a tool touches a matching file. Rulesync dropped it in both directions for the `grokcli` target: a skill authored with `paths` silently became unconditional for Grok, and an import discarded a Grok `paths` value.

## Changes

- `rulesync-skill.ts`: add `paths` (string or string list) to the `grokcli` section schema and type, alongside the two invocation flags already there. It is a section key rather than a root default, matching how `claudecode.paths`, `cursor.paths` and `qwencode.paths` are modelled.
- `grokcli-skill.ts`: accept `paths` in the tool frontmatter schema, emit `grokcli.paths` as authored in `fromRulesyncSkill` (Grok accepts both shapes, so nothing is coerced), and read it back into the `grokcli` section in `toRulesyncSkill`.
- Tests: list and string emit, no bleed from another tool's section, omitted when unset, and import round-trip.
- Docs: `grokcli.paths` in the skills frontmatter reference; `docs-content.ts` regenerated.

## Evidence

- Rendered docs frontmatter table (`paths` — "Gitignore globs. Hidden until a matching file is touched"): https://docs.x.ai/build/features/skills-plugins-marketplaces
- Parser: `parse_skill_paths` / `coerce_path_list` (string comma-split or YAML sequence) / `normalize_skill_paths` in https://github.com/xai-org/grok-build/blob/main/crates/codegen/xai-grok-tools/src/implementations/skills/discovery.rs

This is one mechanical sub-item of the Grok CLI follow-up issue; the remaining items there are design-gated and stay open.

Refs #2498


dyoshikawa:
@dyoshikawa Thank you!

PR #3102, opened 2026-09-17
fix(permissions): stop writing the retired Codex CLI untrusted approval policy

dyoshikawa:
## Summary

Refs #2496

Codex CLI retired `approval_policy = "untrusted"` in `rust-v0.149.0` ([openai/codex#39630](https://github.com/openai/codex/pull/39630)): the value is gone from `config.schema.json` (0.154.0 lists only `on-request`, `never`, and the `granular` table) and `codex-rs/core/src/config/mod.rs` returns `approval_policy = "untrusted" is no longer supported; remove this setting` when a config still sets it. Rulesync's `codexcli.approval_policy` override wrote it verbatim, producing a `config.toml` Codex refuses to start on. The [config reference](https://learn.chatgpt.com/docs/config-file/config-reference) also marks `on-failure` as deprecated (still read as `on-request`).

## Changes

- `src/features/permissions/codexcli-permissions.ts`
  - Generate: an authored `approval_policy: "untrusted"` is skipped with a warning that names the replacement (a user-level `[projects."<path>"] trust_level = "untrusted"` entry, which is per-machine state rulesync does not author). The key then falls back to the existing `config.toml` value or the `on-request` default — the same outcome as the "remove this setting" migration Codex asks for. An existing user-set value in `config.toml` is still never clobbered.
  - Generate: `on-failure` is still written (Codex reads it as `on-request`) but warned about as deprecated.
  - Import: an existing `approval_policy = "untrusted"` in `config.toml` is no longer lifted into the `codexcli` override; a single `warnWithFallback` names the file and the migration path instead of round-tripping a value the next generate would drop again. The other values import unchanged.
  - The approval-policy check lives in a small `isWritableApprovalPolicy` helper to keep `computeCodexcliOverridePatch` under the complexity lint.
- `src/types/permissions.ts`: `untrusted` stays in `CodexApprovalPolicySchema` so an existing `.rulesync/permissions.jsonc` still parses (a hard validation error would be a worse failure mode than the warning); the comments now describe the retired / deprecated status.
- Tests: generate skips `untrusted` (default fill and existing-value cases), warns on `on-failure`, stays silent on `on-request` / `never`; import drops `untrusted` with the migration warning and still lifts the other values.
- `docs/reference/file-formats.md` (Codex CLI-only override paragraph) and the regenerated `src/generated/docs-content.ts`.

## Not in this PR

The other items tracked in #2496 (nested `AGENTS.md` for `codexcli`, `mcp_tool` hook handlers, inline `[hooks]` in `config.toml`, `## Code Review Rules`, `[agents]` / `[[skills.config]]` ownership, Agent Plugins bundles, `approvals_reviewer` default) are design decisions and remain open there.

## Test plan

- [x] `pnpm cicheck` (oxfmt, oxlint, typecheck, 12,333 tests, docs-content drift, cspell, secretlint)
- [x] New unit tests in `codexcli-permissions.test.ts`


dyoshikawa:
@dyoshikawa Thank you!

PR #3103, opened 2026-09-17
fix(permissions): match Takt's companion filter and assignments in runtime mode detection

dyoshikawa:
## Summary

Refs #2421 (Takt upstream follow-up — 2026-08-25 comment, item 1: runtime-mode detection diverges from upstream's companion filter).

Takt's mode detection (`hasActiveProviderContent` in `src/infra/config/runtime-provider/schema.ts`, checked at v0.65.0) first reduces the `provider:` section with `getEffectiveProviderSection`: unless `companion.enabled` is `true` (`DEFAULT_COMPANION_ENABLED = false`), it drops `targets.companions` — top-level and inside every named `assignments` set — plus every profile reachable only from a companion target via `extends`. Only then does it test for content, and a named `assignments` set with `defaults` or agent targets counts as content too.

rulesync's `isRuntimeModeActive` skipped that reduction and did not know about `assignments`, so:

- a `runtime.yaml` that only routes companion reviewers was treated as runtime mode: an authored `takt.provider_options` was dropped with a spurious `Mixed provider configuration detected` warning, although Takt itself is in legacy mode there;
- a section active only through a named `assignments` set was treated as legacy mode, so `provider_options` was written into a `config.yaml` Takt then refuses.

## Changes

- `src/features/permissions/takt-permissions.ts`
  - `effectiveRuntimeProviderSection` mirrors upstream's filter (companion targets, companion-only profile closure over `extends`, agent roots from `defaults`/other `targets`/`assignments`/`auto_routing` router, pool candidates and fallback).
  - `isRuntimeModeActive` runs on the effective section and also counts named `assignments` sets.
  - `mergeTaktRuntimeConfigs` merges `assignments` as a union (project wins per name) and carries `companion.enabled` the way `mergeRuntimeProviderFiles` does (unset when neither file states it; otherwise on only when neither file turns it off).
  - `collectRuntimeProviderOptions` skips profiles Takt ignores, so a companion-only profile's `options` are not lifted on import.
- Tests for each case in `takt-permissions.test.ts` (4 of the new tests fail against the previous implementation).
- `docs/reference/file-formats.md` Takt paragraph updated; `docs-content.ts` regenerated.

The remaining items in #2421 (`assistant.init_files`, companion reviewer files, workflow-wide rules, the runtime `mcp` registry, provider-options presets) need a design decision and are left as they are.

## Verification

- `pnpm cicheck` green (12,343 tests).


dyoshikawa:
@dyoshikawa Thank you!

PR #3104, opened 2026-09-17
feat(skills): warn when a Zed skill name or description breaks Zed's load rules

dyoshikawa:
## Summary

Refs #2415 (Zed gap analysis). This picks up the one remaining mechanical sub-gap: Zed applies rules of its own when it loads a skill, and rulesync used to write a skill Zed would refuse without saying anything.

Per https://zed.dev/docs/ai/skills:

- A skill `name` must be 1–64 characters of lowercase letters, digits and hyphens, with no leading, trailing or consecutive hyphens; "skills with invalid names fail to load and surface an error in the UI".
- A `description` should be at most 1024 characters; "skills with longer descriptions still load, but with a warning".

## Changes

- `src/features/skills/zed-skill.ts`: `fromRulesyncSkill` now accepts the run's `logger` (already passed by `SkillsProcessor`) and reports each violation through `warnWithFallback`, pointing at the `SKILL.md` about to be written — the same pattern `AgentsSkillsSkill.reportSpecViolations` uses. The skill is still written, because the canonical skill is shared with every other target.
- `src/features/skills/zed-skill.test.ts`: 9 new tests (valid skill is silent; each name rule; 64-character name cap; 1024-character description; multiple violations reported in order; fallback logger when none is passed). All 9 fail on `main`.
- `docs/reference/file-formats.md`: new **Zed note** in the skills section; `src/generated/docs-content.ts` regenerated.

## Deliberately not included

- The 50KB cap on the combined names + descriptions of every installed skill spans skills rulesync did not write, so a single skill cannot judge it.
- "Should match the folder name": Zed's docs say *should* and do not state what happens on a mismatch, so no warning is emitted for it (the `agentsskills` target already reports a mismatch under the Agent Skills spec).
- The rest of #2415 (commands-as-skills for Zed, cross-scope `private_files` accumulation warning, projecting `enabledTools`/`disabledTools` into `agent.profiles`) is a design decision and stays open on the issue. The MCP tool-filter drop warning itself already shipped in 6c82f45d.

## Test plan

- [x] `pnpm cicheck` (12,353 tests, docs-content in sync, cspell, secretlint)
- [x] New tests verified to fail without the implementation (`git stash` on `zed-skill.ts`)


dyoshikawa:
@dyoshikawa Thank you!

PR #3109, opened 2026-09-17
feat(goose): write failClosed as Goose's on_failure: block on PreToolUse hooks

dyoshikawa:
## Summary

Follows the `on_failure` item recorded in #2404: Goose v1.48.0 added `on_failure` to command hooks and rulesync dropped the equivalent canonical `failClosed` for Goose in both directions.

- `src/features/hooks/tool-hooks-converter.ts`: new `keywordPassthroughFields` converter kind for a canonical boolean a tool spells as a keyword. `true` → `trueValue`, `false`/absent → omitted (the tool's default); import reads `trueValue` / optional `falseValue` back and drops any other value with a warning. An optional `events` set restricts the field to the canonical events the tool honours it on — a `true` elsewhere is dropped with a warning in both directions. The canonical event name is now threaded through the import path (`toolMatcherEntryToCanonical` → `toolHookToCanonical` → `importAllPassthroughFields`) so the import side can apply the same scope.
- `src/features/hooks/goose-hooks.ts`: registers `failClosed` ⇄ `on_failure` (`"block"` / `"allow"`), `commandOnly`, scoped to `preToolUse`.
- Docs: the `failClosed` field description and the Goose hooks note.

## Evidence (upstream, tag `v1.48.0`+)

- [`crates/goose/src/hooks/mod.rs`](https://github.com/aaif-goose/goose/blob/v1.50.0/crates/goose/src/hooks/mod.rs): `RawCommandAction { command, timeout, on_failure: Option<Value> }`; `#[serde(rename_all = "lowercase")] enum OnFailure { #[default] Allow, Block }`; the loader applies `on_failure` only when `event == HookEvent::PreToolUse` (other events get `OnFailure::Allow`), and on `PreToolUse` a value that is not one of the two keywords fails the whole hooks file (`with_context("reading on_failure in …")?`). `apply_verdict` denies the tool call on a hook failure only when `on_failure == Block && event == PreToolUse`.
- [v1.48.0 release](https://github.com/aaif-goose/goose/releases/tag/v1.48.0) (2026-08-27), PR [#11449](https://github.com/aaif-goose/goose/pull/11449).

Because an off-keyword value on `PreToolUse` makes Goose refuse the file, the emit side only ever writes the documented `"block"` keyword and only on `PreToolUse`.

## Tests

- `src/features/hooks/goose-hooks.test.ts`: generate (`block` on `PreToolUse`, `false` omitted, `true` on `postToolUse` dropped with one warning) and import (`block` / `allow` → `true` / `false`, `"Block"` typo and `on_failure` on `PostToolUse` dropped with warnings).
- `pnpm cicheck` passes (12,373 tests).

Refs #2404


dyoshikawa:
@dyoshikawa Thank you!

Issue #2402, opened 2026-07-27
Follow up GitHub Copilot CLI upstream updates: .claude/commands/ surface, userPromptTransformed event, matcher-aware events, skill invocation flags


dyoshikawa:
## Summary

Copilot CLI has a documented `.claude/commands/*.md` command surface that rulesync does not support at all, plus four smaller drifts: an unmapped `userPromptTransformed` hook event, matchers dropped on four events upstream documents as matcher-aware, the portable `command` / `timeout` hook fields ignored on import, and two skill invocation flags dropped on generate.

## Recent Releases

Source: `gh release list -R github/copilot-cli` and [changelog.md](https://github.com/github/copilot-cli/blob/main/changelog.md). Latest stable at time of research: **v1.0.75 (2026-07-24)**.

| Version | Date | Relevant change |
| --- | --- | --- |
| v1.0.75 | 2026-07-24 | Claude Opus 5 support only |
| v1.0.74 | 2026-07-23 | Open Plugin Spec v1 manifests + `mcp.json`; "Fully honor the skill disable-model-invocation flag" |
| v1.0.72 | 2026-07-20 | `agentStop` `stop_hook_active`; plugin skill/MCP management |
| v1.0.71 | 2026-07-16 | `githubMcpToolsets` / `githubMcpTools` in `settings.json`; mark disabled skills in `skill list` |
| v1.0.70 | 2026-07-09 | Repo-scoped `.github/copilot/settings.json`; `preToolUse` exit code 2 denies |
| v1.0.66 | 2026-06-30 | `@`-imports in AGENTS.md/CLAUDE.md/copilot-instructions; custom agents set reasoning effort; nested Claude-style inline hook groups; `COPILOT_HOME` stops loading `~/.agents/skills` |
| v1.0.63 | 2026-06-15 | `postToolUse` matchers honored |
| v1.0.36 | 2026-04-24 | `~/.claude/` agents/skills/commands no longer loaded (project `.claude/…` still loaded) |
| v0.0.399 | 2026-01-29 | **`.claude/commands/` single-file commands** as a simpler alternative to skills |

Docs were read from the `github/docs` markdown source (authoritative, not the rendered SPA): `content/copilot/reference/copilot-cli-reference/cli-command-reference.md` and `content/copilot/reference/hooks-reference.md`.

### 1. commands — `.claude/commands/*.md` is a native, documented surface

- **Upstream**, verbatim:

  > ### Commands (alternative skill format)
[...]

PR #3112, opened 2026-09-17
feat(copilotcli): support exec/args command hooks and type models/modelPolicy on subagents

dyoshikawa:
## Summary

Ships the two mechanical items recorded on #2402 (the 2026-09-08 and 2026-09-12 re-checks); the remaining items there (`.claude/commands/` surface, plugin packaging, `.mcp.json` precedence warning) are product decisions and stay tracked on the issue.

- **hooks — `exec` / `args` (CLI-only shell-free command form).** The [hooks reference](https://docs.github.com/en/copilot/reference/hooks-reference) documents `exec` ("Executable name or path. Runs the executable directly without a shell. Only supported in Copilot CLI.") and `args`, and says not to combine `exec` with `bash` / `powershell` / `command`. `CopilotCliHookEntrySchema` did not know either field, so an `exec` hook imported as a `command` hook with **no command** and was regenerated empty. The form now maps both ways onto the canonical `command` + `args` exec form already shared with Claude Code and AugmentCode:
  - generate: a canonical `args` list (an empty one counts) writes `exec` + `args` instead of a shell field; a `shell` selector on such a hook is dropped with a warning.
  - import: `exec` lands as `command` + `args` (empty list when upstream omits `args`); `exec` alongside a shell field imports as the `exec` with a warning.
- **subagents — `models` / `modelPolicy` typed.** The [CLI command reference](https://docs.github.com/en/copilot/reference/copilot-cli-reference/cli-command-reference) frontmatter table documents `models: string[]` (priority order, overrides `model`) and `modelPolicy: "preferred" | "required"` (CLI v1.0.81+). They already rode through `looseObject` untyped; they are now declared so a mistyped value is rejected before the agent file is written. `model` stays string-only per the docs (the changelog's `model: [...]` spelling is still unconfirmed, as noted on the issue).
- Docs: the Copilot CLI hooks note and the canonical `args` field description cover the `exec` mapping; the subagent example gains a `copilotcli:` block.

Refs #2402

## Test plan

- [x] `pnpm cicheck` (unit tests incl. new exec-form generate / import / round-trip cases and the typed `models` / `modelPolicy` case)
- [x] Regenerated `src/generated/docs-content.ts`


dyoshikawa:
@dyoshikawa Thank you!

Issue #2401, opened 2026-07-27
Follow up GitHub Copilot upstream updates: edits/urls autoApprove permissions, mcp.json inputs/sandbox clobbered, global hooks scope


dyoshikawa:
## Summary

VS Code's Copilot approval surface now has `chat.tools.edits.autoApprove` and `chat.tools.urls.autoApprove` alongside the terminal map rulesync already emits, `.vscode/mcp.json` gained a `sandbox` section (v1.112) next to the existing `inputs` section that rulesync's whole-file write destroys, and hooks now have a documented user scope at `~/.copilot/hooks`. rulesync follows none of these.

## Recent Releases

- **VS Code v1.111 – v1.114** (Mar – Jun 2026) — [release notes](https://code.visualstudio.com/updates/) — agent customization surfaces (custom instructions, custom agents, prompt files, hooks) and MCP configuration.
- **Copilot in VS Code June 2026 releases** — 2026-07-08 — [GitHub Changelog](https://github.blog/changelog/2026-07-08-github-copilot-in-visual-studio-code-june-2026-releases/).
- **Copilot code review: new configurations and controls** — 2026-06-12 — [GitHub Changelog](https://github.blog/changelog/2026-06-12-copilot-code-review-new-configurations-and-controls/).
- **Copilot code review customization and configurability improvements** — 2026-07-17 — [GitHub Changelog](https://github.blog/changelog/2026-07-17-copilot-code-review-customization-and-configurability-improvements/) — code review now also reads `REVIEW.md`, `GEMINI.md`, `CLAUDE.md` from the PR head branch.

Dry-run baseline (from `/workspace`): `--targets copilot --features "*" --dry-run` → `.vscode/mcp.json`, `.github/copilot-instructions.md`. The `--global` run emits 5 rules + 5 subagents + 41 skills and **no hooks, no permissions**.

### 1. permissions — `chat.tools.edits.autoApprove` not mapped

- **Upstream**: a glob→boolean map with the same shape as the terminal map, e.g. `{ "**/*": true, "**/.env": false }`. — [approvals docs](https://code.visualstudio.com/docs/agents/approvals) (also VS Code v1.104 release notes).
- **rulesync**: `project`, and `src/features/permissions/copilot-permissions.ts` hard-codes `TERMINAL_CATEGORY = "bash"` with a comment stating that other categories are intentionally unmapped.

### 2. permissions — `chat.tools.urls.autoApprove` not mapped

[...]

PR #3113, opened 2026-09-17
fix(copilot): keep VS Code per-OS hook overrides on import and note prompt-file deprecation

dyoshikawa:
## Summary

Two mechanical items recorded on #2401 (2026-08-25 per-OS override keys; 2026-09-12 prompt-file deprecation note). The larger items there (`CopilotPermissionsOverrideSchema` sandbox block, `chat.permissions.default`, MCP `~/.copilot/mcp-config.json` / `.mcp.json` scope, `copilot-plugin` packaging, `REVIEW.md`, dictation files) are product decisions and stay tracked on the issue.

- **hooks — `windows` / `linux` / `osx` survive import.** The [VS Code hooks reference](https://code.visualstudio.com/docs/agents/reference/hooks-reference) documents the three per-OS command overrides next to `command` on the same `.github/hooks/*.json` entries. Generate already passed them through as unknown keys, but `importPassthrough` kept only `cwd` / `env`, so they vanished on an import → generate round trip, and an entry authored with only per-OS overrides (valid: "at least one command property") imported as a hook with **no command** and was regenerated empty. The keys are now typed on `CopilotHookEntrySchema` and preserved on import. No change to what a canonical `shell` selector emits: the [VS Code hooks page](https://code.visualstudio.com/docs/agent-customization/hooks) states VS Code maps `bash` → `osx` / `linux` and `powershell` → `windows`, so the existing `bash` / `powershell` output is read correctly there — this corrects the issue's 2026-08-25 premise that `powershell` is a key VS Code does not read. The `matcher` drop is now commented as by design (VS Code "ignores matcher values").
- **commands — prompt-file deprecation note.** [Prompt files](https://code.visualstudio.com/docs/agent-customization/prompt-files) are "deprecated for Agent Host sessions and aren't loaded by Agent Host. They continue to work with the Local agent for now, but the Local agent will be removed in a future release." `docs/reference/file-formats.md` gains the same kind of note the Codex CLI custom prompts already carry: output unchanged, skills recommended for new reusable instructions.

Refs #2401

## Test plan

- [x] `pnpm cicheck` (new test: per-OS overrides round-trip through import and re-export, including an entry with only per-OS overrides)
- [x] Regenerated `src/generated/docs-content.ts`


dyoshikawa:
@dyoshikawa Thank you!

PR #3114, opened 2026-09-17
feat(cursor): keep skill badge icon/color, legacy globs fallback, and prompt-hook model

dyoshikawa:
## Summary

Closes three mechanical sub-gaps recorded on the Cursor follow-up issue (all re-verified today against the current Cursor docs):

- **Skills — `icon` / `color` badge fields.** The [skills reference](https://cursor.com/docs/skills) documents `icon` ("Icon shown on the badge when the skill is used as a Custom Mode") and `color` ("Badge color when the skill is used as a Custom Mode"). Both are added to `CursorSkillFrontmatterSchema` and to the `cursor` section of the rulesync skill schema, and threaded through `toRulesyncSkill()` / `fromRulesyncSkill()`. They are typed as plain strings, not enums: Cursor documents a fallback to the default lightning badge for unrecognized values, so an enum would reject frontmatter Cursor itself accepts.
- **Skills — legacy `globs` import fallback.** The same page states "The legacy `globs` field is still accepted as a fallback for older skills, but new skills should use `paths`." A pre-existing `.cursor/skills/<name>/SKILL.md` scoped with `globs:` previously lost its scoping on import. Import now reads `globs` as a fallback for `paths` (dropped when `paths` is also present, matching Cursor's precedence); generate keeps emitting `paths` only.
- **Hooks — prompt-hook `model` override.** The [hooks docs](https://cursor.com/docs/hooks) document an "Optional `model` field to override the default LLM model" on prompt-based hooks (`{"type": "prompt", "prompt": "...", "model": "custom-model-name"}`). The canonical hook definition already carries `model` and Claude Code / Qwen Code forward it, but `CursorHooks.fromRulesyncHooks` rebuilt each definition by explicit key pick and dropped it. It is now forwarded for `type: "prompt"` entries only; `command` hooks never receive it.

Docs: the `cursor:` skill example in `docs/reference/file-formats.md` gains the two badge fields plus the `globs` import note, and the canonical `model` hook bullet now lists Cursor.

Refs #2399

## Test plan

- [x] `pnpm cicheck` passes (12385 tests)
- [x] New unit tests: icon/color round-trip (including an unknown icon value being kept), `globs` → `paths` normalization with `paths` precedence and no `globs` on generate, prompt-hook `model` forwarded and dropped from command hooks


dyoshikawa:
@dyoshikawa Thank you!

Issue #3015, opened 2026-09-12
Follow up JetBrains AI Assistant upstream updates: registry-gated .aiassistant/commands and AI Self-Review guidelines carried over from #2600


dyoshikawa:
## Summary

#2600 was closed on 2026-09-12 once the `.ai/mcp/mcp.json` path was confirmed from the plugin registry keys, but two surfaces recorded on that issue's 2026-08-25 and 2026-09-08 comments were never resolved and now have no open tracker: the registry-gated `.aiassistant/commands/**/*.md` custom chat commands, and the AI Self-Review guidelines Markdown file. This issue carries them forward with a fresh re-check against the 2026.3 EAP line.

## Recent Releases

- JetBrains AI Assistant **263.4732.41** (2026-09-11, 2026.3 EAP) and 263.4732.34 / .31 (2026-09-11) — https://plugins.jetbrains.com/plugin/22282-jetbrains-ai-assistant/versions. Bundled `ml-llm/lib/ml-llm.jar!/META-INF/plugin.xml` re-verified on 2026-09-12: `llm.chat.input.language.user.commands.enabled` is still `defaultValue="false"`; a new experimental key `github.aiReview.enabled` (`defaultValue="false"`, "Enable the experimental AI Review tool window", `restartRequired="true"`) appeared; both `llm.mcp.client.{project,global}.mcp.json.path` keys still default to `.ai/mcp/mcp.json`.
- JetBrains AI Assistant **262.10315.187** (2026-09-11, 2026.2 stable) — same page; bug fixes only.
- Documentation: no page under https://www.jetbrains.com/help/ai-assistant/ has a sitemap `lastmod` newer than 2026-08-18, so neither `.aiassistant/commands` nor the AI Review tool window is documented yet.

### commands — `.aiassistant/commands/**/*.md` (project) is `unsupported`, and upstream still ships it off by default

Full analysis is in https://github.com/dyoshikawa/rulesync/issues/2600#issuecomment-5580829605 (2026-09-08) and https://github.com/dyoshikawa/rulesync/issues/2600#issuecomment-5413018625 (2026-08-25) — in short: plain Markdown files under `.aiassistant/commands/` (recursive), name derived from the file basename, no frontmatter, no description, no argument substitution, project scope only, gated by the registry key `llm.chat.input.language.user.commands.enabled` whose default is `false` in every build to date including 263.4732.41.

[...]

PR #3115, opened 2026-09-17
feat(aiassistant): write the apply metadata block on generated rules

dyoshikawa:
## Summary

Refs #3015 (the rules gap recorded in the 2026-09-16 re-check comment).

JetBrains AI Assistant stores the rule type of `.aiassistant/rules/*.md` in the file itself, as a leading metadata block (`---` / `apply: <type>` / `---`, with `instructions:` for `by model decision` and a comma-separated `patterns:` for `by file patterns`). A file without an `apply:` line is typed **Off** when the plugin collects rules for a chat, so the body-only files rulesync generated were never attached automatically.

- `src/features/rules/aiassistant-rule.ts`: generate the block — specific `globs` → `apply: by file patterns` + `patterns: <globs joined by ", ">`; a `description` on a rule with no `globs` → `apply: by model decision` + `instructions: <description>` (flattened to one line); root rule / a list containing a universal glob (`**/*`, `*`) / bare rules → `apply: always`; an explicit `aiassistant.apply` is written verbatim and overrides the derivation (unknown values are warned about, not dropped). Brace alternations in globs (`*.{ts,tsx}`) are expanded, since the plugin splits `patterns` on bare commas; a glob that still contains a comma, or any value containing `---`, is warned about. The block is line-based (`<field>: <value>`, split on the first colon), not YAML, so it is written and parsed by hand with the plugin's own regex.
- Import strips the block and maps it back (`by file patterns` → `globs`, `by model decision` → `description`, `always` → `globs: ["**/*"]`); any `apply` the derivation would not reproduce (`manually`, `off`, an explicit `by file patterns` on a universal glob, or an unrecognized value) is kept as `aiassistant.apply` so a regenerate reproduces the file byte-for-byte.
- `src/features/rules/rulesync-rule.ts`: new optional `aiassistant.apply` frontmatter key.
- `src/utils/brace-aware-list.ts`: `splitBraceAwareList` (extracted from the CodeBuddy adapter, now shared) and `expandBraceAlternations` (capped at 256 patterns per glob, nested groups included).
- `docs/reference/file-formats.md`: example key and a **JetBrains AI Assistant note** describing the format; `src/generated/docs-content.ts` regenerated.

Field names and values come from the plugin (`AiRulesService`, `AiRulesLLMBundle.properties`; `metadataRegex = ^---\s*([\s\S]*?)---\s*`, `toAiRule` falls back to `AiRuleType.OFF`), verified on build 262.10968.75 — the docs page describes the rule type only as an IDE setting.

## Test plan

- [x] `pnpm cicheck` (12409 tests) passes
- [x] `aiassistant-rule.test.ts` covers derivation for each rule type, the explicit override, warnings, block parsing edge cases (BOM, CRLF, `---` inside a value, unclosed block, brace-aware pattern split), and an import → generate round-trip for every rule type
- [x] `brace-aware-list.test.ts` covers the list splitter and brace expansion, including the expansion cap for sibling and nested groups


dyoshikawa:
@dyoshikawa Thank you!

PR #3091, opened 2026-09-16
refactor(pool): sort the pool rules target into alphabetical position

dyoshikawa:
## Summary

Refs #2984 — item 2 (re-sort `pool` into alphabetical position). Items 1 and 3 need a maintainer decision and are untouched; item 4 was resolved upstream (see the 2026-09-16 comment on the issue).

## Changes

- `src/types/tool-target-tuples.ts`: `pool` moves from after `zoocode` to between `pi` and `qwencode` in `rulesProcessorToolTargetTuple`.
- `src/features/rules/rules-processor.ts`: the `pool` entry of `toolRuleFactories` moves to the same position (map order is `getToolTargets()` order).
- `src/types/tool-targets.test.ts`, `src/features/rules/rules-processor.test.ts`: expected-order arrays updated accordingly.

No behavioural change. The generated supported-tools tables are driven by `TOOL_DISPLAY` (a hand-kept presentation order, not alphabetical), so `pnpm run generate:tables` produced no diff. `dsh`, added after the issue was filed, still sits at the end of the tuple; left as-is since it is outside the issue's scope.

## Verification

- `pnpm cicheck` passes (455 files, 12273 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019yLN6CVqeZ8HCrjRb9o7mK

Issue #2984, opened 2026-09-10
Follow-ups from the Pool (Poolside) rules target review


dyoshikawa:
## Background

PR #2976 added Pool (Poolside) as a rules target. The review found no `high`/`critical` issues and the two `mid` findings (an unused `POOL_RULE_FILE_NAME` export and the `XDG_CONFIG_HOME` wording) were fixed by the contributor before the merge. The remaining observations are all `low` severity and were not worth blocking that PR, so they are collected here.

### 1. `pool-rule.ts` is a verbatim copy of `vibe-rule.ts` (refactor candidate)

Applying a mechanical `Vibe` -> `Pool` rename to `src/features/rules/vibe-rule.ts` leaves only two differences against `src/features/rules/pool-rule.ts`: the `POOL_GLOBAL_DIR` import and the JSDoc/`@see` URLs on `getNestedFilePatterns`. The logic is identical, and so are the ten test cases in the two `*.test.ts` files.

That makes two adapters with the same shape: an `AGENTS.md` root, nesting driven by `agentsmd.subprojectPath`, and folding for everything else. A shared base (for example an `AgentsMdFamilyRule` on the `ToolRule` side, parameterised by `globalDir` and `toolTarget`) would remove roughly 180 duplicated lines. Worth deciding before a third AGENTS.md-family target arrives.

### 2. `pool` is appended out of alphabetical order

`rulesProcessorToolTargetTuple` in `src/types/tool-target-tuples.ts` is broadly alphabetical (with `devin` as the existing exception), but `pool` sits after `zoocode` instead of between `pi` and `qwencode`. The expected arrays in `src/types/tool-targets.test.ts` follow that position. `toolRuleFactories` in `src/features/rules/rules-processor.ts` registers `pool` last as well, which is why `TOOL_DISPLAY` and the generated supported-tools tables list it after `ZCode`.

No functional impact — map order simply becomes `getToolTargets()` order, which becomes the generated table row order. Purely a question of whether the lists should be kept sorted.

### 3. `FeatureProcessor.writeAiFiles` has no symlink guard

[...]

PR #3123, opened 2026-09-18
refactor(rules): share the AGENTS.md-family rule adapter and guard writes against symlink escapes

dyoshikawa:
## Summary

Resolves the two remaining items of #2984 (items 2 and 4 were closed out earlier by #3091 and the upstream Windows-path statement).

### Item 1 — shared AGENTS.md-family rule adapter

`pool-rule.ts`, `vibe-rule.ts` and `dsh-rule.ts` were three verbatim copies of the same adapter: an `AGENTS.md` root, nesting driven by `agentsmd.subprojectPath`, and folding of every other non-root rule into the root file. They now extend a new `NestedAgentsmdRule` base (`src/features/rules/nested-agentsmd-rule.ts`) that carries all of that logic; each target only supplies its global directory and tool target through a `getFamily()` override, and keeps the vendor-specific JSDoc describing its discovery walk. The three adapters shrink from ~200 lines each to ~20. The existing per-target tests are unchanged and still pass; a small base-class test covers the family hook.

### Item 3 — symlink guard on every writer

Decision recorded here: the strict `assertWritablePathInsideRoot` used by the sweep paths was **not** adopted for ordinary writes, because it refuses every symbolic link and would break the common dotfiles setup (`~/.claude -> ~/dotfiles/.claude`) under `--global`, as the triage comment on the issue anticipated.

Instead, `writeAiFiles` now calls a new `writablePathEscapesRoot` helper (`src/utils/file.ts`) before reading or writing a target. It resolves the nearest existing ancestor of the output path and skips the file, with a warning, only when that real path sits **outside** the output root. A checked-out repository carrying `AGENTS.md -> ~/.bashrc` or `.cursor -> /etc` can therefore no longer redirect a write, while links that stay inside the root keep working. Paths whose root or intermediate directories do not exist yet pass, since no link can be on the way.

Two gaps found in review are closed in the follow-up commit:

- **Dangling links** were judged by the link's own (in-root) path because `realpath` cannot resolve them. `writablePathEscapesRoot` now follows a dangling link to the target it would create (relative targets resolved from the link's directory, chains capped at 40 hops) and judges that path, so `AGENTS.md -> ~/.zshenv` is refused even before `~/.zshenv` exists.
- The guard lived only in `FeatureProcessor.writeAiFiles`. It is now a shared `refusesWriteOutsideRoot` helper that `DirFeatureProcessor.writeAiDirs` (skills, subagent directories) and the Hermes / Goose config rewrites in the commands processor also call, so `.claude/skills -> /etc` is refused too.
- The Symlinks section of `docs/reference/file-formats.md` now documents the write-side guard.

Refs #2984

## Test plan

- [x] `pnpm cicheck` (12441 tests)
- [x] New tests: `nested-agentsmd-rule.test.ts`; `writablePathEscapesRoot` cases in `file.test.ts`; refusal and in-root-link cases in `feature-processor.test.ts`


dyoshikawa:
@dyoshikawa Thank you!

PR #3124, opened 2026-09-18
fix(hooks): stop reporting emitted override-block events as unsupported

dyoshikawa:
## Summary

Resolves item 2 of #2946 at the shared level rather than with a `zcode`-only flag.

The generic `Skipped hook event(s) for <tool> (not supported): ...` warning checked the tool-specific override block (`config[target].hooks`) for every target unless the factory opted out with `passthroughOverrideEvents`. Most adapters — every one built on the shared `canonicalToToolHooks` converter, plus cursor, copilot, copilotcli, tabnine, qwencode, the settings.json family and zcode — emit unknown override-block keys verbatim (`lookupOwn(...) ?? eventName`), so for 22 targets the warning claimed a drop that never happened. A probe over every factory found:

- **22 targets** emit the key and still warned (false claim) — now silent.
- **10 targets** really drop the key — amp, antigravity-plugin, claudecode-plugin, cline, deepagents, kilo, opencode, pi, reasonix, vibe — still warn.
- **3 targets** report their own drops or passthroughs (hermesagent, kimi-code, the deprecated kiro alias) — no generic echo.
- **kiro-cli / kiro-ide** emit unknown `kiro` block keys verbatim — silent (the earlier `passthroughOverrideEvents` opt-out already had this right).

## Changes

- `ToolHooksFactory.passthroughOverrideEvents` (opt-out, default warn) is replaced by `dropsUnknownOverrideEvents` (opt-in, default leave alone). `unsupportedEventNames` only inspects override-block keys when the flag is set; shared `hooks` keys are always checked as before.
- The flag is set on the ten adapters above; the six former `passthroughOverrideEvents: true` entries (crush, hermesagent, kimi-code, kiro, kiro-cli, kiro-ide) drop it and keep their explanatory comments.
- `HOOKS_OVERRIDE_KEY_ALIASES` is exported so the test can address the shared `kiro` block.
- Tests: a cursor case asserting that an override-only unknown key is emitted without a "not supported" warning, and an `it.each` matrix over `toolHooksFactories` asserting that the generic warning fires **iff** the key is absent from every generated file and the adapter did not warn about it itself (and never alongside an adapter warning). Mutation-checked: moving the flag onto cursor or off amp fails the respective case.

No user-visible file output changes; only the warning goes away where it was wrong.

## Remaining items of #2946

Not touched here — they are recorded on the issue: item 4 (per-tool timeout field selection), items 5/6 (global-scope safety switch / replace-owned-keys design notes), item 7 (nothing to harden per upstream docs), item 8 (symlink handling; partially covered by #3123).

Refs #2946


dyoshikawa:
@dyoshikawa Thank you!

PR #3125, opened 2026-09-18
feat(permissions): add Pool permissions support

dyoshikawa:
## Summary

Adds the `permissions` feature for Pool (`pool` target), the third dimension tracked in #2945.

- **Generate** (project + global): rulesync tool rules become `tools.<name>.allow` / `tools.<name>.deny` lists (`bash` → `shell`, `webfetch` → `web_fetch`, `websearch` → `web_search`, other categories pass through), and `read` / `edit` / `write` rules become `paths.allow` (`{ path }` or `{ path, write: true }`) and `paths.deny` entries in `.poolside/settings.yaml` or `~/.config/poolside/settings.yaml`.
- The file is shared with the mcp feature, so writes go through the shared-config gateway with a `deep-merge` policy: `pool`, `mcp_servers`, `disabled`, and unmanaged tool blocks are preserved; the managed `allow`/`deny` lists and the `paths` block are rebuilt from rulesync each run.
- Rules Pool cannot express are **withheld with a warning** instead of being written weaker than authored: `ask` (Pool has no ask list, so an overlapping `allow` is dropped to keep prompting), all-tools `*` rules (mirrored onto `bash` only), MCP categories (owned by `poolAllow` in `mcp.json`), non-`*` wildcards (`?`, `[]`, `{}`), write-only denies (Pool cannot deny writes without denying reads), and paths whose form does not fit the scope (project settings take relative paths, user settings take absolute / `~` paths).
- **Import**: `tools.<name>` lists map back (deny wins, `disabled: true` → `"*": deny`), `paths.allow` → `read` allow plus `edit`/`write` allow when `write: true`, `paths.deny` → deny for all three file categories.
- Registered `pool` in the permissions processor / tool-target tuples, added the `pool` override key, updated the shared-file derive snapshot, e2e Tool × Feature cases (generate, import, global), docs (`docs/reference/file-formats.md`), and regenerated tables / embedded docs.

Source: https://docs.poolside.ai/settings-file-reference (`tools`, `paths` sections).

Refs #2945

## Test plan

- [x] `pnpm cicheck` green
- [x] `src/features/permissions/pool-permissions.test.ts` (24 cases)
- [x] `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-permissions.spec.ts -t pool`


dyoshikawa:
@dyoshikawa Thank you!

PR #3126, opened 2026-09-18
feat(hooks): add Pool hooks support

dyoshikawa:
## Summary

Adds the **hooks** feature for Pool (Poolside's coding agent CLI), the third of the four dimensions listed in #2945 (rules landed in #2976, MCP in #3091, permissions is in flight as #3125).

Pool keeps its hooks under the `hooks` key of the same YAML settings file as its MCP servers and permissions — `.poolside/settings.yaml` (project) and `~/.config/poolside/settings.yaml` (global, via `--global`) — as a flat per-event list of `{name?, matcher, command, timeout?}` entries ([hooks docs](https://docs.poolside.ai/hooks), [settings file reference](https://docs.poolside.ai/settings-file-reference)).

### Mapping

- Six canonical events map onto Pool's: `preToolUse` ⇄ `PreToolUse`, `postToolUse` ⇄ `PostToolUse`, `beforeSubmitPrompt` ⇄ `UserPromptSubmit`, `preCompact` ⇄ `PreCompact`, `sessionStart` ⇄ `SessionStart`, `stop` ⇄ `Stop`. Other events are skipped with the processor's usual warning; only `command` hooks are written (a command hook without a command is skipped with a warning).
- Pool's schema requires `matcher` on every entry but only tests it on `PreToolUse`/`PostToolUse` (against the tool name: `*`/`""` any, bare name exact, `a|b` list, else regex). The canonical matcher is written on those two events (`*` when absent); on the other events the `*` placeholder is written and an authored matcher is dropped with a warning.
- `timeout` is an integer number of seconds (Pool defaults to 60), so a fractional value is rounded up.
- The event lists of the `hooks` block are owned through the shared-config gateway (`replace-owned-keys` on `hooks`), while `stop_hook_max_continuations` and every sibling key (`pool`, `mcp_servers`, `tools`, `paths`, `sandbox`, ...) survive a regenerate; an empty payload creates no file and the file is never deleted.
- Import reads the six events back (a `*`/empty matcher is dropped, as is a matcher on an event Pool never tests it on) and files any other event key under the `pool.hooks` override so it round-trips verbatim.

### Changes

- `src/features/hooks/pool-hooks.ts` (+ unit tests) and the `pool` entry of the hooks factory map.
- `POOL_HOOK_EVENTS` / event-name tables and the `pool.hooks` override key in `src/types/hooks.ts`; `pool` in the hooks target tuple.
- `hooks` added to the Pool settings declaration of the shared-config gateway (derive snapshot updated).
- Standalone e2e cases at project and global scope, docs (`file-formats.md` per-tool location + Pool note, override-key list), regenerated tables/docs embed.

### Base branch

This branch is stacked on `resolve-issue-2945-pool-permissions` (#3125) because both features share the Pool settings declaration; the PR is opened against that branch so the diff shows only the hooks change, and it will be retargeted to `main` once #3125 merges.

Refs #2945


dyoshikawa:
@dyoshikawa Thank you!

PR #3127, opened 2026-09-18
feat(subagents): add Pool subagents support

dyoshikawa:
## Summary

Adds native **subagents** support for Pool (Poolside), the third of the remaining feature dimensions tracked in #2945 (after permissions in #3125 and hooks in #3126).

Pool has no per-agent files: every subagent is an entry of the `subagents.agents` map in its settings file. The new `PoolSubagent` adapter therefore aggregates all rulesync subagents targeting `pool` into `.poolside/settings.yaml` (project) or `~/.config/poolside/settings.yaml` (`--global`), going through the shared-config gateway that already routes the MCP feature's `mcp_servers` block into the same file.

### Behaviour

- **Generate:** each subagent becomes one agent keyed by its `name`, with `type` (defaults to `in_process`), `description`, `instructions` (the Markdown body, or `pool.instructions`) and whatever the `pool:` frontmatter block adds (`command`, `args`, `env`, `agent_server`, `inherit_agent_config`, `disabled`, …). A missing description on an enabled agent is warned about, since Pool requires one.
- **Ownership:** rulesync owns the whole `subagents` block via `replace-owned-keys`. It preserves `subagents.default` and Pool's built-in `general` entry, replaces every other agent, and retracts the map when no subagent targets Pool (`emitsEmptyAggregate`), so agents from a previous run never linger. Other top-level keys (`pool`, `mcp_servers`, …) are untouched.
- **Guards:** a subagent named `general` is skipped with a warning (reserved built-in), duplicate names keep the last definition with a warning, prototype-pollution keys are neither written nor read, and a settings file whose root is not a mapping fails closed.
- **Import:** `rulesync import --targets pool` fans `subagents.agents` (except `general`) out to `.rulesync/subagents/<sanitized-name>.md`, moving `instructions` into the body and the non-default keys into the `pool:` block; `type: in_process` is omitted on round-trip.
- `ToolSubagentSettablePaths` gains an optional `relativeFilePath` (mirroring `ToolCheckSettablePaths`) so the gitignore derivation narrows to the shared settings file — which is user-managed and therefore stays un-ignored — instead of claiming `**/.poolside/`.

### Files

- `src/features/subagents/pool-subagent.ts` (+ tests): the adapter.
- `src/features/subagents/subagents-processor.ts`: registers `pool`; the factory type gains `logger` and `meta.emitsEmptyAggregate`.
- `src/features/subagents/rulesync-subagent.ts`: `pool:` frontmatter schema.
- `src/features/shared/shared-config-gateway.ts`: `subagents` ownership on Pool's settings declaration.
- `src/constants/pool-paths.ts`, `src/types/tool-target-tuples.ts`, `src/lib/shared-file-derive.test.ts` snapshot.
- `src/e2e/e2e-subagents.spec.ts`: `pool` added to the project and global happy-path matrices, plus a merge-preservation and an import case.
- `docs/reference/file-formats.md`: `pool:` block in the frontmatter example and a Pool note; generated tables and docs content regenerated.

Refs #2945

## Test plan

- [x] `pnpm cicheck` (459 files / 12519 tests, content checks)
- [x] `npx vitest run --config vitest.e2e.config.ts src/e2e/e2e-subagents.spec.ts -t pool` (4 cases)


dyoshikawa:
@dyoshikawa Thank you!

Issue #3131, opened 2026-09-18
generate -c/--config is ignored because the CLI passes `config` but ConfigResolver expects `configPath`


rudironsoni:
## Summary

`rulesync generate -c <path>` does not load the requested config file.

I reproduced this on **v16.37.0** while using two colocated RuleSync source trees. Inspecting the implementation shows a wiring mismatch:

- `generate` declares Commander option `-c, --config <path>`
- the action calls `generateCommand(logger, options as GenerateOptions)` unchanged
- `GenerateOptions` extends `ConfigResolverResolveParams`
- `ConfigResolverResolveParams` expects `configPath`, not `config`

As a result, `ConfigResolver.resolve()` falls back to its default `rulesync.jsonc` path instead of using the file passed with `-c`.

Current `main` appears to contain the same wiring.

## Reproduction

Create two configs whose target sets are visibly different, then run:

```sh
rulesync generate -c path/to/alternate-rulesync.jsonc --dry-run --verbose
```

Observed: generation behaves as though `-c` was not supplied and uses the config resolved from the working directory/default path.

A useful control is to make the alternate config exclude a target that the cwd/default config enables. The supposedly excluded target is still generated.

## Expected

`rulesync generate -c path/to/config.jsonc` loads that file as the base config.

## Likely cause

In `src/cli/program.ts`, the `generate` action forwards Commander options directly:

```ts
.action(
  wrapCommand("generate", "GENERATION_FAILED", async (logger, options) => {
    await generateCommand(logger, options as GenerateOptions);
  }),
);
```

Commander populates `options.config`, but the resolver consumes `configPath`.

Other commands such as `install` explicitly map the option:

```ts
configPath: (options as { config?: string }).config
```

Doing the equivalent mapping for `generate` should restore the documented `-c/--config` behavior.

## Related note

`inputRoots` in a config file is not itself the problem. It is honored when the config is actually loaded. The failure is that `generate -c` never passes the requested config path into the resolver.

PR #3132, opened 2026-09-19
fix: map -c/--config to configPath in generate command

RinZ27:
Running `rulesync generate -c path/to/config.jsonc` silently falls back to the default config because Commander populates `options.config`, but `GenerateOptions` (via `ConfigResolverResolveParams`) expects `configPath`.

Other commands like `install` and `add` already handle this correctly by explicitly mapping the property — noticed `generate` was the odd one out.

Applied the same `Omit + re-map` pattern that `add` uses at line 115, so the wiring stays consistent across all subcommands.


dyoshikawa:
Thank you for the PR, and sorry for the overlap: the same fix landed on `main` via #3133 (closing #3131) shortly after this PR was opened, and it shipped in v16.39.1. Because that commit touches the same lines in `src/cli/program.ts`, this branch now conflicts with `main`.

Since the bug is already fixed and released, I intend to close this PR as superseded rather than asking you to rebase. Thanks again for spotting the inconsistency and for the clear write-up.

PR #3133, opened 2026-09-19
fix(cli): forward generate --config to the config resolver as configPath

### LINKED ISSUE #3131: generate -c/--config is ignored because the CLI passes `config` but ConfigResolver expects `configPath`

dyoshikawa:
## Summary

`rulesync generate -c <path>` / `--config <path>` was silently ignored: Commander stores the option as `options.config`, but the `generate` action passed the raw options object straight to `generateCommand`, whose `ConfigResolver.resolve` only reads `configPath`. The default `rulesync.jsonc` was loaded regardless of the flag, while `add` and `install` already mapped `config` → `configPath` explicitly.

Closes #3131

## Changes

- `src/cli/program.ts`: destructure `config` out of the `generate` options and forward it as `configPath`, matching the `add` / `install` wiring. `rulesync.local.jsonc` overlaying and `--watch`'s config-file watching both derive from the resolved path, so they follow the flag as well.
- `src/cli/program.test.ts`: regression tests that parse `generate --config …`, `generate -c …`, and a bare `generate`, and assert what reaches `generateCommand` (the three fail on `main`).
- `docs/reference/cli-commands.md`: add the missing `--config, -c <path>` row to the `generate` options table (the flag already existed in `--help` but was undocumented there); `src/generated/docs-content.ts` regenerated.

## Verification

- `pnpm cicheck` passes (461 test files / 12581 tests).
- The new tests fail without the `program.ts` change and pass with it.


dyoshikawa:
@dyoshikawa Thank you!

PR #3134, opened 2026-09-19
chore(release): v16.39.1

### LINKED ISSUE #3131: generate -c/--config is ignored because the CLI passes `config` but ConfigResolver expects `configPath`

dyoshikawa:
### Bug Fixes

- fix(cli): forward `generate --config` to the config resolver as `configPath` by @dyoshikawa in https://github.com/dyoshikawa/rulesync/pull/3133
  - `rulesync generate -c/--config <path>` was silently ignored and always loaded the default `rulesync.jsonc`; the flag now selects the base configuration file (with `rulesync.local.jsonc` next to it still overlaying it, and `--watch` following the selected file). The option is now also listed in the `generate` reference. Fixes #3131.

### Other Changes

- ci(deps): Bump anomalyco/opencode/github in the all-actions group by @dependabot in https://github.com/dyoshikawa/rulesync/pull/3128
- chore: update Homebrew formula to v16.39.0 by @dyoshikawa in https://github.com/dyoshikawa/rulesync/pull/3130

## Contributors

- @dyoshikawa
- @dependabot
- Thanks to @rudironsoni for reporting #3131

## Full Changelog

https://github.com/dyoshikawa/rulesync/compare/v16.39.0...v16.39.1


dyoshikawa:
@dyoshikawa Thank you!

PR #3136, opened 2026-09-19
fix(codex): handle unsupported filesystem globs

besdar:
Codex only supports `deny` access for filesystem globs such as `*.ts` or `**/*.tf`. Read and write access is supported only for exact paths or subtree patterns ending in `/**`.

Rulesync previously generated invalid Codex configurations when a canonical permission rule combined a non-trailing glob with read-only access. For example, generated:

```toml
"some/path/to/file/*.ts" = "read"
```

Codex Desktop rejected this configuration during message submission.

This change makes the Codex adapter detect unsupported read/write globs and emit `deny` with a warning instead of generating an invalid configuration. Exact paths and trailing `/**` patterns continue to support `read` and `write` access.

The change also preserves readable, non-writable access for valid `edit`/`write` `ask` and `deny` rules, adds regression coverage, and updates the embedded documentation.


src/features/permissions/codexcli-permissions.ts
```diff
@@ -445,7 +445,7 @@ function convertRulesyncToCodexProfile({
       filesystem,
       workspaceRootFilesystem,
       pattern,
-      access,
+      access: normalizeCodexFilesystemAccess({ pattern, access, logger }),
```
dyoshikawa: This is where the inversion happens: by the time `normalizeCodexFilesystemAccess` runs, a `read` that came from `read: allow` (no write-side rule) and a `read` that came from `write: deny` are indistinguishable, and both become `deny` for a non-trailing glob. `mergeFilesystemCategoryRules` knows which side produced the level, so it could return that (or the normalization could take the original read/write actions) and skip-with-warning for the grant case instead of denying.

src/features/permissions/codexcli-permissions.ts
```diff
@@ -1100,8 +1131,8 @@ function mapReadAction(action: PermissionAction): "read" | "deny" {
   return action === "allow" ? "read" : "deny";
 }
 
-function mapWriteAction(action: PermissionAction): "write" | "deny" {
-  return action === "allow" ? "write" : "deny";
+function mapWriteAction(action: PermissionAction): "write" | "read" {
```
dyoshikawa, line 1179: With this change a write-side-only `deny` emits `"read"`. Because the merge is keyed by the exact pattern string, a broader `read: deny` on a parent path (or `:root`) is not consulted, and Codex gives the deeper entry priority, so the `"read"` entry reopens that subtree. Could you add a warning here for the `deny` → `read` case as well (the `ask` branch already has one), so existing users notice their generated `deny` entries changing?

src/features/permissions/codexcli-permissions.ts
```diff
@@ -1168,6 +1200,12 @@ function mergeFilesystemCategoryRules({
     const readAction = readRules[pattern];
     const writeAction = writeSideRules[pattern];
 
+    if (writeAction === "ask") {
```
dyoshikawa: Minor: this fires before the `readAction` branch, so for `read: deny` + `write: ask` the message says the pattern "will be emitted without write access" although the final level is `deny` for a different reason, and for `edit: { "*.ts": "ask" }` the user gets this warning plus the glob warning for the same pattern. Moving the check below the `readAction` handling (or wording it by outcome) would tidy that up.

src/features/permissions/codexcli-permissions.test.ts
```diff
@@ -384,8 +439,8 @@ enabled = true
 
     const fileContent = codexPermissions.getFileContent();
     expect(fileContent).toContain('[permissions.rulesync.filesystem.":workspace_roots"]');
-    expect(fileContent).toContain('"src/*" = "read"');
-    expect(fileContent).toContain('"docs/*" = "write"');
+    expect(fileContent).toContain('"src/*" = "deny"');
```
dyoshikawa: This expectation pins the allow→deny inversion: the input here is `read: { "src/*": "allow" }` / `edit: { "docs/*": "allow" }`, and the test now asserts both are denied. Whatever fallback we settle on for allow-derived globs, this test should assert that behavior explicitly (skip + warning, or similar) rather than a deny.

docs/reference/file-formats.md
```diff
@@ -1656,10 +1656,10 @@ For Codex CLI, this generates a `rulesync` named profile in `.codex/config.toml`
 
 - `bash`: generates one `prefix_rule(...)` per command pattern in `.codex/rules/rulesync.rules` (`allow` → `allow`, `ask` → `prompt`, `deny` → `forbidden`)
 - `read`: `allow` → `read`, `ask`/`deny` → `deny` in `permissions.<profile>.filesystem`
-- `edit` / `write`: `allow` → `write`, `ask`/`deny` → `deny` in `permissions.<profile>.filesystem`
+- `edit` / `write`: `allow` → `write`, while `ask`/`deny` → `read` in `permissions.<profile>.filesystem` when the path is exact or ends in `/**`; Codex has no path-level write-approval state, so both non-allow write actions preserve read access without write access and `ask` emits a warning. An explicit `read: deny` still takes precedence and emits `deny`. For other globs, Codex supports only `deny`, so Rulesync emits `deny` with a warning rather than generating an invalid profile.
```
dyoshikawa: "An explicit `read: deny` still takes precedence" is only true when the `read: deny` is on the same pattern key. A `read: deny` on a parent path (e.g. `/secrets/**` vs `/secrets/db/**`) does not take precedence — the more specific `"read"` entry wins in Codex. Worth saying "for the same pattern" here so readers don't assume a broader deny protects them.

dyoshikawa (commented):
Thanks for this — the glob detection itself is right. I checked it against Codex's `compile_read_write_glob_path` / `contains_glob_chars_for_platform` (strip one trailing `/**`, then look for `* ? [ ]`) and it matches exactly, and the regenerated `docs-content.ts` is in sync with the markdown. The direction (never emit a profile Codex will refuse to load) is the right one.

What I can't merge as-is is the fallback for *allow*-derived globs. `normalizeCodexFilesystemAccess` only sees the merged access level, so `read: { "**": "allow" }` or `write: { "docs/*.md": "allow" }` now come out as `"**" = "deny"` / `"docs/*.md" = "deny"` — the exact opposite of what the user wrote. `read: { "**": "allow" }` in particular is what `rulesync import --targets pool` produces (and `src/**/*.ts`-style allows are common in hand-written files), so with this change a plain generate would make the whole workspace unreadable in Codex (with a warning, but the generated profile loads fine, so it's easy to miss). Before this PR Codex rejected the config at load time, which was at least loud. I think the safer mapping is: for a merged level that comes from a grant (`write`, or `read` from a read-side `allow` with no write-side restriction), skip the entry with a warning and let the baseline apply (both `:workspace` and `:read-only` already grant read everywhere); only downgrade to `deny` when the level comes from a write-side `deny`/`ask` restriction. Happy to discuss alternatives, but the allow→deny inversion needs to go before merging.

Second, the `mapWriteAction` change: I agree `edit: deny` → `"read"` is the more faithful mapping for a single key, but the merge is per pattern key, so a broader read deny under a different key is no longer honored — `read: { "/secrets/**": "deny" }` + `edit: { "/secrets/db/**": "deny" }` now emits `"/secrets/db/**" = "read"`, and Codex resolves the more specific path first (its `resolve_access` picks the deepest entry, and `:root = "deny"` is excluded from the deny-read matcher), so `/secrets/db` becomes readable inside a tree the user denied. It also changes every existing user's generated `.codex/config.toml` from `deny` to `read` with no warning. Not blocking on its own, but at minimum this deserves a warning on the `deny` → `read` path (like the `ask` one) and a note that the docs sentence about `read: deny` taking precedence only holds for the same pattern key.

I left the smaller points inline. I'll pick up the round-trip asymmetry (`"read"` importing as `read: allow` only, dropping the edit deny) and the Windows device-path `?` edge case as follow-up issues on my side unless you want to fold them in.

src/features/permissions/codexcli-permissions.ts
besdar: Fixed. The merge now preserves whether an emitted access level came from a grant or from a write-side `ask`/`deny` restriction. Unsupported non-trailing grant globs are skipped with a warning so the selected Codex baseline remains effective. Unsupported write-side restrictions still emit `deny`, since Codex cannot represent read-only access for those globs.

src/features/permissions/codexcli-permissions.ts
besdar: Fixed. A write-side `deny` that maps to `read` now emits a warning explaining that a broader `read: deny` can be overridden by the more-specific read-only path.

src/features/permissions/codexcli-permissions.ts
besdar: Fixed. The `ask` warning is now emitted only when the final emitted level is `read`. A read-side denial that already determines the final `deny` produces no misleading `ask` warning, and an unsupported glob with a write-side restriction produces only its fallback warning.

src/features/permissions/codexcli-permissions.test.ts
besdar: Fixed. The test now asserts that unsupported grant-derived globs are omitted and that Rulesync warns that the base permission profile remains in effect. It no longer expects them to become `deny`.

docs/reference/file-formats.md
besdar: Fixed. The documentation now says this precedence applies only when the read denial uses the same pattern. It also explains the warning for a more-specific read-only entry that can override a broader denial.

dyoshikawa:
@besdar Thank you!

